Issue #33, opened 2021-09-22
Improve display of tick/frame info


byorgey:
Right now the ticks per second and ticks per frame are displayed centered at the top of the world view, and the ticks per frame value is updated every frame.  I have in mind several ways this could be improved:

* Only update the ticks/frame value every second or so.
    * Add a counter to the `UIState` to count frames (we already have one to count ticks).
    * Get rid of the `frameTickHist` list, the `frameTickHistSize`, and the `rememberFrameTicks` function.
    * Also store the `TimeSpec` corresponding to the last time we updated the info in the UI.
    * In the `updateUI` function, check if the last time we updated the info in the UI was at least one second ago.  If so, reset the time, compute ticks per frame and frames per second, and reset both counters to zero.
* Display ticks per frame AND frames per second, perhaps using compact abbreviations like TPF and FPS, or t/f and f/s, or something like that.
* We should always show the ticks per second; but we could add a key (probably `f`?) to let the user toggle the display of TPF + FPS on and off. 
* ~~Once #37 is implemented, we could move the display of these things to e.g. the bottom right of the world border instead of centered at the top.~~ (done in https://github.com/byorgey/swarm/commit/aef3d4e7ec1ba935a6016ef24d709686a8188efe)

PR #47, opened 2021-09-24
Reduce framerate to ~100fps

HuwCampbell:
I was seeing CPU usage at 100%, which I believe is mostly due to this hot loop. The game probably shouldn't take up all of one's CPU.

### DIFF
--- a/src/Swarm/App.hs
+++ b/src/Swarm/App.hs
@@ -14,7 +14,7 @@
 
 module Swarm.App where
 
-import           Control.Concurrent   (forkIO)
+import           Control.Concurrent   (forkIO, threadDelay)
 
 import           Brick
 import           Brick.BChan
@@ -46,10 +46,10 @@ appMain = do
     Left errMsg -> T.putStrLn errMsg
     Right s -> do
 
-      -- Just send Frame events as fast as possible.  The game is
-      -- responsible for figuring out how many steps to take each
-      -- frame to achieve the desired speed, regardless of the frame
-      -- rate.
+      -- Send Frame events as at a reasonable rate for 60 fps. The
+      -- game is responsible for figuring out how many steps to take
+      -- each frame to achieve the desired speed, regardless of the
+      -- frame rate.
       --
       -- 5 is the size of the bounded channel; when it gets that big,
       -- any writes to it will block.  Probably 1 would work fine,
@@ -58,7 +58,9 @@ appMain = do
       -- and do another write.
 
       chan <- newBChan 5
-      _ <- forkIO $ forever $ writeBChan chan Frame
+      _ <- forkIO $ forever $ do
+        threadDelay 16_666
+        writeBChan chan Frame
 
       -- Run the app.
 


TristanCacqueray:
This is still quite a busy loop, should the delay be `10_000_000` instead?

TristanCacqueray:
Otherwise it seems like the Swarm.TUI.Controller.handleEvent should handle the delay length as explained in the `Variable speed` section of https://samtay.github.io/posts/introduction-to-brick

TristanCacqueray:
@byorgey it seems like `runFrame` is already computing the frameTime, do you think it could also set a delay TVar to cap the fps so that the busy loop could threadDelay accordingly?

byorgey:
Thanks!  You're right, this is definitely a problem---there's no reason it needs to eat 100% of the CPU in a busy loop when there's not much going on in the game.

At first glance I thought that adding a `threadDelay` in this thread generating `Frame` events would not be the right way to fix this.  But now that I think about it more I think it actually is the right way.  Since this thread is independent of the thread that actually runs the game, it will not artificially limit the amount of computation the game can perform---if there is a lot going on then the game can still end up using 100% CPU if it needs to.  And if the game cannot render 100 FPS then no harm done, this thread will simply fill up the bounded channel with `Frame` events and then block until the next frame.  All this does is make sure that `Frame` events are generated *no faster than* 100 FPS.

Note that [the argument to `threadDelay` is in microseconds](https://hackage.haskell.org/package/base-4.15.0.0/docs/Control-Concurrent.html#v:threadDelay), so a delay of `10_000` would be the correct number to cap things at 100 FPS.  The number `100000` in the PR as it stands would correspond to 10 FPS.  Let's definitely make use of the `NumericUnderscores` extension to make it easier to read the number.

@TristanCacqueray , we could indeed set up a `TVar` so that the maximum framerate is under the control of the game.  However, I'm not sure I understand what the benefit of this would be.  Is it so the user would be able to set the maximum framerate if they want?  Note that the speed at which the game runs is independent of the frame rate.  The user can request a specific number of game ticks per second, and the number of ticks to step the game in a single frame is determined dynamically by the amount of time elapsed since the previous `Frame` event.

byorgey:
Also, 100 FPS seems excessive, is there any reason not to limit the maximum framerate to, say, 60 FPS?

TristanCacqueray:
@byorgey I meant to add a TVar so that the delay could be adjusted dynamically to avoid unnecessary slowdown, e.g. something similar to: https://github.com/TristanCacqueray/swarm/commit/2692607aa5738055890c28e791c3a494fc211018 .

However that does not work because the game runs is indeed independent of the frame rate :) So I guess the frameDelay could be increased by the gameTick when it is idle (e.g. when no robots are active).

TristanCacqueray:
Well it is not so simple to adjust this delay dynamically because when the game transition from idle to busy then we should update the display as fast as possible...

On my system, the game runs at ~40fps, thus the delay would need to be higher to have an effect. Though using `threadDelay 17_000` would help faster system by capping the fps at 60.

byorgey:
Yeah, I don't think it is really worth it to try to figure out some complicated dynamic framerate cap, at least not until there is a demonstrated need to do so.  I'm much more of a mind to just do something simple here like what @HuwCampbell 's PR does.  I think it will clearly benefit those with fast systems without unduly complicating things.

byorgey:
Just as a data point, on my system (2.4 GHz Intel i5) when I put in a `threadDelay 16_666` (thus capping the framerate at 60 FPS), it uses about 50% CPU.

Also, @TristanCacqueray , how do you know it runs at 40fps on your system?  Did you add some code to measure it?  (If so, want to take a crack at #33 ?)

TristanCacqueray:
@byorgey sure I can have a look, I got the number by printing the frameTime in the runFrame.

byorgey:
@HuwCampbell if you change the number in your PR to `16_666` I will merge it.  If we want, we can open a new issue to continue discussing more complex dynamic adjustment of the framerate.

HuwCampbell:
Ahh the joys of being in Australia, put up a PR then wake up to big conversations hours later.

Of course you're right, I was limiting it to 10fps, which did seem fine for gameplay and limited CPU usage to about 15% on startup.

With 60 fps it's closer 10 50%, which is still a bit high in my opinion, but I guess we can refine. I've update the PR and comment.

TristanCacqueray (approved):

TristanCacqueray:
@HuwCampbell thanks! After more investigations it seems like the runFrameUI can be optimized to further reduce the CPU load.

PR #56, opened 2021-09-24
Add continueWithoutRedraw when possible to save cpu cycles

TristanCacqueray:
This change requires #47 and further improve the cpu cost per frame.

### DIFF
--- a/src/Swarm/TUI/Controller.hs
+++ b/src/Swarm/TUI/Controller.hs
@@ -130,7 +130,9 @@ shutdown s = do
 --   this may involve stepping the game any number of ticks (including
 --   zero).
 runFrameUI :: AppState -> EventM Name (Next AppState)
-runFrameUI s = execStateT (runFrame >> updateUI) s >>= continue
+runFrameUI s = do
+  (redraw, newState) <- runStateT (runFrame >> updateUI) s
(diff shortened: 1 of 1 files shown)


src/Swarm/TUI/Controller.hs
```diff
@@ -197,7 +199,7 @@ runGameTick = zoom gameState gameTick
 
 -- | Update the UI.  This function is used after running the
 --   game for some number of ticks.
-updateUI :: StateT AppState (EventM Name) ()
+updateUI :: StateT AppState (EventM Name) Bool
 updateUI = do
 
   loadVisibleRegion
```
TristanCacqueray, line 205: not sure what to do with the result of `loadVisibleRegion`, it seems like it always modifies `gameState . world`

byorgey: `loadVisibleRegion` just makes sure that the world has loaded all tiles that overlap the visible area into its tile cache.  It doesn't affect whether the screen needs to be redrawn.

byorgey:
Seems to work great.  With this and #47 applied, my CPU usage is down to around 5-10% while the game is idle.  We could consider bringing the max frame rate down even more --- I really don't think anyone can possibly notice the difference between 30 FPS and 60 FPS with this kind of game.

PR #57, opened 2021-09-24
Display FPS counter

### LINKED ISSUE #33: Improve display of tick/frame info

TristanCacqueray:
Fixes #33

### DIFF
--- a/src/Swarm/TUI/Controller.hs
+++ b/src/Swarm/TUI/Controller.hs
@@ -170,9 +170,33 @@ runFrame = do
         | lgTPS >= 0 = oneSecond `div` (1 `shiftL` lgTPS)
         | otherwise  = oneSecond * (1 `shiftL` abs lgTPS)
 
+  -- Update TPS/FPS counters every second
+  infoUpdateTime <- use (uiState . lastInfoTime)
+  let updateTime = toNanoSecs $ diffTimeSpec curTime infoUpdateTime
+  when (updateTime >= oneSecond) $ do
+    -- Wait for at least one second to have elapsed
+    when (infoUpdateTime /= 0) $ do
+      -- set how much frame got processed per second
+      frames <- use (uiState . frameCount)
+      uiState . uiFPS .= fromIntegral (frames * fromInteger oneSecond) / fromIntegral updateTime
+
+      -- set how much ticks got processed per frame
+      ticks <- use (uiState . tickCount)
+      uiState . uiTPF .= fromIntegral ticks / fromIntegral frames
+
+      -- ensure this frame gets drawn
+      gameState . needsRedraw .= True
+
+    -- Reset the counter and wait another seconds for the next update
+    uiState . tickCount .= 0
+    uiState . frameCount .= 0
+    uiState . lastInfoTime .= curTime
+
+  -- Increment the frame count
+  uiState . frameCount += 1
+
   -- Now do as many ticks as we need to catch up.
   runFrameTicks (fromNanoSecs dt)
-  zoom uiState rememberFrameTicks
 
 -- | Do zero or more ticks, with each tick notionally taking the given
 --   timestep, until we have used up all available accumulated time.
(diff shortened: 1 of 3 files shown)


TristanCacqueray:
@byorgey is it ok to store the TPS/FPS in the UIState? And should I add a Bool to manage the fps display toggle?

src/Swarm/TUI/Controller.hs
```diff
@@ -164,9 +164,31 @@ runFrame = do
...
+      let calc counterLens valueLens = do
+            count <- use (uiState . counterLens)
+            let countSec = toInteger count * oneSecond
+            uiState . valueLens .= fromIntegral countSec / fromIntegral updateTime
+      calc frameTicks uiTPS
+      calc frameCount uiFPS
+      -- ensure this frame gets drawn
+      gameState . needsRedraw .= True
+    -- Reset the counter and wait another seconds for the next update
+    uiState . frameTicks .= 0
```
byorgey, line 183: Let's change the name of `frameTicks` to `tickCount`, since it no longer counts the ticks in a single frame, but the ticks since the last info widget update.

byorgey (commented):
Yes, storing the TPS/FPS in the UIState makes perfect sense, we need that info every time we redraw the UI.  And yes, if you want to do the fps display toggling, adding a Bool to the UIState would make sense.

src/Swarm/TUI/Model.hs
```diff
@@ -176,26 +179,24 @@ uiInventory :: Lens' UIState (Maybe (Int, BL.List Name InventoryEntry))
 --   error message that is shown on top of the rest of the UI.
 uiError :: Lens' UIState (Maybe (Widget Name))
 
+-- | Computed ticks per milli seconds
+uiTPS :: Lens' UIState Double
```
byorgey, line 183: It's not ticks per second that we want to display (the ticks per second is set directly by the user), but rather the ticks per frame.

TristanCacqueray: Oops, I see :)

TristanCacqueray:
Ok I think this is now ready, though the tpf is different:

- Previously: `8 ticks / s (1.0 ticks/frame)`  (at 30fps)
- Now: `8 ticks / s (0.3 tpf, 29.1 fps)`

Though for 8 ticks/s at 30fps, the new 0.3 ticks/frame seems correct (8 / 30 = 0.2.666)

TristanCacqueray:
Also when trying higher tps value, it seems like we need #67 to ensure the value are properly updated on screen.

byorgey:
Agree the new tpf value seems correct, not sure what was going on before.

This seems good, my only suggestion is to show/hide *both* the tpf and fps values when hitting `f`.  I think normal users will only care about the ticks / second (since that is what they can control).

TristanCacqueray:
Alright, the latest commit hide the whole info box by default.

byorgey:
Thanks, works great!

PR #62, opened 2021-09-25
Further reduce maximum FPS to 30

byorgey:
#47 reduced the maximum framerate to 60 FPS, which helped with CPU usage when the game is idle.  I propose to lower the cap further, to 30 FPS.  There is a reason movies run at 30 FPS.  I know some games run at higher rates than that, and probably it is possible to tell the difference---a game with nice graphics running at 60 FPS probably "feels smoother" than one running at 30.  However, "nice graphics" and "smooth" definitely are not words to describe an ASCII art simulation.  So I am highly skeptical that anyone would be able to tell the difference between running Swarm at 30 FPS or at something higher.  Besides, on my machine at least, when there are robots actively doing stuff and the screen has to be redrawn every frame, performance degrades to about 20 FPS anyway.

@TristanCacqueray , @HuwCampbell , would be interested to hear your feedback!

### DIFF
--- a/src/Swarm/App.hs
+++ b/src/Swarm/App.hs
@@ -46,10 +46,13 @@ appMain = do
     Left errMsg -> T.putStrLn errMsg
     Right s -> do
 
-      -- Send Frame events as at a reasonable rate for 60 fps. The
+      -- Send Frame events as at a reasonable rate for 30 fps. The
       -- game is responsible for figuring out how many steps to take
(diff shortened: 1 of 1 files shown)


HuwCampbell:
Sounds very reasonable to me.

TristanCacqueray (approved):
Thanks, 30fps seems more terminal friendly.

PR #63, opened 2021-09-25
Update UI when game mode is changed

ElderEphemera:
Similarly to cd54fe6, this is needed after #56.


TristanCacqueray (approved):

TristanCacqueray:
Thanks, good catch.

PR #64, opened 2021-09-25
Enable bracketed paste mode to resolve #41

### LINKED ISSUE #41: Handle bracketed paste events in the REPL
Right now, if you paste something into the REPL, it gets sent one keypress event at a time, which in particular means the parser and typechecker get re-run between every single key event.  It's not *too* slow, but the lag is noticeable.

[According to the brick user manual](https://github.com/jtdaugherty/brick/blob/master/docs/guide.rst#paste-support), however, the `vty` library supports a "bracketed paste mode"---if the terminal does---which sends the entire paste as a `ByteString` in a single paste event.  We should definitely look into enabling this.  I'm not yet sure whether we would need to update the REPL event handler; it passes most events on to the input form event handler (provided by `brick`), so it depends on whether the input form event handler knows how to handle paste events.  If it does, this might be as simple as just turning on `vty` bracketed paste support.

samtay:
Tested this out in my environment (MacOS & kitty terminal) and it worked great! Very noticeable lag when pasting full sentences is now gone completely.

Closes #41. Happy to make any changes, however small/stylistic.

### DIFF
--- a/src/Swarm/App.hs
+++ b/src/Swarm/App.hs
@@ -18,6 +18,7 @@ import           Control.Concurrent   (forkIO, threadDelay)
 
 import           Brick
 import           Brick.BChan
+import           Brick.Main
 import qualified Graphics.Vty         as V
 
 import           Control.Monad.Except
@@ -33,7 +34,7 @@ app = App
   { appDraw         = drawUI
   , appChooseCursor = showFirstCursor
   , appHandleEvent  = handleEvent
-  , appStartEvent   = return
+  , appStartEvent   = \s -> s <$ enablePasteMode
   , appAttrMap      = const swarmAttrMap
   }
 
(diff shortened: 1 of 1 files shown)


TristanCacqueray (approved):
That also works on Linux. I thought the previous mode, where pasting looked like it was typed, was a feature :).

I like the new behavior, could you please check this @byorgey

src/Swarm/App.hs
```diff
@@ -18,6 +18,7 @@ import           Control.Concurrent   (forkIO, threadDelay)
 
 import           Brick
 import           Brick.BChan
+import           Brick.Main
```
byorgey, line 21: This new import seems redundant, could you remove it?

byorgey:
Huh, it doesn't seem to work for me. Tried with both urxvt and gnome-terminal.  But oh well, if it works for both of you then I'll just chalk it up to something weird about my setup.  Let's go ahead and merge once the redundant import is removed.

src/Swarm/App.hs
TristanCacqueray: Could we add -Wall to the library ghc-options to show those error?

TristanCacqueray:
That's odd, this work on my gnome-terminal (wayland) and emacs-vte (xorg).

src/Swarm/App.hs
byorgey: Yes, that's probably a good idea.  We might even consider adding `-Wall -Werror` to the CI.  I really like trying to keep things `-Wall`-clean, I have found many bugs over the years by forcing myself to pay attention to warnings.

byorgey:
Thanks!

PR #65, opened 2021-09-25
Restore REPL on Ctr+C

mateoatr:
Currently, if I write something in the REPL and press Ctr+C, the REPL form gets updated and puts three periods (`...`), and if I type something I will just be adding characters to whatever string I had before pressing Ctr+C. I would expect Ctr+C to behave just like in most terminals, that is, delete anything I had written at that moment and restore the it to `> `. This PR does just that.

### DIFF
--- a/src/Swarm/TUI/Controller.hs
+++ b/src/Swarm/TUI/Controller.hs
@@ -272,7 +272,7 @@ stripCmd pty                    = pty
 handleREPLEvent :: AppState -> BrickEvent Name AppEvent -> EventM Name (Next AppState)
 handleREPLEvent s (VtyEvent (V.EvKey (V.KChar 'c') [V.MCtrl]))
   = continue $ s
-          & uiState . uiReplForm %~ updateFormState ""
+      & uiState . uiReplForm %~ updateFormState ""
 handleREPLEvent s (VtyEvent (V.EvKey V.KEnter []))
   = case processTerm' topCtx topCapCtx entry of
       Right t@(ProcessedTerm _ (Module ty _) _ _) ->
--- a/src/Swarm/TUI/Controller.hs
+++ b/src/Swarm/TUI/Controller.hs
@@ -272,7 +272,7 @@ stripCmd pty                    = pty
 handleREPLEvent :: AppState -> BrickEvent Name AppEvent -> EventM Name (Next AppState)
 handleREPLEvent s (VtyEvent (V.EvKey (V.KChar 'c') [V.MCtrl]))
   = continue $ s
-      & gameState . robotMap . ix "base" . machine .~ idleMachine
+          & uiState . uiReplForm %~ updateFormState ""
 handleREPLEvent s (VtyEvent (V.EvKey V.KEnter []))
   = case processTerm' topCtx topCapCtx entry of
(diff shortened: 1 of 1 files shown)


src/Swarm/TUI/Controller.hs
```diff
@@ -272,7 +272,7 @@ stripCmd pty                    = pty
 handleREPLEvent :: AppState -> BrickEvent Name AppEvent -> EventM Name (Next AppState)
 handleREPLEvent s (VtyEvent (V.EvKey (V.KChar 'c') [V.MCtrl]))
   = continue $ s
-      & gameState . robotMap . ix "base" . machine .~ idleMachine
+      & uiState . uiReplForm %~ updateFormState ""
```
TristanCacqueray, line 275: Not so familiar with that part of the code, @byorgey should confirm, but I think we need to keep the idleMachine assignment for the base, e.g.:

 ```suggestion
       & (uiState . uiReplForm %~ updateFormState "")
        . (gameState . robotMap . ix "base" . machine .~ idleMachine)
```

byorgey: Yes, the point of the `idleMachine` assignment is that if the base is in the middle of running a computation but you want it to stop, you can hit Ctrl-C and it will replace its currently running machine with a machine that does nothing.

I am not sure off the top of my head why three dots are displayed when you type something at the REPL and then hit Ctrl-C.  The three dots are displayed when the base is actively running a program.  My guess is that it is "actively running" the `idleMachine` for a tick or two and then realizes that there is nothing to do.  Ideally, no dots would be displayed at all when you Ctrl-C.

byorgey: I see, yes, it is because of the way `idleMachine` is defined: it has to actually execute a `Noop` instruction, but this is silly.  Instead we should just define `idleMachine = Out VUnit []`.  @mateoatr , if you want to change that definition as part of this PR that would be great, or we could do it separately.

byorgey: Hmm, I think we also want to explicitly set `gameState . replStatus .~ REPLDone`, so the `replStatus` doesn't get out of sync with what the base is actually doing.

byorgey:
Hey @mateoatr , is this something you intend to keep working on?  I don't mean to rush you, just trying to figure out if it's something you're still interested in or if someone else should finish it up.  We're grateful for the contribution either way!  If you do want to keep working on it but need some help figuring things out, just let us know, very happy to help.

mateoatr:
> Hey @mateoatr , is this something you intend to keep working on?

Yeah, definitely! Sorry for slowness, many things got in the way on past days, but I'll make the suggested changes and update the PR. Thanks for being so supportive in the repo, it's very cool to be able to collaborate and watch others contributions on such a cool game!

lsmor:
Now that #439 has been merge, I think this can be implemented easily by just calling `resetWithREPLForm (mkReplForm $ CmdPrompt "") s` As it is done when pressed Esc. 

I can take care of this change. @mateoatr is it ok for you? If you prefer to merge yourself you should look at `handleREPLEvent` matching on `EscapeKey`. The code por this should be similar to that.

byorgey:
@lsmor , I think you should go ahead and take care of it, thanks!

PR #67, opened 2021-09-25
Reset the needsRedraw only once per frame

TristanCacqueray:
When multiple ticks happen per frame, the needsRedraw should only be reset once per frame to avoid skipping update if the last tick does not need update.


byorgey:
Oh, very good catch!

Issue #16, opened 2021-09-20
Quasiquoters for Swarm types and terms


byorgey:
We often have cause to write down Swarm types and terms in the Haskell source, but right now this is very tedious and hard to read, due to the need to encode them as ASTs.  It would be really nice to create some quasiquoters so that we could write them in Swarm syntax but still have them checked at compile time.

PR #68, opened 2021-09-25
Add ghc-options and ci flags to prevent compilation warnings

TristanCacqueray:


TristanCacqueray:
I started to fix non-exhaustive pattern but i don't know how to handle the one from Border module `let [rl,rc,rr] = rendered`

byorgey (approved):
Fantastic, thanks so much!  You even found some additional warnings. =D

As a general design principle I am trying as much as possible to make sure that swarm will never, ever crash, even if it is buggy, because if it crashes a user might lose all their progress and be (justifiably) upset.  You found two more places that could cause a crash, and now will crash explicitly with `error`.

* It is very plausible that we might change something about the language which causes the `seedProgram` to fail.  However, once we finish #16 I hope that the `seedProgram` can be processed at *compile* time instead of runtime, which would make this a non-issue.
* `find isTree` can't possibly return `Nothing`, since `enumerate (int' >< int')` is an infinite list.  At worst, if we accidentally change the world generation function so there are no trees, it will just run forever.

So I don't think we need to change anything at the moment, just wanted to write this down while I was thinking about it.

byorgey:
> I started to fix non-exhaustive pattern but i don't know how to handle the one from Border module `let [rl,rc,rr] = rendered`

Oh, good point.  I think I was trying to be too clever there. We should just change it to something like
``` haskell
let renderLabel = render . maybe emptyWidget (vLimit 1)
rl <- renderLabel l
rc <- renderLabel c
rr <- renderLabel r
```

TristanCacqueray:
Yeah I agree `error` shall be avoided, but in this change this is just to handle missing pattern match that would otherwise have failed with another exception.

Latest commit enables the `-Wincomplete-uni-patterns`.

By the way, would you prefer follow-up commits to address comments and then use the `Squash and merge` button?

byorgey:
> Yeah I agree `error` shall be avoided, but in this change this is just to handle missing pattern match that would otherwise have failed with another exception.

Right.

> By the way, would you prefer follow-up commits to address comments and then use the `Squash and merge` button?

I'm not super familiar with the pros and cons of different approaches to merging.  But I think that does sound better.  It means that new contributors do not have to know anything about force-pushing, rebasing, etc., and we still get to keep a relatively clean commit history.  I assume by your question you think the follow-up commits + squash-and-merge would be a superior approach?

TristanCacqueray:
I'm not super familiar either with the GitHub merging options :-) It seems like follow-up commits might be easier to review, and perhaps this is more common. On the other hand, I don't mind the merge commits in the history since they provide additional information, but I think it is important to keep a good commit history.

I think a good solution is to use a bot like mergify to automate the merge process, and we could use the same configuration as dhall-lang (see: https://github.com/dhall-lang/dhall-haskell/blob/master/.mergify.yml ) which uses a squash strategy.

PR #72, opened 2021-09-25
Persist the inventory of the last robot viewed

### LINKED ISSUE #61: Persist the inventory of the last robot viewed
Right now, we conflate two different things: the `ViewCenterRule` determines both how to decide on which world coordinates to render, and also which robot's inventory we are currently viewing.  This means that as soon as the user scrolls the world, the inventory disappears.  This seems kind of silly.  We should just make a new field to store the currently focused robot (i.e. the robot whose inventory we should be viewing); the `view` command will set both the `ViewCenterRule` *and* the currently focused robot.

lsmor:
Fixes #61 

Changes are fairly straightfoward. Just adds an extra field to `VCLocation` constructor with the name of the last robot viewed. The name of the robot is passed on `modifyViewCenter` function.

### DIFF
--- a/src/Swarm/Game/State.hs
+++ b/src/Swarm/Game/State.hs
@@ -116,7 +116,7 @@ data GameState = GameState
   , _focusedRobotName :: Text
   }
 
-let exclude = ['_viewCenter] in
+let exclude = ['_viewCenter, '_focusedRobotName, '_viewCenterRule] in
   makeLensesWith
     (lensRules
        & generateSignatures .~ False
@@ -150,7 +150,26 @@ recipesIn :: Lens' GameState (IntMap [Recipe Entity])
 world :: Lens' GameState (W.World Int Entity)
 
 -- | The current rule for determining the center of the world view.
+--   It updates also, viewCenter and focusedRobotName to keep
+--   everything synchronize.
 viewCenterRule :: Lens' GameState ViewCenterRule
+viewCenterRule = lens getter setter
+  where
+    getter :: GameState -> ViewCenterRule
+    getter = _viewCenterRule
+
+    -- The setter takes care of updating viewCenter and focusedRobotName
+    -- So non of this fields get out of sync.
+    setter :: GameState -> ViewCenterRule -> GameState
+    setter g rule =
+      case rule of
+        VCLocation v2 -> g { _viewCenterRule = rule, _viewCenter = v2}
+        VCRobot txt ->
+          let robotcenter = g ^? robotMap . ix txt <&> view robotLocation -- retrive the loc of the robot if it exist, Nothing otherwise.  sometimes, lenses are amazing... 
+          in case robotcenter of
+               Nothing -> g
+               Just v2 -> g { _viewCenterRule = rule, _viewCenter = v2, _focusedRobotName = txt}
+
 
 -- | The current center of the world view. Note that this cannot be
 --   modified directly, since it is calculated automatically from the
@@ -168,8 +187,10 @@ replStatus :: Lens' GameState REPLStatus
 -- | A queue of global messages.
 messageQueue :: Lens' GameState [Text]
 
--- | The current robot in focus
-focusedRobotName :: Lens' GameState Text
+-- | The current robot in focus. It is only a Getter because
+--   this value should be updated only when viewCenterRule is. 
+focusedRobotName :: Getter GameState Text
+focusedRobotName = to _focusedRobotName
 
 -- | Given a current mapping from robot names to robots, apply a
 --   'ViewCenterRule' to derive the location it refers to.  The result
@@ -201,7 +222,6 @@ modifyViewCenter update g = g
   & case g ^. viewCenterRule of
       VCLocation l -> viewCenterRule .~ VCLocation (update l)
       VCRobot _    -> viewCenterRule .~ VCLocation (update (g ^. viewCenter))
-  & recalcViewCenter
 
 -- | Given a width and height, compute the region, centered on the
 --   'viewCenter', that should currently be in view.
--- a/src/Swarm/Game/State.hs
+++ b/src/Swarm/Game/State.hs
@@ -70,8 +70,8 @@ import           Swarm.Util
 -- | The 'ViewCenterRule' specifies how to determine the center of the
 --   world viewport.
 data ViewCenterRule
-  = VCLocation (V2 Int64) Text  -- ^ The view should be centered on an absolute position. Keeping the last robot viewed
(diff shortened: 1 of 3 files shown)


src/Swarm/Game/State.hs
```diff
@@ -212,7 +212,10 @@ viewingRegion g (w,h) = (W.Coords (rmin,cmin), W.Coords (rmax,cmax))
 --   'viewCenterRule', if any.
 focusedRobot :: GameState -> Maybe Robot
```
TristanCacqueray, line 213: I guess this could be changed to `GameState -> Robot`

TristanCacqueray (approved):
Thanks, that seems to work as expected. I'm not familiar with that part of the code, could you please have a look @byorgey ?

src/Swarm/Game/State.hs
TristanCacqueray: Nevermind, the last robot may not be there anymore.

byorgey:
Thanks!  However, I think this is still confusing two different things: which robot's inventory we should be viewing, and where the world should be centered.  Conceptually, these are two different things.  I think the code would be much simpler if we get rid of the new `Text` field in `VCLocation` again, and then do something like the following:
* Add a new field `focusedRobotName` to the `GameState`.
* The `focusedRobot` function can now simply try looking up this robot name in the `robotMap`.
* Add a line of code to `Swarm.Game.Step` so that the `View` command sets the `focusedRobotName` in addition to the `ViewCenterRule`: https://github.com/byorgey/swarm/blob/6368a9811ae8417d6035548349caf701a2cbd1ac/src/Swarm/Game/Step.hs#L647

lsmor:
@byorgey I undo the changes and modify the `GameState` as you propose (which makes much more sense). However, I'm noticing that some problems may arise from this.

The are two fields chaching information from `viewCenterRule`: `viewCenter` and `focusedRobotName`, so when `viewCenterRule` is updated both fields should be updated too manually. This can lead to inconsistent states in which a programmer updates the `viewCenterRule` but forgets to update any of the other two. The perfect example is commit  [eaf62f2](https://github.com/byorgey/swarm/pull/72/commits/eaf62f24ebe5262dc82a03e6f089d2fa812d455a). 

A posible solution would be to refactor lenses for this two field, making `viewCenter` and `focusedRobotName` read only, and defining a manual setter for `viewCenterRule` so It updates the fields acoordingly. I'm not that familiar with `lenses` to know if this is a nasty hack or a common pattern.

src/Swarm/Game/Step.hs
```diff
@@ -645,6 +645,8 @@ execConst c vs k = do
           (`isJustOr` cmdExn View [ "There is no robot named ", s, " to view." ])
 
         lift . lift $ viewCenterRule .= VCRobot s
+        lift . lift $ focusedRobotName .= s
```
TristanCacqueray, line 648: Another solution would be to use a helper function:

```haskell
setFocusedRobot :: Text -> GameState -> GameState
setFocusedRobot name = (focusedRobotName .~ name) . (viewCenterRule .~ VCRobot name)
```

to replace those two updates with `lift . lift $ modify (setFocusedRobot s)`

src/Swarm/TUI/Controller.hs
```diff
@@ -395,6 +395,7 @@ handleWorldEvent s (VtyEvent (V.EvKey k []))
 handleWorldEvent s (VtyEvent (V.EvKey (V.KChar 'c') [])) = do
   invalidateCacheEntry WorldCache
   continue $ s & gameState . viewCenterRule .~ VCRobot "base"
+               & gameState . focusedRobotName .~ "base"
```
TristanCacqueray, line 398: ... and those two updates would be `gameState %~ setFocusedRobot "base"`

TristanCacqueray (approved):
Thanks, this works as expected.

@byorgey just to make sure, as discussed in #68 , if you are ok with this PR, should we use the GitHub squash option to merge it?

src/Swarm/TUI/Controller.hs
lsmor: Yes, but I think there is a better solution, since `viewCenterRule` shouldn't be use as a setter without updating `viewCenter` and `focusedRobotName` too, so probably the lens foncusing in `viewCenterRule` should be a little bit more complex than that. Something in the line of 

```haskell
viewCenterRule :: Lens' GameState ViewCenterRule
viewCenterRule = lens getter setter
  where
    getter = _viewCenterRule 
    setter gameState rule = gameState {_viewCenterRule = rule, _viewCenter = newCenterFromRule, _focusedRobotName = newRobotNameFromRule}
```

With the definition above, it'd be impossible to call `viewCenterRule .~ someRule` and forget to update the center and the robot name. This would simplify functions `modifyViewCenter`, `execConst` and `handleWorldEvent`. Again, I don't know how much of a good idea this is...

byorgey:
> A posible solution would be to refactor lenses for this two field, making `viewCenter` and `focusedRobotName` get only, and defining a manual setter for `viewCenterRule` so It updates the fields acoordingly. I'm not that familiar with `lenses` to know if this is a nasty hack or a common pattern.

This is a perfect solution.  I don't think it is a nasty hack at all.  You can see it as an instance of https://byorgey.wordpress.com/2021/09/17/automatically-updated-cached-views-with-lens/ , though it is actually a bit more interesting: `focusedRobotName` does not only cache information computed from the `viewCenterRule`, but continues to remember that information even after the `viewCenterRule` has changed to something else (until it changes to `VCRobot` again).  But I don't think this causes any problems.

src/Swarm/TUI/Controller.hs
byorgey: Right, and then only declare `Getter`s for `viewCenter` and `focusedRobotName`.  Then it will be impossible for them to ever get out of sync.

lsmor:
@TristanCacqueray @byorgey Lenses have been refactored as discussed in this PR. Code have been simplified a little bit and now there is no risk to get out of sync.

TristanCacqueray:
Thanks, that seems to work as expected. @byorgey would you like to try adding the `merge me` label?

byorgey:
Looks great, thanks!

byorgey:
@lsmor , congrats on having your first PR accepted to Swarm!  We're really glad to welcome you as part of the community.  In a minute we will send you an invite to the repo (see https://github.com/byorgey/swarm/blob/main/CONTRIBUTING.md#i-have-push-access-to-the-swarm-repository-now-what ).  If you use IRC, feel free to also join the `#swarm` channel on Libera.Chat.

PR #74, opened 2021-09-25
Add panel shortcuts

### LINKED ISSUE #45: Add shortcuts to jump directly to specific panels
I can already tell I am going to be annoyed having to push Tab an unknown number of times when I want to type something at the REPL.  It would be nice to offer some shortcuts for jumping directly to a specific panel. For example, Ctrl-I for info panel, Ctrl-R for REPL, and Ctrl-W for world.

TristanCacqueray:
Fixes #45

### DIFF
--- a/src/Swarm/TUI/Controller.hs
+++ b/src/Swarm/TUI/Controller.hs
@@ -14,6 +14,7 @@
 {-# LANGUAGE LambdaCase         #-}
 {-# LANGUAGE NumericUnderscores #-}
 {-# LANGUAGE OverloadedStrings  #-}
+{-# LANGUAGE PatternSynonyms    #-}
 {-# LANGUAGE TemplateHaskell    #-}
 {-# LANGUAGE TypeApplications   #-}
 
@@ -83,6 +84,11 @@ import           Swarm.TUI.List
 import           Swarm.TUI.Model
 import           Swarm.Util
 
+-- | Pattern synonyms to simplify brick event handler
+pattern ControlKey, MetaKey :: Char -> BrickEvent n e
+pattern ControlKey c = VtyEvent (V.EvKey (V.KChar c) [V.MCtrl])
+pattern MetaKey c    = VtyEvent (V.EvKey (V.KChar c) [V.MMeta])
+
 -- | The top-level event handler for the TUI.
 handleEvent :: AppState -> BrickEvent Name AppEvent -> EventM Name (Next AppState)
 handleEvent s (AppEvent Frame)
@@ -96,13 +102,23 @@ handleEvent s (VtyEvent (V.EvKey (V.KChar '\t') [])) = continue $ s & uiState . 
 handleEvent s (VtyEvent (V.EvKey V.KBackTab []))     = continue $ s & uiState . uiFocusRing %~ focusPrev
 handleEvent s (VtyEvent (V.EvKey V.KEsc []))
   | isJust (s ^. uiState . uiError) = continue $ s & uiState . uiError .~ Nothing
-handleEvent s (VtyEvent (V.EvKey (V.KChar 'q') [V.MCtrl])) = shutdown s
-handleEvent s ev =
-  case focusGetCurrent (s ^. uiState . uiFocusRing) of
-    Just REPLPanel  -> handleREPLEvent s ev
-    Just WorldPanel -> handleWorldEvent s ev
-    Just InfoPanel  -> handleInfoPanelEvent s ev
-    _               -> continueWithoutRedraw s
+handleEvent s ev = do
+  -- intercept special keys that works on all panels
+  case ev of
+    ControlKey 'q'      -> shutdown s
+    MetaKey 'w'         -> setFocus s WorldPanel
+    MetaKey 'e'         -> setFocus s InfoPanel
+    MetaKey 'r'         -> setFocus s REPLPanel
+    _anyOtherEvent ->
+      -- and dispatch the other to the focused panel handler
+      case focusGetCurrent (s ^. uiState . uiFocusRing) of
+        Just REPLPanel  -> handleREPLEvent s ev
+        Just WorldPanel -> handleWorldEvent s ev
+        Just InfoPanel  -> handleInfoPanelEvent s ev
+        _               -> continueWithoutRedraw s
+
+setFocus :: AppState -> Name -> EventM Name (Next AppState)
+setFocus s name = continue $ s & uiState . uiFocusRing %~ focusSetCurrent name
 
 -- | Shut down the application.  Currently all it does is write out
 --   the updated REPL history to a @.swarm_history@ file.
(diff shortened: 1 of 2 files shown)


TristanCacqueray:
Note that I could not use `ctrl-i` for the info panel as it seems to be converted to a `\t` at a lower level.
Perhaps that can be fixed using a custom vti config?

src/Swarm/TUI/Controller.hs
```diff
@@ -83,6 +84,8 @@ import           Swarm.TUI.List
 import           Swarm.TUI.Model
 import           Swarm.Util
 
+pattern ControlEvent c = VtyEvent (V.EvKey (V.KChar c) [V.MCtrl])
```
byorgey, line 87: Nice idea!  We should make these for normal key events too (though in a separate commit).

src/Swarm/TUI/Controller.hs
```diff
@@ -96,14 +99,20 @@ handleEvent s (VtyEvent (V.EvKey (V.KChar '\t') [])) = continue $ s & uiState .
 handleEvent s (VtyEvent (V.EvKey V.KBackTab []))     = continue $ s & uiState . uiFocusRing %~ focusPrev
 handleEvent s (VtyEvent (V.EvKey V.KEsc []))
   | isJust (s ^. uiState . uiError) = continue $ s & uiState . uiError .~ Nothing
-handleEvent s (VtyEvent (V.EvKey (V.KChar 'q') [V.MCtrl])) = shutdown s
+handleEvent s (ControlEvent 'q') = shutdown s
+handleEvent s (ControlEvent 'g') = setFocus s InfoPanel
+handleEvent s (ControlEvent 'r') = setFocus s REPLPanel
+handleEvent s (ControlEvent 'w') = setFocus s WorldPanel
```
byorgey, line 105: I am starting to wonder whether it is a mistake to bind Ctrl-Q to quit and Ctrl-W to something else.  You just know someone is going to accidentally quit when they meant to switch to the world panel.  At the very least we should add a confirmation dialog when the user quits (which is probably a good idea anyway).

I don't know, maybe rather than trying to come up with mnemonic keys for the panels, we should just bind them to Ctrl-J,K,L or something like that?

TristanCacqueray: Good point, I first tried `ctrl-m` for map, but this is not working for the same reason as `ctrl-i`.

The Info panel could also be mapped to `ctrl-e`, as `e` is often used for inventory in such game.

Another solution would be to use F1/F2/... keys, which might be easier to use without ctrl modifier.

byorgey: Ah, yeah, I think I like the idea of using Fn keys.

src/Swarm/TUI/Controller.hs
TristanCacqueray: Good, I wasn't sure at first, but since pattern synonyms are already used elsewhere I think it's worth enabling the extension here too.

src/Swarm/TUI/Controller.hs
TristanCacqueray: Alright I'll update the match for F keys.

amalloy: It's not so easy to type function keys on all keyboards, and even when it is leaving the home row kinda sucks. What about just assigning to, say, Ctrl-1, Ctrl-2 etc? Or Alt-anything?

byorgey: That's true.  Well, Ctrl-1 has the same problem as Ctrl-W: it is right next to Ctrl-Q.  We could also try Alt but that is a bit tricky too as far as I understand it, because what ends up being called "Alt" varies a lot depending on your keyboard and setup etc.

byorgey: Of course the real solution is to let the user assign their own keyboard shortcuts, but that sounds like too much work right now. =)

TristanCacqueray: The latest commit use the Meta modifiers with `e` for the info, `m` for the world and `r` for the repl. I think using Meta is common for such shortcuts, and it seems to match the left alt key.

byorgey (changes requested):
This is definitely converging to something nice, I think.  The shortcuts work great for me (both Alt keys work in my case).

One other thing to consider (but quite possibly as a separate issue) is how to advertise these shortcuts to players beyond just e.g. sticking them in the tutorial.  I had been thinking to add them somewhere on the border of each panel, but I tried that and it looked kind of ugly.  Maybe we need to add a global "help" key (e.g. Ctrl-H) that opens a pop-up dialog window listing key shortcuts and anything else helpful we can think to fit in there.

src/Swarm/TUI/Controller.hs
```diff
@@ -83,6 +84,12 @@ import           Swarm.TUI.List
 import           Swarm.TUI.Model
 import           Swarm.Util
 
+-- | Pattern synonyms to simplify brick event handler
+pattern ControlKey c = VtyEvent (V.EvKey (V.KChar c) [V.MCtrl])
```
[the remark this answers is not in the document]
TristanCacqueray: That's good :)

src/Swarm/TUI/Controller.hs
```diff
@@ -96,13 +103,23 @@ handleEvent s (VtyEvent (V.EvKey (V.KChar '\t') [])) = continue $ s & uiState .
...
-    Just REPLPanel  -> handleREPLEvent s ev
-    Just WorldPanel -> handleWorldEvent s ev
-    Just InfoPanel  -> handleInfoPanelEvent s ev
-    _               -> continueWithoutRedraw s
+handleEvent s ev = do
+  -- intercept special keys that works on all panels
+  case ev of
+    ControlKey 'q'      -> shutdown s
+    MetaKey 'e'          -> setFocus s InfoPanel
+    MetaKey 'm'          -> setFocus s WorldPanel
```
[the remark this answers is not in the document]
TristanCacqueray: I picked `m` as in map, which is usually the key used to show a map, while `w` often mean *move forward*.

TristanCacqueray:
Thank you for the feedback. For the help we could also use `F1` (as it is the default for htop or mutt).

I find `m` more natural for the WorlPanel, but having the toggle next to each other is also good, thus I switched back to `w` in the last commit.

src/Swarm/TUI/Controller.hs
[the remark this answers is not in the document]
byorgey: Ah, `m` for map, I see. I am so used to thinking of it as the "world" that I didn't even think of the word "map". =)

byorgey (approved):
Looks good to me!  Thanks for putting up with all the bikeshedding =)

TristanCacqueray:
You are welcome, I think bikeshedding is necessary for such change.

PR #78, opened 2021-09-25
add a quasiquoter for terms

### LINKED ISSUE #16: Quasiquoters for Swarm types and terms

byorgey:
~~Also modify the `seedProgram` so it will throw a Fatal exception in
case of an error instead of crashing~~ The `seedProgram` can now only fail at compile time! (cf discussion on #68 ).

Closes #16.

### DIFF
--- a/src/Swarm/Game/Step.hs
+++ b/src/Swarm/Game/Step.hs
@@ -15,6 +15,7 @@
 {-# LANGUAGE FlexibleContexts  #-}
 {-# LANGUAGE LambdaCase        #-}
 {-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE QuasiQuotes       #-}
 {-# LANGUAGE RankNTypes        #-}
 {-# LANGUAGE TypeApplications  #-}
 {-# LANGUAGE TypeOperators     #-}
@@ -26,7 +27,7 @@ import           Control.Lens              hiding (Const, from, parts)
 import           Control.Monad.Except
 import           Control.Monad.State
 import           Data.Bool                 (bool)
-import           Data.Either               (fromRight, rights)
+import           Data.Either               (rights)
 import           Data.Int                  (Int64)
 import           Data.List                 (find)
 import qualified Data.Map                  as M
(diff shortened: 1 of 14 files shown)


src/Swarm/Game/Step.hs
```diff
@@ -436,28 +438,22 @@ evalConst c vs k = do
...
-      ]
+-- | A system program for a "seed robot", to regrow a growable entity
+--   after it is harvested.
+seedProgram :: Text -> Either Text ProcessedTerm
+seedProgram thing = processParsedTerm [tmQ|
+  let repeat : int -> cmd () -> cmd () = \n.\c.
+    if (n == 0) {} {c ; repeat (n-1) c}
+  in {
+    r <- random 500;
+    repeat (r + 100) wait;
```
TristanCacqueray, line 449: not sure why, but inserting bad code here doesn't fail at compile time.

TristanCacqueray (approved):
That seems to work, and it's good to see how a quasiquote can access haskell variables.

byorgey:
> That seems to work, and it's good to see how a quasiquote can access haskell variables.

Yes, I had fun learning about it.  I'd never written any quasiquoters before.

src/Swarm/Game/Step.hs
byorgey: It fails at compile time if there is a *parse* error.  That is all the quasiquoter does.  Typechecking and elaborating the term still happens at runtime (see the call to `processParsedTerm`). I thought a bit about how to make a quasiquoter that did the entire processing pipeline but it made my head hurt. But now that you mention it I want to try a little harder before merging this.

byorgey: Ask and ye shall receive!  With the latest commit it now it runs the entire processing pipeline at compile time, so it will fail at compile time on any kind of error, whether a parse error or type error.

TristanCacqueray: That's wonderful, thanks, my test is now properly failing.

PR #80, opened 2021-09-26
add flake.nix

wizzup:
using flake template from https://input-output-hk.github.io/haskell.nix/tutorials/getting-started-flakes/

I am using this for myself for a couple days and it working fine.
BTW, I don't know any other template that simpler than this. This one may seem overkill.

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,6 +78,14 @@ The recommended way to install Swarm at the moment is as follows:
    tutorial](TUTORIAL.md) to help get you started.  Eventually there
    will be an in-game tutorial.
 
+## Nix flake (experimental)
+
+Nix flake user can run `swarm` binary using this command:
+
+    $ nix run github:byorgey/swarm
+
+Please see iohk's `haskell.nix` [guide](https://input-output-hk.github.io/haskell.nix/tutorials/getting-started-flakes/) on how to setting binary cache.
+
 Contributing
 ============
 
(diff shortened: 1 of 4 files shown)


TristanCacqueray:
Thank you, that seems to work as expected, in particular nix build uses the correct hsnoise version.

Though I'm not sure the maintainers are ready to support another build system, could we have an extra CI job to prevent regression?

byorgey:
Hi, thanks for the contribution!  On the one hand, as @TristanCacqueray says, I don't want to have another thing to support at this point.  I don't really understand how this works (why does it need `cardano-shell`? why are there multiple things called `ghc`, and with different versions than what we are testing on?), nor do I really want to put in the time to do so.  Having it live in the repo implicitly promises some level of official support that I don't want to be responsible for.

On the other hand I definitely see the value in providing this for people who want to use it. What if we put the instructions on how to set this up in the project wiki, and link to that from the README?

TristanCacqueray:
If I understand correctly, nix flakes need to be part of the project to be useful, but without CI (and associated cache) this will eventually bitrot. Moreover it seems like flake.lock are still experimental and the one that is proposed definitely contains unrelated dependencies.

byorgey:
OK, I am going to close this, but feel free to post this info about how to set it up somewhere (such as on the [swarm wiki](https://github.com/byorgey/swarm/wiki)---which I have now enabled and set to be editable by anyone---or elsewhere), and we will be happy to link to it from the README.

PR #83, opened 2021-09-26
Add Help screen

TristanCacqueray:


TristanCacqueray:
This is just a first draft.

byorgey:
I like the approach so far.

TristanCacqueray:
Well it's not the prettiest thing, but I think that's functional enough for review.

Please let me know if you have suggestions!

src/Swarm/TUI/Controller.hs
```diff
@@ -120,6 +124,14 @@ handleEvent s ev = do
 setFocus :: AppState -> Name -> EventM Name (Next AppState)
 setFocus s name = continue $ s & uiState . uiFocusRing %~ focusSetCurrent name
 
+toggleModal :: AppState -> Modal -> EventM Name (Next AppState)
+toggleModal s modal =
+  continue $ s
+      & gameState . paused %~ not
```
byorgey, line 130: First off, I totally agree the game should automatically pause if you open the help.  However, I'm not sure this works correctly. What if you pause the game and then hit F1?  Wouldn't that *unpause* the game and show the help modal?  We could fix that by setting the `paused` state based on the state of the modal.  But then there is another problem; what if you have manually paused the game and then open the help?  I would argue that the game should remain paused when you close the help.  So maybe `paused` can't be a `Bool` any more, but should be something like `data RunStatus = Running | ManualPause | AutoPause` so we can tell the difference.

src/Swarm/TUI/View.hs
```diff
@@ -142,15 +144,71 @@ drawTPS s = hBox (tpsInfo : rateInfo)
...
+           , hCenter $ mkTable baseCommands
+           , padTop (Pad 1) $ hCenter $ txt "Robot commands"
+           , hCenter $ mkTable robotCommands
+           ]
+    baseCommands =
+      [ ("build <name> <commands>", "Create a robot")
+      , ("make <name>", "Craft an item")
+      ]
+    robotCommands =
+      [ ("move", "Move one step in the current directory")
```
byorgey, line 199: s/directory/direction/

src/Swarm/TUI/View.hs
```diff
@@ -142,15 +144,71 @@ drawTPS s = hBox (tpsInfo : rateInfo)
...
+      , ("Meta-r", "focus on the REPL")
+      ]
+    helpCommands =
+      vBox [ hCenter $ txt "Base commands"
+           , hCenter $ mkTable baseCommands
+           , padTop (Pad 1) $ hCenter $ txt "Robot commands"
+           , hCenter $ mkTable robotCommands
+           ]
+    baseCommands =
+      [ ("build <name> <commands>", "Create a robot")
```
byorgey, line 195: I'm not sure we want to make a distinction in the help between "base commands" and "robot commands".  There is actually no distinction as far as the game is concerned, other than the fact that they start out with different sets of devices.  But, for example, it is already pretty easy to craft another workbench which you can install in a robot so that it can use the `make` command.  And the base itself can execute the `grab` command (actually if you try this, you realize that it is sitting on top of a tree =).  In the future, we can imagine having robots `build`ing more robots.  I also think it would be fun if eventually you can do something like install treads on your base so it can `move`.

byorgey (changes requested):

src/Swarm/TUI/Controller.hs
TristanCacqueray: Good point, that was an oversight. We also needs to need to skip the event panel dispatch when a modal is displayed, e.g. to avoid accidental unpause while looking the help.

The last commit implements your RunStatus suggestion.

src/Swarm/TUI/View.hs
TristanCacqueray: Alright, that also simplifies the layout :)

TristanCacqueray:
Also note that I used follow-up commits here so that we can try the squash strategy with mergify once this is ready.

src/Swarm/TUI/Controller.hs
```diff
@@ -413,7 +421,7 @@ handleWorldEvent s (VtyEvent (V.EvKey (V.KChar 'c') [])) = do
 handleWorldEvent s (VtyEvent (V.EvKey (V.KChar 'p') [])) = do
   curTime <- liftIO $ getTime Monotonic
   continue $ s
-      & gameState . paused %~ not
+      & gameState . runStatus .~ (if s ^. gameState . runStatus == Running then ManualPause else Running)
```
byorgey, line 424: I think this could be more naturally written as `gameState . runStatus %~ (\stat -> if stat == Running then ManualPause else Running)`.  Well, at least it seems more natural to me. =)

src/Swarm/TUI/View.hs
byorgey: Great. =)  And maybe just call it "commands" instead of "base commands".

byorgey:
> Also note that I used follow-up commits here so that we can try the squash strategy with mergify once this is ready.

Great. And yes, I do like how that allows you to view all the changes together OR just the delta from the initial PR.

byorgey:
Oh, and one more thing: we should add F1 to the keys menu in between the panels, so users know how to get help.

src/Swarm/TUI/Controller.hs
TristanCacqueray: Oh indeed, thanks!

src/Swarm/TUI/View.hs
TristanCacqueray: Done.

TristanCacqueray:
In the last commit, I actually replaced the `[^q] quit` with `[F1] help` to keep the noise low. But I don't mind adding the quit shortcut back to the menu if you prefer.

byorgey (approved):

src/Swarm/TUI/Controller.hs
```diff
@@ -120,6 +125,21 @@ handleEvent s ev = do
...
 setFocus s name = continue $ s & uiState . uiFocusRing %~ focusSetCurrent name
 
+toggleModal :: AppState -> Modal -> EventM Name (Next AppState)
+toggleModal s modal =
+  continue $ s & case s ^. uiState . uiModal of
+    Nothing -> (uiState . uiModal ?~ modal)   . ensurePause
+    Just _  -> (uiState . uiModal .~ Nothing) . maybeUnpause
+  where
+    -- Set the game to AutoPause if needed
+    ensurePause
```
byorgey, line 135: Hmm, actually, this doesn't quite work yet.  Notice in `handleWorldEvent` when we process a manual pause, we also reset the `lastFrameTime`, which is necessary since otherwise, once the game is unpaused, it thinks the last frame took a really long time and that it needs to catch up by doing a bunch of ticks all at once.  When we `maybeUnpause` here we don't do that, so after hitting F1, waiting for a little while, then hitting F1 again you can see the game jump ahead to where it would have been if it hadn't been paused the whole time.

I wonder if there is a good way to encapsulate this functionality so we don't have to remember to do this in every single place in the code that might want to unpause.

byorgey (changes requested):

byorgey:
> In the last commit, I actually replaced the `[^q] quit` with `[F1] help` to keep the noise low.

No, I think that makes sense.

src/Swarm/TUI/Controller.hs
TristanCacqueray: Oops, good catch. I don't know how to properly encode this functionality... How about we use a `Lens' AppState -> RunStatus` which reset the lastFrameTime to 0, and in runFrameUI we could consider such value as curTime (and skip the catch up phase)?

byorgey: Hmm, but I'm not sure we want to reset the `lastFrameTime` every time we look up the `RunStatus`.  Maybe for now let's just manually reset the `lastFrameTime` in `maybeUnpause` and defer the question to later of whether we can refactor it in a better way.

byorgey (approved):

byorgey (changes requested):
Whoops, I hit approve again, then decided to try it out one last time before merging, but it still doesn't work.  I am not sure why.

Manually pausing works fine; but after auto-pausing with the help dialog the game tries to catch up.  I have been staring at the code, checking and re-checking that I have the right branch checked out and built, etc., and I can't figure it out.

TristanCacqueray:
I just rebased the change to solve a merge conflict, and showing/hidding the help while a robot is working seems to work correctly. What's the issue?

byorgey:
> I just rebased the change to solve a merge conflict, and showing/hidding the help while a robot is working seems to work correctly. What's the issue?

Ah, yes, I just updated to the rebased version and now everything works properly.  No idea what was going on before.

byorgey (approved):

Issue #79, opened 2021-09-26
Missing semicolon between def yield confusing error


TristanCacqueray:
**Describe the bug**
This code works:

```
def m1 { move } end
```

But this code:

```
def m1 { move } end
def m2 { m1; m1 } end
```

... fails with `Definitions may only be at the top level: def m1 = move end`

**To Reproduce**
Write the code in a file and load it with `run()`. 

**Expected behavior**
The error should indicate the missing semicolon.
Also when running a file, perhaps the last semicolon shall be required, so that the first program also fails.

byorgey:
Hmm, yes, I think what is happening is that it successfully parses as `(def ... end) (def ... end)`, i.e. the application of one definition to another.  Of course, this makes no sense.   As a first pass, we should definitely put in a special case in the parser to check for this and emit an error suggesting a missing semicolon.   But perhaps there are more general things we could do as well.

TristanCacqueray:
Perhaps the language could support haskell style layout so that it automatically insert a semi-colon on new line if the following line is not indented?

By the way, is there a plan to add tasty validation for such issue?

byorgey:
Parsing for Haskell-style layout is very tricky.  I would not be at all confident in our ability to either get it right or explain it adequately to users.

Hmm, *however*, perhaps we could add one special case: a semicolon can be automatically inserted following an `end` token?

We definitely need some test suites, see #12 .

byorgey:
Closed by #86 .

PR #86, opened 2021-09-26
Add parser test and fix missing semicolon between defs

TristanCacqueray:


TristanCacqueray:
Here is an initial attempt to fix #79 , though the REPL now fails with `Bad machine state in stepRobot: FApp of non-function`. I guess this is not correct because it would require another special case in `stepCEK`. Is there a better solution?

byorgey:
Right, I think the better solution would be to fix it in the *parser*.  Perhaps wrapping an extra postprocessing step around `parseExpr`:
``` haskell
parseExpr = fixDefMissingSemis $ makeExprParser ...
```
And the `fixDefMissingSemis` looks for `TApp (TDef ...) (TDef ...)` and turns it into a `TBind` instead of a `TDef`.

Note that will only work for two definitions; it won't work for three or more consecutive definitions with no semicolons.  That gets a little awkward because application associates to the left but bind associates to the right.  We'd have to make a function that looks for nested `TApp (TApp (TApp ... TDef) TDef) TDef` and returns a list of them, then does a `foldr` with `TBind`.

TristanCacqueray:
I see, thanks for the suggestion. I guess another solution would be to add a tokenizer stage where we could ensure that `semicolon` happens after an `end`.

byorgey:
Yeah, but making a separate tokenizer stage would be a lot of extra complexity just to allow users to omit a few semicolons.

byorgey:
Just to entertain a completely different possibility: we could make curly braces required.  That is, a semicolon by itself doesn't create a new command out of two commands, but instead the only way to make a "compound command" is to write `{ cmd1 ; cmd2 ; .... ; cmdn }`.  This means:
* Definitions would no longer need an `end` token, because the syntax is `def name : type = command`, and you can tell for sure where the command ends because it is either a single expression or a compound command enclosed by curly braces.  The problem before is that you couldn't tell the difference between a semicolon which continued the command which was the body of the definition, and a semicolon which separated the command from the next command.
* However, it would mean that if you want to write multiple definitions in one file, you have to surround the whole thing by curly braces.
* It also doesn't really help the issue of forgetting semicolons in between definitions.

I don't know, I guess I'm not convinced that would really be any better than the status quo.
*

TristanCacqueray:
I kind of like the current `def ... end` structure and I think it matches well the robot theme of the game.

My only issue was the requirement for trailing semicolon between defs, and thinking about it more, we could also support compound command separated by line return, kind of the opposite of the required curly brace you suggest :). Like that this code would also be valid:

```
def m2 =
    move
    move
end
```

Though, if I understand correctly, this would be complicated to support with the current grammar parser...

I am also not convinced that this, or the required curly brace, would be any better than the status quo.

byorgey:
Well, it's not so much that it would be complicated to support, but it would mean that newlines can't be used anywhere else --- like you couldn't even have the two branches of an `if` on separate lines.  I think that's pretty much a non-starter.  Anyway, it's definitely been helpful contemplating some alternatives; in the end, I think what we have is a pretty good compromise, even if it requires a hack or two. =)

PR #87, opened 2021-09-27
Add entity growth rates and parameterize the `seedProgram`

### LINKED ISSUE #23: Parametrized growth rates
Entities which have the `growable` property, when harvested via the `grab` command, spawn a system robot representing a "seed", which takes a while to grow through two phases and then replaces itself with another copy of the originally harvested entity.  However, the length of time that it takes to grow is hard-coded into the program for the seed robot.  This should be configurable, by specifying some data in `entities.yaml`.

There are a lot more details yet to work out here; just putting this issue here as a placeholder for now.

byorgey:
* Add antiquoted int variables to the term AST, for including them in quasiquotes
* Add a `Maybe GrowthRate` field to `Entity`, which optionally records the min and max time an entity should take for each regrowth stage
* Make up some growth times for existing growable entities (very open to arguments that these should be higher or lower etc.)
* Parameterize the `seedProgram` on the growth rate

Closes #23 .

### DIFF
--- a/data/entities.yaml
+++ b/data/entities.yaml
@@ -5,6 +5,7 @@
   description:
   - A tree.
   properties: [portable, growable]
+  growth: [500, 600]
 
 - name: branch
   display:
@@ -128,6 +129,7 @@
   description:
   - A flower.
   properties: [portable, growable]
+  growth: [30, 50]
 
 - name: lambda
   display:
@@ -140,6 +142,7 @@
   - '  def thrice : cmd () -> cmd () = \c. c;c;c end'
   - defines the function 'thrice' which repeats a command three times.
   properties: [portable, growable]
+  growth: [100, 200]
   capabilities: [lambda]
 
 - name: curry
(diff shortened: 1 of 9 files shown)


TristanCacqueray (approved):
The default growth times look good to me.

TristanCacqueray:
It seems like the mergify configuration worked as expected: the commits got squashed using the PR comment as commit message, and rebased without a merge commit.

byorgey:
Yep, looks good, though actually I think the commit message is the concatenation of the commit messages from the squashed commits?

TristanCacqueray:
Oh you are right, I though the default would be to use the PR title+body. To get that we need to set `commit_message: title+body` in the merge configuration.

byorgey:
> To get that we need to set `commit_message: title+body` in the merge configuration.

I think that probably makes more sense, since the titles of some of the individual commits may not be sensible to include.  We just have to make sure that the PR body contains a good description for the squashed commit.

PR #106, opened 2021-09-29
add `scanner` device, `scan` and `upload` commands

### LINKED ISSUE #96: `scanner` device and `scan` command
As discussed in #49 , we should add some sort of scanner device that robots can use to find out about items in the world, with a way for them to report their findings back to the base.

There are definitely some details to work out; here are some of my current thoughts:
* Since this is an extremely basic action which is one of the very first things you will need to do, `scanner` devices should be installed on robots by default (or, if we implement #35 , the base should start out with a supply of them).
* It would be very annoying if the `scan` command returns a value which you have to specifically transmit to the base (as in `info <- scan forward; upload "base" info) because that would allow you to scan only one thing at a time before returning to the base: robots can't use complex data structures like lists at the start of the game, so threre would be no way to store multiple scanned info values.  However, it would be nice if robots can just wander around scanning things before returning to the base later.
* Therefore, for the type of `scan` I propose `scan : dir -> cmd ()` which scans in a particular direction and *adds information about the scanned entity to the robot's inventory* (with a count of zero).  (It might be nice if `scan` also returned the name of the scanned thing, as in `scan : dir -> cmd string`, but then we have to worry about what it should return if there is nothing to scan.)
* There should also be a command `upload : string -> cmd ()` which uploads some scan results info another robot.  This would just transfer *knowledge of* all things in the robot's inventory to the target robot (without actually `give`ing any of them).  Newly acquired knowledge would show up as entries in the inventory with a count of 0.

As discussed in #49, we might also consider drawing entities using question marks until the base knows about them.

byorgey:
Closes #96.

- Delete the `rangefinder` entity.
- A `scanner` now provides the capability to execute `blocked`.
- Add a new direction, `down`, to be used for scanning the current
  cell (i.e. `scan down`).  Can also be used with `drill` in the future.
- `scan` and `upload` commands work as described in #96.

This PR does not implement showing things as question marks, that can
be decided and implemented separately.


byorgey:
I'd actually like to update the tutorial to match before merging this.

TristanCacqueray (approved):

PR #107, opened 2021-09-29
stack: update to latest lts-18.12 for ghc-8.10.7

juhp:


byorgey:
Last I checked, `lsp-haskell` did not yet support GHC 8.10.7. Has it been updated?

juhp:
Oh I see
Do you mean https://github.com/emacs-lsp/lsp-haskell ?

Anyway no rush to merge this - just though it would be better to target 8.10.7 if possible

stack.yaml
```diff
@@ -1,73 +1,10 @@
...
-# User packages to be built.
-# Various formats can be used as shown in the example below.
-#
-# packages:
-# - some-directory
-# - https://example.com/foo/bar/baz-0.0.2.tar.gz
-#   subdirs:
-#   - auto-update
-#   - wai
-packages:
```
byorgey, line 32: Is the `packages` field not required?  I guess `.` is the default?

byorgey:
Sorry, I actually just meant the Haskell Language Server in general. (I do use it via that emacs package but that's irrelevant.)  However, I just double-checked and GHC 8.10.7 is now supported.

stack.yaml
byorgey: Aha, I found where in the documentation it says that it is indeed optional.

byorgey (approved):

byorgey:
@juhp , congrats on having your first PR accepted to Swarm!  We're really glad to welcome you as part of the community.  In a minute we will send you an invite to the repo (see https://github.com/byorgey/swarm/blob/main/CONTRIBUTING.md#i-have-push-access-to-the-swarm-repository-now-what ).  If you use IRC, feel free to also join the `#swarm` channel on Libera.Chat.

PR #108, opened 2021-09-29
Add basic editor integration with LSP

### LINKED ISSUE #82: Add a command line argument to validate and/or load an initial program file into the base
When working on robot code written in a file, it is not super convenient to use the in-game repl.

The CLI could features an argument parser to handle `--load file` argument to perform the initial `run` command.

A `--check` argument could also be used to just validate the code without running the game.

TristanCacqueray:
This change improves the user experience when editing swarm code outside of the REPL with:

- CLI for a `format` command to quickly parse and type check a file.
- Basic LSP server to provide IDE feedback on save, e.g. visual squiggly for errors.
- Initial emacs mode.

Fixes #82 #110


TristanCacqueray:
I'm not so sure about the one-liner output though, should this use the regular megaparsec error bundle text output?

src/Swarm/Language/Parse.hs
```diff
@@ -345,3 +348,22 @@ fully p = sc *> p <* eof
...
 readTerm = runParser (fully parseTerm)
+
+-- | A lower-level readTerm which returns the megaparsec bundle error
+--   for precise error reporting.
+readTerm' :: Text -> Either ParserError Term
+readTerm' = parse (runReaderT (fully parseTerm) DisallowAntiquoting) ""
+
+-- | A utility for converting a ParserError into a one line message:
+--   <line-nr>: <error-msg>
+showShortError :: ParserError -> String
```
byorgey, line 359: What's the purpose of showing a short one-line error instead of the whole pretty-printed parse error bundle? (I'm not implying it's bad, just trying to understand.)

app/Main.hs
```diff
@@ -1,6 +1,111 @@
...
 
 main :: IO ()
-main = appMain
+main = do
+  args <- getArgs
+  case args of
+    []              -> appMain
+    ["--lsp"]       -> lspMain
+    ["--check", fp] -> check fp
+    _               -> putStrLn "usage: swarm [--check fp]"
```
byorgey, line 111: See, you already forgot to add `--lsp` to the usage message.  That something a command-line parsing library would handle for us. =)

app/Main.hs
```diff
@@ -1,6 +1,111 @@
...
+  case err of
+    Nothing -> exitSuccess
+    Just e  -> do
+      putStrLn $ fp <> ":" <> e
+      exitFailure
 
 main :: IO ()
-main = appMain
+main = do
+  args <- getArgs
```
byorgey, line 106: We're going to quickly wish we were using a proper command-line handling library, might as well bite the bullet now.  I'm personally familiar with and fond of `optparse-applicative` but don't have strong opinions.  But if that's something you'd rather not tackle now, I'm fine with merging this as-is and splitting out another issue to convert to a command-line handling library.

byorgey (changes requested):
I'd really rather put all this code somewhere besides `app/Main.hs`.  e.g. maybe the LSP-specific code can go in `Swarm.Language.LSP`; maybe the `check` and command-line arguments code can go in `Swarm.App` (or in a new module?).  In general my philosophy is to put absolutely everything in the library and just have `app/Main.hs` contain a single line of code; that way all the code is more easily reusable (in case we want multiple front ends, or someone else wants to build on it, etc).

byorgey:
I'm trying to test it out --- I have it built, and the `swarm-mode.el` successfully installed where emacs can find it.  So I opened a file with swarm code and enabled `swarm-mode` --- now what?  I can also manually start `lsp-mode` but it doesn't seem to start a connection.

src/Swarm/Language/Parse.hs
TristanCacqueray: I thought this would help with non-lsp editor integration, like when running `M-x compile`, emacs automatically understand the `filename:line:message` format. Though now that we can have LSP, I think we should just use that for IDE.

TristanCacqueray:
After evaluating the `swarm-mode.el` file, running `M-x lsp` in a swarm-mode buffer should be enough. Though you need the `swarm` command in your path, e.g. by running `ln -s $(pwd)dist-newstyle/build/x86_64-linux/ghc-8.10.4/swarm-0.1.0.0/x/swarm/noopt/build/swarm/swarm ~/.local/bin`

byorgey:
It works! Very cool!!

TristanCacqueray:
Thank you for the prompt feedback, I've switched it to draft until this uses optparse-applicative along with a `Swarm.Language.LSP` module.
Also it seems like the source line/column is not properly calculted.

app/Main.hs
TristanCacqueray: Sure I'll integrate the cli handling library.

app/Main.hs
TristanCacqueray: Actually that was on purpose, I don't think it should appear on the usage doc since this flag is made for IDE. If that's ok with you, i'll use the `hidden` mod from `optparse-applicative`.

byorgey: Ah, OK, makes sense. =)

TristanCacqueray: Actually I used subcommand and I'm not sure how to hide it.

src/Swarm/Language/Parse.hs
TristanCacqueray: Instead of `--check`, i rename the feature `format`, where it currently prints the file when it is valid.

app/Main.hs
byorgey: I don't think it's a big deal whether it's hidden or not.

TristanCacqueray:
Please let me know if the CLI and initial doc is correct. Note that the `format` command is more of a place holder, but I think it's better to use this name instead of `check`.

Also it seems like you'll have to restart your emacs to replace the new lsp client (that uses `swarm lsp` instead of `swarm --lsp`).

byorgey (approved):
I'll let you merge this if you think it's ready. Could you also make an issue or two to track whatever else we still need to improve (instructions for more editors?).

TristanCacqueray:
Thanks! Here are two follow-up issues ^ and #124 .

Let's give it a day and I'll add the mergify label.

byorgey:
This is fantastic, thanks for your work on this!

TristanCacqueray:
Thank you for the kind words, swarm is a great project to learn Haskell and its many libraries!

PR #125, opened 2021-09-30
TUI: Add j,k vim bindings for infopane control

ndzik:
Adds `j` to move "downwards" and `k` to move "upwards".

I figured I cannot be the only terminal enthusiast who loves to never leave the homerow. I realize this is opinionated but I could not live without it (':

PS: I did not use any formatter, so I aligned stuff the way ~~my~~ I thought would be okay for now.

BTW: I just love everything about this project :rocket:

### DIFF
--- a/src/Swarm/TUI/View.hs
+++ b/src/Swarm/TUI/View.hs
@@ -239,7 +239,7 @@ drawMenu isPaused viewingBase mode
       [ ("c", "recenter") | not viewingBase ]
 
     keyCmdsFor (Just InfoPanel)  =
-      [ ("↓↑/Pg{Up,Dn}/Home/End", "navigate")
+      [ ("↓↑/Pg{Up,Dn}/Home/End/jk", "navigate")
       , ("Enter", "make")
       ]
     keyCmdsFor _ = []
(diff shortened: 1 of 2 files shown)


src/Swarm/TUI/View.hs
```diff
@@ -239,7 +239,7 @@ drawMenu isPaused viewingBase mode
       [ ("c", "recenter") | not viewingBase ]
 
     keyCmdsFor (Just InfoPanel)  =
-      [ ("↓↑/Pg{Up,Dn}/Home/End", "navigate")
+      [ ("↓↑/Pg{Up,Dn}/Home/End/{j,k}", "navigate")
```
TristanCacqueray, line 242: Should we use the same help as the world panel, e.g. `↓↑ / jk` ?

TristanCacqueray (approved):
Thanks!

src/Swarm/TUI/View.hs
ndzik: I also could not decide. I have no problems with a version using less characters. Will update it!

byorgey (approved):

byorgey:
> BTW: I just love everything about this project :rocket:

Thanks, glad you like it! =)

byorgey:
@ndzik  , congrats on having your first PR accepted to Swarm!  We're really glad to welcome you as part of the [community](https://github.com/byorgey/swarm/blob/main/COMMUNITY.md).  In a minute we will send you an invite to the repo (see https://github.com/byorgey/swarm/blob/main/CONTRIBUTING.md#i-have-push-access-to-the-swarm-repository-now-what ).  If you use IRC, feel free to also join the `#swarm` channel on Libera.Chat.

Issue #17, opened 2021-09-20
Give `view` a limited broadcast range


byorgey:
_This issue used to be about adding an `antenna` entity to enable `view`, but see comments below for some reasons why I no longer think we should do that._

I'm thinking there can be an "Antenna" entity.

* An antenna provides a `broadcast` capability, with a corresponding `broadcast` command.
* Only robots (besides `base`) which have executed `broadcast` can be viewed with the `view` command.
* The recipe for an antenna could involve, say, 5 copper wires, + maybe something else?  Depends on how hard it should be to obtain early in the game.  It would probably be pretty helpful for debugging etc. so probably shouldn't be too hard to get.
* Perhaps later, an antenna can have a limited broadcast range, and there can be more difficult ways to craft more powerful antennas.  But probably for a first crack we can just make antennas have an unlimited range.

Anrock:
Would also be nice to have API for antenna to allow some semi-autonomous robots maybe.

byorgey:
Hi @Anrock , thanks for the idea!  However, I'm not sure I understand what you mean.  What kind of API do you imagine?  And can you explain what you mean by "semi-autonomous"?

Anrock:
> What kind of API do you imagine?

In Colobot there were "Exchange posts" buildings, where robots could store/retrieve info. API was basically a wrapper for string-string map, like `store("topic", "some value", distance)`. When executed it would put "some value" as value for "topic" key in all exchange posts in `distance` radius. Same for `retrieve` command.

So if we replace "exchange post" with "any entity with antenna" that would go as API for swarm. Distance argument in Colobot was there since every robot had depletable battery and distance affected how much energy was spent for a broadcast, we can drop that since bots seems to have indefinite batteries in swarm.

Further improvement of API could include storing something more interesting than strings for values, for example scan data or functions. Some kind of callback/notification mechanism when some broadcast is received seems to be good to have too.

> And can you explain what you mean by "semi-autonomous"?

If understand correctly, in Swarm once robot is built it will execute its program and there is no way to alter it or control robot dynamically. With antennas you can write something like that by broadcasting commands from base and writing an interpreter on bot side to execute those commands. Using base REPL it can be done interactively.

byorgey:
[...]

Issue #126, opened 2021-09-30
Inventory search


byorgey:
As you progress through the game, most likely the inventory will get very long and it will be annoying to have to scroll through the whole thing to find that one item for which you forgot the syntax or need the recipe or whatever.  We should add an incremental search mode, so that when the inventory is highlighted you can hit the `/` key and then start typing a few characters of the thing you're looking for to narrow the list based on your search term.  Maybe `brick` already has something like this built in for lists? I am not sure.

fryguybob:
I've wanted to explore `brick` for a while, so I started working on this.

byorgey:
Here's another idea: if we got rid of the j/k key navigation in the inventory panel (though I know that would make some people sad), what if simply typing in the inventory panel causes it to narrow?  I'm imagining something like

- There is a notion of a "current search string"
- Every time you hit a key that denotes a character, it appends to the current search string.
- The inventory list is filtered according to some sort of (fuzzy?) search based on the search string.
- The current search string could be displayed somewhere, maybe along the bottom edge of the panel
- There is some other key (Esc, perhaps?) to clear the current search string.

I think this would be fairly simple and wouldn't require coding any kind of independent "search mode".

byorgey:
Note that #151 contains an initial (but currently abandoned) attempt at this.

byorgey:
@kostmo I see you assigned yourself to this issue a long time ago --- had you written any code for it yet?  Or did you have any ideas beyond what's recorded here?  I might work on it if you're not.  The lack of inventory search is something that annoys me often. :smile:

Issue #35, opened 2021-09-22
Starting base with a large but finite supply of basic robot parts


byorgey:
Right now, the base can build infinitely many robots with basic devices: a detonator, grabber, solar panel, and treads.  However, it is a bit unrealistic that we can make infinitely many of these.  Well, not that the game is realistic, but you know what I mean.  It might be worth considering instead:

* Do not any longer install those devices on robots by default; they must be available in the inventory just like any other device.
* Start the base out with a large number (100? 500? 1000?  maybe the number depends on the difficulty level? maybe different numbers of different devices, e.g. fewer detonators than treads?) of these devices, so that at the start of the game the user can construct a bunch of basic robots "for free", but will eventually have to think about producing more of these devices.
* In order to make this work, we also have to come up with recipes for these devices, and think carefully about how to balance the depth and difficulty.

byorgey:
Maybe there should even be an entity for "robot chassis" with its own recipe, and building a robot requires one of those along with whatever devices it needs.

abyo:
- I'll agree that having to install those devices manually would allow for more customization. If we assume that #49 add fog of war, we can use a robot for exploration only (no need for a `grabber`)? Besides, assuming #22 allowing us to have full control on the robot, we don't necessarily need a `detonator` since we can re-use it for multiple collecting tasks.
- When it comes to the number of robots/devices by default, if we can re-use the robot for multiple tasks (`program` command?), I'd suggest having no more than 50 (more or less depending on difficulty) robots and fewer `grabber` devices maybe?
- I suggested a recipe in #93 but not making too easy to produce would be great in my opinion.

byorgey:
This becomes especially important if we restrict the use of the `view` command until later, as in #17 .  Then there would be no way to see and read about all the devices that get installed on robots by default.  If they were simply in the base's inventory to start then you could read about them before installing them on some robots.

byorgey:
I think we should first generalize challenges to be a way to describe the starting conditions for any game (see #296) --- then we don't have to have magic numbers built into the code to say what the base's starting inventory should be; we would just have a file describing classic mode, including the base's starting inventory.  Or even multiple files with different levels of difficulty.

TristanCacqueray:
A related confusing thing is that default devices installed on build can be salvage infinitely. In particular, it's a bit odd that the base does not have `treads`, but they can easily be acquired by salvaging a nearby robot early on. To avoid that, could we change the `salvage` output to filter out the default devices?

[...]

PR #128, opened 2021-10-01
Add scrolling for bottom-left info panel and logger device for message logs

### LINKED ISSUE #21: `salvage` command
There ought to be a way to retrieve all the inventory and devices from a "dead" robot.  I'm imagining something like `salvage : string -> cmd ()`, where `salvage x` will try to salvage a robot named `x`.  If successful, all of `x`'s inventory and installed devices go into the inventory of the robot that executed `salvage`, and `x` itself disappears.

Some issues to work out:

* What is the range of the `salvage` command?  I'm thinking probably you have to be in the very same cell.
* What if you don't know the name of the robot you want to salvage? There has to be some other way to discover the name(s) of nearby robot(s).  That should be orthogonal to `salvage`, but it's still worth thinking about them together.
* We'll need to add a new `salvage` capability.  What device should confer this capability, and what should its recipe be?  How difficult should it be to construct?

### LINKED ISSUE #36: Wrapping and scrolling for messages
Currently, when displaying the message queue, very simple logic is used to decide how to display the messages and how many to display.  This often leads to messages being unreadable, especially in a smaller terminal.  Also, we choose a static number of messages to display based on the height of the panel, but this is wrong if any of the messages are more than one line high; this leads to messages being cut off the bottom of the panel.

* We ought to use something like `txtWrap` to display the actual messages, so they wrap to whatever width the info panel happens to be.
* Instead of the current logic displaying the last n messages, we should render all the messages into a `brick` `Viewport`.  As a bonus, we should provide some sort of mechanism to allow the user to scroll it.  Note, this kind of scrolling ability will also be needed for entity descriptions, which can  be cut off at the bottom if they are too long and/or the window is too small.

### LINKED ISSUE #39: Add logger device
We should add a new kind of device (a `logger`, recipe: 1 log) which
allows logging messages via a `log` command (i.e. there will be a
`log` capability, provided by a `logger`, and required to execute the
`log` command).  This will also be the mechanism by which uncaught
exceptions are reported; on any robot without a logger, such uncaught
exceptions will just be silently discarded. See this [comment in `Swarm.Game.Step`](https://github.com/byorgey/swarm/blob/main/src/Swarm/Game/Step.hs#L378).

This will require adding a message queue field to the `Robot` record
type.  We should also get rid of the global message queue in the
`GameState`.

We'll have to rethink what the `say` command does---perhaps it should
broadcast a message that will be added to the log of nearby robots?

### LINKED ISSUE #40: Persist entity descriptions in the UI when switching to other panels
Right now, a description of the highlighted inventory entity is shown in the bottom portion of the info panel whenever the info panel is focused.  However, this disappears as soon as you switch away from the info panel.  This feels annoying, though, because an item description might contain something you want to refer to while typing a command at the REPL: some examples of a language feature, for example, or a list of recipes.

The thing that replaces the entity description when not focused on the
info panel is the global message log.  Perhaps the answer is to make
message logs per-robot, and to associate them with a logger device: see #39 .  Then the bottom part of the info
panel will always show info about the currently selected inventory
item, and if you want to see the message log, you can just select that
device in your inventory.

byorgey:
The original main goals of this PR are to (1) solve #36 , that is, make the bottom-left info area fully wrapped and scrollable; and (2) solve #40 , that is, persist the info in that bottom-left info area even when switching to other panels.  In order to address those I ended up making a rather large constellation of related changes, which in the end I hope work well with each other.  Eager to hear others' feedback either from looking at the code or just trying out the game with this PR.

- In order to make the bottom-left area scrollable, split it out into its own panel.  Now we can focus it and then use normal keybindings to scroll it.
- The bottom-left info panel now indicates whether it has additional content scrolled off the top or bottom.
- The bottom-left panel used to display a global list of log/error messages when the inventory was not focused.  So where will those messages be viewable now, if the information displayed in the bottom-left panel persists?  My answer: there is a special `logger` device (#39) which, when focused, now displays the log messages.
    - This means that logs are now per-robot instead of global.
    - The base starts with a `logger` installed.
    - A `log` command can be used to write to a logger.
    - Uncaught exceptions also now write to a logger (instead of using `say`), or are simply discarded if there is no `logger` device installed.
    - The first time a robot's inventory is viewed after a message has been appended to its log, the `logger` device will be automatically focused and the log scrolled all the way to the bottom to show the new entry.  This means that e.g. typing `move` at the REPL will immediately show an exception in the lower left.
- Add a `salvage` command, making it possible to access the log of a dead robot after the fact, once we implement #17 .

There are still a few issues to address which I think can be done in separate PRs:

- The global message log still exists, and the `say` command still appends to it, but there is no way to view it.  Obviously we need to do something with this --- either get rid of it, or figure out a sensible way to view it, etc.
- Exception messages should probably be reformatted to not include any newlines, now that we know they will be wrapped.

Closes #36.  Closes #40.  Closes #39.  Closes #21.


byorgey:
@xsebek , I know this messes up your distance-measuring functions that used `say`.  Would love to hear your thoughts on that.

data/recipes.yaml
```diff
@@ -9,6 +9,11 @@
   out:
   - [4, board]
 
+- in:
+  - [100, log]
```
xsebek, line 13: I hope this is a joke otherwise I might need to go cut down a forest or two :fearful: :evergreen_tree: :evergreen_tree: :evergreen_tree:

src/Swarm/TUI/Model.hs
```diff
@@ -148,14 +158,17 @@ data REPLHistItem
...
---   between the robot's /inventory/ and its /installed devices/.
-data InventoryEntry
+--   can either have an entity with a count in the robot's inventory,
+--   an entity installed on the robot, or a labelled separator.  The
+--   purpose of the separators is to show a clear distinction between
+--   the robot's /inventory/ and its /installed devices/.
+data InventoryListEntry
   = Separator Text
   | InventoryEntry Count Entity
+  | InstalledEntry Entity
```
xsebek, line 168: Nice improvement, the `1` entity installed was a bit confusing :+1:

xsebek (approved):

byorgey:
> I would just point out, that `install` is hidden under grabber below `--- . . . ----`, so maybe link to that or explicitly add the command here for new players?

Good point.  Updated the `logger` description to try to make this more clear.

> Do you want to remove the say command here, or shall we experiment with the log for a bit and then see if it's a good replacement?

Actually, I was kind of thinking that we might redesign `say` to retain some of its original character, in terms of broadcasting a message.  Like maybe the device needed for `say` is a speaker, and using `say` broadcasts a message to all robots within a certain radius.  (Implementing this might get tricky.)

byorgey:
I've now added a `salvage` command (which, as usual, turned out to be more interesting than I thought --- we didn't already have a way to look up robots by their location, and of course we don't want to have to scan through all robots to see where they are, so...).  This should help with debugging robots that crashed while carrying a `logger`.

I will just update the tutorial tomorrow and then I think this should be ready to merge, if others agree.  But very happy to entertain further questions or suggestions of course.  Thankful for the feedback so far which has made this a lot better already.

xsebek:
@polux well I think players will bother sending robots to `salvage` if they want to get the crashed robots inventory, save the logs for later and clean up. IMHO reasons enough, but we will have to find the right balance when adding antenna #17 and game modes #4.

byorgey:
> how come I can not give away installed devices (e.g. the devices are welded in), but salvage takes them out just fine?

Because salvaging is incredibly destructive, and can only be done by slicing the robot into bits with a high-powered plasma cutter? (Hmm, that gives me an idea for a new device...)

> Also, it currently gives an infinite supply of basic parts

But you already have an infinite supply of basic parts! You cannot do anything different with them than you could do before. This is really a problem with the way `build` works (and will be fixed by #35 ), not with `salvage` per se. 

> Viewing the salvaged robot is weird, fallback to base?

Good catch, I will look into that.

xsebek:
Huh, I can not argue against that reasoning. :slightly_smiling_face:

My point was, that not being able to salvage installed parts would leave the base-on-wheels and infinite-build problems for later. Really we could enable full salvage then, but right now in the "Classic/Creative" modes, it clutters my inventory and is surprisingly stronger than "give" which is provided by grabber too.

byorgey:
> My point was, that not being able to salvage installed parts would leave the base-on-wheels and infinite-build problems for later.

Right, but salvaging installed parts is going to be critical once we implement #35 , I'd rather just have it work the right way from the beginning.  Inventory clutter is definitely a problem that we're going to have to figure out, and it will only get worse once #164 is merged.  In other words just changing the way salvage works is not really going to fix it.  #126 will help, and maybe there can be ways for the user to filter the view, e.g. move some items to a different "I don't care" section at the bottom, toggle whether entities with count 0 are shown, etc.

byorgey:
> surprisingly stronger than "give" which is provided by grabber too.

Yeah, I agree, it is too strong, but we can fix that later.

xsebek:
> how to parse that last sentence

I was a bit worried, that you were planning to fix everything here, glad to hear it is ready :slightly_smiling_face:
Thanks for explaining your plans with relation to other issues, I guess only saw the early game point of view and did not see why it should change :wink:

PR #129, opened 2021-10-01
Add some benchmark

polux:
Add a benchmark measuring the time it takes to run 1000 ticks of the growing 10, 20, or 30 trees.


bench/Benchmark.hs
```diff
@@ -0,0 +1,77 @@
...
+  rn2s <- replicateM numTrees (randomRIO (0, 99))
+  let robots = zipWith3 mkTreeBot [V2 x 0 | x <- [0 ..]] rn1s rn2s
+  Right initState <- runExceptT initGameState
+  execStateT (mapM_ addRobot robots) initState
+
+-- | Runs numGameTicks ticks of the game.
+runGame :: Int -> GameState -> IO ()
+runGame numGameTicks = evalStateT (repeatM numGameTicks gameTick)
+  where
+    repeatM 0 m = m
```
byorgey, line 59: Is there a particular reason to define our own `repeatM` instead of using `replicateM_`?

bench/Benchmark.hs
```diff
@@ -0,0 +1,77 @@
...
+treeProgram :: Integer -> Integer -> ProcessedTerm
+treeProgram rn1 rn2 =
+  [tmQ|
+  let repeat : int -> cmd () -> cmd () = \n.\c.
+    if (n == 0) {} {c ; repeat (n-1) c}
+  in {
+    repeat ($int:rn1 + 500) wait;
+    appear "|";
+    repeat ($int:rn2 + 500) wait;
+    place "tree";
```
byorgey, line 32: It looks like each tree program will take 1000 + rn1 + rn2 + 3 or so ticks to run to completion, right?  So it seems odd to me that we are only running for 1000 ticks, when none of the tree programs will be finished.  But maybe that's on purpose?

polux: Good point, I forgot there were two waits! I'll change the constant to 300, even if that's different from the real value in the tree entity, because the benchmark would take too long otherwise.

bench/Benchmark.hs
polux: No! I hoogled for it but couldn't find it.

byorgey (approved):

PR #131, opened 2021-10-01
Add cabal to HLS configuration

xsebek:
- allow HLS to use Cabal instead of Stack
  - Cabal works fine, and I do not see any advantage to using Stack
  - someone using Stack should test this does not switch him to Cabal :warning:  
- ignore `cabal.project.local` so I can put there `ghc-options: -haddock`
  - this flag could lead to errors before GHC 9, but it gives me [docs on hover](https://haskell-language-server.readthedocs.io/en/latest/features.html)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,9 @@
 dist-newstyle/
 TAGS
 
+cabal.project.local
+hie.yaml
+
 .stack-work/
 stack.yaml.lock
 .hie/
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -199,6 +199,11 @@ already comfortable contributing to open source, you can skip to the [Convention
    rather than working in your own fork. For more information, see
    [I have push access to the Swarm repository, now what?](#i-have-push-access-to-the-swarm-repository-now-what).
 
+As a note, if you want to use the Stack tool with [HLS](https://github.com/haskell/haskell-language-server),
+check the `hie.yaml.stack` file. Pure Cabal should work fine with the defaults,
+but you might want to [add a GHC flag](https://haskell-language-server.readthedocs.io/en/latest/features.html)
+to build docs into `cabal.project.local`.
+
 ### Conventions
 
 We follow a few conventions to help keep everyone on the same page.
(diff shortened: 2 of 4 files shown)


TristanCacqueray (approved):
Thanks!

byorgey:
>     * someone using Stack should test this does not switch him to Cabal

I checked and it does actually switch to using Cabal by default, which in my experience at least uses a lot more memory for some reason.

I am not sure what a good solution is here.  Maybe we should provide two files in the repo, `hie.yaml.stack` and `hie.yaml.cabal`, and add `hie.yaml` to `.gitignore`, with instructions to copy whichever one you want to `hie.yaml`?

TristanCacqueray:
It seems like HLS can now works without the `hie.yaml` file, could we simply drop it?

xsebek:
> could we simply drop it?

Good point, it seems cabal works without it :man_facepalming:

byorgey:
> could we simply drop it?

Well, I would still use one personally, because I want it to use stack instead of cabal, and it sounds like @xsebek would still use one because they want something other than the default options. So in the end I think this amounts to the same thing as my suggestion.  Providing `hie.yaml.stack` and `hie.yaml.cabal` would be a convenience for new contributors who don't already know how to get it set up.  We could make sure they stay updated if each person e.g. just creates `hie.yaml` as a symlink to the one they prefer.

xsebek:
Oh, in that case, I think it is easier to:
1. add cabal config to `hie.yaml`
2. add a comment to hie.yaml for stack users to delete the Cabal part
3. ignore `hie.yaml` so further changes by users do not get committed

If we ever add another target, we can always add the `hie.yaml` file manually and we do not add clutter with weird files.

byorgey:
That sounds good to me.  If you add a comment to the `hie.yaml` and also add it to the `.gitignore` we can merge this.

xsebek:
Never mind my previous idea; it turns out that is not how Git works. :man_facepalming: 
@byorgey please check this makes sense and then we can merge :slightly_smiling_face:

byorgey:
Oh, I see, `.gitignore` only determines which untracked files it should ignore.

byorgey (approved):

PR #137, opened 2021-10-01
Annotate the AST with source location to improve error message

TristanCacqueray:
This change modifies the `Swarm.Language.Syntax.Term` data type to include the source location so that we can include a line number/column to type errors


TristanCacqueray:
@byorgey so if I understand correctly, c0a3dfe and its follow-ups (ignoring the unrelated fourmolu change) is the way to go to get initial source locations attached to type error. Could you please double check I got this right?

byorgey:
I'll take a look tomorrow!

TristanCacqueray:
It seems like the change could be made less intrusive by renaming the new Term constructor (e.g. `SApp`, `SBind`, ...) and providing pattern synonyms `TApp`, `TBind` to works with Term without syntax.

TristanCacqueray:
Since this is quite a complicated change, I based this change on the fourmolu version of the code, so we should merge #146 first and rebase this to drop the top commit.

byorgey (approved):
See my comments on #145 .

TristanCacqueray:
Removing draft status as this seems to work as expected. Though I haven't test the diagnostic extensively...

test/Unit.hs
```diff
@@ -100,5 +101,6 @@ prettyConst =
         )
     ]
  where
+  mkOp c t1 t2 = TApp (TApp (TConst c) t1) t2
```
xsebek, line 104: Let's fix the examples to use `mkOp'` instead :slightly_smiling_face:

src/Swarm/Language/Syntax.hs
```diff
@@ -397,18 +396,21 @@ data Syntax = Syntax {sLoc :: Location, sTerm :: Term}
 data Location = Location {locStart :: Int, locEnd :: Int}
   deriving (Eq, Show, Data)
 
-emptyLoc :: Location
-emptyLoc = Location 0 0
+instance Semigroup Location where
+  Location s1 e1 <> Location s2 e2 = Location (min s1 s2) (max e1 e2)
+
+instance Monoid Location where
+  mempty = Location 0 0
```
byorgey, line 403: This is not a valid `Monoid`, since `mempty <> x /= x`, e.g. `Location 0 0 <> Location s e = Location (min 0 s) (max 0 e) = Location 0 e /= Location s e`.  `min` and `max` do not have identity elements, so I don't think there is any valid way to define `mempty`. 

I see we are using `mempty` in various places where we need some default `Location` which hasn't been specified yet.  Why don't we instead define `Location = NoLoc | Location Int Int` and then define `mempty = NoLoc`, and define the combining operation accordingly.  This is the standard way to make a `Semigroup` without an identity element into a `Monoid`, by adding a new special value to serve as `mempty`.  We could also do this by wrapping `Location` in `Maybe`, but it seems nicer to add our own custom constructor.

byorgey (changes requested):

test/Unit.hs
TristanCacqueray: Oops, I forgot to remove the `mkOp'` definition, oh well let's use it, in case it is useful elsewhere.

src/Swarm/Language/Syntax.hs
TristanCacqueray: Oops I see, I knew this was too easy. Thanks for the tip.

byorgey (approved):

byorgey:
Tried it out and it works great!

byorgey:
@TristanCacqueray this looks good to me.  Go ahead and merge it if you think it's ready!

TristanCacqueray:
Alright, I think that's a good to have refactor. Thank you for all the feedback!

PR #138, opened 2021-10-01
Add noop keyword

### LINKED ISSUE #101: Add an explicit `noop` command
Right now the only syntax to use for a no-op is `{}`.  We should add a case to the parser to also allow writing `noop` explicitly.

7h3w1zz:
Adds noop keyword to the player-facing language. Previous syntax
(`{}`) is retained.  Only parsing of the language is affected, the
representation remains the same regardless of which syntax was used.
The new keyword is now the canonical version of the command, used in
pretty-printing and the like.

Fixes #101

I'm not certain about which syntax to use for the syntax info.  I decided to
go with the new `noop` keyword instead of the old `{}` because it is more
explicit and clearer.

### DIFF
--- a/src/Swarm/Language/Parse.hs
+++ b/src/Swarm/Language/Parse.hs
@@ -70,7 +70,7 @@ type ParserError = ParseErrorBundle Text Void
 reservedWords :: [String]
 reservedWords =
   [ "left", "right", "back", "forward", "north", "south", "east", "west", "down"
-  , "wait", "selfdestruct", "move", "turn", "grab", "place", "give", "make"
+  , "wait", "noop", "selfdestruct", "move", "turn", "grab", "place", "give", "make"
   , "build", "run", "getx", "gety", "scan", "upload", "blocked"
   , "random", "say", "view", "appear", "create", "ishere"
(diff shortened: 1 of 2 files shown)


byorgey (approved):
Looks great, thanks!  `"noop"` is the correct thing to use for the syntax info, if we want it to pretty-print as `noop` instead of `{}`.

Looking at this diff reminds me that the `reservedWords` which are constants should really be computed from `map constInfo allConst`, so we don't have to remember to update `reservedWords` every time we add a new constant like this.  Maybe that's something you'd like to tackle (in a separate PR)?

byorgey:
@7h3w1zz , congrats on having your first PR accepted to Swarm!  We're really glad to welcome you as part of the [community](https://github.com/byorgey/swarm/blob/main/COMMUNITY.md).  In a minute we will send you an invite to the repo (see https://github.com/byorgey/swarm/blob/main/CONTRIBUTING.md#i-have-push-access-to-the-swarm-repository-now-what ).  If you use IRC, feel free to also join the `#swarm` channel on Libera.Chat.

7h3w1zz:
Thanks for the warm welcome.  I'll take a look at the reserved words.

PR #142, opened 2021-10-02
Compatibility with hashable-1.3.4

danidiaz:
The code has an orphan instance `Hashable (Map k v)` but, starting from [hashable-1.3.4.0](https://hackage.haskell.org/package/hashable-1.3.4.0/docs/Data-Hashable.html#t:Hashable), that instance already exists in hashable itself, making compilation fail.

I disabled it for that case using a CPP directive.

### DIFF
--- a/src/Swarm/Game/Display.hs
+++ b/src/Swarm/Game/Display.hs
@@ -57,8 +57,6 @@ type Priority = Int
 
 -- Some orphan instances we need to be able to derive a Hashable
 -- instance for Display
-instance (Hashable k, Hashable v) => Hashable (Map k v) where
-  hashWithSalt = hashUsing M.assocs
 instance Hashable AttrName
 
 -- | A record explaining how to display an entity in the TUI.
--- a/stack.yaml
+++ b/stack.yaml
@@ -2,6 +2,7 @@ resolver: lts-18.12
 extra-deps:
 - brick-0.64
 - word-wrap-0.5
+- hashable-1.3.4.0
 - git: https://github.com/colinhect/hsnoise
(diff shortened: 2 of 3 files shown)


TristanCacqueray:
Thank you for your contribution!

@byorgey would it makes sense to update the min bound of hashable to 1.3.4.0 for such case?

byorgey:
Yes, let's just change the lower bound of `hashable` instead of using CPP.  I can't think of any reason we need to support older versions.

byorgey:
Then we can also remove the `-fno-warn-orphans` option.

danidiaz:
swarm currently uses the Stack resolver lts-18.12, which has [version 1.3.0.0](https://www.stackage.org/lts-18.12/package/hashable-1.3.0.0) of hashable. Could it be a problem?

byorgey:
> swarm currently uses the Stack resolver lts-18.12, which has [version 1.3.0.0](https://www.stackage.org/lts-18.12/package/hashable-1.3.0.0) of hashable. Could it be a problem?

Good point.  Let's just add `hashable-1.3.4.0` as an `extra-dep`.

byorgey (approved):

byorgey:
@danidiaz  , congrats on having your first PR accepted to Swarm!  We're really glad to welcome you as part of the [community](https://github.com/byorgey/swarm/blob/main/COMMUNITY.md).  In a minute we will send you an invite to the repo (see https://github.com/byorgey/swarm/blob/main/CONTRIBUTING.md#i-have-push-access-to-the-swarm-repository-now-what ).  If you use IRC, feel free to also join the `#swarm` channel on Libera.Chat.

PR #145, opened 2021-10-02
[Preview] Annotate the AST with source location to improve error message

TristanCacqueray:
This is just a preview of #137 until the main branch is fourmolu'ed

### DIFF
--- a/src/Swarm/Language/Parse.hs
+++ b/src/Swarm/Language/Parse.hs
@@ -227,29 +227,38 @@ parseConst = asum $ map alternative consts
   consts = filter isUserFunc allConst
   alternative c = c <$ reserved (syntax $ constInfo c)
 
-parseTermAtom :: Parser Term
+ploc :: Parser Term -> Parser Syntax
+ploc pterm = do
+  start <- getOffset
+  term <- pterm
+  end <- getOffset
+  pure $ Syntax (Location start end) term
+
+parseTermAtom :: Parser Syntax
 parseTermAtom =
-  TUnit <$ symbol "()"
-    <|> TConst <$> parseConst
-    <|> TVar <$> identifier
-    <|> TDir <$> parseDirection
-    <|> TInt <$> integer
-    <|> TString <$> stringLiteral
-    <|> TBool <$> ((True <$ reserved "true") <|> (False <$ reserved "false"))
-    <|> TLam <$> (symbol "\\" *> identifier)
-      <*> optional (symbol ":" *> parseType)
-      <*> (symbol "." *> parseTerm)
-    <|> TLet <$> (reserved "let" *> identifier)
-      <*> optional (symbol ":" *> parsePolytype)
-      <*> (symbol "=" *> parseTerm)
-      <*> (reserved "in" *> parseTerm)
-    <|> TDef <$> (reserved "def" *> identifier)
-      <*> optional (symbol ":" *> parsePolytype)
-      <*> (symbol "=" *> parseTerm <* reserved "end")
+  ploc
+    ( TUnit <$ symbol "()"
+        <|> TConst <$> parseConst
+        <|> TVar <$> identifier
+        <|> TDir <$> parseDirection
+        <|> TInt <$> integer
+        <|> TString <$> stringLiteral
+        <|> TBool <$> ((True <$ reserved "true") <|> (False <$ reserved "false"))
+        <|> TLam <$> (symbol "\\" *> identifier)
+          <*> optional (symbol ":" *> parseType)
+          <*> (symbol "." *> parseTerm)
+        <|> TLet <$> (reserved "let" *> identifier)
+          <*> optional (symbol ":" *> parsePolytype)
+          <*> (symbol "=" *> parseTerm)
+          <*> (reserved "in" *> parseTerm)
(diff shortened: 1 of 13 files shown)


src/Swarm/Language/Parse.hs
```diff
@@ -286,36 +297,49 @@ mkStmt (Just x) = Binder x
 --     App (App (TDef a) (TDef b)) (TDef x)
 --   This function fix that by converting the Apps into Binds, so that it results in:
 --     Bind a (Bind b (Bind c))
-fixDefMissingSemis :: Term -> Term
+fixDefMissingSemis :: Syntax -> Syntax
 fixDefMissingSemis term =
   case nestedDefs term [] of
     [] -> term
-    defs -> foldr1 (TBind Nothing) defs
+    -- TODO: figure out what should be the Syntax Location of this rewrite
```
TristanCacqueray, line 304: I guess there is nothing to do here because we shouldn't need location of the bind term (since we already have the one for t1 and t2)

src/Swarm/Language/Parse.hs
```diff
@@ -286,36 +297,49 @@ mkStmt (Just x) = Binder x
...
       (++)
-      [ Map.singleton 9 [InfixL (TApp <$ string "")]
+      [ Map.singleton 9 [InfixL (exprLoc2 $ SApp <$ string "")]
       , binOps
       , unOps
-      , Map.singleton 2 [InfixR (TPair <$ symbol ",")]
+      , Map.singleton 2 [InfixR (exprLoc2 $ SPair <$ symbol ",")]
       ]
 
+exprLoc2 :: Parser (Syntax -> Syntax -> Term) -> Parser (Syntax -> Syntax -> Syntax)
```
TristanCacqueray, line 326: Not so sure about these `exprLoc` functions, it's just adapter for `makeExprParser` usage. I guess we could attach location around the argument parser, but it does not seems useful.

src/Swarm/Language/Parse.hs
```diff
@@ -346,7 +370,7 @@ unOps = Map.unionsWith (++) $ mapMaybe unOpToTuple allConst
     pure $
       Map.singleton
         (fixity ci)
-        [assI (TApp (TConst c) <$ symbol (syntax ci))]
+        [assI (exprLoc1 $ SApp (noLoc $ TConst c) <$ symbol (syntax ci))]
```
TristanCacqueray, line 373: ditto, I think it's fine to use empty location here too.

src/Swarm/Language/Syntax.hs
```diff
@@ -361,10 +375,67 @@ constInfo c = case c of
...
+
 -- | Make infix operation (e.g. @2 + 3@) a curried function
 --   application (@((+) 2) 3@).
-mkOp :: Const -> Term -> Term -> Term
-mkOp c = TApp . TApp (TConst c)
+mkOp :: Const -> Syntax -> Syntax -> Syntax
+mkOp c s1@(Syntax l1 _) s2@(Syntax l2 _) = Syntax newLoc newTerm
+ where
+  -- The new syntax span both terms
+  newLoc = Location (locStart l1) (locEnd l2)
```
byorgey, line 387: Use a `Semigroup` instance for `Location` here.  The code as-is assumes that `s1` came before `s2`, which is probably always true now but seems an unnecessary assumption.  If we make a generic, symmetric `Semigroup` instance then the order does not matter.

byorgey (approved):
This looks good!  I have a couple small suggestions & questions, but nothing showstopping.

src/Swarm/Language/Parse.hs
TristanCacqueray: Alright, thank you for the suggestion. I've added a Monoid Location instance.

src/Swarm/Language/Syntax.hs
TristanCacqueray: Good point, thanks!

src/Swarm/Language/Parse.hs
```diff
@@ -227,57 +229,66 @@ parseConst = asum $ map alternative consts
   consts = filter isUserFunc allConst
   alternative c = c <$ reserved (syntax $ constInfo c)
 
-parseTermAtom :: Parser Term
+ploc :: Parser Term -> Parser Syntax
+ploc pterm = do
+  start <- getOffset
+  term <- pterm
+  end <- getOffset
```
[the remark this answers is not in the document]
TristanCacqueray: You are right, the current diagnostic includes trailing whitespace...

It seems like grace stopped recording the end position when adding location to type error, see this commit: https://github.com/Gabriel439/grace/commit/19b0178739361aa50493da47dd452483f73ce009#diff-ecb68a04196887b0821f476eb551f1c7e33ee26c780d305750ddd97adcf8cf12 . Though I haven't check how it reports error spanning multiple lines...

I guess we could do the same and simply compute the end position it as it is currently done for parse error (e.g. by measuring the length of the word after the start position), or we could `dropWhileEnd whitespace` right before computing the line/column number. I'll give that second approach a try as it seems more effective.

TristanCacqueray:
Thank you for the feedback, let me close that temporary PR now that the original one can be rebased on the main branch.

PR #149, opened 2021-10-02
Implicitly quantify over free type variables

### LINKED ISSUE #148: Implicitly quantify polymorphic types
`def id : a -> a = \x. x end` looks like it should work, but it turns out this function is unusable, since it expects the specific type `a` as input which does not unify with anything.  Instead, we would like it to be equivalent to writing `def id : forall a. a -> a = \x. x end`.

byorgey:
Also, if an explicit `forall` is given, require all type variables to be bound.

Fixes #148.

We might want to wait until merging #137 and then rebase this on top, since some changes might be required.

### DIFF
--- a/src/Swarm/Language/Parse.hs
+++ b/src/Swarm/Language/Parse.hs
@@ -1,5 +1,3 @@
------------------------------------------------------------------------------
------------------------------------------------------------------------------
 {-# LANGUAGE FlexibleInstances #-}
 {-# LANGUAGE NamedFieldPuns #-}
 {-# LANGUAGE OverloadedStrings #-}
@@ -53,6 +51,7 @@ import qualified Text.Megaparsec.Char.Lexer as L
 import qualified Text.Megaparsec.Pos as Pos
 
 import Data.Foldable (asum)
+import qualified Data.Set as S
 import Swarm.Language.Syntax
 import Swarm.Language.Types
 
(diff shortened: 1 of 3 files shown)


TristanCacqueray (approved):
I fixed the merge conflict. The change looks good to me, though I'm not familiar with quantification nor cata.

byorgey:
Thanks!

You are familiar with quantification already, if you understand how `forall` works in Haskell (and how it gets added implicitly).  That is, when you write a type like `a -> b -> a` in Haskell, it really means `forall a b. a -> b -> a`.  This is just doing the same thing, though perhaps you never thought about how to implement that. =)  It just goes through the whole type and looks for variables.

`cata` is a generic fold for recursive things defined via `Fix`.  Happy to explain it sometime but not in a GitHub issue comment. =)

TristanCacqueray:
Thanks again for the insightful comments. I meant I'm not familiar enough to verify the implication of such change :) I did read this useful article about quantification https://wasp-lang.dev/blog/2021/09/01/haskell-forall-tutorial , but this is still a bit too advanced for me...

Similarly I looked at https://github.com/recursion-schemes/recursion-schemes#flowchart-for-choosing-a-recursion-scheme , but haven't been able to find a good use for such thing. Thank you for the offer!

PR #151, opened 2021-10-03
Inventory search

fryguybob:
This gets some basic inventory filtering based on a prefix search term.  When focus is on the inventory pane and the <kbd>/</kbd> key is pressed it accepts a search term that can be ended with <kbd>Enter</kbd> to make the search stay <kbd>Esc</kbd> to clear the search.  It is workable, but a bit clunky.  Ideas are welcome :D.


src/Swarm/TUI/Controller.hs
```diff
@@ -507,6 +509,10 @@ handleInfoPanelEvent s (VtyEvent (V.EvKey V.KEnter [])) = do
               & gameState . replStatus .~ REPLWorking mkTy Nothing
               & gameState . robotMap . ix "base" . machine .~ initMachine mkPT topEnv
         _ -> continueWithoutRedraw s
+
+handleInfoPanelEvent s (VtyEvent (V.EvKey (V.KChar '/') []))
+  = continue $ s & uiState . uiSearching .~ True
```
TristanCacqueray, line 514: nit: there is no feedback when entering search mode, should we display an indicator indicating the ui is waiting for input? Also perhaps we should add the key to the help bar, e.g. next to the navigate tooltip `[/] search`.

src/Swarm/TUI/Controller.hs
```diff
@@ -516,3 +522,29 @@ handleInfoPanelEvent s (VtyEvent ev) = do
...
+-- | Handle user input events searching in the info panel.
+handleSearchingEvent :: AppState -> BrickEvent Name AppEvent -> EventM Name (Next AppState)
+handleSearchingEvent s (VtyEvent (V.EvKey V.KEsc []))
+  = continue $ s & uiState . uiSearching .~ False
+                 & uiState . uiSearch .~ ""
+                 & uiState . uiInventory . _Just . _1 .~ 0
+
+handleSearchingEvent s (VtyEvent (V.EvKey V.KEnter []))
+  = continue $ s & uiState . uiSearching .~ False
+                 & uiState . uiInventory . _Just . _1 .~ 0
```
TristanCacqueray, line 535: Should we remove the search on enter? otherwise it seems like we need to search again and press escape to clear the search.

src/Swarm/TUI/Controller.hs
```diff
@@ -516,3 +522,29 @@ handleInfoPanelEvent s (VtyEvent ev) = do
...
+
+handleSearchingEvent s (VtyEvent (V.EvKey (V.KChar c) []))
+  = continue $ s & uiState . uiSearch %~ (`T.snoc` c)
+                 & uiState . uiInventory . _Just . _1 .~ 0
+
+handleSearchingEvent s (VtyEvent (V.EvKey V.KBS []))
+  = continue $ s & uiState . uiSearch %~ safeTextInit
+                 & uiState . uiInventory . _Just . _1 .~ 0
+
+handleSearchingEvent s _ = continueWithoutRedraw s
```
TristanCacqueray, line 545: I think we should handle navigation event (e.g. up/down) so that we can select the desired item while searching.

src/Swarm/TUI/Model.hs
```diff
@@ -308,15 +321,19 @@ populateInventoryList :: MonadState UIState m => Maybe Robot -> m ()
...
+        = (\case { [] -> []; xs -> Separator label : xs })
+        . map mkInvEntry
+        . filter (T.isPrefixOf search . view entityName . snd)
+        . sortOn (view entityName . snd)
+        . elems
+      allItems = (r ^. robotInventory . to (itemList "Inventory"))
+              ++ (r ^. installedDevices . to (itemList "Installed devices"))
+      items
+        | T.null search = allItems
+        | otherwise     = (Separator $ "Search: " <> search) : allItems
```
TristanCacqueray, line 336: when searching, i think it would be nice to adjust the currently selected item to skip the separator, otherwise when starting a search, the `Installed devices` separator is selected.

TristanCacqueray (changes requested):
Thank you for the change, this is looking great!

I think the KEnter handler needs to be changed to avoid being stuck in search mode.

src/Swarm/TUI/Controller.hs
TristanCacqueray: Alternatively, we could merge the KEsc handler with the global one, so that even when not in search mode, pressing Esc would still clear the search.

src/Swarm/TUI/Controller.hs
fryguybob: This bothered me as well.  I'll have to dig in to `brick` more to figure out how to do what I want which is blinking cursor on search `Separator`.  It of course doesn't have to be a separator, but that already existed.  It also has been bothering me that the cursor continues to blink in the repl pane when focus is in another pane.

src/Swarm/TUI/Controller.hs
fryguybob: I liked the idea of a "sticky" search, but perhaps that is not a wanted feature.  I agree it is awkward to have to enter search and hit <kbd>Esc</kbd> to clear it.

src/Swarm/TUI/Controller.hs
fryguybob: Oh, I think I can pass those events on to the pane and it might do the right thing.

src/Swarm/TUI/Model.hs
fryguybob: Yeah, this is a bit weird.

byorgey:
I'm going to close this for now, but of course the code is still here if anyone wants to revive it.

PR #155, opened 2021-10-03
Case insensitivity bug

fryguybob:
The `Inventory` `counts` and `byName` maps could get out of sync due to missing case insensitivity on deletion.  Perhaps a better fix would be to use something like [https://hackage.haskell.org/package/case-insensitive](https://hackage.haskell.org/package/case-insensitive).


fryguybob:
Oops, this was based on the wrong branch.

PR #157, opened 2021-10-03
bug fix: byName out of sync with counts due to case insensitivity.

fryguybob:
The `Inventory` `counts` and `byName` maps could get out of sync due to missing case insensitivity on deletion.  Perhaps a better fix would be to use something like [https://hackage.haskell.org/package/case-insensitive](https://hackage.haskell.org/package/case-insensitive).

### DIFF
--- a/src/Swarm/Game/Entity.hs
+++ b/src/Swarm/Game/Entity.hs
@@ -531,7 +531,7 @@ deleteCount k e (Inventory cs byN) = Inventory cs' byN'
   newCount = lookup e (Inventory cs' byN)
 
   byN'
-    | newCount == 0 = M.adjust (IS.delete (e ^. entityHash)) (e ^. entityName) byN
+    | newCount == 0 = M.adjust (IS.delete (e ^. entityHash)) (T.toLower $ e ^. entityName) byN
     | otherwise = byN
 
   removeCount :: Maybe (Count, a) -> Maybe (Count, a)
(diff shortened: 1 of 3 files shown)


xsebek:
I am not familiar with this code, but since you say it is a *bug fix*, could you please describe how to reproduce it with the current code? Maybe we could extend our test suite :slightly_smiling_face:

fryguybob:
The `Inventory` data type has two maps:

```haskell
data Inventory = Inventory
  { counts :: IntMap (Count, Entity) -- main map
  , byName :: Map Text IntSet -- Mirrors the main map; just
  -- caching the ability to
  -- look up by name.
  }
```

Entries with a hash in the `byName` map should exist with non-zero count in the `counts` map.  Case insensitivity is maintained by inserting into `byName` with `T.toLower`, but this was not done for deletion.  I encountered this by `install`ing a `"3D printer"` not realizing that it would automatically install due to the use of `build`.  The use of `build` removed my one `"3D printer"` from inventory (`delete`) which removed the printer from the `counts` but not from `byName` as the entry in `byName` was `"3d printer"`.

It can be recreated by running a program like this when you have one `3D printer` in inventory:

```.swarm
def buildDude =
    d <- build "dude" noop;
    install d "3D printer"
end;
```
Expected behavior is to give an error about no `3D printer` available.  Currently the program crashes due to `IntMap.!`.

fryguybob:
I'll push a unit test in a second.

xsebek:
Right, I had to do it twice to get
```
swarm: IntMap.!: key 2273174733132554414 is not an element of the map
```
Nice catch @fryguybob and thanks for the fix and test!

byorgey:
@xsebek note the mergify bot requires a `merge me` label *and* at least one approving review to merge (click on "Files changed", then click on "Review changes".  You do not necessarily need to write comments there unless you want to, you can just check the box for approval).

xsebek (approved):

xsebek:
Thanks, @byorgey! That is a bit surprising workflow when the PR is already perfect :sweat_smile:

PR #158, opened 2021-10-03
Handle DocumentDidChange LSP event

TristanCacqueray:
This change enables the LSP server to report diagnostic on
the virtual file, without waiting for a save event.

Not sure how to correctly configure emacs (and why it is not needed
for HLS), but this only works when `flycheck-check-syntax-automatically`
has `idle-change`, e.g.:

  (setq flycheck-check-syntax-automatically '(save mode-enable idle-change))


byorgey (approved):
Works great!

PR #159, opened 2021-10-03
Accept filename without .sw extension in run command

### LINKED ISSUE #135: Ability to omit `.sw` extension from argument to `run`
The `run` command could automatically add a `.sw` extension if given an argument with no extension, e.g. `run("mydefs")` will load `mydefs.sw`.  `run("mydefs.sw")` should still work too.

twitu:
Closes #135

### DIFF
--- a/src/Swarm/Game/Step.hs
+++ b/src/Swarm/Game/Step.hs
@@ -862,17 +862,26 @@ execConst c vs k = do
         flagRedraw
         return $ Out (VString (newRobot' ^. robotName)) k
       _ -> badConst
+    -- run can take both types of text inputs
+    -- with and without file extension as in
+    -- "./path/to/file.sw" and "./path/to/file"
     Run -> case vs of
       [VString fileName] -> do
-        mf <- liftIO $ readFileMay (into fileName)
+        mf <- liftIO $ readFileMay (into filePath)
 
-        f <- mf `isJustOr` cmdExn Run ["File not found:", fileName]
+        f <- mf `isJustOr` cmdExn Run ["File not found:", filePath]
 
         t <-
           processTerm (into @Text f) `isRightOr` \err ->
-            cmdExn Run ["Error in", fileName, "\n", err]
+            cmdExn Run ["Error in", filePath, "\n", err]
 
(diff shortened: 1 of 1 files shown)


src/Swarm/Game/Step.hs
```diff
@@ -862,17 +862,26 @@ execConst c vs k = do
...
+        f <- mf `isJustOr` cmdExn Run ["File not found:", filePath]
 
         t <-
           processTerm (into @Text f) `isRightOr` \err ->
-            cmdExn Run ["Error in", fileName, "\n", err]
+            cmdExn Run ["Error in", filePath, "\n", err]
 
         return $ initMachine' t empty k
+       where
+        hasExtension = T.isSuffixOf ".sw" fileName
```
xsebek, line 880: This could surprise a user that does not know about the `.sw` extension and has for example a file `moves` or `zigzag.swarm` - maybe consider adding a `doesFileExist` above `readFileMay`? :slightly_smiling_face:

xsebek (approved):
Thanks for the PR @twitu, I have just one nitpick, but otherwise it looks good to me :+1:

src/Swarm/Game/Step.hs
TristanCacqueray: Perhaps we should pick the first existing file from [fileName, fileName <> ".sw"], and keep the provided fileName in the error message.

twitu: I've made the required changes.

TristanCacqueray (approved):
Thanks!

byorgey:
@twitu  , congrats on having your first PR accepted to Swarm!  We appreciate the contributions and are really glad to welcome you as part of the [community](https://github.com/byorgey/swarm/blob/main/COMMUNITY.md).  In a minute we will send you an invite to the repo (see https://github.com/byorgey/swarm/blob/main/CONTRIBUTING.md#i-have-push-access-to-the-swarm-repository-now-what ).   And this is the part of the standard welcome message where I would invite you to join us on IRC, but that's unnecessary in this case. =)

PR #160, opened 2021-10-03
Add hint for using repl history

### LINKED ISSUE #59: Add hints for up/down keys to scroll history in the REPL
Right now, while in the REPL panel, the key menu only displays
```
 [^q] quit  [Tab] cycle panels  [Enter] execute
```
but it ought to also note that the up and down arrows can be used to scroll the history.  So maybe something like this:
```
 [^q] quit  [Tab] cycle panels  [↑↓] history  [Enter] execute
```

twitu:
Closes #59


TristanCacqueray (approved):
Thanks!

PR #164, opened 2021-10-03
leave discovered entities in the inventory forever

### LINKED ISSUE #163: Leave discovered entities in inventory
Once an entity has been discovered (via `scan`, `upload`, or `grab`), it should never leave the inventory.  If you use up all of them, then of course you will have 0, but there should still be an entry with a number of 0 (just as currently happens with `scan` and `upload`).  As an optional nice touch, there could be a key to toggle the inventory display between showing all discovered entities, and showing only the ones with a nonzero count.

See https://github.com/byorgey/swarm/issues/49#issuecomment-933013945 for discussion.

byorgey:
Once an entity has been discovered, it stays in the inventory, even if its count becomes 0.

I think I just deleted the code @fryguybob carefully fixed in #157... =)

Fixes #163.


byorgey:
This actually breaks a test but our CI doesn't run the test suite!  I will fix that shortly.

byorgey:
Please merge https://github.com/byorgey/swarm/pull/165 first, then we can decide what to do about the failing test here.

fryguybob:
The change is ok, but we also need to visit uses of `lookupByName` as they are currently assumed to return empty list when there is zero of an entity in inventory.  With this patch if a robot has had a resource, it can keep giving forever.

byorgey:
> we also need to visit uses of `lookupByName` as they are currently assumed to return empty list when there is zero of an entity in inventory

Very good catch!  I think this should be fixed now.

fryguybob (approved):
Looks good!

PR #169, opened 2021-10-04
Add Nix support

Skyfold:
In particular this adds support for:
- Note: I uploaded built versions of the development environment and build dependencies to cachix (It takes a long time to build them yourself)
  - setup cachix and run `cachix use swarm` before you build
- docker
  - `nix-build --attr docker`
  - `docker load -i result`
  - `docker run -it swarm:latest`
  - If you want to load files you can mount the project directory: `docker run -it -v $PWD:/tmp swarm:latest`
    - then run("/tmp/example/wander.sw") will work
    - I could get .swarm_history to work, but I need swarm to read it from a specific path instead of the current directory (cannot mount directly to /)
- build dynamically linked executable
  - `nix-build --attr exe`
  - `./result/bin/swarm`
- nix development environment `nix-shell`
  - comes with: hls, hoogle (with only swarm deps), fourmolu, hlint, ghcid
  - Note: you have to remove cabal.project file for cabal new-repl to find `dependent-sum` and `th-extras`. 
    - I could not figure out why cabal cannot find only those two dependencies from the cabal.project file when ghc-pkg shows the correct versions available. Regardless, you can remove the cabal.project file safely when using nix to build and develop, the default.nix file specifies the dependencies in cabal.project.

Note: I tried to get a statically linked version of swarm building, but failed to get ghc 8.10.7 to build with musl.

What this pull request is missing is how installing/developing with Nix should be presented in the README.md and any integration with your current CI setup. Let me know what you think.


TristanCacqueray:
Thank you for the contribution @Skyfold. Though we already rejected such addition, please see https://github.com/byorgey/swarm/pull/80 for the motivation.

On the other hand, if this can be used to produce a static executable, then it might be worth to consider nix again, at least for the release process, though we are not there yet.

byorgey:
We do really appreciate your hard work in putting this together, and realize it could be useful for others who might want to build swarm with Nix.  But as Tristan alluded to, we don't want to have to support this going forward.  However, I think I would be willing to accept it if you can promise to support it in perpetuity (to the best of your ability), i.e. updating it if it needs to be updated, responding to reports of breakage, etc.

byorgey:
I'm going to close this.  @Skyfold , please do post this on the [Swarm wiki](https://github.com/byorgey/swarm/wiki) or elsewhere!

Issue #143, opened 2021-10-02
Add a `drill` command


byorgey:
Currently you can construct a drill but not use it for anything.  Current ideas:

* add a constant `drill : dir -> cmd ()`, requiring a `drill` capability which is of course granted by a `drill`
* drilling a `boulder` will make it disappear and maybe add some rocks to your inventory?
* drilling `down` on certain tiles bearing copper ore should yield more; see https://github.com/byorgey/swarm/issues/18
* see https://github.com/byorgey/swarm/issues/18#issuecomment-928358067 for an idea of what can happen when you drill a mountain.

xsebek:
As an alternative, it could be `drill : dir -> cmd string` giving `"dirt"` on any normal grassy tile. I prefer this style so I can simply check what I got.

As a bonus, we could go for Factorio style replacing water with [Landfill](https://wiki.factorio.com/Landfill), crafted from *X* dirt.

Some interactions to consider:
- drilling water - I guess that is not safe for the drill :smirk: 
- drilling normal entities that can be `grab`ed - could add them to inventory (seems too easy), leave them alone or destroy them

xsebek:
I hacked together a quick implementation and will push when I get back on wi-fi 🙂 

My idea was to add recipes with drill as a required entity, so it renders nicely and adding new entities that can be drilled is easy.

PR #170, opened 2021-10-04
Add seed option to CLI

### LINKED ISSUE #13: Deterministic `random` values
Values generated by the `random` command should actually be deterministically based on the world seed.  We could have some kind of PRNG that gets stored in the `GameState`.  Right now the random values are just generated in the `IO` monad.

### LINKED ISSUE #14: Allow generating worlds from different seeds
Right now, in `Swarm.Game.WorldGen`, some seed values are simply hardcoded.  Instead, especially once we implement #4 , when the user starts a new game they should be able to specify a seed value or have one randomly generated.  Then the seed value should in turn deterministically generate a unique world.

### LINKED ISSUE #90: Tree at base?
**Describe the bug**

I'm not sure if this is intended or not, but upon starting in a new world, `grab` issued at the base will grab a `"tree"`, adding a tree to the base's inventory. The map does not appear to change. Eventually the base tree will regrow and can be `grab`bed again.

**To Reproduce**

Start the game. Enter `grab` at the REPL.

Or just ask `ishere "tree"`.

**Expected behavior**

Either there's no tree at base, or the map shows it ... or maybe it's just an Easter egg.

xsebek:
- adds CLI option `--seed` and propagates it to `testWorld2` (Closes #14)
- moves the base in the tree shade (Closes #90)
- makes the `random` command depend on initial seed (Closes #13)


xsebek:
Also messes with the `genBiome` settings to get more mountains :sweat_smile: :mountain: 

The main thing to improve is to make `--seed` visible in `--help`, but even better would be to also add biome threshold values loading from some configuration YAML file.

xsebek:
Resolves:
- #90 
- #14

Related (but not solved yet):
- #13

xsebek:
Seed 1 is one of the funny ones :laughing: But think how satisfying it would be to [`drill`](#143) through.

I think these values should be left to the users to tweak and it would help us choose the "best" ones too - we could run:
```sh
while inotifywait -e close_write swarm_conf.yaml; do killall swarm; swarm; done
```

But maybe it is best to make another issue and focus on the `--seed` here :sweat_smile: I will revert the values and try to get the parser right :wink:

byorgey:
Yeah, it makes sense to just add the `--seed` option in one PR and fiddle with world values in another.

> Seed 1 is one of the funny ones laughing But think how satisfying it would be to [`drill`](#143) through.

True. :grin:

xsebek:
The `--help` would now read:
```
Swarm game - pre-alpha version

Usage: swarm [COMMAND | (-s|--seed ARG)]
  To play the game simply run without any command.

Available options:
  -s,--seed ARG            Seed for world generation
  -h,--help                Show this help text

Available commands:
  format                   Format a file
  lsp                      Start the LSP
```

xsebek:
The `random` code may need to be improved - I do not understand Lenses or PRGs that well, and the code uses both. :worried: 
But it seems to give the same numbers every time - `random 6` is 0,0,0,3,4...

TUTORIAL.md
```diff
@@ -307,6 +308,24 @@ you want to reload the definitions.  Eventually, there will be a way
 to both save and load commands, but this is better than nothing for
 now.
 
+World generation
+----------------
+
+If you do not like the starting place of the base, there is a way
+to start somwhere else. *In a different world!*
```
byorgey, line 315:
```suggestion
to start somewhere else. *In a different world!*
```

TUTORIAL.md
```diff
@@ -307,6 +308,24 @@ you want to reload the definitions.  Eventually, there will be a way
...
+
+```bash
+$ swarm --seed $RANDOM
+```
+
+You can specify the *world seed* leading to a radically different
```
byorgey, line 321:
```suggestion
You can specify the *world seed* leading to radically different
```

byorgey (approved):
Looks good to me.  A few minor editing suggestions in the tutorial.

byorgey:
The random code looks fine.  There's probably some clever way to generate a random number and update the `StdGen` in the `GameState` in one line of code, but I don't know what it is off the top of my head, and it would probably be unreadable.

PR #172, opened 2021-10-04
Add robot self naming capability

twitu:
Adds the following -
* `mirror` entity that adds `self` capability
* `self` returns a callee robot's name

I have added the names and description that I saw fit. Please recommend better names as you see fit.


twitu:
Closing pull request to fix historical commits showing up.

PR #174, opened 2021-10-04
Add self naming capability

### LINKED ISSUE #113: Command for a robot to find out its own name
Right now there's no way for a robot to know what its own name is; I propose adding `myname : cmd string` for this purpose. This seems like it might be useful.  For example, if robot A builds robot B which is supposed to fetch something and bring it back to A, robot B needs to be provided with the name of robot A so it can `give` to it.

twitu:
Adds the following -

* mirror entity that adds self capability
* self returns a callee robot's name

I have added the names and description that I saw fit. Please recommend better names as you see fit.

Here are a few examples on using `self`.

```
self  -- this will return "base"
build "fetch" { move; move; name <- self; move; } -- this will print "fetch" and store "fetch" in `name` for later use
```

Closes #113

### DIFF
--- a/data/entities.yaml
+++ b/data/entities.yaml
@@ -411,7 +411,7 @@
     char: 'U'
   description:
   - "With a mirror, robots can reflect on themselves and see their own name"
-  - "A mirror enables the `self` command which returns the calling robots
+  - "A mirror enables the `whoami` command which returns the callee robot's
      name as text"
   properities: [portable]
-  capabilities: [self]
+  capabilities: [whoami]
--- a/data/entities.yaml
+++ b/data/entities.yaml
@@ -404,3 +404,14 @@
     knowledge to another nearby robot; for example, 'upload \"base\"'."
   properties: [portable]
   capabilities: [scan, sensefront]
+
+- name: mirror
+  display:
+    attr: device
+    char: 'U'
+  description:
+  - "With a mirror, robots can reflect on themselves and see their own name"
+  - "A mirror enables the `self` command which returns the calling robots
+     name as text"
+  properities: [portable]
+  capabilities: [self]
--- a/data/entities.yaml
+++ b/data/entities.yaml
@@ -410,8 +410,8 @@
     attr: device
     char: 'U'
   description:
-  - "With a mirror, robots can reflect on themselves and see their own name"
-  - "A mirror enables the `whoami` command which returns the callee robot's
-     name as text"
+  - "With a mirror, robots can reflect on themselves and see their own name."
+  - "A mirror enables the 'whoami' command, which returns the robot's
+     name as a string."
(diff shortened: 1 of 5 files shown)


xsebek:
Nice work @twitu! :+1: 

I imagine the recipe for the mirror could use water and copper (iron or silver would be cooler) adding another "cooper plate" entity could come in handy later.
I think `self` is confusing as it's not a verb - perhaps we could use Linux `whoami` or something like `getname` to line up with `getx`?

twitu:
Yeah `self` is pretty confusing. Although I'm inclined towards `whoami` especially since we might find linux machines on the map. But I guess it's somewhat ambiguous as to what it does, so I'll stick with `getname`.

TristanCacqueray:
`whoami` sounds like a good name for this command.

byorgey:
I like `whoami` too.  In fact while we're at it, I think we should also replace `getx` and `gety` with a `whereami` command. =)

twitu:
I'm not very sure of what would be a good recipe for mirror. These are a couple of ideas I have -

* 1 metal sheet + 1 silver (realistic)
* 1 linux console + 1 sysadmin ( 1 sysadmin = 8 bit(0) + 2 rocks) (funny)

data/entities.yaml
```diff
@@ -404,3 +404,14 @@
...
   capabilities: [scan, sensefront]
+
+- name: mirror
+  display:
+    attr: device
+    char: 'U'
+  description:
+  - "With a mirror, robots can reflect on themselves and see their own name"
+  - "A mirror enables the `whoami` command which returns the callee robot's
+     name as text"
```
byorgey, line 415:
```suggestion
  - "With a mirror, robots can reflect on themselves and see their own name."
  - "A mirror enables the 'whoami' command, which returns the robot's
     name as a string."
```

byorgey (approved):
Looks good, I just have some small suggested edits to the mirror description.

We can go ahead and merge without a recipe (it would still be possible to use in creative mode and we can add a recipe later), or we can try to decide on a recipe first, it's up to you.  I will leave some thoughts re: recipe in a separate comment.

byorgey:
For recipes, if we do the realistic one, we would have to add `silver` and figure out where you can get that.  Probably it's going to be rare and/or difficult to mine.

Your idea for a `sysadmin` entity is funny. =)  And what is a `Linux console` and where can you get one?  Or did you mean the `Linux` entity?  Or maybe you have to combine `Linux` with something else?

twitu:
Yes I meant `Linux` entity with `sysadmin` as a possible recipe for `mirror`. @xsebek suggestions are also good ones.

However I was thinking we can postpone adding a recipe `mirror` for later, when we can see more use cases where `whoami` will be used.

Issue #66, opened 2021-09-25
Add notification when game is not up-to-date


abyo:
Right now whenever the game is having an update, we have no notification of it and that can create problems for people following the tutorial. I ran the game right after I saw the blog post and didn't tried it until next day. After reading the tutorial I got an error while typing this line.

```
build "fetch" {turn west; m8; thing <- grab; turn back; m8; give "base" thing; selfdestruct}
```

![](https://cdn.discordapp.com/attachments/871732319449395240/891253192644526111/update_error.png)

This happened because my game wasn't up-to-date and in this case it was because of the give command modification (23f69a1). After updating I could type the command just fine.

![](https://cdn.discordapp.com/attachments/871732319449395240/891253194817167360/update_solve.png)

It still took me a few minutes to understand that the error wasn't on my side. So I'm suggesting to implement a text that would tell the user to update his game. **Note that this problem isn't going to happen whenever we get in-game tutorial (#25) but it might still be relevant in the future to notify the user of new updates.**

byorgey:
Ah, great idea!  I'm thinking something like:
1. Use https://hackage.haskell.org/package/githash to automatically compile in the hash of the latest git commit.
2. Add some code to `Swarm.TUI.Model.initUIState` which uses the GitHub API to request the hash of the most recent commit from the repo on github and check if it is the same as the has obtained locally.  If not, initialize the `uiError` widget so that it will pop up a dialog box warning you that you are running an out-of-date version.

byorgey:
I just thought of a small wrinkle: it would be very annoying while developing and testing on some branch to have this always pop up because the latest commit on my branch does not match the latest commit on `main`.  So I guess it should also get the name of the current branch, and ask for the latest commit on the remote branch with the same name.

And there should probably also be a command-line flag `--no-update-check` or something to disable this behavior, since there's probably going to be some annoying case we didn't think of.

byorgey:
For access to the github API, see https://hackage.haskell.org/package/github .

byorgey:
See the above linked commit to see how to use `githash` to get the current branch and commit hash.

xsebek:
> And there should probably also be a command-line flag --no-update-check or something to disable this behavior, since there's probably going to be some annoying case we didn't think of.

@byorgey how about adding a widget in the main menu instead that would share this wonderful news?

Something like:
```
   v<^vv<<@   ^^     vv    >^v^T    ^^^v<     <><      v   >
[...]

PR #176, opened 2021-10-04
Only update robots that are not waiting.

### LINKED ISSUE #20: `wait` command and waiting robot queue
Right now, the only way for a robot to wait for a certain amount of time is to execute a program like `repeat 500 wait`, but this means the robot is doing a bunch of computation steps every tick (evaluating the recursive `repeat` command, subtracting one from a number, checking whether it is zero...).  Instead, there should be a way for a robot to register its intent to sleep for a certain amount of time, and have the game engine wake it up later.  This will save a lot of work --- for example, I have noticed that the game slows down noticeably when harvesting a large forest, generating a large number of system seed robots which are all busily waiting for a long time to grow.

The idea is as follows.  Note that #19 needs to be implemented first.

* Change the `wait` command to take an `int` argument, denoting the duration of waiting.
* Right now the `wait` command requires no capabilities, but perhaps with this change it should.  I propose adding a `time` capability, which can be provided by a new `clock` entity (recipe: some gears + some springs (in turn made from copper wire)?).
    * While we're at it, perhaps we can also add `getTime : cmd int`, also requiring the `time` capability.
* Add a counter to the `GameState` which tracks the number of ticks since the game started.
* Add a priority queue of waiting robots to the `GameState`, keyed by the tick at which the robot should wake up.
* When a robot executes a `wait` command, remove it from the `activeRobots` queue and place it in the priority queue (after calculating the time at which it should be woken).
* Each tick, before iterating through the active robots, check the priority queue and move any robots which should be woken on this time step into the active robot queue.

polux:
Resolves #20 by introducing an active robots set and a waiting robots queue as discussed in the issue.

Growing trees benchmark before change:
```
benchmarking run 1000 game ticks/10 trees
time                 141.1 ms   (137.3 ms .. 146.4 ms)
                     0.997 R²   (0.995 R² .. 1.000 R²)
mean                 140.1 ms   (138.5 ms .. 142.2 ms)
std dev              3.807 ms   (2.703 ms .. 5.189 ms)
                   
benchmarking run 1000 game ticks/20 trees
time                 298.9 ms   (281.7 ms .. 326.0 ms)
                     0.989 R²   (0.972 R² .. 0.998 R²)
mean                 301.7 ms   (295.0 ms .. 311.0 ms)
std dev              13.93 ms   (10.02 ms .. 19.92 ms)
                   
benchmarking run 1000 game ticks/30 trees
time                 470.9 ms   (432.0 ms .. 513.0 ms)
                     0.991 R²   (0.978 R² .. 1.000 R²)
mean                 450.1 ms   (441.9 ms .. 468.4 ms)
std dev              18.85 ms   (5.941 ms .. 30.21 ms)
variance introduced by outliers: 11% (moderately inflated)
```

After change:
```
benchmarking run 1000 game ticks/10 trees
time                 4.666 ms   (4.529 ms .. 4.955 ms)
                     0.963 R²   (0.899 R² .. 0.999 R²)
mean                 4.447 ms   (4.361 ms .. 4.668 ms)
std dev              539.2 μs   (185.5 μs .. 1.097 ms)
variance introduced by outliers: 80% (severely inflated)
                   
benchmarking run 1000 game ticks/20 trees
time                 6.955 ms   (6.826 ms .. 7.122 ms)
                     0.996 R²   (0.993 R² .. 0.999 R²)
mean                 6.855 ms   (6.795 ms .. 6.936 ms)
std dev              262.8 μs   (197.0 μs .. 340.6 μs)
variance introduced by outliers: 25% (moderately inflated)
                   
benchmarking run 1000 game ticks/30 trees
time                 10.53 ms   (9.738 ms .. 11.47 ms)
                     0.954 R²   (0.914 R² .. 0.988 R²)
mean                 9.504 ms   (9.226 ms .. 9.974 ms)
std dev              1.211 ms   (809.9 μs .. 1.948 ms)
variance introduced by outliers: 76% (severely inflated)
```


example/cat.sw
```diff
@@ -11,7 +11,7 @@ let randdir : cmd dir = {
 
 forever {
   n <- random 20;
-  repeat (10 + n) wait;
+  wait (10 + n);
```
byorgey, line 14: Good catch.  We should probably add something to the CI so it runs some kind of check on every file in `example/` (but we can do that separately).

src/Swarm/Game/State.hs
```diff
@@ -354,3 +383,37 @@ emitMessage :: MonadState GameState m => Text -> m ()
...
+  time <- use ticks
+  mrns <- waitingRobots . at time <<.= Nothing
+  case mrns of
+    Nothing -> return ()
+    Just rns -> internalActiveRobots %= S.union (S.fromList rns)
+
+deleteRobot :: MonadState GameState m => Text -> m ()
+deleteRobot rn = do
+  mrobot <- robotMap . at rn <<.= Nothing
+  let robot = fromMaybe (error "expected robot to exist") mrobot
```
byorgey, line 410: As a general principle we should never, ever have calls to `error` in the codebase, even for things that "should never" happen: a bug in the code could then cause the game to crash, meaning a player would lose all their progress.  For this function in particular, I don't see any need for this anyway.  If the robot doesn't exist, can't we just do nothing?

src/Swarm/Game/Step.hs
```diff
@@ -68,26 +68,19 @@ evalStepsPerTick = 100
...
+  wakeUpRobotsDoneSleeping
+  robotNames <- use activeRobots
+  time <- use ticks
   forM_ robotNames $ \rn -> do
     mr <- uses robotMap (M.lookup rn)
     case mr of
       Nothing -> return ()
       Just curRobot -> do
-        curRobot' <- tickRobot curRobot
+        curRobot' <- tickRobot time curRobot
```
byorgey, line 79: Why do we thread the `time` as a parameter through all these functions that have access to the `GameState`?  Can't we just move the `use ticks` to the `stepCEK` function where it is actually needed?

byorgey:
Oh, also, when a robot finishes its program, we should remove it from `activeRobots`.  That way we don't have to loop over all the "dead" robots every tick.  I'll leave it up to you whether you want to do that in this PR or do it as a separate change.

byorgey:
Also also, empirically, it feels so snappy!  It no longer bogs down at all when harvesting a big forest. :tada:

src/Swarm/Game/State.hs
```diff
@@ -354,3 +383,37 @@ emitMessage :: MonadState GameState m => Text -> m ()
...
+  -- Currently the only way to delete a robot is to self-destruct, so
+  -- deleteRobot should always be called on an active robot. But we prepare
+  -- for the future.
+  -- We could blindly delete the robot from both waitingRobots and activeRobots
+  -- but deleting from waitingRobots is costly, even more so if you don't know
+  -- the key.
+  case waitingUntil robot of
+    Nothing -> internalActiveRobots %= S.delete rn
+    Just time -> waitingRobots . ix time %= filter (/= rn)
```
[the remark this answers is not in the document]
polux: I didn't think of that optim and this reason why it wouldn't work but good point :)

src/Swarm/Game/Step.hs
polux: Good point, I didn't connect the dots and realize that the time was available in the ambient monad, will fix.

src/Swarm/Game/Step.hs
```diff
@@ -413,9 +413,9 @@ takesTick c = isCmd c && (c `notElem` [Selfdestruct, Noop, Return, GetX, GetY, B
 --   trying to use a function constant without the proper capability
 --   (for example, trying to use `if` without having a conditional
 --   device).  Any other exceptions constitute a bug.
-evalConst :: (MonadState GameState m, MonadIO m) => Const -> [Value] -> Cont -> StateT Robot m CEK
-evalConst c vs k = do
-  res <- runExceptT $ execConst c vs k
+evalConst :: (MonadState GameState m, MonadIO m) => Integer -> Const -> [Value] -> Cont -> StateT Robot m CEK
```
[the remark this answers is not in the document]
polux: :+1: will fix

polux:
> Oh, also, when a robot finishes its program, we should remove it from activeRobots. That way we don't have to loop over all the "dead" robots every tick. I'll leave it up to you whether you want to do that in this PR or do it as a separate change.

Ah good point, will maybe do in another PR for readability.

src/Swarm/Game/State.hs
polux: I suppose in that case you could argue that calling deleteRobot for a robot that doesn't exist is not a bug, so I'll do what you suggested. But my question still stands: what should we do when we discover that an invariant is broken at runtime?

byorgey (approved):

byorgey:
Before we merge this, it seems like an appropriate place to discuss something I've been wondering about.  This change is going to break some user programs, because the type of `wait` has changed.  Now, in one sense, I don't care: we are definitely not making any guarantees about stability at this point, and if someone is trying out a pre-alpha game then they should not be surprised if it changes, sometimes in breaking ways, from one commit to the next.  However, it would be nice to have some consistent mechanism for communicating these kinds of changes, so that someone can easily see what major changes have happened since the last time they played, and maybe read some tips on how to port their existing code and so on.

* At this stage, we could use the `CHANGELOG` to list and describe big changes, organized not by release, but by commit.  Not every single change needs to be listed there, just bigger changes that users might care about (if they care about the nitty-gritty then can of course look at the Git history itself).
* But how would users know to look there?  One idea might be to have the game, on startup, save the hash of the HEAD commit it was compiled from; every time it starts up, if it notices the commit has changed, it could pop up a message notifying the user of any changes since the previous commit (or just tell them to read the CHANGELOG).  This is related to #66 .
* The above idea would be some work for someone to implement, though.  In the meantime, is there anything even simpler we could do?

polux:
It seems to me there are two different scenarios in your question:
  1. I want to load a saved game / re-use definitions from an old version of the game
  2. I want to be made aware of new language features.

Regarding 1: Theoretically in that particular case, we could have kept `wait` around as an alias and introduced a new `waitNTicks`. It's not worth paying the price of a worse name right now but my point is that in many cases we may get away with being backwards compatible. We could also imagine refactoring existing definitions when loading a game, if there is a consistent way to do it, and assuming they are saved. Many games and languages go out of their way to be backwards compatible and I think that after swarm grows a reasonable user base it should too. Breaking backwards compatibility should be a rare enough even once we reach stability that it would warrant a blog post or something like that. But in the meantime we could fall back to #2.

Regarding 2: I think your solution of displaying the changelog when starting a new version is quite standard and it sounds like a good solution to me. I can't think of a simpler way of notifying users in the meantime.

byorgey:
> Many games and languages go out of their way to be backwards compatible and I think that after swarm grows a reasonable user base it should too.

I definitely agree with that. We can draw a line in the sand at some point, call it "1.0", and after that work a lot harder to be backwards compatible.

TristanCacqueray:
Rendering the changelog on start would be great. Then for such language change, perhaps we could try to support the old syntax in the `fixDefMissing` to rewrite legacy usage and display a deprecation notice or something.

src/Swarm/Game/State.hs
polux: :+1:

byorgey:
> perhaps we could try to support the old syntax in the `fixDefMissing`

I don't think we can do it at that level.  Just looking at syntax, if you see a `wait` command not applied to anything, there's no way to know whether it is a use of the old command, or whether e.g. they are passing the new command to a higher-order function. In general I think we would have to store a version number along with saved code so we know when we are parsing something from an older version.

example/cat.sw
polux: Created #180

Mr-Andersen:
>  However, it would be nice to have some consistent mechanism for communicating these kinds of changes

@byorgey, if you create a release for each breaking change, those who "watch" them (via the button on top of the page) would be notified via email.

PR #179, opened 2021-10-04
Add entity for reprogramming

### LINKED ISSUE #22: `program` command
`program : string -> cmd () -> cmd ()` is like `build`, except it operates on an existing robot in the same cell instead of creating a new robot.  The robot being programmed must (1) be idle and (2) have the necessary capabilities to execute the program.

What entity should provide this ability?

twitu:
Closes #22

### DIFF
--- a/data/entities.yaml
+++ b/data/entities.yaml
@@ -404,3 +404,17 @@
     knowledge to another nearby robot; for example, 'upload \"base\"'."
   properties: [portable]
   capabilities: [scan, sensefront]
+
+- name: flash memory
+  display:
+    attr: device
+    char: '§'
+  description:
+  - "With a flash memory device, robots can be reprogrammed using the
+     `reprogram` command.
+  - "Only robots that have completed executing their previous command
+     can be reprogrammed"
+  - "The robot must possess enough capabilities to run the new command
+     otherwise it will fail"
+  properties: [portable]
+  capabilities: [reprogram]


twitu:
Reprogram requires a `flash memory` entity which gives it `CReprogram` capability required to run the `reprogram` command.

`Reprogram` is very similar to `Build` in terms of implementation. There are two other considerations -

* Should both the target robot and robot being programmed be required to have certain capabilities? For e.g. the caller robot needs `flash memory` and the target robot needs a capability a receiver entity of some kind?
* Should it be possible to reprogram from a distance? Right now there is no limitation, the base can reprogram any robot however far away. I like this capability but it can be kept restricted to creative mode only. It could also unlock other possibilities like `bluetooth` or `radio` making it possible reprogram at any distance while `flash memory` allows reprogramming adjacent robots.

byorgey:
>     * Should both the target robot and robot being programmed be required to have certain capabilities?

I don't think the robot being programmed needs any capabilities.  That way no robot is ever "lost": if a robot crashes or just gets left somewhere, you can always send another robot out to reprogram it.  I suppose if a device was required for both, you could just send robot 1 out to first install a device on robot 2 and then reprogram it, but it seems redundant.

>     * Should it be possible to reprogram from a distance? Right now there is no limitation, the base can reprogram any robot however far away.

That doesn't seem very consistent with other things in the game.  I think you should need to be in the same cell, or perhaps up to a distance of 1 cell away.  Yes, perhaps other devices later will allow reprogramming at a greater distance.  And being able to reprogram at any distance in creative mode makes sense.

src/Swarm/Game/Step.hs
```diff
@@ -803,6 +803,64 @@ execConst c vs k = do
...
+        otherRobot <-
+          robotNamed otherRobotName
+            >>= (`isJustOr` cmdExn Reprogram ["There is no robot named", otherRobotName, "."])
+
+        -- check if robot has completed executing it's current command
+        _ <-
+          finalValue (otherRobot ^. machine)
+            `isJustOr` cmdExn
+              Reprogram
+              ["You cannot reprogram a robot that has not completed it's current command"]
```
byorgey, line 821:
```suggestion
              ["You cannot reprogram a robot that has not completed its current command"]
```

src/Swarm/Game/Step.hs
```diff
@@ -803,6 +803,64 @@ execConst c vs k = do
...
+
+        -- check if robot has all devices to execute new command
+        (mode == Creative || S.null missingDevices)
+          `holdsOr` cmdExn
+            Reprogram
+            [ "the target robot does not have required devices:\n"
+            , commaList (map (^. entityName) (S.toList missingDevices))
+            ]
+
+        -- check that current robot is not trying to reprogram self
```
byorgey, line 849: This check seems unnecessary.  In fact it can never actually run, because we would have already failed the check about the target robot having completed its command.  If we want this as a separate error message, we would have to move this check *before* the check to see if the robot has completed its command.

src/Swarm/Game/Step.hs
```diff
@@ -803,6 +803,64 @@ execConst c vs k = do
...
+        let -- Standard devices that are always installed.
+            -- XXX in the future, make a way to build these and just start the base
+            -- out with a large supply of each?
+            stdDeviceList = ["treads", "grabber", "solar panel", "detonator", "scanner"]
+            stdDevices = S.fromList $ mapMaybe (`lookupEntityName` em) stdDeviceList
+
+            -- Find out what capabilities are required by the program that will
+            -- be run on the other robot, and what devices would provide those
+            -- capabilities.
+            (caps, _capCtx) = requiredCaps (snd (otherRobot ^. robotCtx)) cmd
```
byorgey, line 832: I don't think it makes sense to use the `otherRobot`'s `robotCtx` to check this, especially if we are later going to reset it to empty.  Honestly I am not quite sure what the right thing to do is, I will have to think about it a bit...

byorgey (changes requested):
This is looking good so far!  I have a number of suggestions that I think will simplify the code in `Swarm.Game.Step`.

src/Swarm/Game/Step.hs
twitu: I think we can perform this check against an empty set, if we assume that reprogram requires all devices to already be preinstalled on the target robot, and that reprogrammed robot is starting it's computation afresh.

byorgey: I'm not sure that would be correct.  The context given as an argument to `requiredCaps` is supposed to map names of definitions to the capabilities required by those definitions.  If we pass the empty context to `requiredCaps` then it might return fewer capabilities than it should, if the given command refers to some definitions in the environment.  That would lead to a situation where the `reprogram` command is successful and the reprogrammed robot starts running its program, but later crashes due to not having the right capabilities to execute something.  In that situation we would rather have it fail right away (simply as a kindness to the user).

The environment used for the reprogrammed robot (the `e` on line 862, `In cmd e [FExec]`) came from the `VDelay` value on line 807 (`VDelay _ cmd e`), which in turn came from the environment active in the parent robot when the `delay` was processed.  That means the reprogrammed robot will now know about all the definitions the parent robot knew about (which is exactly as it should be).  That means the `capCtx` we should be passing to `requiredCaps` here is the one stored in the *parent* robot's `robotCtx` (i.e. `rctx@(_, capCtx) <- use robotCtx`).

This also means the `robotCtx` of the reprogrammed robot should be set to `rctx`, the same context as its parent (in case it, in turn, reprograms another robot, etc.).

It probably doesn't matter as much what the `robotEnv` is set to, since that is used more for the `base` and its ability to make new definitions, but on principle we might as well just set it to the same thing as its parent robot.

byorgey: Probably this also means `robotEnv` is a bad name!  I'm open to better suggestions.

src/Swarm/Game/Step.hs
```diff
@@ -864,10 +864,13 @@ execConst c vs k = do
...
-        -- update other robot's CEK machine and environment
-        lift . lift $ robotMap . at otherRobotName . _Just . machine .= In cmd e [FExec]
-        lift . lift $ robotMap . at otherRobotName . _Just . robotEnv .= empty
-        lift . lift $ robotMap . at otherRobotName . _Just . robotCtx .= (empty, empty)
+        -- update other robot's CEK machine, environment and context
+        -- the childRobot inherits the parent robot's environment
+        -- and context which collectively mean all the variables
+        -- declared in the parent robot
+        lift . lift $ robotMap . at childRobotName . _Just . machine .= In cmd e [FExec]
+        lift . lift $ robotMap . at childRobotName . _Just . robotEnv .= e
```
byorgey, line 872: This should probably not be `e`, but rather should be literally whatever is stored in the `robotEnv` of the parent robot.  The difference is that the current environment `e` might have additional local bindings beyond what is in the `robotEnv`.  Like if the base has defined a command `c` but then this `reprogram` command was run like `let x = 3 in reprogram ... ` Now the environment `e` will contain both `c` and `x`.  However in this case the definition of `x` is supposed to be local to this computation, we wouldn't want to keep it around for later programs run on the same robot.

As discussed elsewhere, it ultimately doesn't really matter right now, but we might as well do "the most sensible thing" in case it turns out to matter with changes we make later!

byorgey (approved):
This is looking great!  And from my simple tests it seems to work as expected.

PR #181, opened 2021-10-04
REPL ignores Enter while working

### LINKED ISSUE #139: Don't cancel current base computation when entering another one
**Describe the bug**
When you hit Enter at the REPL, there is no check for whether the base is currently running a computation or not; it simply replaces any running computation with what you entered.

**To Reproduce**
This is easiest to reproduce if you crank the tick speed down to e.g. 2 or 4 seconds per tick.
* Start some kind of computation that will take the base a few ticks and have an observable effect, e.g.
    ```
    build "x" move; build "y" {turn north; move}; build "z" {turn west; move}; build "w" {turn south; move}
    ```
* While the above is running, type another command, e.g. `grab`, and hit Enter.  You will not be able to see what you type, since the REPL will be displaying `...`.
* The original program will not run to completion, but will be replaced by the new program you typed.

**Expected behavior**
I don't think it's a problem that you can still type even while the REPL is displaying `...`; in fact, it's kind of nice since you can already get started typing the next command without having to wait for the previous one to complete.  However, hitting Enter while the base is currently in the middle of a computation should do nothing.  You should be required to Ctrl+C if you want to interrupt a running computation.

Fixing this should be as easy as adding a condition in the REPL event handler---when an Enter key event is received, we must first check the `REPLStatus` to see whether the base is in the middle of a computation.  If so, simply ignore the Enter key event.

Alexander-Block:
When you hit Enter at the REPL the REPL event handler now checks whether the base is currently running a computation. If this is the case, then the Enter key event is ignored. 

Fixes #139

### DIFF
--- a/src/Swarm/TUI/Controller.hs
+++ b/src/Swarm/TUI/Controller.hs
@@ -339,20 +339,21 @@ handleREPLEvent s (VtyEvent (V.EvKey (V.KChar 'c') [V.MCtrl])) =
     s
       & gameState . robotMap . ix "base" . machine .~ idleMachine
 handleREPLEvent s (VtyEvent (V.EvKey V.KEnter [])) =
-  case processTerm' topCtx topCapCtx entry of
-    Right t@(ProcessedTerm _ (Module ty _) _ _) ->
-      continue $
-        s
-          & uiState . uiReplForm %~ updateFormState ""
-          & uiState . uiReplType .~ Nothing
-          & uiState . uiReplHistory %~ (REPLEntry True entry :)
-          & uiState . uiReplHistIdx .~ (-1)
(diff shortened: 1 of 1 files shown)


byorgey (approved):
Looks great, thanks!

Technically, in the case that we ignore the Enter key press, we could use `continueWithoutRedraw` instead of `continue`, since nothing has changed and the screen does not need to be redrawn. However, that would be an extremely minor optimization and probably not even worth the extra code. I only mention it for future reference. =)

Alexander-Block:
@byorgey Thanks for the hint about the use of ``continueWithoutRedraw``. I am not really familiar with the brick library yet and will keep that in mind.

PR #182, opened 2021-10-04
Abstract Const commands out of reserved words list

7h3w1zz:
Now that `constInfo` provides the name of Const values, there is no
need to re-list each command in the reserved words list.  Also
represent reserved words as Text instead of String.

### DIFF
--- a/src/Swarm/Language/Parse.hs
+++ b/src/Swarm/Language/Parse.hs
@@ -36,10 +36,9 @@ module Swarm.Language.Parse (
 
 import Control.Monad.Reader
 import Data.Bifunctor
-import Data.Char
 import qualified Data.List.NonEmpty (head)
 import Data.Maybe (fromMaybe, mapMaybe)
-import Data.Text (Text, index)
+import Data.Text (Text, index, toLower)
 import Data.Void
 import Witch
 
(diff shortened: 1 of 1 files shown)


7h3w1zz:
Talked about in #138

byorgey:
Looks great.  While we're at it, can we also autogenerate the list of direction names, via `map show [minBound .. maxBound :: Direction]` or something like that?

I wonder if there is a way we could also autogenerate the names of types like `bool`, `int`, and so on...

7h3w1zz:
The directions in the reserved list could be abstracted in a similar way.  Either by using the pretty print code or by creating another function similar to `constInfo`.  The other term-level constructs (`def`, `let ... in ...`, etc.) would be more difficult to do.

byorgey:
Oh, right, we can't use `show` for `Direction` since the names of the constructors don't match the syntax.  Huh, maybe it does make sense to make a function similar to `constInfo`.  It could contain the syntax and also the info needed for `toDirection`, `fromDirection`, and `applyTurn`.  Then we could rewrite the parser and pretty printer to make use of that info, so if we ever add any new `Direction` values we only have to add them in one place.  But I think this is getting to be a bigger change than I thought.  I'll just merge this as-is, and make a separate issue for the `Direction` stuff.

byorgey (approved):

xsebek:
Thanks, @7h3w1zz, for finishing this after me. :+1:  (Sorry that I left it there, some TH error made me think it's untouchable.)

PR #185, opened 2021-10-05
Clear the error widget on successful REPL entry

### LINKED ISSUE #76: Clear the error widget on successfull REPL entry
When an error is displayed, it seems like the only way to remove from the screen it is by pressing `Esc`.

When a new REPL entry is validated, any previous error should be removed from the screen.

Alexander-Block:
This pull request combines two changes:

1. When the user hits "Enter" in the REPL and the entry is validated, any previous error is removed from the screen.
2. When the user hits "Enter" while the base is actively running a computation, this is ignored now without triggering a redraw. This change is based on the review comment https://github.com/byorgey/swarm/pull/181#pullrequestreview-770727562 . While using ``continueWithoutRedraw`` for this case is more verbose than factoring out ``continue`` as I did before and the performance impact is most certainly negligible, I think that this solution is semantically cleaner. Of course I am happy to roll back this change if there is a disagreement about this point. 

Closes #76

### DIFF
--- a/src/Swarm/TUI/Controller.hs
+++ b/src/Swarm/TUI/Controller.hs
@@ -339,21 +339,23 @@ handleREPLEvent s (VtyEvent (V.EvKey (V.KChar 'c') [V.MCtrl])) =
     s
       & gameState . robotMap . ix "base" . machine .~ idleMachine
 handleREPLEvent s (VtyEvent (V.EvKey V.KEnter [])) =
-  continue $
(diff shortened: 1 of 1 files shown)


twitu:
I tried checked this out, it makes the user experience much better, especially for small screens.

byorgey (approved):

byorgey:
Thanks!

PR #186, opened 2021-10-05
Add drill command

### LINKED ISSUE #18: Copper mines
Right now there is very limited and non-renewable copper available on the surface, but it's going to be an important ingredient in many recipes.  The description for copper says you can get more by drilling, but that needs to be implemented.  My ideas are:

* Create a new kind of terrain, "copper seam", which looks the same as stone.
* Add a `drill` command, with a corresponding capability provided by a drill entity.  At first `drill` won't take any parameters, though perhaps later it will take a direction as an argument (to allow for drilling through boulders and mountains etc.)
* Executing `drill` on a copper seam cell has a small percentage chance of adding copper to the robot's inventory.

### LINKED ISSUE #143: Add a `drill` command

xsebek:
- adds drill capability (from drill entity) (Closes #143)
- adds drilling recipes
- adds mountain tunnel entity
- adds copper vein entity (not portable) (Closes #18)


xsebek:
At the moment, `drill` throws on any problem - this is currently easily solved with `try`, but that should eventually be harder to obtain, so it probably does not matter. :man_shrugging: I was thinking about drowning the drill, but that would be unprecedented :hand_over_mouth:

xsebek:
@byorgey how would I make the mountain show on top? I could not figure it out :smile: 

Screenshot from 2021-10-05 21-46-25

src/Swarm/TUI/Attr.hs
```diff
@@ -36,6 +36,7 @@ swarmAttrMap =
     , (woodAttr, fg (V.rgbColor @Int 139 69 19))
     , (flowerAttr, fg (V.rgbColor @Int 200 0 200))
     , (copperAttr, fg V.yellow)
+    , (copperAttr, fg V.brightYellow)
```
byorgey, line 39: Is one of these supposed to be `copperAttr'`?

data/recipes.yaml
```diff
@@ -48,6 +48,29 @@
   out:
   - [1, drill]
 
+- in:
```
byorgey, line 51: Oh, I think I get it --- the idea is that even though you can never actually have a `boulder`, `mountain`, or `copper vein` in your inventory, once you `scan` them you can see the recipe which hints that you should use a `drill`?  I like that a lot.

data/entities.yaml
```diff
@@ -66,11 +66,19 @@
 - name: mountain
   display:
     attr: snow
-    char: '∧'
+    char: '▲'
```
byorgey, line 69: I am not convinced whether these characters will be supported in all terminal fonts.  Perhaps we should come up with an official policy re: what subset of Unicode the game uses, and perhaps pair that with some recommendations of what terminal font(s) will work well.

src/Swarm/Game/Step.hs
```diff
@@ -671,6 +672,53 @@ execConst c vs k = do
...
+        loc <- use robotLocation
+        rDir <- use robotOrientation
+
+        let nextLoc = loc ^+^ applyTurn d (rDir ? V2 0 0)
+        em <- lift . lift $ use entityMap
+        drill <- lookupEntityName "drill" em `isJustOr` cmdExn Drill ["Drill does not exist?!"]
+        nextE <- entityAt nextLoc >>= (`isJustOr` cmdExn Drill ["There is nothing to drill", "in the direction", "of robot", rname <> "."])
+        -- TODO: this would lose the drill when trying to drill water
+        --       but I guess it would not work well with try which
+        --       might eventually be provided by some time-machine
```
byorgey, line 688: Agreed.  It would not break anything, but it does seem a bit weird to have an error be partly recoverable but also have irrecoverable consequences.

src/Swarm/Game/Step.hs
```diff
@@ -671,6 +672,53 @@ execConst c vs k = do
...
+
+        inRs <- lift . lift $ use recipesIn
+
+        let recipes = filter drilling (recipesFor inRs nextE)
+            drilling = any ((== drill) . snd) . view recipeRequirements
+
+        not (null recipes) `holdsOr` cmdExn Drill ["There is no way to drill a", nextE ^. entityName <> "."]
+
+        -- add and remove drill as it is requiered in recipe - the drilled entity is replaced at the end
+        let inv' = insert drill $ insert nextE inv
```
byorgey, line 704: It is a bit weird to add a drill to the inventory just to satisfy the recipe.  How about we modify the code for running a recipe so that it looks for the `recipeRequirements` in either the inventory *or* the `installedDevices`?

src/Swarm/Game/Step.hs
```diff
@@ -671,6 +672,53 @@ execConst c vs k = do
...
+            `isJustOr` cmdExn Drill ["You don't have the ingredients\n to drill", indefinite (nextE ^. entityName) <> "."]
+
+        let (out, down) = L.partition ((`hasProperty` Portable) . snd) outs
+            downAmount = sum $ map fst down
+
+        (downAmount <= 1) `holdsOr` cmdExn Drill ["Bad recipe - can not place more then one unmovable entity."]
+
+        robotInventory .= delete drill (L.foldl' (flip (uncurry insertCount)) invTaken out)
+        _ <-
+          if downAmount == 0
```
byorgey, line 716: It would make me feel better to do a `case down` instead --- that way we don't have to use `head`.

byorgey (changes requested):

byorgey:
> EDIT: one thing I thought about is that this gives unlimited copper without any waiting

We had also talked about drilling producing copper only with some small percentage, or possibly having the vein disappear with some small percentage.  Having it take more than one tick to drill could also make sense.

data/entities.yaml
```diff
@@ -66,11 +66,19 @@
...
   description:
   - A mountain.
   properties: [unwalkable]
 
+- name: mountain tunnel
+  display:
+    attr: snow
+    char: '△'
```
byorgey, line 77:
```suggestion
    char: '△'
    priority: 11
```

src/Swarm/Game/Step.hs
xsebek: Right, now that I think about it, that could help with the furnace too. :+1:

data/recipes.yaml
xsebek: Yes, I think I will add a note to the YAML explaining this assumption :slightly_smiling_face: The recipes do look nice :heart_eyes: 
Screenshot from 2021-10-05 23-18-32

data/entities.yaml
xsebek: Good point, the current chat is a bit confusing with robots, so I would personally prefer pointing people in the direction of better fonts. Though we could add a fallback `--ascii` option for DOS users that would pick `ascii: ^` in the YAML.

byorgey: Yeah, I originally had filled triangle chars for robots, but then I saw someone try it and those chars were double wide in their font =(

data/entities.yaml
xsebek: I added a check in `drawLoc`, but it does not seem to apply - something is not right yet :confused:

data/recipes.yaml
xsebek: Actually, we are missing a map from requirements to recipes, so `drill` and `furnace` do not have recipes with them shown. I am thinking about leaving that improvement for later Issue/PR :sweat_smile:

data/entities.yaml
xsebek: I do not have the energy and time to do it right, so I found the characters `A` and `Å` instead. :mountain:

src/Swarm/Game/Step.hs
```diff
@@ -704,24 +699,22 @@ execConst c vs k = do
...
 
-        robotInventory .= delete drill (L.foldl' (flip (uncurry insertCount)) invTaken out)
-        _ <-
-          if downAmount == 0
-            then updateEntityAt nextLoc (const Nothing)
-            else updateEntityAt nextLoc (const (Just . snd . head $ down))
+        robotInventory .= L.foldl' (flip (uncurry insertCount)) invTaken out
         flagRedraw
-        return $ Out VUnit k
+        finishCookingRecipe recipe
```
xsebek, line 717: Recipes can now take time (using `Waiting`), though they do change the world first.
Not sure if that weird effect can be avoided, but at least it can not be cancelled with `Ctrl+c` :smile:

src/Swarm/TUI/View.hs
```diff
@@ -382,8 +388,10 @@ drawRecipe e inv (Recipe ins outs reqs) =
         , joinableBorder (Edges True False True True)
         , hLimit 2 hBorder
         ]
-  inLen = length ins
+  inLen = length ins + length times
   outLen = length outs
+  times = [(fromIntegral time, timeE) | time /= 1]
+  timeE = mkEntity (defaultEntityDisplay '.') "tick" [] []
```
xsebek, line 394: I wanted to add time to the recipe, so I created an ad-hoc entity.

It would be funnier to use a _real entity_, say "grain of time", which would actually be obtainable in the end game and used to build the time machine for `throw`. :hourglass_flowing_sand:

src/Swarm/Game/Step.hs
xsebek: I see you noticed too @byorgey :smile: maybe it is not a feature after all :bug: 
- #194

byorgey: Oh! I noticed independently.  No, I do not think it is a feature.  I don't see why I shouldn't be allowed to change my mind about waiting. =)

byorgey: I definitely think all the effects need to happen *after* the waiting period finishes, otherwise it seems weird (if the recipe already worked right away, why do we still need to wait?) and open to glitches that circumvent the waiting period.

I think the way to do this is to add a new kind of continuation stack frame.  The continuation stack says "what to do next" once you are done evaluating the current thing; in this case we want to `wait` while remembering that after done waiting, we need to make some changes to the game state.
* Create a new constructor for the `Frame` type in `Swarm.Game.CEK`, e.g. `FFinishRecipe`.  Shove whatever information is needed into it to actually carry out the effects of a recipe (e.g. I guess we need some things to take away from the robot's inventory and some things to add, and maybe also something about how to update the world?).
* Add a case to the `stepCEK` function: if we see `Out v (FFinishRecipe blah blah : k)`, then do the stuff (insert and delete from the inventory, modify the world, etc.) and finally return `Out v k`.
* Now in `finishCookingRecipe`, return something like `Out VUnit (FFinishRecipe blah blah : k)` instead of `Out VUnit k`.

src/Swarm/TUI/View.hs
```diff
@@ -283,17 +283,23 @@ drawWorld g =
...
-      withAttr (r ^. robotDisplay . displayAttr) $
-        str [lookupDisplay ((r ^. robotOrientation) >>= toDirection) (r ^. robotDisplay)]
-    Nothing -> drawCell coords (g ^. world)
+
+  drawLoc :: W.Coords -> Widget Name
+  drawLoc coords =
+    let (ePrio, eWidget) = drawCell coords (g ^. world)
+     in case M.lookup (W.coordsToLoc coords) robotsByLoc of
+          Just r
+            | ePrio > 10 -> eWidget
```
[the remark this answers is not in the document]
byorgey: Actually I think that will fix a current bug where if you `place` an entity on top of a seed robot, the seed robot gets drawn on top of the entity.

src/Swarm/Game/Recipe.hs
```diff
@@ -78,6 +81,9 @@ recipeInputs :: Lens' (Recipe e) (IngredientList e)
 -- | The outputs from a recipe.
 recipeOutputs :: Lens' (Recipe e) (IngredientList e)
 
+-- | The outputs from a recipe.
```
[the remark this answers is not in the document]
xsebek: Fixed, good catch :+1:

src/Swarm/Game/Step.hs
```diff
@@ -913,6 +954,10 @@ execConst c vs k = do
           [ "Bad application of execConst:"
           , from (prettyCEK (Out (VCApp c (reverse vs)) k))
           ]
+  finishCookingRecipe r = do
+    let recTime = pred $ r ^. recipeTime
```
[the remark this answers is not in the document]
xsebek: Well I am certainly not against it, as now I can just always return waiting. :slightly_smiling_face:

src/Swarm/TUI/View.hs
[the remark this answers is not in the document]
xsebek: It did (with one more little change) and now it all works nicely. :heart_eyes:

src/Swarm/TUI/View.hs
xsebek: It does not look that great in requirements:

Screenshot from 2021-10-09 00-56-06

It makes some sense for it to be in inputs (it *takes time* to finish) so I will just make it look a little nicer and be done with it.

src/Swarm/Game/Step.hs
```diff
@@ -697,30 +713,68 @@ execConst c vs k = do
...
+          listToMaybe (rights (map makeRecipe recipes))
+            `isJustOr` cmdExn Drill ["You don't have the ingredients\n to drill", indefinite (nextE ^. entityName) <> "."]
+
+        let (out, down) = L.partition ((`hasProperty` Portable) . snd) outs
+
+        let changeWorld = case down of {
+            [] -> W.update (W.locToCoords nextLoc) (const Nothing);
+            [de] -> W.update (W.locToCoords nextLoc) (const (Just $ snd de));
+            _ -> id -- throwError $ cmdExn Drill ["Bad recipe - can not place more then one unmovable entity."]
+          } :: W.World Int Entity -> W.World Int Entity
```
xsebek, line 774: Fun fact, `case x of {...\n...}` breaks the formatter. :laughing:

data/entities.yaml
```diff
@@ -60,17 +60,26 @@
     attr: rock
     char: '@'
   description:
-  - A rock.
+  - A rock. Can be drilled to smaller rocks relatively fast.
```
byorgey, line 63: I'd like to give a stronger hint that you can use a drill to get this out of the way.  Also, the description `A rock` should have been changed before when I changed the name.
```suggestion
  - A very large, impassable rock. A drill can be used to get rid of it and turn it into smaller rocks.
```

src/Swarm/Game/CEK.hs
```diff
@@ -253,3 +262,31 @@ prettyFrame FLoadEnv {} = "loadEnv"
...
+--
+-- NOTE: we can not use GameState and Robot directly, as it
+-- would create a cyclic dependency. The alternative is
+-- making CEK, Cont and Frame polymorphic which just muddies
+-- the picture too much for one little game feature.
+--
+-- BEWARE: the types do not follow normal laws for Show and Eq
+--------------------------------------------------------------
+
+newtype WorldFun = WorldFun
```
byorgey, line 278: I'm pretty sure we already have something called `WorldFun` in `Swarm.Game.World`.  Perhaps we should call this something else.  `WorldUpdate`, perhaps?  We could also change the name of `RobotFun` to match.

src/Swarm/Game/Step.hs
```diff
@@ -271,6 +277,15 @@ stepCEK cek = case cek of
     if wakeupTime == time
       then stepCEK cek'
       else return cek
+  Out v (FImmediate wf rf : c) -> do
```
byorgey, line 280: Let's stay consistent with the rest of the code and use `k` for the continuation instead of `c`.  It might seem like a bit of a weird convention if you've never seen it before, but it is extremely common to represent continuations with the variable `k` (perhaps because the word for it starts with k in German?).  And `c` is used for lots of other things already.

src/Swarm/Game/Step.hs
```diff
@@ -697,30 +716,63 @@ execConst c vs k = do
...
-        inv' <-
-          listToMaybe (rights (map (make inv) recipes))
+        let makeRecipe r = (,r) <$> make (inv, ins) r
+        (inv', recipe) <-
+          listToMaybe (rights (map makeRecipe recipes))
             `isJustOr` cmdExn Make ["You don't have the ingredients to make", indefinite name, "."]
 
-        robotInventory .= inv'
-        return $ Out VUnit k
+        finishCookingRecipe recipe (WorldFun Right) (RobotFun $ const inv')
```
byorgey, line 743: I don't think `const inv'` is right.  What if another robot `give`s this robot something while it is waiting for its recipe to complete?  When the recipe completes the inventory will be replaced with `inv'` and we will lose the thing that the other robot gave.  Instead I think we just have to remember some things to delete and some things to add.

src/Swarm/Game/Step.hs
```diff
@@ -697,30 +716,63 @@ execConst c vs k = do
...
+        let makeRecipe r = (\(i, o) -> (i, o, r)) <$> make' (insert nextE inv, ins) r
+        (invTaken, outs, recipe) <-
+          listToMaybe (rights (map makeRecipe recipes))
+            `isJustOr` cmdExn Drill ["You don't have the ingredients\n to drill", indefinite (nextE ^. entityName) <> "."]
+
+        let (out, down) = L.partition ((`hasProperty` Portable) . snd) outs
+            inv' = L.foldl' (flip (uncurry insertCount)) invTaken out
+            changeWorld = changeWorld' nextE nextLoc down
+
+        finishCookingRecipe recipe (WorldFun changeWorld) (RobotFun $ const inv')
```
byorgey, line 775: Same goes here for `const inv'`.

data/entities.yaml
```diff
@@ -81,6 +90,14 @@
     be found by drilling beneath the surface.
   properties: [portable]
 
+- name: copper vein
```
byorgey, line 93: I was kind of imagining that the copper and iron veins would be *inside* the mountain tunnels, i.e. first you have to drill into the mountains and once you're in there you can find some veins, and it's not visually obvious where you have to drill to find the copper.

One idea would be something like this:
1. Change `copper vein` to look the same as a `mountain` and be `unwalkable`.  In other words a `copper vein` is now a special kind of mountain which has copper inside.
2. `drill`ing a `copper vein` produces a `copper mine`, which looks the same as a `mountain tunnel` but can be drilled to get copper.

What do you think?

src/Swarm/Game/Step.hs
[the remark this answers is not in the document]
byorgey: I still think `pred` is wrong, it should just be `recTime = r ^. recipeTime`.  For example, suppose `r ^. recipeTime = 1`.  Then we want to do `Waiting (time + 1)`, not `Waiting (time + 0)`, since the first one means we should resume on the next tick, whereas the second one does not even make sense, it means we want to wait and continue running once the time reaches... now.

data/entities.yaml
```diff
@@ -81,6 +90,14 @@
     be found by drilling beneath the surface.
```
byorgey, line 90: I guess this comment about finding additional copper beneath the surface is not true, right?  We should change it to say that you can find additional copper by drilling beneath the mountains.

src/Swarm/TUI/View.hs
```diff
@@ -499,13 +506,18 @@ drawRecipe e inv (Recipe ins outs reqs) =
...
     | missing = withAttr invalidFormInputAttr $ txtLines nm
     | otherwise = txtLines nm
    where
     -- Split up multi-word names, one line per word
     nm = ingr ^. entityName
     txtLines = vBox . map txt . T.words
 
+-- | Ad-hoc entity to represent time - only used in recipe drawing
+timeE :: Entity
+timeE = mkEntity (defaultEntityDisplay '.') "time piece" [] []
```
byorgey, line 519: `time piece` sounds very weird to me.  Can we just call it `ticks`?

byorgey: I still don't really like this way of drawing time as an input, but we don't need to solve it now, we can keep playing with better layouts later.

byorgey (approved):
I made a bunch more suggestions, but I don't want to drag this out unnecessarily.  At this point I think it is pretty good and it's really just details that are left.  We can keep polishing it if you want, or we can just merge it so everyone can start playing with it, and continue polishing the details later.

byorgey:
Also, you mentioned that it seems too easy to get a lot of copper.  What about adding a *success probability* field to `Recipe` (which is 1 by default, of course)? If you attempt to make a recipe, then it always takes the specified amount of time, but at the end of the time, we randomly determine whether it was successful or not, according to the specified probability.  If it was unsuccessful then nothing in your inventory or the world changes.  Then we could say that drilling for copper has a low probability of success.  In the long run, "copper takes 420 ticks to mine" and "copper takes 42 ticks to mine, with a 10% success rate" are mathematically equivalent, but the latter seems more fun.

Maybe you could build fancier drills later that give you a success modifier, e.g. that add 5 or 10% to the success rate.

src/Swarm/Game/CEK.hs
xsebek: Somehow I forgot about `WorldFun`, now it is fixed.

src/Swarm/Game/Step.hs
xsebek: Ich habe es geändert.

data/entities.yaml
xsebek: Fixed.

src/Swarm/Game/Step.hs
byorgey: Indeed, we should probably make this an official invariant: stuff cannot *leave* an active robot's inventory unless under its explicit control.

xsebek:
Thanks @byorgey for helping me finish this :slightly_smiling_face:

PR #187, opened 2021-10-06
More benchmarks

polux:
Improve the benchmark and add more scenarios.

Improvement are:
- Run each benchmark for 10 seconds to get enough samples in more costly scenarios
- Run benchmark on a blank terrain to avoid crashing into mountain and boulders
- Take advantage of the seed in tests using randomness

New scenarios are:
- A bunch of robots doing nothing. Will be useful for measuring the impact of removing idle robots from the active robots set.
- A bunch of robots moving in circles.
- A bunch of robots moving forward. This surprisingly is slower than moving in circles.

### DIFF
--- a/bench/Benchmark.hs
+++ b/bench/Benchmark.hs
@@ -3,21 +3,28 @@
 
 module Main where
 
+import Control.Lens ((&), (.~))
 import Control.Monad (replicateM_)
 import Control.Monad.Except (runExceptT)
 import Control.Monad.State (evalStateT, execStateT)
-import Criterion.Main (bench, bgroup, defaultMain, whnfIO)
+import Criterion.Main (Benchmark, bench, bgroup, defaultConfig, defaultMainWith, whnfAppIO)
+import Criterion.Types (Config (timeLimit))
+import Data.Int (Int64)
 import Linear.V2 (V2 (V2))
 import Swarm.Game.CEK (initMachine)
 import Swarm.Game.Robot (Robot, mkRobot)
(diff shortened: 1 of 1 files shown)


polux:
@byorgey explained on IRC why movers are seemingly slower than circlers:

> 17:28 byorgey polux: I think it might be because the movers are doing more CEK machine steps per tick than the circlers.
> 17:28 byorgey polux: much of the time, the circlers just have to advance to the next command in a sequence, whereas the movers have to reduce a recursive function application every tick
> [...]
> 17:32 byorgey each tick will do as many CEK machine steps as needed to get around to executing an "external command" (like a move or turn), OR a fixed maximum number of CEK machine steps, whichever comes first
> 17:33 byorgey the intention is that robots will mostly stay "in sync" without having to think a lot about the exact number of CEK machine steps needed, but to also ensure that a robot stuck in an infinite loop can't freeze the UI

xsebek:
Fantastic, thanks @polux! Could we maybe add an Action to run benchmarks on every commit in main and aggregate the results somewhere?

byorgey:
> Could we maybe add an Action to run benchmarks on every commit in main and aggregate the results somewhere?

That sounds valuable indeed, but it can probably go in a separate PR.

polux:
> Well, yes, once they reach a tile that hasn't been loaded into the tile cache, the whole thing is loaded. I don't know if that's what the difference is.

That seems like a lot given the simplicity of the terrain function here, but maybe the update of the map is what is costly. I'll try to profile it.

PR #198, opened 2021-10-08
Use type family for creating robot context

### LINKED ISSUE #190: Refactor to have a single `Context` type
I just spent 15 minutes writing a detailed description for this issue and then GitHub ate it. :sob: I will fill it in again later when I am no longer upset.  In the meantime, here is a link to the comment that triggered this issue: https://github.com/byorgey/swarm/pull/179#discussion_r723638726

twitu:
Closes #190

### DIFF
--- a/src/Swarm/Game/Context.hs
+++ b/src/Swarm/Game/Context.hs
@@ -1,0 +1,62 @@
+-----------------------------------------------------------------------------
+-----------------------------------------------------------------------------
+{-# LANGUAGE DataKinds #-}
+{-# LANGUAGE DeriveDataTypeable #-}
+{-# LANGUAGE DeriveFoldable #-}
+{-# LANGUAGE DeriveFunctor #-}
+{-# LANGUAGE DeriveTraversable #-}
+{-# LANGUAGE TypeFamilies #-}
+
+module Swarm.Game.Context where
+
+import Control.Applicative
+import Data.Data
+import Data.Functor.Identity
+import Data.Map
+import Data.Set
+import Swarm.Game.Value
+import Swarm.Language.Capability
+import Swarm.Language.Types
+
+-- | A context is a mapping from variable names to things.
+newtype Ctx t = Ctx {unCtx :: Map Var t}
+  deriving (Eq, Show, Functor, Foldable, Traversable, Data)
+
+-- | Different phases of the program
+data Phase
+  = -- | Command parsing
+    ParseCommand
+  | -- | Checking capabilities for a definition
+    CapabilityCheck
+  | -- | Evaluating a Term
+    EvaluateTerm
+  deriving (Eq, Ord, Show)
+
+-- A context is a mapping from variable name to information
+-- about the variable. There can be multiple "things" that
+-- independently store information about the variable
+-- they are bundled together in the record
+newtype Context (p :: Phase) = Context (Map Var (VarCtx p))
+data VarCtx p = VarCtx
+  { -- Set of capaibilities required to compute a variable
+    varCaps :: CapsForPhase p (Set Capability)
+  , -- Type of the value the variable stores
+    varType :: TypeForPhase p Polytype
+  , -- Value of the variable
+    varVal :: ValForPhase p Value
+  }
+
+type family CapsForPhase (p :: Phase) :: * -> * where
+  CapsForPhase 'ParseCommand = Maybe
+  CapsForPhase 'CapabilityCheck = Maybe
+  CapsForPhase 'EvaluateTerm = Const ()
+
+type family TypeForPhase (p :: Phase) :: * -> * where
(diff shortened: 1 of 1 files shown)


src/Swarm/Game/Robot.hs
```diff
@@ -74,8 +75,10 @@ data Robot = Robot
   , _systemRobot :: Bool
   , _selfDestruct :: Bool
   , _tickSteps :: Int
+  , _robotContext :: VarContextStore
```
byorgey, line 78: Eventually, the idea is that this replaces the `robotCtx` and `robotEnv`.

byorgey:
This is great work so far, but I am starting to wonder whether all this stuff with type families etc. is more complex than we need.

After taking another quick look through the codebase, I *think* there are only two places where we need to carry around multiple types of context that go together.

- The base robot needs all three context types: it needs to remember the type, capabilities, and value of each name that has been defined.
- When the user types something at the REPL, it is passed to `processTerm'` or something similar; the base peels off just the type and capability contexts and passes them together to `processTerm'` (it does not need to know the values of definitions).

In all other cases we have just one context.

- During type inference we have a context mapping variable names to their types. We never need to know anything about capabilities or values.
- During capability checking we have a context mapping variable names to capability sets.  We don't need to know about types (we assume everything type checks) nor about values.
- During evaluation of the CEK machine, we have an environment, i.e. a context mapping variable names to their values.  We don't need to know anything about types at runtime, because once everything type checks the actual runtime behavior does not depend on knowing anything about types.  We do actually need to know about capabilities when encountering a `Build` command, but we don't need to keep track of them along with the environment, we simply look up the capability context from the `robotCtx`.

So I wonder if there's a good way to keep all the code for type checking, capability checking, and the CEK machine runtime the same, but just build something that makes it easier to carry around multiple contexts together.

byorgey:
> Next is the part about combining all three contexts into a single data type to pass around. The challenge here is that the three contexts are used in very different places, so the single data type will often have to be split up. This makes it difficult to use a single variable everywhere.

Yes, I agree. When I said "build something that makes it easier to carry around multiple contexts together" I had in mind just a data type that holds multiple contexts, i.e. multiple maps.  This doesn't give us any guarantee that they have the same keys, but it just makes it a little easier to pass them around together or remember to update them all appropriately.  In fact I think simply creating such a data type and putting one in a field in `Robot` (replacing the `robotCtx` and `robotEnv`) would already be a nice improvement over the status quo.

src/Swarm/Game/Robot.hs
```diff
@@ -52,14 +52,26 @@ import Linear
...
 import Swarm.Game.Display
 import Swarm.Game.Entity hiding (empty)
 import Swarm.Game.Value as V
 import Swarm.Language.Capability
 import Swarm.Language.Context
 import Swarm.Language.Types (TCtx)
 
+-- | 'RobotContext' is a record that stores all the information
+--   related to all the variables defined for a 'Robot'.
+data RobotContext = RobotContext
```
byorgey, line 64: Yes, this looks like the kind of thing I had in mind in my most recent comment.

twitu:
The lens library is absolutely fascinating and it's been a great fun on learning how to use them.

I have made the changes required to have a single record holding all the information for stored definitions. I have changed the names of the fields to reflect how I think they're being used. I have also made the updates where the previous fields were being used.

Please suggest better names or more idiomatic usage of lenses, as you see fit.

src/Swarm/TUI/Controller.hs
```diff
@@ -390,7 +390,7 @@ handleREPLEvent s (VtyEvent (V.EvKey (V.KChar 'c') [V.MCtrl])) =
       & gameState . robotMap . ix "base" . machine .~ idleMachine
 handleREPLEvent s (VtyEvent (V.EvKey V.KEnter [])) =
   if not $ s ^. gameState . replWorking
-    then case processTerm' topCtx topCapCtx entry of
+    then case processTerm' topTypeCtx topCapCtx entry of
```
byorgey, line 393: What do you think about changing `processTerm'` to just take a `RobotCtx` instead of separate type and cap contexts?  It's true that it doesn't really need the `valCtx` but I don't think it would hurt anything.  It means this code would be simpler because we wouldn't have to project out all the individual contexts in order to pass them to `processTerm'`.

twitu: I think that we should keep `processTerm'` the same., firstly because projecting the two individual contexts is not really that bad.

But secondly and more importantly is that while `RobotContext` fields' types are the same as the arguments to `processTerm'`, I think they mean different things. `RobotContext` is stores information of all the definitions for the duration of the lifetime of a robot but `processTerm'` is more general and can start with any context. I feel that the current types make this more evident.

PR #200, opened 2021-10-08
Remove idle robots from the active robots set

### LINKED ISSUE #19: Active robot queue
This is an idea to make the interpreter a bit more efficient, especially if there are lots of "dead" robots lying around.

* In addition to the `robotMap`, add `activeRobots :: [Text]` to the `GameState`, representing the names of robots which are actively running code.
* The `addRobot` function should add a robot to both the `robotMap` and the `activeRobots` list.
* Each tick, we iterate over every robot in `activeRobots` (instead of over every robot in the `robotMap`), filter out any which are inactive, tick the rest, and rebuild the list.

This way we only spend time iterating over the robots which will actually do some computation.

polux:
Resolves #19. This is a big win for idle robots, but decreases the performance of active robots a bit.

Benchmark results on the main branch:

```
benchmarking run 1000 game ticks/idlers/40
time                 18.93 ms   (18.63 ms .. 19.34 ms)
                     0.999 R²   (0.998 R² .. 1.000 R²)
mean                 18.82 ms   (18.69 ms .. 19.00 ms)
std dev              419.6 μs   (284.8 μs .. 642.9 μs)

benchmarking run 1000 game ticks/trees/40
time                 10.27 ms   (10.21 ms .. 10.32 ms)
                     1.000 R²   (0.999 R² .. 1.000 R²)
mean                 10.16 ms   (10.12 ms .. 10.21 ms)
std dev              142.0 μs   (105.9 μs .. 211.4 μs)

benchmarking run 1000 game ticks/circlers/40
time                 260.2 ms   (257.9 ms .. 262.4 ms)
                     1.000 R²   (1.000 R² .. 1.000 R²)
mean                 257.5 ms   (256.5 ms .. 258.9 ms)
std dev              1.817 ms   (1.235 ms .. 2.674 ms)
variance introduced by outliers: 11% (moderately inflated)

benchmarking run 1000 game ticks/movers/40
time                 785.9 ms   (620.8 ms .. 987.0 ms)
                     0.992 R²   (0.975 R² .. 1.000 R²)
mean                 847.7 ms   (802.5 ms .. 892.8 ms)
std dev              57.17 ms   (30.38 ms .. 69.01 ms)
variance introduced by outliers: 20% (moderately inflated)
```

Benchmark results after removing idle robots from the active robots set:

```
benchmarking run 1000 game ticks/idlers/40
time                 888.4 μs   (877.9 μs .. 904.7 μs)
                     0.998 R²   (0.996 R² .. 1.000 R²)
mean                 870.3 μs   (865.6 μs .. 878.8 μs)
std dev              22.49 μs   (14.50 μs .. 39.52 μs)
variance introduced by outliers: 17% (moderately inflated)

benchmarking run 1000 game ticks/trees/40
time                 4.075 ms   (4.023 ms .. 4.118 ms)
                     0.999 R²   (0.999 R² .. 1.000 R²)
mean                 3.993 ms   (3.974 ms .. 4.016 ms)
std dev              68.87 μs   (55.92 μs .. 87.58 μs)

benchmarking run 1000 game ticks/circlers/40
time                 362.5 ms   (264.8 ms .. 492.7 ms)
                     0.902 R²   (0.838 R² .. 1.000 R²)
mean                 285.1 ms   (266.2 ms .. 340.9 ms)
std dev              47.45 ms   (1.859 ms .. 68.36 ms)
variance introduced by outliers: 41% (moderately inflated)

benchmarking run 1000 game ticks/movers/40
time                 815.1 ms   (642.8 ms .. 1.042 s)
                     0.991 R²   (0.973 R² .. 1.000 R²)
mean                 866.1 ms   (819.0 ms .. 913.2 ms)
std dev              58.40 ms   (28.79 ms .. 73.29 ms)
variance introduced by outliers: 20% (moderately inflated)
```

Idlers are much faster. Trees are also faster because they become idle after they're done growing. Circlers and movers are slower. Inlining the active robot check helps with the circlers:

```
benchmarking run 1000 game ticks/idlers/40
time                 941.6 μs   (932.4 μs .. 951.3 μs)
                     0.999 R²   (0.999 R² .. 1.000 R²)
mean                 933.1 μs   (926.8 μs .. 939.9 μs)
std dev              25.30 μs   (20.66 μs .. 32.12 μs)
variance introduced by outliers: 19% (moderately inflated)

benchmarking run 1000 game ticks/trees/40
time                 3.925 ms   (3.901 ms .. 3.948 ms)
                     1.000 R²   (0.999 R² .. 1.000 R²)
mean                 3.902 ms   (3.888 ms .. 3.919 ms)
std dev              57.08 μs   (47.38 μs .. 76.42 μs)

benchmarking run 1000 game ticks/circlers/40
time                 266.8 ms   (260.1 ms .. 275.4 ms)
                     0.999 R²   (0.998 R² .. 1.000 R²)
mean                 263.2 ms   (261.6 ms .. 266.1 ms)
std dev              3.170 ms   (1.914 ms .. 4.360 ms)
variance introduced by outliers: 11% (moderately inflated)

benchmarking run 1000 game ticks/movers/40
time                 829.2 ms   (692.9 ms .. 1.023 s)
                     0.994 R²   (0.982 R² .. 1.000 R²)
mean                 862.0 ms   (822.4 ms .. 901.8 ms)
std dev              50.46 ms   (21.92 ms .. 66.59 ms)
variance introduced by outliers: 19% (moderately inflated)
```

I'm not sure why the movers don't seem to benefit from the inlining.

The last commit that removes non-sensical tests (waiting after self destruction) has not effect on the performance, but cleans up the logic.


polux:
> I had not considered the possibility that this would slow down active robots, but I guess it makes sense.

It also came as a surprise to me :)

> Do you think it is worth merging this? I definitely lean towards yes, but in a sense it depends on one's play style. If you throw robots around like candy and leave lots of dead robots everywhere then it is going to help a lot. If you are fastidious about cleaning up then it might actually slow things down a tiny bit.

I also lean towards yes because unless your code is guaranteed exception-free you'll probably end up with a lot of dead robots as you spawn tons of them. 

> Are there other things we could mark INLINE that would speed it up?

I'm guessing yes, but maybe the low hanging fruits are already inlined by GHC. I suggest we profile before adding more.

> Also, why would trees become inactive after growing? Don't the seed robots self-destruct when they're done? (If not, they should!)

You're right, that is really weird. I do not understand why we're seeing a gain in the trees benchmark. We should make sure we understand what is going on here before we merge this. I may have introduced some bug.

polux:
So turns out that we were not removing the base from the active robots set. I thought that would explain the results but adding back the base does not bring the trees benchmark up to 10ms:

```
benchmarking run 1000 game ticks/idlers/40
time                 1.562 ms   (1.544 ms .. 1.580 ms)
                     0.996 R²   (0.989 R² .. 0.999 R²)
mean                 1.604 ms   (1.573 ms .. 1.655 ms)
std dev              156.6 μs   (110.7 μs .. 217.2 μs)
variance introduced by outliers: 73% (severely inflated)

benchmarking run 1000 game ticks/trees/40
time                 5.170 ms   (5.071 ms .. 5.249 ms)
                     0.993 R²   (0.985 R² .. 0.997 R²)
mean                 5.453 ms   (5.342 ms .. 5.670 ms)
std dev              525.5 μs   (325.1 μs .. 792.5 μs)
variance introduced by outliers: 66% (severely inflated)

benchmarking run 1000 game ticks/circlers/40
time                 289.4 ms   (280.7 ms .. 297.1 ms)
                     0.999 R²   (0.996 R² .. 1.000 R²)
mean                 294.3 ms   (289.0 ms .. 304.9 ms)
std dev              10.44 ms   (4.598 ms .. 16.00 ms)
variance introduced by outliers: 12% (moderately inflated)

benchmarking run 1000 game ticks/movers/40
time                 1.038 s    (795.2 ms .. 1.187 s)
                     0.994 R²   (0.981 R² .. 1.000 R²)
mean                 1.017 s    (956.9 ms .. 1.052 s)
std dev              59.06 ms   (24.04 ms .. 81.26 ms)
variance introduced by outliers: 19% (moderately inflated)
```

I suppose that the circlers and movers benchmarks end up even worse because of the string comparison. But I don't have an explanation for the better trees benchmark. Maybe this is due to b69871d.

polux:
> I guess I actually don't see the problem with removing the base from the activeRobots set, as long as we add it back when the user enters something at the REPL

Good point! This is better than testing for `robotName /= "base"` at every iteration like I do in 999f89f5f3c89392a40c0fcff0acaa0ae350e173.

> In particular we need to add a line of code right here.

Done in 8b56c375ba2bdfcd19983411a07ac6d4882fae99. The final benchmark is:

```
benchmarking run 1000 game ticks/idlers/40
time                 926.7 μs   (920.0 μs .. 935.9 μs)
                     0.999 R²   (0.997 R² .. 1.000 R²)
mean                 931.1 μs   (924.3 μs .. 944.9 μs)
std dev              37.99 μs   (24.95 μs .. 68.24 μs)
variance introduced by outliers: 35% (moderately inflated)

benchmarking run 1000 game ticks/trees/40
time                 4.384 ms   (4.102 ms .. 4.731 ms)
                     0.978 R²   (0.966 R² .. 0.999 R²)
mean                 4.123 ms   (4.071 ms .. 4.228 ms)
std dev              262.2 μs   (99.22 μs .. 425.7 μs)
variance introduced by outliers: 47% (moderately inflated)

benchmarking run 1000 game ticks/circlers/40
time                 268.1 ms   (265.7 ms .. 272.7 ms)
                     1.000 R²   (0.999 R² .. 1.000 R²)
mean                 267.4 ms   (265.9 ms .. 269.0 ms)
std dev              2.471 ms   (1.518 ms .. 3.294 ms)
variance introduced by outliers: 11% (moderately inflated)

benchmarking run 1000 game ticks/movers/40
time                 790.6 ms   (605.8 ms .. 1.011 s)
                     0.991 R²   (0.970 R² .. 1.000 R²)
mean                 876.4 ms   (822.4 ms .. 951.5 ms)
std dev              71.32 ms   (5.385 ms .. 89.37 ms)
variance introduced by outliers: 21% (moderately inflated)
```

The newly introduce `activateRobot` method has an unchecked precondition that the robot being activated is not part of the `waitingRobots` queue. This makes me a bit uncomfortable. Here are some unorganized thoughts about this:

- I cannot imagine a scenario where it would make sense to call this function on a waiting robot: it would immediately go back to waiting.
- I can imagine a scenario where you would call `activateRobot` on a non-waiting robot that is no more waiting but is still in the waiting queue though: if one were to change the program of a waiting robot. So it may make sense to consider that case in `activateRobot` and try to remove the robot from the `waitingRobots` queue. But then, we couldn't look it up by wake up time (because the CEK machine wouldn't be in state `Waiting wakeupTime cek` anymore) and so removing it from waitingRobots would be pretty expensive. 
- For that reason, I am tempted to say that modifying the program of a robot should already remove the robot from the `waitingRobots` queue and possibly add it back to the `activeRobots` one. More generally we have this problem that there is an unwritten (my bad) invariant that robots in the `waitingRobots` queue are in state `Waiting _ _` but we don't enforce it everywhere. Or maybe the invariant should be weaker, more on that below.
- We could maybe use [assertions](https://downloads.haskell.org/ghc/latest/docs/html/users_guide/exts/assert.html) for fonctions with preconditions, or for checking that invariants are maintained in stateful functions, and make sure not to run our tests in -O mode.

More on what the invariant should be. In #176 you wrote:

> I was going to ask whether we really needed to delete from waitingRobots at all; couldn't we just notice that the robot no longer exists whenever we get around to its time and thus ignore it? But then I realized that another robot with the same name could have been created in the meantime...

Maybe that is fine still: this other robot with the same name created in the meantime can be one of three different states by the time we call `wakeUpRobotsDoneSleeping` :
  - Idle, in which case `gameTick` will immediately remove it back from the `activeRobots` set;
  - Waiting, in which case `gameTick` will immediately put it back in the `waitingRobots` queue, with the right date this time;
  - Running, in which case it will continue running.

So maybe "robots in the `waitingRobots` queue are in state `Waiting _ _`" is too strong an invariant. Maybe `waitingRobots` is just a heuristic, and the real invariant is "Non-idle and non-waiting robots are included in the `activeRobots` set". What do you think?

byorgey:
Yes, it seems to me that nothing bad happens if `waitingRobots` gets out of sync with other things.  It just helps us save some computation.

We currently never change the program of a waiting robot (the `reprogram` command requires a robot to be idle), but even if we did in the future, I think your analysis is correct that nothing would break even if we changed the program of a waiting robot.  We would just have to make sure to put a robot in the `activeRobots` set when we change its program.

polux:
This is the final benchmark after the merge with main:

```
benchmarking run 1000 game ticks/idlers/40
time                 879.9 μs   (868.9 μs .. 890.5 μs)
                     0.989 R²   (0.972 R² .. 1.000 R²)
mean                 918.6 μs   (878.6 μs .. 993.7 μs)
std dev              211.5 μs   (42.60 μs .. 356.2 μs)
variance introduced by outliers: 95% (severely inflated)

benchmarking run 1000 game ticks/trees/40
time                 4.340 ms   (4.149 ms .. 4.729 ms)
                     0.973 R²   (0.945 R² .. 1.000 R²)
mean                 4.233 ms   (4.176 ms .. 4.376 ms)
std dev              309.9 μs   (145.3 μs .. 577.4 μs)
variance introduced by outliers: 54% (severely inflated)

benchmarking run 1000 game ticks/circlers/40
time                 281.5 ms   (261.4 ms .. 296.3 ms)
                     0.995 R²   (0.983 R² .. 1.000 R²)
mean                 303.7 ms   (292.3 ms .. 330.5 ms)
std dev              23.79 ms   (7.706 ms .. 34.59 ms)
variance introduced by outliers: 14% (moderately inflated)

benchmarking run 1000 game ticks/movers/40
time                 852.9 ms   (740.2 ms .. 1.088 s)
                     0.991 R²   (0.982 R² .. 1.000 R²)
mean                 942.9 ms   (888.9 ms .. 1.035 s)
std dev              91.27 ms   (29.06 ms .. 123.0 ms)
variance introduced by outliers: 22% (moderately inflated)
```

Looks pretty good to me! If you don't have any additional remark, I think this is ready to be merged.

PR #220, opened 2021-10-12
Refactor getx and gety into whereami command

### LINKED ISSUE #178: Replace `getX` and `getY` with `whereami` command
I like `whoami` too.  In fact while we're at it, I think we should also replace `getx` and `gety` with a `whereami` command. =)

_Originally posted by @byorgey in https://github.com/byorgey/swarm/issues/174#issuecomment-933465471_

twitu:
Closes #178


twitu:
Updating example files which use `getX` and `getY`.

PR #223, opened 2021-10-12
Delay type

### LINKED ISSUE #150: Error when applying function through build
**Describe the bug**
`build` command fails after passing through identity function 

**To Reproduce**
```
def id = \a. a end

id build "" {} <- this errors
id ( build "" {} ) <- this works, however
```

**Expected behavior**
x and (id x) should always be equivalent

byorgey:
Closes #150 .  The idea is to make explicit in the type system when evaluation of a computation should be delayed.  This gives the user fine-grained control over selective laziness (for example, once we have sum types and recursive types, one could use this to define lazy infinite data structures).  It also allows us to guarantee that certain commands such as `build` and `reprogram` delay evaluation of their arguments, and lets the user e.g. define their own modified versions of `build` without compromising those guarantees.

I'm very open to bikeshedding on the syntax.  I have initially chosen `{{ ... }}` to denote a delayed computation, and `! (...)` to denote forcing.  I think it's important to have nice-looking syntax for `delay` since the user is going to be typing it a lot.  I am more on the fence about whether we should have a prefix operator for forcing, or just use the word `force`.

- Add a new type constructor `delay`
- `TDelay` is now written with special double brackets `{{...}}`; `{{t}} : delay ty` whenever `t : ty`
- Add a special case to the parser: `{{}}` parses as `delay noop`
- `force : delay ty -> ty` is now written as a prefix `!`, e.g. `{{5}} : delay int` and `!{{5}} : int`
- change the types of `build` and `reprogram` to require a delayed program, e.g. `build : string -> delay (cmd a) -> cmd string`
- don't elaborate Build and Reprogram with extra Delay wrappers since one is now required by the type

### DIFF
--- a/src/Swarm/Language/Syntax.hs
+++ b/src/Swarm/Language/Syntax.hs
@@ -51,6 +51,7 @@ module Swarm.Language.Syntax (
   pattern TLet,
   pattern TDef,
   pattern TBind,
+  pattern TDelay,
 
   -- * Terms
   Var,
@@ -325,7 +326,7 @@ isCmd c = case constMeta $ constInfo c of
 -- | Function constants user can call with reserved words ('wait',...).
 isUserFunc :: Const -> Bool
 isUserFunc c = case constMeta $ constInfo c of
-  ConstMFunc {} -> c /= Force
+  ConstMFunc {} -> True
   _ -> False
 
 -- | Information about constants used in parsing and pretty printing.
@@ -368,7 +369,7 @@ constInfo c = case c of
   If -> functionLow 3
   Fst -> functionLow 1
   Snd -> functionLow 1
-  Force -> functionLow 1 -- TODO: make internal?!
+  Force -> unaryOp "!" 8 P
   Not -> functionLow 1
   Neg -> unaryOp "-" 7 P
   Add -> binaryOp "+" 6 L
@@ -452,9 +453,13 @@ pattern TLet v pt t1 t2 = SLet v pt (STerm t1) (STerm t2)
 pattern TDef :: Var -> Maybe Polytype -> Term -> Term
 pattern TDef v pt t = SDef v pt (STerm t)
 
--- | Match a TDef without syntax
+-- | Match a TBind without syntax
 pattern TBind :: Maybe Var -> Term -> Term -> Term
 pattern TBind v t1 t2 = SBind v (STerm t1) (STerm t2)
+
+-- | Match a TDelay without syntax
+pattern TDelay :: Term -> Term
+pattern TDelay t = SDelay (STerm t)
 
 -- | COMPLETE pragma tells GHC using this set of pattern is complete for Term
 {-# COMPLETE TUnit, TConst, TDir, TInt, TAntiInt, TString, TAntiString, TBool, TVar, TPair, TLam, TApp, TLet, TDef, TBind, TDelay #-}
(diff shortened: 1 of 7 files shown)


byorgey:
Once we're all on the same page re syntax etc, I will update the tutorial and example programs.

polux:
This made me realize a few things and I'm wondering if we should do this at all. Adding my thoughts/questions to #150 so the conversation doesn't get fragmented.

src/Swarm/Language/Syntax.hs
```diff
@@ -509,14 +514,14 @@ data Term
     SDef Var (Maybe Polytype) Syntax
   | -- | A monadic bind for commands, of the form @c1 ; c2@ or @x <- c1; c2@.
     SBind (Maybe Var) Syntax Syntax
-  | -- | Delay evaluation of a term.  Swarm is an eager language, but
-    --   in some cases (e.g. for @if@ statements and recursive
-    --   bindings) we need to delay evaluation.  The counterpart to
-    --   @delay@ is @force@, where @force (delay t) = t@.  Note that
-    --   'Force' is just a constant, whereas 'TDelay' has to be a
-    --   special syntactic form so its argument can get special
+  | -- | Delay evaluation of a term, written @{{...}}@.  Swarm is an
```
polux, line 517: Should be `@{...}@` here and below.

src/Swarm/Language/Typecheck.hs
```diff
@@ -322,8 +322,8 @@ infer (Syntax _ (TAntiString _)) = return UTyString
 infer (Syntax _ (TBool _)) = return UTyBool
 -- To infer the type of a pair, just infer both components.
 infer (Syntax _ (SPair t1 t2)) = UTyProd <$> infer t1 <*> infer t2
--- delay t has the same type as t.
-infer (Syntax l (TDelay t)) = infer (Syntax l t)
+-- if t : ty, then  {{t}} : delay ty.
```
polux, line 325: `{{t}}` -> `{t}`

src/Swarm/Game/CESK.hs
```diff
@@ -1,54 +1,55 @@
...
--- Module      :  Swarm.Game.CEK
+-- Module      :  Swarm.Game.CESK
 -- Copyright   :  Brent Yorgey
 -- Maintainer  :  byorgey@gmail.com
 --
 -- SPDX-License-Identifier: BSD-3-Clause
 --
 -- The Swarm interpreter uses a technique known as a
--- <https://matt.might.net/articles/cek-machines/ CEK machine>.
+-- <https://matt.might.net/articles/cesk-machines/ CESK machine>.
```
polux, line 12: Thought this was the result of a sed, but no! This URL truly exists :)

src/Swarm/Game/CESK.hs
```diff
@@ -150,23 +161,54 @@ data Frame
...
--- | The overall state of a CEK machine, which can actually be one of
---   three kinds of states. The CEK machine is named after the first
+------------------------------------------------------------
+-- Store
+------------------------------------------------------------
+
+-- | 'Store' represents a store, indexing integer locations to 'Cell's.
+data Store = Store {next :: Int, mu :: IntMap Cell} deriving (Show, Eq)
+
+-- | A memory cell can be in one of three states.
```
polux, line 171: Maybe add a comment explaining what Blackhole is?

polux (changes requested):
Nice! I only have a few nits.

I reviewed the PR because if I understand the github UI correctly you did request a review. But the step function is not modified yet. Is it because you'd like to stage the review? (Which is very much welcome if that is the case: it's always easier to review small chunks.) Or is it because you're planning on modifying the step function in a different PR?

byorgey:
@polux I think I requested a review when I initially opened the PR.  I didn't specifically request one just now, but I appreciate the feedback, and you're right, reviewing in small chunks is helpful!  I do plan to update the step function in this PR.  Right now a store exists but is never used for anything.

PR #224, opened 2021-10-13
Add sum types

### LINKED ISSUE #46: Add sum types to the language
Somehow I added product types but haven't gotten around to adding sum types yet!

byorgey:
Adds sum types (e.g. `int + bool`) along with constants `inl : a -> a + b`, `inr : b -> a + b`, and `case : a + b -> (a -> c) -> (b -> c) -> c`.

I considered whether to automatically wrap the arguments to `case` in `delay`, like the arguments to `if`, but they are already functions anyway so we get a lot of laziness for free.  It would only make a difference if you wrote something like `case foo (let x = blah in \y. ...) (...)` in which case the `let x = blah` is going to be evaluated eagerly, *before* we know which branch we are going to take.  I don't think it's worth bothering about.  In the normal case that you write `case foo (\y. ...) (\z. ...)` then only one of the two functions gets run at all, without us having to do anything special with `delay`.

I also considered whether to make special syntax for case expressions, like `case foo of { inl x -> ... ; inr y -> ... }` but doing it as a simple built-in function is both easier to implement and feels like it fits better with the aesthetic of the language so far.

Closes #46 .


src/Swarm/Game/Value.hs
```diff
@@ -43,6 +46,8 @@ data Value where
   VDir :: Direction -> Value
   -- | A boolean.
   VBool :: Bool -> Value
+  -- | An injection into a sum type.  False = left, True = right.
+  VInj :: Bool -> Value -> Value
```
polux, line 50: Any reason you went with that representation instead of `VInl` and `VInlr`?

polux (commented):
LGTM, I'm exited for this! Can't wait to play with streams :) Jus left one comment that is more of a question to satisfy my curiosity than something to address.

src/Swarm/Game/Value.hs
byorgey: Merely to cut down on the number of pattern matches needed later, e.g. when reducing a `case` there's just a single code path which uses the `Bool` to select between the continuations.  It really doesn't make that much of a difference, and I could be convinced it would be cleaner/more transparent to use separate `VInl` and `VInr`.

byorgey: I tried changing it but then I realized there's an even better reason to do it this way: implementing comparison (`inl _` is always less than `inr _`, and when the tags match the underlying values are compared) is *much* nicer when we use `VInj`: we can just `compare` the bools, and then `compare` the values if they are equal.  With `VInl` and `VInr` we need to tediously list out the four possible cases.

src/Swarm/Game/Value.hs
```diff
@@ -1,6 +1,7 @@
 -----------------------------------------------------------------------------
 -----------------------------------------------------------------------------
 {-# LANGUAGE GADTs #-}
+{-# LANGUAGE PatternSynonyms #-}
```
xsebek, line 4: Is this still necessary when you have `VInj s v`?

test/Unit.hs
```diff
@@ -125,3 +141,77 @@ prettyConst =
...
+            "nested 2"
+            ("(\\x : int + bool + string. case x (\\q. 1) (\\s. case s (\\y. 2) (\\z. 3))) (inr (inl false))" `evaluatesTo` VInt 2)
+        , testCase
+            "nested 2"
+            ("(\\x : int + bool + string. case x (\\q. 1) (\\s. case s (\\y. 2) (\\z. 3))) (inr (inr \"hi\"))" `evaluatesTo` VInt 3)
+        ]
+    ]
+ where
+  evaluatesTo :: Text -> Value -> Assertion
+  evaluatesTo tm val = do
```
xsebek, line 200: This is really nice, I would like to steal this in #239 :innocent:

src/Swarm/Language/Parse.hs
```diff
@@ -377,8 +378,8 @@ runParser p t = first (from . errorBundlePretty) (parse (runReaderT p DisallowAn
 runParserTH :: (Monad m, MonadFail m) => (String, Int, Int) -> Parser a -> String -> m a
 runParserTH (file, line, col) p s =
   case snd (runParser' (runReaderT (fully p) AllowAntiquoting) initState) of
-    Left err -> fail $ errorBundlePretty err
-    Right e -> return e
+    Prelude.Left err -> fail $ errorBundlePretty err
+    Prelude.Right e -> return e
```
xsebek, line 382: I assume this is leftover too, since you do not define new `Right`, right? :thinking:

xsebek (approved):
LGTM, I played with this earlier and it seemed to work fine.

<sub>However, I have not yet found a good use for sum types, so my testing was fairly limited. :sweat_smile: </sub>

src/Swarm/Game/Value.hs
byorgey: Nope, good catch. Fixed now.

src/Swarm/Language/Parse.hs
byorgey: Another good catch!

PR #230, opened 2021-10-15
Generate direction helper methods from consolidated direction information in `DirInfo` records

### LINKED ISSUE #183: Consolidate `Direction` info
As discussed in the comments of #182, we should make a function `dirInfo :: Direction -> DirInfo`, similar to `constInfo`, which returns all the information about a `Direction`:
* Its concrete syntax, as `Text`
* A field `Maybe (V2 Int64)`, saying whether it corresponds to an absolute direction (for implementing `toDirection` and `fromDirection`)
* A field `applyTurn :: V2 Int64 -> V2 Int64`

Once we have this, we can
1. automatically generate the directions in the list of `reservedWords`
2. automatically generate `parseDirection`
3. automatically generate the `PrettyPrec` instance

This way, if we ever add any new `Direction` values (e.g. northwest etc.?) we only have to add them in one place.

twitu:
Closes #183

### DIFF
--- a/src/Swarm/Language/Pretty.hs
+++ b/src/Swarm/Language/Pretty.hs
@@ -35,7 +35,6 @@ import Control.Unification.IntVar
 import Swarm.Language.Capability
 import Swarm.Language.Context
 import Swarm.Language.Syntax
-import Swarm.Language.Syntax (DirInfo (dirSyntax))
 import Swarm.Language.Typecheck
 import Swarm.Language.Types
 


src/Swarm/Language/Syntax.hs
```diff
@@ -86,27 +90,59 @@ import Swarm.Language.Types
...
+    dirApplyTurn :: V2 Int64 -> V2 Int64
+  }
+
+allDirs :: [Direction]
+allDirs = [minBound .. maxBound]
+
+-- | TODO Information about all directions
+dirInfo :: Direction -> DirInfo
+dirInfo d = case d of
+  Lft -> DirInfo "left" Nothing (\(V2 x y) -> V2 (- y) x)
```
xsebek, line 115: Maybe we could prefix direction with `D`, so then you could do the same `lowShow` trick as with `Const` and avoid having to write `South -> DirInfo "south"`.

twitu: Ohk so it needs a "D" prefix like `DLeft`, `DRight` because Prelude exports `Left` and `Right`

src/Swarm/Language/Syntax.hs
```diff
@@ -112,15 +113,19 @@ allDirs = [minBound .. maxBound]
...
-  Down -> DirInfo "down" Nothing (const $ V2 0 0)
+  DLeft -> DirInfo directionSyntax Nothing (\(V2 x y) -> V2 (- y) x)
+  DRight -> DirInfo directionSyntax Nothing (\(V2 x y) -> V2 y (- x))
+  DBack -> DirInfo directionSyntax Nothing (\(V2 x y) -> V2 (- y) (- x))
+  DForward -> DirInfo directionSyntax Nothing id
+  DNorth -> DirInfo directionSyntax (Just north) (const north)
+  DSouth -> DirInfo directionSyntax (Just south) (const south)
+  DEast -> DirInfo directionSyntax (Just east) (const east)
+  DWest -> DirInfo directionSyntax (Just west) (const west)
+  DDown -> DirInfo directionSyntax Nothing (const $ V2 0 0)
```
xsebek, line 124: @twitu actually why is `down` not treated as other cardinal directions here? In `toDirection` it is. 😕 

It would be nice to avoid all repetition with expressive helpers, that would just take the necessary info - in this case `west` is all that is different in treatment of other cardinal directions.

twitu: I have fixed the treatment of `down` and have created a helper function `cardinalDirs :: Map (V2 Int64) Direction` to generate a mapping  for `toDirection`. This makes all the direction related function auto generated from the `dirInfo`.

However I just wanted to check if the results of `cardinalDirs` will be memoized because it's a constant function? Otherwise a new `Map` will be created each time `toDirection` is called.

src/Swarm/Language/Syntax.hs
```diff
@@ -85,51 +91,42 @@ import Swarm.Language.Types
...
+  DForward -> DirInfo directionSyntax Nothing id
+  DNorth -> DirInfo directionSyntax (Just north) (const north)
+  DSouth -> DirInfo directionSyntax (Just south) (const south)
+  DEast -> DirInfo directionSyntax (Just east) (const east)
+  DWest -> DirInfo directionSyntax (Just west) (const west)
+  DDown -> DirInfo directionSyntax (Just down) (const $ V2 0 0)
+ where
+  -- name is generate from Direction data constuctor
+  -- e.g. DLeft becomes "left"
+  directionSyntax = toLower . T.tail . from . show $ d
```
xsebek, line 129:
```suggestion
dirInfo d = case d of
  DLeft -> relative (\(V2 x y) -> V2 (- y) x)
  DRight -> relative (\(V2 x y) -> V2 y (- x))
  DBack -> relative (\(V2 x y) -> V2 (- y) (- x))
  DForward -> relative id
  DNorth -> cardinal north
  DSouth -> cardinal south
  DEast -> cardinal east
  DWest -> cardinal west
  DDown -> cardinal down
 where
  -- name is generate from Direction data constuctor
  -- e.g. DLeft becomes "left"
  directionSyntax = toLower . T.tail . from . show $ d
  cardinal d = DirInfo directionSyntax (Just d) (const d)
  relative cd = DirInfo directionSyntax Nothing cd
```

xsebek: Cool, so this is how GitHub PR suggestion works :smile: Sorry about this perfectionism @twitu, I just wanted to try this out :wink:

src/Swarm/Language/Syntax.hs
xsebek: I would guess the Map *will be fully evaluated* once `toDirection` is called - but if you are worried about performance, you can add a benchmark* or use `Map.Strict`.

*) I do not think it will show any difference though

twitu: By "fully evaluated" do you mean re-calculated? And why so? Why won't the performance difference show up? Sorry if I'm pestering you with this, I'm not sure what to search to understand this.

xsebek: No I meant it should be computed once and then it will be ready to use for every operation after. :+1: 
Since you have given it a name `cardinalDirs` it should be treated that way as it does not depend on input of `toDirection`.
With `Map.Lazy` it can be in a partially evaluated state - say it finds the vector at beginning of the `allDirs` list and then stops. 

In any case the overhead is unlikely to be measurable. :wink:

src/Swarm/Language/Syntax.hs
```diff
@@ -147,6 +146,37 @@ east = V2 1 0
...
+applyTurn :: Direction -> V2 Int64 -> V2 Int64
+applyTurn = dirApplyTurn . dirInfo
+
+-- | Mapping from heading to their corresponding cardinal directions
+--   only directions with a 'dirAbs` value are mapped
+cardinalDirs :: M.Map (V2 Int64) Direction
+cardinalDirs =
+  M.fromList
+    . map (\d -> (fromJust . dirAbs . dirInfo $ d, d))
+    . filter (isJust . dirAbs . dirInfo)
```
xsebek, line 165: You can make this shorter and safer (needs `TupleSection` above and change of imports to `(fromMaybe, mapMaybe)`)
```suggestion
    . mapMaybe (\d -> (,d) <$> dirAbs (dirInfo d))
```

src/Swarm/Language/Syntax.hs
```diff
@@ -85,51 +91,44 @@ import Swarm.Language.Types
...
+  DSouth -> cardinal south
+  DEast -> cardinal east
+  DWest -> cardinal west
+  DDown -> cardinal down
+ where
+  -- name is generate from Direction data constuctor
+  -- e.g. DLeft becomes "left"
+  directionSyntax = toLower . T.tail . from . show $ d
+  cardinal d = DirInfo directionSyntax (Just d) (const d)
+  relative cd = DirInfo directionSyntax Nothing cd
```
xsebek, line 131: Sorry, my mistake, I just checked it out in editor and HLint is not happy about using the same `d` here:
```suggestion
  cardinal v2 = DirInfo directionSyntax (Just v2) (const v2)
  relative vf = DirInfo directionSyntax Nothing vf
```

xsebek (approved):

PR #237, opened 2021-10-16
Fix operator parsing

xsebek:
- add check that operator is not followed by another "operator"

### DIFF
--- a/src/Swarm/Language/Parse.hs
+++ b/src/Swarm/Language/Parse.hs
@@ -4,6 +4,7 @@
 {-# LANGUAGE TypeApplications #-}
 {-# LANGUAGE TypeFamilies #-}
 {-# LANGUAGE TypeSynonymInstances #-}
+{-# LANGUAGE LambdaCase #-}
 
 -- |
 -- Module      :  Swarm.Language.Parse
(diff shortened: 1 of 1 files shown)


byorgey:
Looks good to me. Can we also add some tests?

src/Swarm/Language/Parse.hs
```diff
@@ -346,7 +348,16 @@ unOps = Map.unionsWith (++) $ mapMaybe unOpToTuple allConst
...
+        [assI (exprLoc1 $ SApp (noLoc $ TConst c) <$ operatorString (syntax ci))]
+
+operatorString :: Text -> Parser Text
+operatorString n = (lexeme . try) (string n <* notFollowedBy operatorSymbol)
+
+operatorSymbol :: Parser Text
+operatorSymbol = T.singleton <$> oneOf opChars
+ where
+  isOp = \case { ConstMFunc {} -> False; _ -> True } . constMeta
+  opChars = concatMap (from . syntax) . filter isOp $ map constInfo allConst
```
byorgey, line 360: Should we call `nub` on `opChars`?  If I'm reading this correctly, it may end up with multiple copies of the same character for characters that occur in multiple operators (`=`, `<`, etc.).  I doubt it makes a huge difference but I suppose `oneOf` might have to do some redundant work if `opChars` has duplicates.

byorgey (approved):

PR #246, opened 2021-10-19
Cap ticks in model

### LINKED ISSUE #245: Counter overflow
When I hold the '>' button, the speed counter increases so that it reaches negative values, and then completely to zero.
[video](https://radikal.ru/video/7VY65tw1BSn)

xsebek:
- fixes  #245


src/Swarm/TUI/Model.hs
```diff
@@ -258,6 +265,15 @@ uiFPS :: Lens' UIState Double
 -- | The base-2 logarithm of the current game speed in ticks per
 --   second.
 lgTicksPerSecond :: Lens' UIState Int
+lgTicksPerSecond = lens _lgTicksPerSecond safeSetLgTicks
+ where
+  maxTicks = 5 -- 2^(-5) == 1/32
```
byorgey, line 270: The `ticksPerFrameCap` is ticks per *frame*, not ticks per second.  There is no reason to limit the max number of ticks per second to 32.  I know my computer can easily handle higher than that depending on how many robots there are.

src/Swarm/TUI/Model.hs
```diff
@@ -258,6 +265,15 @@ uiFPS :: Lens' UIState Double
...
+  minTicks = 1 - floor (logBase 2 (fromIntegral (maxBound :: Int)) :: Float)
```
byorgey, line 271: This seems unnecessarily complicated.  Why not just have it be something arbitrary like -30 and be done with it?

src/Swarm/TUI/Controller.hs
```diff
@@ -241,7 +241,7 @@ runFrame = do
   runFrameTicks (fromNanoSecs dt)
 
 ticksPerFrameCap :: Int
-ticksPerFrameCap = 30
+ticksPerFrameCap = 32
```
byorgey, line 244: There's no reason to change this.  As I mention below, this number has nothing to do with the ticks per second.

src/Swarm/TUI/Model.hs
xsebek: I do not know *what value* makes sense, so I left it both in (one less then) maximum. If the players want to skip ahead from the bronze age to Rome or just [don't want to use the pause](https://xkcd.com/1172/), I will not stop them. :smirk:

src/Swarm/TUI/Model.hs
```diff
@@ -255,9 +262,19 @@ uiTPF :: Lens' UIState Double
...
 uiFPS :: Lens' UIState Double
 
--- | The base-2 logarithm of the current game speed in ticks per
---   second.
+-- | The base-2 logarithm of the current game speed in ticks/second.
+--   Note that we cap this value to the range of +/- log2 INTMAX.
 lgTicksPerSecond :: Lens' UIState Int
+lgTicksPerSecond = lens _lgTicksPerSecond safeSetLgTicks
+ where
+  maxLog = floor (logBase 2 (fromIntegral (maxBound :: Int)) :: Float)
```
byorgey, line 270: What do you think about using `finiteBitSize` from `Data.Bits` instead of `logBase 2` etc? Seems cleaner to me.

PR #252, opened 2021-10-20
Only append entry to REPL history if it is not the same as last entry

### LINKED ISSUE #75: Remove duplicate entries in the REPL history
When repeating a command, the REPL history gets cluttered with repeated elements, rendering the history navigation tedious.

The history should only contains unique command.

To avoid the confusion, the history index could be annotated to avoid confusion when the line does not change when pressing up.

Alexander-Block:
Closes #75. 
A REPL entry is only appended to the REPL history if it is not a duplicate of the last entry.


byorgey (approved):
Works great, thanks!

PR #253, opened 2021-10-21
Store history in XDG data directory

### LINKED ISSUE #136: Store REPL history file in $XDG_DATA_HOME
why not to save history file somewhere in `$HOME`?
not in `pwd`

xsebek:
- fixes #136 
- changes the history file format to simple text line per `REPLEntry`
- refactors out a `REPLHistory` type and its pure functions
  - adds tests for REPL logic (move to different previous entry,...)

### DIFF
--- a/src/Swarm/TUI/Controller.hs
+++ b/src/Swarm/TUI/Controller.hs
@@ -31,6 +31,7 @@ module Swarm.TUI.Controller (
   handleREPLEvent,
   validateREPLForm,
   adjReplHistIndex,
+  TimeDir (..),
 
   -- ** World panel
   handleWorldEvent,
@@ -151,15 +152,12 @@ toggleModal s modal = do
 --   the updated REPL history to a @.swarm_history@ file.
 shutdown :: AppState -> EventM Name (Next AppState)
 shutdown s = do
-  let hist = mapMaybe getNewEntry (s ^. uiState . uiReplHistory)
+  let hist = mapMaybe getREPLEntry $ getLatestREPLHistoryItems maxBound history
   liftIO $ (`T.appendFile` T.unlines hist) =<< getSwarmHistoryPath True
-  let s' = s & uiState . uiReplHistory . traverse %~ markOld
+  let s' = s & uiState . uiReplHistory %~ restartREPLHistory
   halt s'
  where
-  markOld (REPLEntry _ d e) = REPLEntry False d e
-  markOld r = r
-  getNewEntry (REPLEntry True _ t) = Just t
-  getNewEntry _ = Nothing
+  history = s ^. uiState . uiReplHistory
 
 ------------------------------------------------------------
 -- Handling Frame events
@@ -320,7 +318,7 @@ updateUI = do
     -- result as a REPL output, with its type, and reset the replStatus.
     REPLWorking pty (Just v) -> do
       let out = T.intercalate " " [into (prettyValue v), ":", prettyText (stripCmd pty)]
-      uiState . uiReplHistory %= (REPLOutput out :)
+      uiState . uiReplHistory %= addREPLItem (REPLOutput out)
       gameState . replStatus .= REPLDone
       pure True
 
@@ -403,8 +401,7 @@ handleREPLEvent s (VtyEvent (V.EvKey V.KEnter [])) =
           s
             & uiState . uiReplForm %~ updateFormState ""
             & uiState . uiReplType .~ Nothing
-            & uiState . uiReplHistory %~ prependReplEntry
-            & uiState . uiReplHistIdx .~ (-1)
+            & uiState . uiReplHistory %~ addREPLItem (REPLEntry entry)
             & uiState . uiError .~ Nothing
             & gameState . replStatus .~ REPLWorking ty Nothing
             & gameState . robotMap . ix "base" . machine .~ initMachine t topValCtx topStore
@@ -422,13 +419,10 @@ handleREPLEvent s (VtyEvent (V.EvKey V.KEnter [])) =
   topStore =
     fromMaybe emptyStore $
       s ^? gameState . robotMap . at "base" . _Just . robotContext . defStore
-  prependReplEntry replHistory
-    | firstReplEntry replHistory == Just entry = REPLEntry True True entry : replHistory
-    | otherwise = REPLEntry True False entry : replHistory
 handleREPLEvent s (VtyEvent (V.EvKey V.KUp [])) =
-  continue $ s & adjReplHistIndex (+)
+  continue $ s & adjReplHistIndex Older
 handleREPLEvent s (VtyEvent (V.EvKey V.KDown [])) =
-  continue $ s & adjReplHistIndex (-)
+  continue $ s & adjReplHistIndex Newer
 handleREPLEvent s ev = do
   f' <- handleFormEvent ev (s ^. uiState . uiReplForm)
   continue $ validateREPLForm (s & uiState . uiReplForm .~ f')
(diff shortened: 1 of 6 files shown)


xsebek:
The new history file path is:
```bash
echo "${XDG_DATA_HOME:-$HOME/.local/share}/swarm/history"
```
The quick and easy way to migrate old history using `ghci`:
```haskell
data R = REPLEntry Bool String deriving Read
new_history = "<output of shell command above>"
readFile ".swarm_history" >>= pure . unlines . map (\(REPLEntry _ t) -> t) . read >>= writeFile new_history
```

src/Swarm/TUI/Controller.hs
```diff
@@ -150,16 +151,15 @@ toggleModal s modal = do
 --   the updated REPL history to a @.swarm_history@ file.
 shutdown :: AppState -> EventM Name (Next AppState)
 shutdown s = do
+  let hist = mapMaybe getNewEntry (s ^. uiState . uiReplHistory)
+  liftIO $ (`T.appendFile` T.unlines hist) =<< getSwarmHistoryPath True
```
byorgey, line 155: Ah yes, this is much better. This way it won't break next time we change the format of the `REPLHistory`.

src/Swarm/Util.hs
```diff
@@ -83,27 +94,38 @@ maximum0 :: (Num a, Ord a) => [a] -> a
...
+readFileMay :: FilePath -> IO (Maybe String)
+readFileMay = catchIO . readFile
+
+-- | Safely attempt to (efficiently) read a file.
+readFileMayT :: FilePath -> IO (Maybe Text)
+readFileMayT = catchIO . T.readFile
+
+-- | Turns any IO error into Nothing.
+catchIO :: IO a -> IO (Maybe a)
+catchIO act = (Just <$> act) `catchIOError` (\_ -> return Nothing)
```
byorgey, line 117: Oh good, this is much better than my `readFileMay` function.

src/Swarm/Util.hs
```diff
@@ -83,27 +94,38 @@ maximum0 :: (Num a, Ord a) => [a] -> a
...
+
+-- | Get path to swarm history, optionally creating necessary
+--   directories. This could fail if user has bad permissions
+--   on his own $HOME or $XDG_DATA_HOME which is unlikely.
+getSwarmHistoryPath :: Bool -> IO FilePath
```
byorgey, line 122: In the future we may want to put other things in the swarm data dir (e.g. save files).  How about generalizing this to `getSwarmPath` which takes the name of the wanted file as a `FilePath` argument, instead of hard-coding the file name `"history"`?

src/Swarm/TUI/Controller.hs
```diff
@@ -449,22 +445,37 @@ validateREPLForm s =
...
   validate = setFieldValid (isRight result) REPLInput
 
+data TimeDir = Newer | Older deriving (Eq, Ord, Show)
+
 -- | Update our current position in the REPL history.
-adjReplHistIndex :: (Int -> Int -> Int) -> AppState -> AppState
-adjReplHistIndex (+/-) s =
-  s & uiState . uiReplHistIdx .~ newIndex
-    & (if curIndex == -1 then saveLastEntry else id)
+adjReplHistIndex :: TimeDir -> AppState -> AppState
```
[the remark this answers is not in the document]
Alexander-Block: I agree that this looks a lot simpler than before. However, I propose to refactor this function even further. I mainly see a problem with the two conditionals 
```haskell
if s ^. replIx == s ^. replLen then saveLastEntry else id
``` 
and
```haskell
if s ^. replIx /= ns ^. replIx then showNewEntry else id
```

src/Swarm/TUI/Controller.hs
```diff
@@ -450,21 +444,27 @@ validateREPLForm s =
...
 -- | Update our current position in the REPL history.
-adjReplHistIndex :: (Int -> Int -> Int) -> AppState -> AppState
-adjReplHistIndex (+/-) s =
-  s & uiState . uiReplHistIdx .~ newIndex
-    & (if curIndex == -1 then saveLastEntry else id)
-    & (if newIndex /= curIndex then uiState . uiReplForm %~ updateFormState newEntry else id)
+adjReplHistIndex :: TimeDir -> AppState -> AppState
+adjReplHistIndex d s =
+  ns
+    & (if s ^. replIx == s ^. replLen then saveLastEntry else id)
```
Alexander-Block, line 450: The condition `s ^. replIx == s ^. replLen` expresses that the input buffer is the current focus of the `replHistory` belonging to the `uiState` of `s`. I can see that immediately now, because I know the indexing convention for `REPLHistory`. Should the controller as a client of `REPLHistory` really know and care about this convention (which incidentally was changed for this new implementation)? I can imagine myself wondering in six months what is going on here because I forgot that the REPL index being the REPL length means focus on the input buffer.

I therefore propose the following change:
 ```suggestion
    & (if replAtInputBuffer (s ^. repl) then saveLastEntry else id)
```
This uses a new (probably badly named) function 
```haskell
replAtInputBuffer :: REPLHistory -> Bool
replAtInputBuffer replHist = replHist ^. replIndex == replHist ^. to replLength
```
which in turn can be moved to `Model.hs`, which should be the right place for internals of `REPLHistory`.

src/Swarm/TUI/Controller.hs
```diff
@@ -450,21 +444,27 @@ validateREPLForm s =
...
+    & (if s ^. replIx /= ns ^. replIx then showNewEntry else id)
```
Alexander-Block, line 451: This line should probably express that a new entry should be shown when the entry currently focused in the REPL history changes. If not, then nothing should happen. 
Here we use a REPL index as a kind of proxy. However, wouldn't the following be clearer and even semantically more correct?
```suggestion
    & (if oldEntry /= newEntry then showNewEntry else id)
```
Here `oldEntry` can be defined in analogy to `newEntry` as follows:
```haskell
oldEntry = s ^. repl . to (fromMaybe replLast . getCurrentItemText)
```

src/Swarm/TUI/Model.hs
```diff
@@ -20,20 +20,33 @@ module Swarm.TUI.Model (
...
+  -- ** REPL
   REPLHistItem (..),
   firstReplEntry,
+  replItemText,
+  isREPLEntry,
+  getREPLEntry,
+  REPLHistory,
+  replSeq,
+  replStart,
+  replIndex,
```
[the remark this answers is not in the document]
Alexander-Block: Having thought a little bit more about this I disagree with your assessment to some extent. I fully agree that if we accept that we need a `Getter` for `replIndex` it is not worth the trouble to just export a `Getter` as opposed to a full lens. 
However, I do think that we neither should need nor pragmatically do need to export even a `Getter` for `replIndex`. 

Pragmatically outside of `Model.hs` we only actually use `replIndex` in `adjReplHistIndex` in `Controller.hs` as well as in `drawREPL` in `View.hs`.  For the former function I propose a refactoring below that does not rely on `replIndex`. For the latter function we only need `replIndex` for a debugging functionality to print the index. So the lens could be replaced by a function exported from `Model.hs` returning a textual representation of the REPL index.

Conceptually I see no big difference between exporting `replIndex` and `replSeq` because both work in tandem. Without knowing about the internals of the storage of `REPLHistItem`s in `REPLHistory` like, e.g., the indexing scheme, the information contained in `replIndex` should be quite meaningless to a client. An example is its first use in `adjReplHistIndex` where it is compared agains the history length to determine if the input buffer is currently focused. I think there should rather be a dedicated function for this in `Model.hs` as part of the public contract around `REPLHistory`.

src/Swarm/TUI/Model.hs
```diff
@@ -20,20 +20,36 @@ module Swarm.TUI.Model (
...
-  firstReplEntry,
+  replItemText,
+  isREPLEntry,
+  getREPLEntry,
+  REPLHistory,
+  replIndex,
+  newREPLHistory,
+  addREPLEntry,
+  addREPLOutput,
+  replLength,
```
Alexander-Block, line 34: I think that if we agree on the proposed refactoring for `adjReplHistIndex` there would be no need to export `replLength` anymore.

Alexander-Block (approved):

Alexander-Block:
I noticed that the status of my review was still "requesting changes"; the reason for this status was the bug I noticed and which has since been fixed. There are still unresolved conversations, but these contain only comments which may or may not lead to changes in this PR and should probably not prevent a merge.

byorgey:
I think it looks good too, though I seem to recall @xsebek wanting to add some tests first?

Alexander-Block:
@xsebek has added unit tests (cf `testModel` in `Unit.hs`). They look fine to me, although I didn't explicitly commented on them.

xsebek:
@byorgey You might be thinking about #270, where I would indeed like to write some tests first :slightly_smiling_face: 

@Alexander-Block sorry for the delay, I actually had the changes you proposed implemented for almost a week now, just did not get around to test and push them :wink:

PR #256, opened 2021-10-21
Use fused-effects for the game logic

### LINKED ISSUE #228: Try using `fused-effects` library
A comment in the `stepCEK` function says:

> It's a little unsatisfactory the way we handle having both `Robot` and `GameState` in different states (by having one be concrete and the other accessible via `lift`).  Having them both be capability constraints is what we really want, but it's not possible to do that with `mtl`.  Ultimately we may want to switch to an effects library.  I am hesitant to use`polysemy` because of performance issues.  Perhaps `fused-effects` would work.  I really want to use `eff` but seems like it's not ready yet.

Besides state effects for `Robot` and `GameState`, we also make use of:
* Randomness. Currently we simply store a `StdGen` in the `GameState` and use it manually.  But this is exactly the kind of thing an effects library can help with: we declare that functions which need randomness make use of a randomness effect, and then we later interpret it using an ambient state effect for `GameState`.
* IO.  The only thing we use this for is for interpreting the `run` function.  But it would be really nice to abstract this to use some kind of "filesystem" effect, which we can later interpret into `IO`, or something else in the future (e.g. in certain modes (challenge mode?) we might not want to allow loading arbitrary files from the filesystem, so we provide a fake "filesystem" that contains certain preallowed files).

After some discussion in IRC it seems like `fused-effects` is probably the way to go.  It supposedly has good performance, and (unlike `capability`) allows us to do the kind of local dispatch of effects that we want.

polux:
Resolves #228.

I've redefined the missing lens operators in `Utils.hs` in order to keep the diff of the function bodies minimal. Hopefully that will make merges and rebases less painful. Most of the type signatures are modified though, we can't avoid this.

### DIFF
--- a/src/Swarm/Game/Step.hs
+++ b/src/Swarm/Game/Step.hs
@@ -131,8 +131,6 @@ gameTick = do
 -- Some utility functions
 ------------------------------------------------------------
 
--- | Perform operation on the monad with game state.
-
 -- | Set a flag telling the UI that the world needs to be redrawn.
 flagRedraw :: (Has (State GameState) sig m) => m ()
 flagRedraw = needsRedraw .= True
@@ -152,10 +150,7 @@ entityAt loc = zoomWorld (W.lookupEntityM @Int (W.locToCoords loc))
 
 -- | Modify the entity (if any) at a given location.
 updateEntityAt ::
-  (Has (State GameState) sig m, Has (State Robot) sig m, Has (Throw Exn) sig m) =>
-  V2 Int64 ->
-  (Maybe Entity -> Maybe Entity) ->
-  m ()
+  (Has (State GameState) sig m) => V2 Int64 -> (Maybe Entity -> Maybe Entity) -> m ()
 updateEntityAt loc upd = zoomWorld (W.updateM @Int (W.locToCoords loc) upd)
 
 -- | Get the robot with a given name (if any).


src/Swarm/Game/Recipe.hs
```diff
@@ -117,11 +119,11 @@ resolveRecipes em = (traverse . traverse) (\t -> maybe (Failure [t]) Success (lo
 
 -- | Given an already loaded 'EntityMap', try to load a list of
 --   recipes from the data file @recipes.yaml@.
-loadRecipes :: MonadIO m => EntityMap -> m (Either Text [Recipe Entity])
-loadRecipes em = runExceptT $ do
-  fileName <- liftIO $ getDataFileName "recipes.yaml"
-  res <- liftIO $ decodeFileEither @[Recipe Text] fileName
-  textRecipes <- res `isRightOr` (from . prettyPrintParseException)
+loadRecipes :: (Has (Lift IO) sig m) => EntityMap -> m (Either Text [Recipe Entity])
```
byorgey, line 122: Eventually I'd love to rewrite things like this that use `IO` in terms of some finer-grained custom effects.  E.g. in this case we could have a custom effect for loading data files.  But we don't have to do it in this PR.  In fact it might be nicer to just get something simple that works and then worry about fancy details later.

src/Swarm/Game/Step.hs
```diff
@@ -125,64 +132,62 @@ gameTick = do
 ------------------------------------------------------------
 
 -- | Perform operation on the monad with game state.
```
byorgey, line 134: Shouldn't this comment be deleted too?

src/Swarm/Game/Step.hs
```diff
@@ -125,64 +132,62 @@ gameTick = do
...
 -- | Get the entity (if any) at a given location.
-entityAt :: MonadState GameState m => V2 Int64 -> ExceptT Exn (StateT Robot m) (Maybe Entity)
-entityAt loc = doOnGame $ zoomWorld (W.lookupEntityM (W.locToCoords loc))
+entityAt :: (Has (State GameState) sig m) => V2 Int64 -> m (Maybe Entity)
+entityAt loc = zoomWorld (W.lookupEntityM @Int (W.locToCoords loc))
 
 -- | Modify the entity (if any) at a given location.
 updateEntityAt ::
-  MonadState GameState m =>
+  (Has (State GameState) sig m, Has (State Robot) sig m, Has (Throw Exn) sig m) =>
```
byorgey, line 155: Does this function really need `Has (State Robot) sig m` and `Has (Throw Exn) sig m)`?

src/Swarm/Game/Step.hs
```diff
@@ -125,64 +132,62 @@ gameTick = do
...
+robotNamed nm = use (robotMap . at nm)
 
 -- | Manhattan distance between world locations.
 manhattan :: V2 Int64 -> V2 Int64 -> Int64
 manhattan (V2 x1 y1) (V2 x2 y2) = abs (x1 - x2) + abs (y1 - y2)
 
 -- | Generate a uniformly random number using the random generator in
 --   the game state.
-uniform :: (MonadState GameState m, UniformRange a) => (a, a) -> ExceptT Exn (StateT Robot m) a
+uniform :: (Has (State GameState) sig m, UniformRange a) => (a, a) -> m a
```
byorgey, line 171: Eventually, it would also be nice to write this in terms of some kind of randomness effect, which we could then interpret in terms of `State GameState`.

src/Swarm/Game/Recipe.hs
polux: Agreed, my goal for this PR was to convert the code to fused-effects with the smallest code changes possible. We can then have one PR per new effect that we introduce. It sure would be great if logging could be an effect for instance.

src/Swarm/Game/Step.hs
polux: Good catch! Done.

src/Swarm/Game/Step.hs
polux: Good catch again, it doesn't.

Issue #84, opened 2021-09-26
In game command reference


Anrock:
**Is your feature request related to a problem? Please describe.**
So far the only way to know about command is either tutorial file or checking part description on installed parts.
This isn't really convenient because:
1) Command descriptions are sort of scattered in flavour text
2) You have to have an entity with specific part to check it description (can't check move\turn commands if you don't have a robot for example)
3) Some commands are not described like "view" or "back" argument of turn.
4) Tutorial file sometimes goes out of sync with the game

**Describe the solution you'd like**
Centralized command reference which is available globally (can also be combined with syntax reference and maybe map legend for #49)
Command entry should consist of:
command (function) name, effects, type signature, required item/capability.

Preferably reference should be autogenerated to avoid synchronization issue like with tutorial file.

**Describe alternatives you've considered**
Leaving things as is.

**Additional context**
Check out [Colobot game](https://colobot.info/) which is pretty similar game about robot programming. It has a [SatCom](https://colobot.info/SatComOnline/master/cbot) system which is basically an ingame wiki with programming language syntax, API and list of entities.

TristanCacqueray:
Thank you for writing down this issue and providing the *Colobot* example, that's super valuable.

I actually started to put together a Help panel in #83 with a simple list of command. Though I'm not sure this should be the place for such centralized command reference since all the command are not available by default.

byorgey:
Thanks for making this issue.  I think some kind of in-game language reference like this will be super valuable, and perhaps it could be incorporated into the help panel.

One thing we should consider is how comprehensive the in-game language reference should be.  Right now (though I could be convinced otherwise) I don't think it should cover absolutely everything.  You want to leave some fun things for players to discover for themselves, and also don't want it to be overwhelming for new players.  However, I am totally OK with having a completely comprehensive language reference externally to the game, on a wiki or `readthedocs.io` or something like that.

Anrock:
> Though I'm not sure this should be the place for such centralized command reference since all the command are not available by default.

I was thinking it should be unlockable. So command entries should be added only after you have obtained a corresponding part or a recipe for crafting it is known.

byorgey:
Thinking more about this.  Right now `generateModal` in `TUI.View` creates the (static) help window.  I guess we can just pass along the `GameState` to the `generateModal` function so that it can figure out what help to display based on discovered devices.

[...]

PR #265, opened 2021-10-29
Add `has` and `count` commands

### LINKED ISSUE #88: `has` command
We should add a command `has : string -> cmd bool` which determines whether the robot has any of the named entity in its inventory.  This would be useful for implementing things like `giveAll`:
```
def whileC : forall a. cmd bool -> cmd a -> cmd () = \test. \body. 
  b <- test; 
  if b (body ; whileC test body) {} 
end;
def giveAll : string -> string -> cmd () = \other. \thing.
  whileC (has thing) { 
    give other thing 
  } 
end
```

byorgey:
Closes #88.

Currently, these commands require no capability and no specific device (you already have an inventory, so why wouldn't you be able to look in it to see what you have?).  However, that makes it impossible to discover these commands without reading the source or tutorial.  Maybe we should make an "inventory" device?  I'm not sure.  Suggestions welcome.  Or maybe once we implement #84 it won't matter since players can discover commands that way.

### DIFF
--- a/data/entities.yaml
+++ b/data/entities.yaml
@@ -265,6 +265,7 @@
   description:
   - An 8-bit counter.
   properties: [portable]
+  capabilities: [count]
 
 - name: Linux
   display:
--- a/data/entities.yaml
+++ b/data/entities.yaml
@@ -263,7 +263,13 @@
     attr: device
     char: 'C'
   description:
-  - An 8-bit counter.
+  - "Increment and reset, what more could you want?"
+  - |
+    A counter enables the command 'count : string -> cmd int',
+    which counts how many occurrences of an entity are currently
+    in the inventory.  Note also the 'has' command, which returns
+    a bool instead of an int and does not require any special
+    device.
   properties: [portable]
   capabilities: [count]
 
--- a/data/entities.yaml
+++ b/data/entities.yaml
@@ -267,9 +267,9 @@
   - |
     A counter enables the command 'count : string -> cmd int',
     which counts how many occurrences of an entity are currently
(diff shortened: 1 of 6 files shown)


data/entities.yaml
```diff
@@ -265,6 +265,7 @@
   description:
   - An 8-bit counter.
   properties: [portable]
+  capabilities: [count]
```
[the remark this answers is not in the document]
byorgey: Good catch, done.

byorgey:
> As a stop-gap measure, could you please add `has` to help?

Good idea, done!

data/entities.yaml
```diff
@@ -263,8 +263,15 @@
...
     char: 'C'
   description:
-  - An 8-bit counter.
+  - "Increment and reset, what more could you want?"
+  - |
+    A counter enables the command 'count : string -> cmd int',
+    which counts how many occurrences of an entity are currently
+    in the inventory.  Note also the 'has' command, which returns
+    a bool instead of an int and does not require any special
+    device.
```
xsebek, line 272: I think this change would make the has/count situation clearer. Then again understanding text is subjective :slightly_smiling_face: 
```suggestion
    in the inventory.  This is an upgraded version of the 'has'
    command, which returns a bool instead of an int and does
    not require any special device.
```

xsebek (approved):
Looks like it is ready to land. :slightly_smiling_face:  :rocket:

data/entities.yaml
byorgey: I like it, thanks.

PR #266, opened 2021-10-29
Don't make `detonator` a standard device

### LINKED ISSUE #260: Don't make detonator standard
`smyds` in IRC related accidentally self-destructing `base`.  It made me reflect that we don't really need `selfdestruct` so much any more, now that we have `salvage`.  We can leave the command and `detonator` in, but just don't install `detonator` as standard any more.

- [ ] remove `detonator` from list of standard devices
- [ ] update tutorial to not mention `selfdestruct`

We can also think about a recipe for detonator, but that can come later.

byorgey:
Closes #260.

### DIFF
--- a/TUTORIAL.md
+++ b/TUTORIAL.md
@@ -76,11 +76,6 @@ always returns a string which is the name of the robot that was built;
 it may be different than the name you specified if there is already
 another robot with that name.
 
-Note that if you don't want a robot to hang around after completing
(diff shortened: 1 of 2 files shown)


xsebek (approved):
LGTM, makes #35 need one less recipe. :smirk:

byorgey:
Well, in terms of #35, I think whether we want to start the base with any detonators at all is a separate discussion.

PR #270, opened 2021-10-29
Add location to type matching errors

### LINKED ISSUE #268: Report error in def body only
Currently typing errors are shown as a problem of the expression - which usually means the whole file.

Good error:

Bad error:

In principle we should be able to report the error in the body of definition where the error occurred. The previous `def`s, `let`s and commands must have been fine as we checked them. They could of course lead to the failure in some later part, but only indirectly.

At minimum we should be able to split at `def`s as those must be at global scope and each individually typecheck.

xsebek:
- closes #268 
- computes the location for `SBind` and operators

### DIFF
--- a/src/Swarm/Language/Parse.hs
+++ b/src/Swarm/Language/Parse.hs
@@ -213,13 +213,17 @@ parseConst = asum $ map alternative consts
   consts = filter isUserFunc allConst
   alternative c = c <$ reserved (syntax $ constInfo c)
 
+-- | Add 'Location' to a parser
+parseLocG :: Parser a -> Parser (Location, a)
+parseLocG pa = do
+  start <- getOffset
+  a <- pa
+  end <- getOffset
+  pure (Location start end, a)
+
 -- | Add 'Location' to a 'Term' parser
 parseLoc :: Parser Term -> Parser Syntax
-parseLoc pterm = do
-  start <- getOffset
-  term <- pterm
-  end <- getOffset
-  pure $ Syntax (Location start end) term
+parseLoc pterm = uncurry Syntax <$> parseLocG pterm
 
 parseTermAtom :: Parser Syntax
 parseTermAtom =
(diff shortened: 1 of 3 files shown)


xsebek:
There might be other escape routes or a better way to annotate the error, but his seems to help. :man_shrugging:

byorgey:
This looks OK, but is there any reason not to add a single call to `addLocToTypeError` at the very top of `inferModule`?  As it is, it will only catch any errors that happen specifically on the two lines where you added it.

xsebek:
@byorgey nope, I was just afraid you would beat me to it, so I pushed the first thing that worked :smile:

byorgey:
> nope, I was just afraid you would beat me to it

lol, I was actually working on it when you pushed this, so well done :laughing:

byorgey (approved):

src/Swarm/Language/Parse.hs
```diff
@@ -329,13 +329,13 @@ parseExpr = fixDefMissingSemis <$> makeExprParser parseTermAtom table
 exprLoc2 :: Parser (Syntax -> Syntax -> Term) -> Parser (Syntax -> Syntax -> Syntax)
```
byorgey, line 329: We should update the comment too.

src/Swarm/Language/Typecheck.hs
```diff
@@ -280,8 +280,10 @@ inferModule s@(Syntax _ t) = case t of
   -- correct context when checking the right-hand side in particular.
   SBind mx c1 c2 -> do
     -- First, infer the left side.
-    Module cmda ctx1 <- inferModule c1
-    a <- decomposeCmdTy cmda
+    (a, ctx1) <- (`catchError` addLocToTypeErr c1) $ do
```
byorgey, line 283: Why do we need this here?  Just curious.

xsebek: Oh, I was just going crazy from bind chains - turns out they were missing location themselves and could not help to limit the error scope. I will make this simple again :+1:

xsebek:
The error scope for type mismatch seems to be reasonable now. :partying_face:
But I better add some unit tests now that it is no longer a two line change :sweat_smile:

src/Swarm/Language/Parse.hs
xsebek: Done and I made it a local definition too :+1:

byorgey:
LGTM!

PR #273, opened 2021-10-30
allow ' in variable names

### LINKED ISSUE #269: Allow `'` in variable names
But we could indeed allow `'` in variable names, I don't think there's any particular reason not to.  People coming from Haskell will expect it.

_Originally posted by @byorgey in https://github.com/byorgey/swarm/issues/267#issuecomment-954960523_

Alexander-Block:
Closes #269 

- The parser now accepts  `'` in a variable name as long as it does not start with `'` (to rule out potentially problematic variable names like `'a'`).
- Unit tests for this are added.


byorgey (approved):
LGTM!

PR #274, opened 2021-11-02
Wrap errors shown in game

### LINKED ISSUE #241: Wrap errors shown in game
I find it annoying when the Error window does not show the full text of the exception.

For example:

Screenshot from 2021-10-16 17-47-29

Here we should wrap the list of commands and make the window larger.

Alexander-Block:
Closes #241 .

This PR implements two changes:
- The error message window uses as much horizontal space as necessary now, wrapping lines when no further horizontal expansion is possible. 
- The parser groups "built-in user functions" and "direction constants" under labels of the same names. This leads to more concise parse error messages.


src/Swarm/TUI/View.hs
```diff
@@ -171,9 +170,12 @@ chooseCursor s locs = case s ^. uiState . uiModal of
...
 
--- | The error dialog window.
-errorDialog :: Dialog ()
-errorDialog = dialog (Just "Error") Nothing 80
+-- | Render the error dialog window with a given error message
+renderErrorDialog :: Text -> Widget Name
+renderErrorDialog err = renderDialog (dialog (Just "Error") Nothing requiredWidth) errContent
+ where
+  errContent = txtWrapWith indent2 {preserveIndentation = True} err
+  requiredWidth = 2 + maximum (textWidth <$> T.lines err)
```
xsebek, line 178: Shouldn't the width and height be capped at some value so it does not cover the whole screen in worst case?

Alexander-Block: Thank you for this feedback.

I think you are right about these caps. I intentionally decided not to cap the width because covering the whole screen horizontally did not seem too bad to me. However, now that you point it out explicitly, I can imagine that it can be bothersome on a comparatively big screen. Since we already have a build-in cap in `renderModal` of 500 for the width, I propose to reuse this also for the error message window. This will still allow for a horizontal coverage of a comparatively small screen. 

I am not quite sure about a good cap for the height. In fact I think that we should be able to scroll up and down in the error message window in case we have error messages big enough to cover a significant part of the screen vertically and decide to cap the height, but I think that implementing this would be a whole different task. This would probably also be helpful for the help window and future modal windows, though. I can write an issue for this, if you also agree that this is desirable.

xsebek: 500 is an insane width :eyes: 
Screenshot from 2021-11-02 17-20-35
If you can see the number in REPL below, that is 400 digits long. I would say `swarm` is not playable at this zoom level because we are not refreshing the cells and REPL efficiently.

I do not have a particular stance toward the UI, maybe @byorgey has some vision? :slightly_smiling_face:

byorgey: 500 does seem kind of ridiculous, I have no idea why that number was chosen for `renderModal` in the first place.  What about something like capping it at either 100 characters or 90% of the available width, whichever is smaller?

byorgey (approved):

byorgey:
Just looking through open PRs.  I see this one has been open for a while --- I am going to merge it.  We can continue adjusting the modal width later.

PR #276, opened 2021-11-04
Hide unknown entities with `?`

### LINKED ISSUE #49: Mechanisms for world exploration & discovery
It's easy to move the map around, but I couldn't tell what features were interesting or how to interact with them. The tutorial explains trees, which is fine, and I guessed the yellowish C was copper, but to do anything with copper I need a furnace, which needs stone. Where does that come from? I can't point at a tile on the map and ask "is this stone?", and there are multiple grey glyphs which plausibly could be.

xsebek:
- closes #49 

**TODO:**
- [x] update TUTORIAL.md
- [x] add a creative mode check


xsebek:
The work was already done by @byorgey, I just added the check in `drawCell`:

byorgey (approved):
Looks good to me.  I guess in creative mode you should just be able to see everything?  Of course right now that enables you to pretty easily "cheat" by switching to creative mode quickly and then back, but eventually there won't be a keyboard shortcut for switching modes, you'll only be able to pick the mode when you start a new game.

xsebek:
@byorgey I slightly changed the tutorial, could you please take a look? :slightly_smiling_face:

src/Swarm/TUI/View.hs
```diff
@@ -323,11 +327,17 @@ drawWorld g =
                 str [lookupDisplay ((r ^. robotOrientation) >>= toDirection) (r ^. robotDisplay)]
           Nothing -> eWidget
 
--- | Draw a single cell of the world.
-drawCell :: W.Coords -> W.World Int Entity -> (Int, Widget Name)
-drawCell i w = case W.lookupEntity i w of
-  Just e -> (e ^. entityDisplay . displayPriority, displayEntity e)
+-- | Draw a single cell of the world, either hiding those that current robot does not know,
+--   or hiding all/none depending on Left value (True/False).
+drawCell :: W.World Int Entity -> Either Bool Robot -> W.Coords -> (Int, Widget Name)
```
byorgey, line 332: How about making a custom data type for this, like `data Hiding = HideAll | HideNone | HideUnknownTo Robot`?  Stuff like `Either Bool` is just confusing, and you always have to read the documentation to remember what `Left True` is supposed to represent and so on.

TUTORIAL.md
```diff
@@ -190,46 +195,48 @@ arrow, like so:
...
+
+Let's build one more robot called `"runner"`. It will get renamed
+to something else to avoid a name conflict, but we can capture its
+name in a variable using the above syntax.
+Then we can use the `view` command to focus on it instead of the base:
 ```
-name <- build "mover" {m8; m8; m4}; view name
+r <- build "runner" { turn west; m4; m }; view r
 ```
 Note that `base` executes the `view name` command as soon as it
```
byorgey, line 209:
```suggestion
Note that `base` executes the `view r` command as soon as it
```

TUTORIAL.md
```diff
@@ -190,46 +195,48 @@ arrow, like so:
...
 ---------
 
 So what is all this stuff everywhere?  Let's find out!  When you
 `build` a robot, by default it starts out with a `scanner` device,
-which you may have noticed in `mover1`'s inventory.  You can `scan`
+which you may have noticed in `runner1`'s inventory.  You can `scan`
 items in the world to learn about them, and later `upload` what you
 have learned to the base.
 
 Let's build a robot to learn about those green `T` things to the west:
```
byorgey, line 234:
```suggestion
Let's build a robot to learn about those green `?` things to the west:
```

TUTORIAL.md
```diff
@@ -190,46 +195,48 @@ arrow, like so:
...
 ```
 build "s" {turn west; m4; move; scan west; turn back; m4; upload "base"}
 ```
-The `turn` command causes a robot to turn, of course. It takes a
-direction as an argument, which can be either an absolute direction
+The `turn` command we used to turn the robot takes a direction as an
+argument, which can be either an absolute direction
 (`north`, `south`, `east`, or `west`) or a relative direction
 (`forward`, `back`, `left`, `right`, or `down`).
```
byorgey, line 242: The below should say "walk on top of a `?` to learn about it".

TUTORIAL.md
```diff
@@ -406,7 +411,7 @@ lakes or in the middle of a plain. Either way, you have established
 your base in the shade of what you assume is a tree and now can send
 out robots to explore!
 
-![World generated with seed 16](./images/world16.png)
+![World generated with seed 16](images/tutorial/world16.png)
```
byorgey, line 414: Should we also regenerate this image to have `?`s?

byorgey (approved):

TUTORIAL.md
xsebek: Good point, I tried to kill two birds with one stone and used Creative mode that is in next section :slightly_smiling_face:

src/Swarm/TUI/View.hs
xsebek: I had exactly those thoughts when clicking "Push" :smile:

xsebek: Thanks for the name, done :heavy_check_mark:

PR #278, opened 2021-11-05
Use mutable arrays for entities

### LINKED ISSUE #30: Switch entity tiles to use *mutable* arrays
Right now we store the world as a triple: a collection of immutable terrain tiles, a collection of immutable entity tiles, and a map specifying which cells have been updated from their original entity values.  However, there is no good reason that we can't just use *mutable* arrays for entity tiles.  Running the game already takes place in a monad; we just have to make sure that it supports the mutable array operations.  This should hopefully make the implementation simpler and speed up the game as well.

twitu:
Closes #30

### DIFF
--- a/src/Swarm/Game/Step.hs
+++ b/src/Swarm/Game/Step.hs
@@ -150,7 +150,7 @@ zoomWorld n = do
   return a
 
 -- | Get the entity (if any) at a given location.
-entityAt :: (Has (State GameState) sig m) => V2 Int64 -> m (Maybe Entity)
+entityAt :: (Has (State GameState) sig m, Has (Lift IO) sig m) => V2 Int64 -> m (Maybe Entity)
 entityAt loc = zoomWorld (W.lookupEntityM @Int (W.locToCoords loc))
 
 -- | Modify the entity (if any) at a given location.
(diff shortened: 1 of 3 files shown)


xsebek:
If possible, I would really prefer using `ST` - with a quick look, this should be possible with the instances ([1](https://hackage.haskell.org/package/base-4.13.0.0/docs/Control-Monad-ST-Safe.html#t:ST),[2](https://hackage.haskell.org/package/fused-effects-1.1.1.1/docs/Control-Effect-Lift.html#t:Lift)):
```haskell
Monad (ST s)
Monad m => Algebra (Lift m) (LiftC m)
```
Which should allow you to write
```
Has (Lift (ST s)) sig m
```
Honestly I am not sure what `s` is, maybe we get it when creating the array with [`newListArray`](https://hackage.haskell.org/package/array-0.5.4.0/docs/Data-Array-MArray.html#v:newListArray)?

If you could point to where you run into trouble, I could take a look once I finish my PRs. :slightly_smiling_face:

twitu:
@xsebek the IO Array was just an experiment to see how it fits. In fact I don't think we've concluded whether to add mutability. So this is more like an experiment to benchmark the performance difference.

Coming back to the IO implementation part. It's pretty difficult to implement since IO infects everything. I've added more details in [my comment](https://github.com/byorgey/swarm/issues/30#issuecomment-961966274) in the issue.

I will be trying out implementing this using ST Array next.

byorgey:
Going to close this since it seems like we have come to consensus to not go in the direction of mutable data structures; see the discussion on #30 .

byorgey:
Going to close this since it seems like we have come to consensus to not go in the direction of mutable data structures; see the discussion on #30 .

PR #279, opened 2021-11-06
Filter paths that run CI

### LINKED ISSUE #221: Add rule to trigger github ci only when code files change
Currently github ci triggers for every commit even if it means changes in example and README files. This can be improved by only triggering ci on code file changes.

This can be fixed by adding `rules` to `haskell-ci.yml`. An example take from this [SO answer](https://stackoverflow.com/questions/40712275/gitlab-ci-how-to-trigger-a-build-only-if-changes-happen-on-particular-set-of-f) is given below.

```
  rules:
    - changes:
      - /*.{java, py} # ... or whatever your file extension is
```

N/A

**Additional context**
Currently there are no end to end tests for examples (they might not be relevant) right now. When tests for examples are added they can also be added to changed files rules.

xsebek:
- only build and test when `.hs`, `.cabal` or `stack.yaml` files change
- closes #221


xsebek:
Cool, it seems like the rules apply to the PR itself :slightly_smiling_face:

.github/workflows/haskell-ci.yml
```diff
@@ -14,8 +16,17 @@
...
-  - push
-  - pull_request
+  push: 
+    paths:
+      - '**.hs'
+      - '**.cabal'
+      - 'stack.yaml'
+  # reuse the same filter for pull-requests
+  pull_request:
+    paths:
```
xsebek, line 26: Surely we could avoid this duplicity with YAML anchors, right? [Well, no.](https://github.community/t/support-for-yaml-anchors/16128) :disappointed:

.github/workflows/haskell-ci.yml
```diff
@@ -14,8 +16,17 @@
 #
 name: Haskell-CI
 on:
-  - push
-  - pull_request
+  push: 
+    paths:
+      - '**.hs'
+      - '**.cabal'
+      - 'stack.yaml'
```
TristanCacqueray, line 23: Should it also keep `**.sw` and `cabal.project` ? And what about `**.yaml` for static tool configs (e.g. fourmolu.yaml) and entities if they are used in test?

TristanCacqueray: Or is there a `irrelevant-path` option we could set to `**.md` instead?

xsebek: There is, but I prefer adding .sw and cabal as I plan to add editor support which would waste Haskell CI.

xsebek: Not sure about .yaml, maybe “data/**.yaml” and “*.yaml” but not “**.yaml” would work?

TristanCacqueray (approved):
Thanks!

byorgey (approved):
I guess since you requested my review it won't actually merge it until I approve?

xsebek:
@byorgey Sorry to bother you, I shall try this again then. 😅

xsebek:
So the issue is that mergify needs the jobs to be successful. I will try to find a good solution, maybe I could make a dummy job with paths-ignore that would always pass and then use mergify `or`? 😕

xsebek:
> Configuration changed. This pull request must be merged manually 

Right, hopefully it works :worried:

PR #283, opened 2021-11-07
Add the CI file check to mergify

xsebek:
I made a mistake in #279, for PRs that modify both .hs an .md the dummy would pass :roll_eyes: 

Copying the logic to `.mergify.yml` should fix this. :crossed_fingers:


xsebek:
The used syntax is described here:
- https://docs.mergify.com/configuration/#regular-expressions
- https://docs.mergify.com/conditions/#combining-conditions-with-operators

Issue #212, opened 2021-10-12
Opaque robot type


byorgey:
Currently, all commands which need to identify a robot (e.g. `install`, `give`, `reprogram`) take a `string` identifying the robot by name.  We also store robots by name internally in the `robotMap`.  However, this has several disadvantages: it's not very type safe (for example, are you supposed to `give "base" "tree"` or `give "tree" "base"`?  And if you misspell the name of a robot, it fails at runtime instead of typechecking time), and necessitates frequent, relatively costly string comparisons.

The basic idea is to add a new type `robot` which is opaque to the user but is internally represented by `Int`.  Here are my current thoughts:

* Add a new base type `robot`.
* Change the type of any commands that currently take or return the name of a robot:
    * `give : robot -> string -> cmd ()`
    * `install : robot -> string -> cmd ()`
    * `reprogram : robot -> cmd a -> cmd ()`
    * build : cmd a -> cmd robot`
* Note that robots can still *have* a name, which is displayed when they are `view`ed.  But the name no longer needs to be unique.
* However, note that the `build` command doesn't take a name any more; in many cases players are probably fine with giving it a descriptive *variable* name (e.g. `lambda_getter <- build get_lambda`), but never plan to `view` it so don't really care what its "display name" is.
* However, we can add a command `setname : string -> cmd ()` which lets a robot set its display name.
* Internally, we can add a new type of `Value`, `VRobot :: Int -> Value`.  It can pretty-print as something like `r105` or `<105>`, it doesn't really matter that much.
* What about robots that want to e.g. `give` something back to, say, the base?  How can they get a reference to it?  I propose adding a special variable `parent : robot` which is always a reference to a robot's parent (i.e. the robot that built it).  So the code for building a robot to get you a thing might look like this:
  ```
  child <- build {move; thing <- grab; turn back; move; give parent thing}
  ```
  While we're at it we could also add a special global variable `base : robot` which is a reference to the base.
* We might also want to add a way to get the results of previous commands entered at the REPL, so if you did `build` and forgot to bind the result to something, it's not lost forever.

byorgey:
I *think* it would be safe to return a handle to a non-existent robot (but I could be convinced otherwise).  We would just have to make sure that robot handles are never reused, but that should be easy.

byorgey:
Agreed!  `meet : cmd (() + robot)` sounds great.  I presume if there is more than one robot there, you just get an arbitrary one?  Maybe later there could be a `meetAll : cmd (list (robot))`.

PR #286, opened 2021-11-17
async as

xsebek:
- use the `async` library to check on `as` computation until it is done


byorgey:
I don't know if you saw [my other recent comment](https://github.com/byorgey/swarm/pull/285#issuecomment-971741855) but I am skeptical of this whole approach.  `as` will primarily (exclusively?) be used when checking win conditions for challenges.  Such win condition checking programs are run *to completion* on *every tick*.  So if an `as` command forks off a computation into another thread, it seems to me it won't help: the CESK machine will simply sit there busy waiting 8 ticks at a time until the computation in the other thread finishes.

xsebek (commented):
Hey, @byorgey I got the background `as` to work :slightly_smiling_face: 

There a few things to iron out, mainly around printing, but the core of the change is pretty small (new async value, step (as&poll), IO cancel).

xsebek:
@byorgey right, did not see your comment here :slightly_smiling_face: 

>  win condition checking programs are run to completion on every tick.

Sure, it is easy that way, but I thought the original idea was to delegate it to some invisible system robot?
That would also allow multiple win conditions in a "do X, then do Y" style.
Also this makes it easy to add a TTL check, so that winning condition does not cycle.

Either way it should not stop us from making our godly robots more omnipotent and I do not see much overhead with this async change (we are passing around current immutable state, which is cheap). :angel:

byorgey:
> Sure, it is easy that way, but I thought the original idea was to delegate it to some invisible system robot? 

I think I considered that briefly but decided against it since I don't see much benefit, besides possibly saving a tiny bit of code because we already have code to run programs contained in robots.  But in the end I think we would need a bunch of special cases either way.  A win condition checker is fundamentally different than a robot.

Actually I'm not even sure I see how you could delegate the win checking to an invisible system robot.  How would that work, exactly?  What program would it run?  Some kind of infinite loop?  And how would it signal that you have won?

> That would allow multiple win conditions in a "do X, then do Y" style. 

Can you elaborate on how that would work?  I am honestly curious, I can't imagine how it would work.  Or at least I don't see how running the win condition checking in a special system robot vs as a separate program would make this any easier to support.

> Also this makes it easy to add a time check, so that winning condition does not cycle.

Right, but to me this doesn't really seem like a problem worth solving.

byorgey:
Stepping back, part of me thinks this is cool, but part of me wonders whether it is extra complication for little benefit.  I would be happy to be wrong, but it feels to me like a solution without a problem.  Can you explain exactly what problem this is supposed to be solving?

xsebek:
@byorgey Thinking about it some more, your `as` is so generic already, that I think there is not much this brings to challenges directly. There is still the indirect benefit of making this REPL command less dangerous, which could come handy when testing.

Long term I would love to take advantage of more CPUs (CESK only runs on one) in game and the `as` command seemed as it could work and benefit from this change :thinking:   I am kind of interested in this oracle power and what other time traveling is possible :slightly_smiling_face: 

> Can you elaborate on how that would work?

I am not sure, but would the CESK machine in robot turn `{cmd bool}` into `Up VBool`? I kind of hoped we would put the robot inside and only check if it finished/restart it on failure.
If that did not work, I can imagine a more direct signalling with e.g. `create "unlock part 1"` or #212 style `renameSelf "done"`.

With the more direct signalling, other robots could wait&check if others finished. My dream tutorial looks like this:

> Robot R1 is waiting for the base to get tree, then it signals first check point is finished.
> After that R2 takes over and checks for boat in inventory, then reprograms the base to move to another island and signals.
> R3 then checks for stone and copper ore in inventory...

src/Swarm/Game/CESK.hs
```diff
@@ -288,13 +289,14 @@ idleMachine :: CESK
...
+cancel cesk = (\s -> Out VUnit s []) <$> s'
  where
-  s' = resetBlackholes $ getStore cesk
-  getStore (In _ _ s _) = s
-  getStore (Out _ s _) = s
-  getStore (Up _ s _) = s
+  s' = resetBlackholes <$> getStore cesk
+  getStore :: CESK -> IO Store
+  getStore (In _ _ s _) = pure s
+  getStore (Out v s _) = (case v of VAsync (GAsync _ as) -> Async.cancel as; _ -> pure ()) >> pure s
```
byorgey, line 298: This isn't guaranteed to cancel all currently running async computations, because there are lots of other places `VAsync` values could be hiding: in the environment, in a closure, ...  I don't know how important this is.  Perhaps not that important, since if an async computation keeps running but nothing ever asks for its value it only wastes some CPU cycles, it won't cause any crashes or anything like that.  If it is important, we would have to do something like put references to currently outstanding async computations in the store, so that we can find them all when we want to cancel them.

src/Swarm/Game/Step.hs
```diff
@@ -913,12 +919,52 @@ execConst c vs s k = do
...
         -- capable of executing any commands; the As command
         -- already requires "God" capability.
-        v <-
-          evalState @Robot (r & systemRobot .~ True) . evalState @GameState g $
-            runCESK (Out prog s [FApp (VCApp Force []), FExec])
+
+        -- -----------------------------------------------------------------------------
+        -- TODO: move this code out
+        -- Slightly edited copy of Unit.hs version of runCESK:
+        let runCESK' :: CESK -> StateT Robot (StateT GameState IO) (Either Text Value)
```
byorgey, line 926: Don't we already have this function in `Step.hs`?  Hmm, or maybe that's only in my `challenge-mode` branch?  Feel free to copy it: https://github.com/byorgey/swarm/pull/285/files#diff-bb4a5ea570cea9f815cdd1d92b33d4b6bd8daaa7d74c7014916c642c159e7d80R147-R173

src/Swarm/Game/Step.hs
```diff
@@ -913,12 +919,52 @@ execConst c vs s k = do
...
+
+        va <-
+          sendIO . Async.async $
+            either VString id -- TODO: that is not right...
+              <$> evalAs
 
         -- Return the value returned by the hypothetical command.
-        return $ Out v s k
+        i <- gensym <+= 1
+        return $ Out (VAsync $ GAsync i va) s (FApp (VCApp Poll []) : k)
```
byorgey, line 949: Conceptually, it probably makes more sense to make a new kind of `Frame` called `FPoll` rather than making `Poll` a primitive function.

src/Swarm/Game/Step.hs
```diff
@@ -913,12 +919,52 @@ execConst c vs s k = do
...
+      [VAsync ga@(GAsync i va)] -> do
+        let fs :: Show a => a -> Text
+            fs = from . show
+        rn <- use robotName
+        time <- use ticks
+        -- DEBUG ------------------------------------------------------
+        mv <- sendIO $ Async.poll va
+        -- let mv = Nothing :: Maybe (Either SomeException Value)
+        let (msg, cesk) = case mv of
+              Nothing -> ("in progress", Waiting (time + 8) (Out (VAsync ga) s (FApp (VCApp Poll []) : k)))
```
byorgey, line 961: I wonder if it makes sense to use some kind of exponential backoff (up to a maximum), e.g. if it hasn't finished after 8 steps, then the next time we wait 16, then 32, ... up to a maximum of 64 or something like that?  For a long-running computation, checking every 8 time steps seems like a lot of polling, especially if we are writing to the robot log every time we check.  We could put the waiting time to use as an argument of `FPoll`, i.e. `FPoll n` means "check if the async computation is done, and if not, wait `n` time steps until checking again".

src/Swarm/Game/Step.hs
```diff
@@ -1313,12 +1359,19 @@ evalCmp c v1 v2 = decideCmp c $ compareValues v1 v2
...
-compareValues = \v1 -> case v1 of
+compareValues v1 = case v1 of
   VUnit -> \case VUnit -> return EQ; v2 -> incompatCmp VUnit v2
-  VInt n1 -> \case VInt n2 -> return (compare n1 n2); v2 -> incompatCmp v1 v2
   VString t1 -> \case VString t2 -> return (compare t1 t2); v2 -> incompatCmp v1 v2
   VDir d1 -> \case VDir d2 -> return (compare d1 d2); v2 -> incompatCmp v1 v2
   VBool b1 -> \case VBool b2 -> return (compare b1 b2); v2 -> incompatCmp v1 v2
+  VInt n1 -> \case
+    VInt n2 -> return (compare n1 n2)
+    VAsync (GAsync n2 _) -> return $ compare n1 (fromIntegral n2)
```
byorgey, line 1369: I don't understand this at all.  Why would we compare a `VInt` against the number in a `GAsync`?

src/Swarm/Game/Step.hs
```diff
@@ -1313,12 +1359,19 @@ evalCmp c v1 v2 = decideCmp c $ compareValues v1 v2
...
+    v2 -> incompatCmp v1 v2
+  VAsync (GAsync n1 _) -> \case
+    VAsync (GAsync n2 _) -> return $ compare n1 n2
+    VInt n2 -> return $ compare (fromIntegral n1) n2
+    v2 -> incompatCmp v1 v2
```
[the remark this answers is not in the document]
byorgey: Yes, I don't think there is any reason for `VAsync` values to be comparable.

src/Swarm/Game/Value.hs
```diff
@@ -80,8 +82,19 @@ data Value where
   VDelay :: Term -> Env -> Value
   -- | A reference to a memory cell in the store.
   VRef :: Int -> Value
+  -- | A computation running in another thread.
+  VAsync :: GAsync Value -> Value
   deriving (Eq, Show)
 
+data GAsync a = GAsync
+  { asyncIndex :: Int
```
byorgey, line 90: What is the `asyncIndex` for?  Just to have a way to `Show` it?

src/Swarm/Language/Syntax.hs
```diff
@@ -323,12 +323,15 @@ data Const
     AppF
   | -- God-like sensing operations
 
-    -- | Run a command as if you were another robot.
+    -- | Hypothetically run a command as if you were another robot.
+    --   Make sure this is the LAST command!
```
byorgey, line 327: What does "Make sure this is the LAST command" mean?

src/Swarm/Language/Syntax.hs
```diff
@@ -323,12 +323,15 @@ data Const
...
     As
+  | -- | Check if the running command has finished.
+    Poll
   deriving (Eq, Ord, Enum, Bounded, Data, Show)
 
 allConst :: [Const]
-allConst = [minBound .. maxBound]
+allConst = [minBound .. As] -- no Poll in code!
```
byorgey, line 334: Right, see, this is telling us that we really should make poll a type of `Frame` instead of a `Const`. =)

src/Swarm/Language/Syntax.hs
```diff
@@ -598,6 +602,11 @@ data Term
     --   be a special syntactic form so its argument can get special
     --   treatment during evaluation.
     SDelay DelayType Syntax
+  | -- | Type for values that can not be directly created in code.
+    --   They are shown as e.g. @(__r1__)@, like C-style internal mangled
+    --   variables. With custom character it is slightly better then
+    --   using TInt.
+    TNonCode Char Int
```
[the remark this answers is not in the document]
byorgey: No, I don't like this.  `VAsync` and `VRef` do not show up in the surface language, so there should not be a way to represent them in `Syntax`.  Pretty-printing `Value`s via `valueToTerm` was a horrible lazy hack in the first place; we should be trying to get rid of technical debt, not letting my lazy decisions in the past snowball into even more technical debt for ourselves.  We should either just do something simple and stupid for now in `valueToTerm` for `VAsync` (like we were doing for `VRef`), or bite the bullet and write a proper `Value` pretty-printer.

src/Swarm/Language/Typecheck.hs
```diff
@@ -384,6 +384,7 @@ infer s@(Syntax l t) = (`catchError` addLocToTypeErr s) $ case t of
     ty2 <- maybe id (`withBinding` Forall [] a) mx $ infer c2
     _ <- decomposeCmdTy ty2
     return ty2
+  TNonCode {} -> return UTyInt -- TODO: probably wrong, but I do not understand this code
```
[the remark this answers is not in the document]
byorgey: There is nothing sensible to do here because there is no way to know what the type of a `TNonCode` is.  Another reason that we should get rid of it. =)

src/Swarm/Language/Typecheck.hs
```diff
@@ -476,6 +477,7 @@ inferConst c = toU $ case c of
   Exp -> arithBinT
   AppF -> [tyQ| (a -> b) -> a -> b |]
   As -> [tyQ| string -> {cmd a} -> cmd a |]
+  Poll -> [tyQ| cmd a |] -- TODO: poll is NOT exported as command
```
byorgey, line 480: Right, `Poll` does not have any sensible type---yet another reason it should be a `Frame` instead of a `Const`.

src/Swarm/TUI/Controller.hs
```diff
@@ -401,28 +402,37 @@ stripCmd pty = pty
...
-            & gameState %~ execState (activateRobot "base")
-      Left err ->
-        continue $
-          s
-            & uiState . uiError ?~ txt err
-    else continueWithoutRedraw s
+handleREPLEvent s = \case
+  (VtyEvent (V.EvKey (V.KChar 'c') [V.MCtrl])) ->
+    let cancel' = do
+          cm <- let m = s ^. gameState . robotMap . to (! "base") . machine in cancel m
```
byorgey, line 408: Using `(! "base")` will crash if there is no robot named `"base"` which is a big no-no!  I will have to think about how to rewrite this code.

byorgey (changes requested):
I have tried to give some useful feedback even though I am still on the fence about whether it is worth merging this.  I guess it is worthwhile seeing how good we can make this before making a final decision about merging it.

I am not sure it is a good idea to always run `as` in an async computation.  Or perhaps we should have two versions, normal `as` which runs in an async computation, and `unsafeAs` which doesn't.  However (vague brainstorming) it might also be cool to put this more in the user's control, i.e. some kind of built-in `promise` type with `async` and `await` commands...?

byorgey:
> Long term I would love to take advantage of more CPUs (CESK only runs on one) in game

Agreed! Though I think we could also do this more simply by running multiple robots in parallel.

> I am not sure, but would the CESK machine in robot turn {cmd bool} into Up VBool? I kind of hoped we would put the robot inside and only check if it finished/restart it on failure.

Yes, that would work, but it would mean we have to specially handle that robot to check its return value and restart it.  We wouldn't really get much benefit from treating it as a robot as opposed to just a special standalone CESK machine.

> Robot R1 is waiting for the base to get tree, then it signals first check point is finished.
After that R2 takes over and checks for boat in inventory, then reprograms the base to move to another island and signals.
R3 then checks for stone and copper ore in inventory...

Yes, this sounds really cool, but I wonder if we could do this more simply by adding a mechanism for mutable state, aka "static variables".  If there is a CESK machine we run on every tick, when it notices that the first check point is finished, it sets a mutable variable in the store; when we run it on the next tick, we pass the same store to it so it can have state that persists across runs.  It then sees that the first checkpoint is complete so it only checks for the second checkpoint, and so on.

xsebek:
@byorgey I went with the `future` type and `async`, `await`, `poll` and `cancel` commands:

Screenshot from 2021-11-25 16-02-35

<sub>(Oops, the poll type is actually wrong, let me fix that...)</sub>

src/Swarm/Game/CESK.hs
xsebek: I think I do not want to cancel *all async computations* - the ones currently being run by base are enough.

We could store the async values in GameState and add some way to cancel all of them in the UI.

src/Swarm/Game/Step.hs
xsebek: Thanks, I will do so during rebase :slightly_smiling_face:

src/Swarm/Game/Step.hs
xsebek: I went the other way and made `poll` a real command. :man_shrugging:

src/Swarm/Game/Step.hs
xsebek: I let `await` take `int` waiting period, but it would be fun to give it `int -> int` function instead :smile:

src/Swarm/Game/Step.hs
xsebek: Fixed :slightly_smiling_face:

src/Swarm/Game/Value.hs
xsebek: I got rid of it, `Async` has hash and `Ord`, that is all I need :slightly_smiling_face:

src/Swarm/Language/Syntax.hs
xsebek: Deleted. All commands are now commands, as they should be. :sweat_smile:

src/Swarm/Language/Syntax.hs
xsebek: I went full commands instead. :slightly_smiling_face:

src/Swarm/TUI/Controller.hs
xsebek: Well I have no idea, let me know if you think of something :wink:

xsebek:
Hmm, `await` does not really work - it does not get run again by the machine and instead return after waiting once:
```
> a <- async "base" {testLIST_ALL; return 1}; await a
await async {force testLIST_ALL; return 1} : int
```
```
> def await0 = \f. e <- poll f; case e (\u. await0 f) return; end
> a <- async "base" {testLIST_ALL; return 1}; await0 a
1 : int
```
I tried with changing it to `functionLow` and it did not help :confused: 
Maybe the second form is not that bad and `await` is not really needed.

src/Swarm/Game/Step.hs
```diff
@@ -1311,12 +1375,13 @@ evalCmp c v1 v2 = decideCmp c $ compareValues v1 v2
...
 --   compared, or @Nothing@ if they cannot.
 compareValues :: Has (Throw Exn) sig m => Value -> Value -> m Ordering
-compareValues = \v1 -> case v1 of
+compareValues v1 = case v1 of
   VUnit -> \case VUnit -> return EQ; v2 -> incompatCmp VUnit v2
   VInt n1 -> \case VInt n2 -> return (compare n1 n2); v2 -> incompatCmp v1 v2
   VString t1 -> \case VString t2 -> return (compare t1 t2); v2 -> incompatCmp v1 v2
   VDir d1 -> \case VDir d2 -> return (compare d1 d2); v2 -> incompatCmp v1 v2
   VBool b1 -> \case VBool b2 -> return (compare b1 b2); v2 -> incompatCmp v1 v2
+  VAsync _ a1 -> \case VAsync _ a2 -> return $ compare a1 a2; v2 -> incompatCmp v1 v2
```
byorgey, line 1384: I don't really think comparing `VAsync` values makes much sense.  What would you use this for?  And what would it mean for one to be less than another?

src/Swarm/Game/Value.hs
```diff
@@ -80,8 +84,17 @@ data Value where
   VDelay :: Term -> Env -> Value
   -- | A reference to a memory cell in the store.
   VRef :: Int -> Value
+  -- | A computation running in another thread.
+  VAsync :: Term -> GAsync Value -> Value
   deriving (Eq, Show)
 
+-- | Type for asynchronous computation which may raise an Exn.
+newtype GAsync a = GAsync {asyncComp :: Async.Async (Either Exn a)}
```
byorgey, line 92: Do we really need the `GAsync` type?  What's the benefit of defining `GAsync` as opposed to inlining its definition directly into `VAsync` (i.e. `VAsync :: Term -> Async.Async (Either Exn Value) -> Value`)?  It kind of feels like unnecessary indirection to me.

src/Swarm/Language/Syntax.hs
```diff
@@ -530,6 +544,10 @@ pattern TBind v t1 t2 = SBind v (STerm t1) (STerm t2)
 pattern TDelay :: DelayType -> Term -> Term
 pattern TDelay m t = SDelay m (STerm t)
 
+-- | Match a TDelay without syntax
```
byorgey, line 547: This comment needs to be updated.

What is the purpose of `SFuture` / `TFuture`?  It seems like it can never actually result from parsing surface syntax, right?  Is it only for pretty-printing values?  As I think I mentioned before I really don't want to add extra stuff to the AST just to be able to pretty-print values.

src/Swarm/Language/Syntax.hs
```diff
@@ -323,8 +324,17 @@ data Const
     AppF
   | -- God-like sensing operations
 
-    -- | Run a command as if you were another robot.
+    -- | Hypothetically run a command as if you were another robot.
     As
+  | -- | Hypothetically run a command as if you were another robot,
```
byorgey, line 329: `async` doesn't actually have the sense of "as if you were another robot", does it?

src/Swarm/TUI/Controller.hs
byorgey: I mean one simple thing is just to wrap this in a check: first check if the `robotMap` has a key `"base"` and if so do the cancelling stuff.

byorgey (changes requested):

src/Swarm/Game/Step.hs
```diff
@@ -920,6 +927,60 @@ execConst c vs s k = do
...
+        va <- sendIO $ Async.async evalAs
+
+        return $ Out (VAsync t $ GAsync va) s k
+      _ -> badConst
+    Await -> case vs of
+      [va@(VAsync t (GAsync a))] -> do
+        time <- use ticks
+        mv <- sendIO $ Async.poll a
+        return $ case mv of
+          Nothing -> Waiting (time + 8) (Out va s (FApp (VCApp Await []) : k))
```
byorgey, line 963: I think an `FExec` is needed, so that the `Await` is re-executed once it is applied.
```suggestion
          Nothing -> Waiting (time + 8) (Out va s (FApp (VCApp Await []) : FExec : k))
```

xsebek:
I do not think we need this now, but if we ever want to explore this direction we can look here for tips. :slightly_smiling_face:

PR #288, opened 2021-11-25
Add challenge world seed

xsebek:
This is a WIP to fix missing world seed option in challenge mode.

From [#285 comment](https://github.com/byorgey/swarm/pull/285#issuecomment-974187422):
>  - [ ] Ability to specify a world seed instead of a default cell content, so we can have challenges that take place in random worlds a la classic mode

It works, but the code needs some polishing :sweat_smile: 

Screenshot from 2021-11-25 13-17-50

### DIFF
--- a/src/Swarm/Game/Challenge.hs
+++ b/src/Swarm/Game/Challenge.hs
@@ -32,7 +32,7 @@ module Swarm.Game.Challenge (
   challengeWin,
 ) where
 
-import Control.Arrow ((***))
+import Control.Arrow ((***), Arrow (first))
 import Control.Lens hiding (from)
 import Data.Array
 import Data.HashMap.Strict (HashMap)
@@ -44,12 +44,14 @@ import GHC.Int (Int64)
 import Linear.V2
 import Witch (from, into)
 
+import GHC.Base ((<|>))
 import Swarm.Game.Entity
 import Swarm.Game.Robot (Robot)
 import Swarm.Game.Terrain
 import Swarm.Game.World
 import Swarm.Language.Pipeline (ProcessedTerm)
 import Swarm.Util.Yaml
+import Swarm.Game.WorldGen (testWorld2FromArray)
 
 -- | A 'Challenge' contains all the information to describe a
 --   challenge.
@@ -57,7 +59,7 @@ data Challenge = Challenge
   { _challengeName :: Text
   , _challengeSeed :: Maybe Int
   , _challengeEntities :: EntityMap
-  , _challengeWorld :: WorldFun Int Entity
+  , _challengeWorld :: EntityMap -> WorldFun Int Entity
   , _challengeRobots :: [Robot]
   , _challengeWin :: ProcessedTerm
   }
@@ -69,7 +71,7 @@ instance FromJSONE EntityMap Challenge where
     em <- liftE (buildEntityMap <$> (v .:? "entities" .!= []))
     Challenge
       <$> liftE (v .: "name")
-      <*> liftE (v .:? "seed")
+      <*> liftE (v .:? "seed") -- TODO: avoid two seeds
       <*> pure em
       <*> withE em (mkWorldFun (v .: "world"))
       <*> withE em (v ..: "robots")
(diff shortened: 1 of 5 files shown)


src/Swarm/Game/Challenge.hs
```diff
@@ -146,5 +150,15 @@ mkWorldFun pwd = E $ \em -> do
...
-  let defTerrain = (fromEnum *** (>>= (`lookupEntityName` em))) (defaultTerrain wd)
-  return $ worldFunFromArray arr defTerrain
+  case defaultTerrain wd of
+    Left seed -> return $ \entities ->
+      let emPlus = em <> entities
+          arr2 = bimap toEnum (fmap (^. entityName)) <$> arr
+       in (lkup emPlus <$>) . first fromEnum <$> testWorld2FromArray arr2 seed
+    Right def ->
+      let defTerrain = (fromEnum *** (>>= (`lookupEntityName` em))) def
+       in return . const $ worldFunFromArray arr defTerrain
```
xsebek, line 160: Actually it might be a good idea to unify this, so that challenges could use the entities which we load anyway. :thinking:

src/Swarm/Game/Challenge.hs
```diff
@@ -44,10 +44,12 @@ import GHC.Int (Int64)
 import Linear.V2
 import Witch (from, into)
 
+import GHC.Base ((<|>))
```
byorgey, line 47: Probably makes more sense to import this from `Control.Applicative`.

src/Swarm/Game/Challenge.hs
```diff
@@ -57,7 +59,7 @@ data Challenge = Challenge
   { _challengeName :: Text
   , _challengeSeed :: Maybe Int
   , _challengeEntities :: EntityMap
-  , _challengeWorld :: WorldFun Int Entity
+  , _challengeWorld :: EntityMap -> WorldFun Int Entity
```
byorgey, line 62: I don't understand the purpose of introducing an `EntityMap` argument here.  That is what the whole `FromJSONE` infrastructure is supposed to be taking care of.  An `EntityMap` will be available at the time we read a challenge, so we can immediately use it to build the `WorldFun`.

src/Swarm/Game/Challenge.hs
```diff
@@ -146,5 +150,15 @@ mkWorldFun pwd = E $ \em -> do
       . mapM toEntity
       . concat
       $ grid
-  let defTerrain = (fromEnum *** (>>= (`lookupEntityName` em))) (defaultTerrain wd)
-  return $ worldFunFromArray arr defTerrain
+  case defaultTerrain wd of
+    Left seed -> return $ \entities ->
```
byorgey, line 154: Where will the extra entities be coming from?

src/Swarm/Game/Challenge.hs
byorgey: I'm not sure I know what you're referring to.  It might be a good idea to unify what?

byorgey (changes requested):
Sorry it took me a while to get around to this, I was travelling for the US holiday.  

Generally looks good but I have a few questions/suggestions.

src/Swarm/Game/Challenge.hs
xsebek: I thought the normal entities are not available at this point.

I probably just got confused by the code 😅

byorgey: I agree it is a little confusing.  I should add some more comments.  Basically, in the `FromJSONE EntityMap Challenge` instance, any custom entities (defined in an `entities` field) are parsed, and those are passed to `mkWorldFun` via `withE` (https://github.com/byorgey/swarm/blob/0623c07f13b108c7b3991fa3b6fd5de31e0592de/src/Swarm/Game/Challenge.hs#L74).  However, `withE` *combines* the given `EntityMap` of custom entities with the ambiently available one (of normal entities) that is passed in when `parseJSONE` is run.  Maybe `appendE` would be a better name?

byorgey:
Hey @xsebek , I am just getting around to looking at this again.  I'd love to get the challenge mode stuff all merged.  Do you think you will have a chance to clean this up, or would you like me to take a shot at it?

xsebek:
I am still studying for my exams, so I likely will not have much time in the next 1-2 weeks.

@byorgey If you want to finish this PR I would not object at all 👍

byorgey:
@xsebek OK, will do!  Good luck with your exams!

PR #290, opened 2021-12-01
generalize `scan` to return `() + string`

### LINKED ISSUE #267: Generalize `scan` to return `() + string`
Given current capability checking, could we not make `scan'` the default so that robots without capability could still use it, but not directly access the result? They could still `upload` to others and all the stuff they do now. 

_Originally posted by @xsebek in https://github.com/byorgey/swarm/issues/88#issuecomment-954857227_

byorgey:
Players can ignore the return value until they have the capability to
deal with sum types.  `scan` still has the effect of adding
information to one's inventory.

Closes #267.

### DIFF
--- a/src/Swarm/Game/Step.hs
+++ b/src/Swarm/Game/Step.hs
@@ -827,11 +827,13 @@ execConst c vs s k = do
         orient <- use robotOrientation
         let scanLoc = loc ^+^ applyTurn d (orient ? zero)
         me <- entityAt scanLoc
-        case me of
-          Nothing -> return ()
-          Just e -> robotInventory %= insertCount 0 e
+        res <- case me of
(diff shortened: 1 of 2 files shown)


xsebek (approved):
Looks nice :slightly_smiling_face: 

Currently sum types are accesible from beginning. Maybe we could mention the return type in the `.yaml` or do you prefer to leave it unspoken until there is some capability for that?

byorgey:
Yeah, sum types are available now but they're not supposed to be.  Eventually they will require some kind of device.  I looked at mentioning the return type of `scan` in the tutorial but there wasn't any obvious place that needed updating, and I thought it would be simpler to just leave it out and let people discover it later.

PR #291, opened 2021-12-01
allow only whitespace when reading a term

### LINKED ISSUE #275: Comments alone are invalid
#### Describe the bug

Inputs like
```c
// commented out expression because I forgot to "run somefile" first
```
are invalid, while
```c
()// commented out expression because I forgot to "run somefile" first
```
is fine. Ideally such hacks would not be required to use comments.

#### Expected behavior

Ideally even empty line would be valid.

#### Additional context

This is also annoying in editors with LSP when starting a file with a comment which is invalid until some code is added too.

byorgey:
- The user can now enter only whitespace (including comments) at the
  REPL.
- A new file which is blank / only comments is no longer invalid from
  the perspective of LSP.

Closes #275.

### DIFF
--- a/src/Swarm/Game/Step.hs
+++ b/src/Swarm/Game/Step.hs
@@ -1158,11 +1158,13 @@ execConst c vs s k = do
 
         f <- msum mf `isJustOrFail` ["File not found:", fileName]
 
-        t <-
+        mt <-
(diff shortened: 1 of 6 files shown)


xsebek (approved):
Works great, thanks! :partying_face:

PR #293, opened 2021-12-01
add `format` function and string concatenation operator `++`

### LINKED ISSUE #248: Show command and string concatenation
I would like `log` to have signature `a -> cmd ()` instead of current `string -> cmd ()`.
The use case I have in mind is to log an `int`, for example in this loop:
```haskell
def iterate = \c.\b. bn <- c b; iterate c b end
build "loop" {iterate {\i. move; turn back; log i; return (i+1)} 0}
```
But we can print any type, so I see no reason why we should not allow that. :thinking: 

The alternative would be to provide a `show` function or some other functions operating on string.

byorgey:
Adds a new function `format : a -> string` and an operator `++ : string -> string -> string`.

Closes #248.

### DIFF
--- a/src/Swarm/Game/Step.hs
+++ b/src/Swarm/Game/Step.hs
@@ -1181,6 +1181,12 @@ execConst c vs s k = do
     Mul -> returnEvalArith
     Div -> returnEvalArith
     Exp -> returnEvalArith
+    Fmt -> case vs of
+      [v] -> return $ Out (VString (prettyValue v)) s k
+      _ -> badConst
+    Concat -> case vs of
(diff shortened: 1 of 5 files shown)


xsebek (approved):
I really like how small the change is. :slightly_smiling_face:  

Regarding names, I could imagine using `++` like Haskell as it looks more like `+`, which we probably do not want to overload.
I would prefer full `Format` so it is easier to understand.

byorgey:
@xsebek re: names, makes sense.  How does it look now?

xsebek:
@byorgey looks good to me :slightly_smiling_face:

PR #298, opened 2022-01-29
Make seed robots self-destruct when an exception is raised

noahyor:
I have added a try statement on line 551 of Step.hs to self-destruct even if place command is obstructed.

### DIFF
--- a/src/Swarm/Game/Step.hs
+++ b/src/Swarm/Game/Step.hs
@@ -548,12 +548,14 @@ evalConst c vs s k = do
 seedProgram :: Integer -> Integer -> Text -> ProcessedTerm
 seedProgram minTime randTime thing =
   [tmQ|
-    r <- random (1 + $int:randTime);
-    wait (r + $int:minTime);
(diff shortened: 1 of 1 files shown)


byorgey:
Thanks!  And just to give more context for anyone reading this in the future, I think the issue this was fixing is that if you harvested a tree and then put an item down in the same spot, the seed can't regrow but then leaves behind a dead "stalk".  This way it disappears whether it's able to regrow or not.  This gives an effective method of "clearing" forest that you don't want to grow back.

byorgey (approved):

byorgey:
@noahyor , congrats on having your first PR accepted to Swarm!   We appreciate the contribution and we're really glad to welcome you as part of the [community](https://github.com/byorgey/swarm/blob/main/COMMUNITY.md).  In a minute we will send you an invite to the repo (see https://github.com/byorgey/swarm/blob/main/CONTRIBUTING.md#i-have-push-access-to-the-swarm-repository-now-what ).

PR #301, opened 2022-02-01
Hide zero inventory

### LINKED ISSUE #282: Add toggle key for hiding 0 in inventory
From #49 - let's add a toggle key that would hide/show entities with count 0.
This could shorten the length of the inventory in some situations.

As an alternative we could:
- add a separate list of "Known entities" below installed devices
- abandon this idea and revive #126 instead

noahyor:
This should resolve #282 .

### DIFF
--- a/src/Swarm/TUI/Controller.hs
+++ b/src/Swarm/TUI/Controller.hs
@@ -578,7 +578,7 @@ handleRobotPanelEvent s (VtyEvent (V.EvKey V.KEnter [])) = do
     Just (_, Separator _) -> continueWithoutRedraw s
     Just (_, InventoryEntry _ e) -> makeEntity s e
     Just (_, InstalledEntry e) -> makeEntity s e
-handleRobotPanelEvent s (VtyEvent (V.EvKey (V.KChar '0' )[])) = do
+handleRobotPanelEvent s (VtyEvent (V.EvKey (V.KChar '0') [])) = do
   continue $ s & (uiState . uiShowZero %~ not) . (uiState . uiInventoryShouldUpdate .~ True)
 handleRobotPanelEvent s (VtyEvent ev) = do
   let mList = s ^? uiState . uiInventory . _Just . _2
--- a/src/Swarm/TUI/Controller.hs
+++ b/src/Swarm/TUI/Controller.hs
@@ -298,13 +298,15 @@ updateUI = do
   let focusedRobotHash = view inventoryHash <$> fr
   -- The hash of the focused robot (if any)
 
+  shouldUpdate <- use (uiState . uiInventoryShouldUpdate)
   -- If the hashes don't match (either because which robot (or
   -- whether any robot) is focused changed, or the focused robot's
   -- inventory changed), regenerate the list.
   inventoryUpdated <-
-    if listRobotHash /= focusedRobotHash
+    if listRobotHash /= focusedRobotHash || shouldUpdate
       then do
         zoom uiState $ populateInventoryList fr
+        (uiState . uiInventoryShouldUpdate) .= False
         pure True
       else pure False
 
@@ -576,6 +578,8 @@ handleRobotPanelEvent s (VtyEvent (V.EvKey V.KEnter [])) = do
     Just (_, Separator _) -> continueWithoutRedraw s
     Just (_, InventoryEntry _ e) -> makeEntity s e
     Just (_, InstalledEntry e) -> makeEntity s e
+handleRobotPanelEvent s (VtyEvent (V.EvKey (V.KChar '0' )[])) = do
+  continue $ s & (uiState . uiShowZero %~ not) . (uiState . uiInventoryShouldUpdate .~ True)
 handleRobotPanelEvent s (VtyEvent ev) = do
   let mList = s ^? uiState . uiInventory . _Just . _2
   case mList of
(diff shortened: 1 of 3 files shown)


src/Swarm/Game/Step.hs
```diff
@@ -548,12 +548,14 @@ evalConst c vs s k = do
...
-    place $str:thing;
+    try {
+      r <- random (1 + $int:randTime);
+      wait (r + $int:minTime);
+      appear "|";
+      r <- random (1 + $int:randTime);
+      wait (r + $int:minTime);
+      place $str:thing;
+    } {};
     selfdestruct
```
xsebek, line 559: Your main branch seems to be behind? Please sync your main with byorgey/main. :slightly_frowning_face: 

I usually branch from byorgey/main itself, but you should have a button in your repo that says **"Fetch upstrem"**.

That might do the trick, if not, rebase the last commit on your main :slightly_smiling_face:

src/Swarm/TUI/Model.hs
```diff
@@ -378,6 +382,12 @@ uiModal :: Lens' UIState (Maybe Modal)
 -- | A togle to show the FPS by pressing `f`
 uiShowFPS :: Lens' UIState Bool
 
+-- | A togle to show inventoy items whith count 0 by pressing `0`
+uiShowZero :: Lens' UIState Bool
+
+-- |
+uiInventoryShouldUpdate :: Lens' UIState Bool
```
xsebek, line 389: Missing comment :wink:

src/Swarm/TUI/Controller.hs
```diff
@@ -298,13 +298,15 @@ updateUI = do
   let focusedRobotHash = view inventoryHash <$> fr
   -- The hash of the focused robot (if any)
 
+  shouldUpdate <- use (uiState . uiInventoryShouldUpdate)
   -- If the hashes don't match (either because which robot (or
   -- whether any robot) is focused changed, or the focused robot's
   -- inventory changed), regenerate the list.
   inventoryUpdated <-
-    if listRobotHash /= focusedRobotHash
+    if listRobotHash /= focusedRobotHash || shouldUpdate
```
xsebek, line 306: I wonder, would it not be enough to check something like `showZero /= listShowsZero`?
I am basing that on `listRobotHash /= focusedRobotHash`, but I have not looked into the code. :slightly_smiling_face: 

If there is some good reason, you can mention it in the `uiInventoryShouldUpdate` docstring :wink:

xsebek (changes requested):
Please merge the #302 - we use CI to avoid different code styles (which would produce diffs).

Also, the description of the PR will be used as a squashed commit by Mergify at the end.
But for this small change it is probably fine :slightly_smiling_face:

src/Swarm/TUI/Model.hs
noahyor: What do you think I should change it to?

src/Swarm/Game/Step.hs
xsebek: Yes, but you did not update *your* main branch and then you branched from it :man_shrugging: 

It probably would not cause any issue when merging, but I would prefer keeping PRs clean of such noise. :slightly_smiling_face:

src/Swarm/TUI/Model.hs
xsebek: Probably just mention what toggles it (and that it is used when viewing same robots inventory?), also see my next comment.

src/Swarm/Game/Step.hs
xsebek: See below, the blue branch is this one, white is your main and orange is byorgey/main:

Screenshot from 2022-02-01 20-47-34

What I would like is, if the white was same as blue (that is your fech upstream into your main) and this branch would go from that updated main instead of the branch you used in #298 (I guess use git rebase).

Sorry to pester you about this, if you don't know how to use rebase, don't worry about it and the merge will hopefully work.
But then please switch back to branching from main. :smile:

src/Swarm/TUI/Controller.hs
byorgey: There isn't anywhere that we currently remember something like `listShowsZero`.  I suppose we could add that as another field of the `uiInventory`, but the `uiInventory` is already complicated enough.  I can imagine that in the future we might add other features that would require the inventory list to be updated, in which case we can reuse `uiInventoryShouldUpdate`, whereas something like `showZero /= listShowsZero` is specific to just this toggle.

src/Swarm/TUI/Model.hs
```diff
@@ -375,9 +379,15 @@ uiError :: Lens' UIState (Maybe Text)
 --   top of the UI, e.g. for the Help screen.
 uiModal :: Lens' UIState (Maybe Modal)
 
--- | A togle to show the FPS by pressing `f`
+-- | A toggle to show the FPS by pressing `f`
 uiShowFPS :: Lens' UIState Bool
 
+-- | A toggle to show inventoy items whith count 0 by pressing `0`
```
byorgey, line 385:
```suggestion
-- | A toggle to show or hide inventory items with count 0 by pressing `0`
```

byorgey (approved):

byorgey:
@xsebek Do you have time to take a quick look at this again and make sure the requested changes look good?  We can't merge it until you approve. =)

byorgey:
I'm just going to go ahead and merge this manually, it looks to me like all the requested changes have been addressed.

PR #307, opened 2022-02-21
Add boolean operations like 'And' and 'Or'

### LINKED ISSUE #189: Logical operators
Right now `not` is the only built-in logical operator.  Of course, given `if` expressions it is possible to define logical operators, but it would be nice to have some infix logical operators built-in.  Probably just `&&` and `||` suffices.

noahyor:
Closes #189

### DIFF
--- a/src/Swarm/Game/Step.hs
+++ b/src/Swarm/Game/Step.hs
@@ -1241,6 +1241,12 @@ execConst c vs s k = do
     Gt -> returnEvalCmp
     Leq -> returnEvalCmp
     Geq -> returnEvalCmp
+    And -> case vs of
+      [VBool a, VBool b] -> return $ Out (VBool (a && b)) s k
+      _ -> badConst
+    Or -> case vs of
+      [VBool a, VBool b] -> return $ Out (VBool (a || b)) s k
+      _ -> badConst
     Add -> returnEvalArith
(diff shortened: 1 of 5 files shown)


noahyor:
No implementation for short-circuiting

byorgey:
This is looking good to me.

In the discussion at #189 someone mentioned using `bool -> {bool} -> bool` as the type of `&&` and `||`, to make short-circuiting explicit.  However, I think I prefer doing it this way (where the built-in `&&` and `||` are strict in both arguments and have type `bool -> bool -> bool`):

- They are much simpler to use.  It would be annoying being forced to put in curly braces all the time when using `&&` and `||`, when in the cast majority of cases we probably don't care about short-circuiting.
- Users who want short-circuiting operators can easily implement them using delay/force, so there's no need for them to be built in, e.g.:
  ```
  and : bool -> {bool} -> bool
  and = \p. \q. if p {force q} {false}
  ```

I think before merging it might be nice to add a capability for these operators and a device that conveys the capability.  (We should probably put the same capability on `not` as well.)

noahyor:
I'll see if I can get those capabilities in there.

byorgey (approved):

byorgey:
I think I'll go ahead and merge this; we can easily add capabilities later.

PR #317, opened 2022-03-17
Make the `salvage` command take one tick to transfer each item

### LINKED ISSUE #202: Make `salvage` command less powerful
Once #128 is merged, we will have a `salvage` command.  However, I think it is overpowered as it stands, since it lets you transfer an arbitrary number of items instantly. 

Imagine a scenario where you want a robot to harvest a bunch of stuff and then transfer it all to the base.  Currently, this involves some slightly interesting difficulties: 
* It takes time: you have to make it sit there and `give` items one at a time.
* Once we implement #115 , you will be able to mitigate the time required with `box`es, but that requires some work to make the boxes and write the programs to pack, transfer, and unpack the boxes, and boxes will probably have limited capacity.
* If you don't know exactly how many items there are, you need to obtain some capabilities to be able to write a little loop to deal with however many there are. (Actually, we don't yet have the inventory-sensing commands to make this work, but we will.)

However, the `salvage` command lets you completely sidestep all these issues too easily: just have the base run `salvage` on the harvester robot, and get its entire inventory all at once!  And if you were planning to have the harvester robot return and do more harvesting, no problem: just build a replacement from the parts you just salvaged.

I propose to nerf the `salvage` command a bit, so that it will still be useful for recovering resources and debugging in the early game, but won't make other stuff too easy.  In particular:
1. Instead of taking a single tick to transfer everything, `salvage` will take one tick to transfer each inventory item from the salvaged robot.  (It's actually a bit tricky to figure out how to make this work; it would be easier to just make it `wait` for an amount of time corresponding to the inventory size, then transfer everything all at once when it's done, but it would be more fun and realistic to actually see it transferring items one at a time.)
2. Perhaps, when `salvaging`, each transferred item has a small probability of being destroyed instead.  This way you can't just keep recycling robots forever via `salvage`.

byorgey:
I initially thought this would be tricky, but it's not: all we need to
do is reprogram the salvaged robot to give the salvaging robot one
item at a time.

Closes #202.

### DIFF
--- a/src/Swarm/Game/Step.hs
+++ b/src/Swarm/Game/Step.hs
@@ -359,6 +359,10 @@ stepCESK cesk = case cesk of
     return $ Up (Fatal (T.append "Antiquoted variable found at runtime: $str:" v)) s k
   In (TAntiInt v) _ s k ->
     return $ Up (Fatal (T.append "Antiquoted variable found at runtime: $int:" v)) s k
+  -- Normally it's not possible to have a TRobot value in surface
+  -- syntax, but the salvage command generates a program that needs to
+  -- refer directly to the salvaging robot.
+  In (TRobot rid) _ s k -> return $ Out (VRobot rid) s k
   -- Function constants of arity 0 are evaluated immediately
   -- (e.g. parent, self).  Any other constant is turned into a VCApp,
   -- which is waiting for arguments and/or an FExec frame.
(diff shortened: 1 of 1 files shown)


byorgey:
Hmm, looks like we have to re-enable the CI workflow somehow?

PR #320, opened 2022-03-26
Add a dialog box to confirm quit

### LINKED ISSUE #77: Confirmation dialog when quitting
It could be very upsetting for the user if they quit when they didn't mean to.  We should add a confirmation dialog that pops up asking the user whether they are sure they want to quit, with "Quit" and "Cancel" buttons (and Cancel highlighted by default of course).

I am not sure of the best way to do this off the top of my head.  It would probably mean using some kind of `brick` form or list for the buttons. Probably there is an example of doing this kind of thing either in one of the examples in the `brick` repository, or in a project using `brick`.

byorgey:
Closes #77 .  When hitting Ctrl-Q, now a dialog box opens up allowing you to confirm you want to quit or cancel.  The machinery for dealing with pop-up windows is somewhat generalized as well, which should make it easy to add other similar pop-up windows as needed in the future.

### DIFF
--- a/src/Swarm/TUI/Controller.hs
+++ b/src/Swarm/TUI/Controller.hs
@@ -106,15 +106,13 @@ handleEvent s (VtyEvent (V.EvKey V.KEsc []))
   | isJust (s ^. uiState . uiError) = continue $ s & uiState . uiError .~ Nothing
   | isJust (s ^. uiState . uiModal) = continue $ s & uiState . uiModal .~ Nothing
 handleEvent s (VtyEvent vev)
-  | isJust (s ^. uiState . uiModal) = do
-    s' <- s & uiState . uiModal . _Just . modalDialog %%~ handleDialogEvent vev
-    continue s'
+  | isJust (s ^. uiState . uiModal) = handleModalEvent s vev
 handleEvent s (VtyEvent (V.EvKey (V.KChar '\t') [])) = continue $ s & uiState . uiFocusRing %~ focusNext
 handleEvent s (VtyEvent (V.EvKey V.KBackTab [])) = continue $ s & uiState . uiFocusRing %~ focusPrev
 handleEvent s ev = do
   -- intercept special keys that works on all panels
   case ev of
-    ControlKey 'q' -> toggleModal s QuitModal -- shutdown s
+    ControlKey 'q' -> toggleModal s QuitModal
     MetaKey 'w' -> setFocus s WorldPanel
     MetaKey 'e' -> setFocus s RobotPanel
     MetaKey 'r' -> setFocus s REPLPanel
--- a/src/Swarm/TUI/Controller.hs
+++ b/src/Swarm/TUI/Controller.hs
@@ -194,14 +194,24 @@ helpWidget = (helpKeys <=> fill ' ') <+> (helpCommands <=> fill ' ')
       , hCenter $ mkTable baseCommands
       ]
   baseCommands =
-    [ ("build <name> {<commands>}", "Create a robot")
-    , ("make <name>", "Craft an item")
+    [ ("build {<commands>}", "Create a robot")
+    , ("make \"<name>\"", "Craft an item")
     , ("move", "Move one step in the current direction")
     , ("turn <dir>", "Change the current direction")
     , ("grab", "Grab whatver is available")
-    , ("give <robot> <item>", "Give an item to another robot")
-    , ("has <item>", "Check for an item in the inventory")
+    , ("give <robot> \"<item>\"", "Give an item to another robot")
+    , ("has \"<item>\"", "Check for an item in the inventory")
     ]
+
+handleModalEvent :: AppState -> V.Event -> EventM Name (Next AppState)
+handleModalEvent s ev = case ev of
+  V.EvKey V.KEnter [] ->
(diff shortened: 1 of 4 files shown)


src/Swarm/TUI/Controller.hs
```diff
@@ -149,6 +152,65 @@ toggleModal s modal = do
   -- the world event handler for the KChar 'p'.
   resetLastFrameTime curTime = uiState . lastFrameTime .~ curTime
 
+generateModal :: ModalType -> Modal
```
byorgey, line 155: This should probably move to `View.hs`?

src/Swarm/TUI/Model.hs
```diff
@@ -314,6 +334,7 @@ data UIState = UIState
   , _uiScrollToEnd :: Bool
   , _uiError :: Maybe Text
   , _uiModal :: Maybe Modal
+  , _uiDialog :: Maybe (Dialog Bool)
```
byorgey, line 337: I think this isn't actually used.

src/Swarm/TUI/Model.hs
```diff
@@ -282,11 +288,24 @@ replIndexIsAtInput repl = repl ^. replIndex == replLength repl
...
+data Modal = Modal
+  { _modalType :: ModalType
+  , _modalDialog :: Dialog Bool
+  , _modalWidget :: Widget Name
+  }
+
+makeLenses ''Modal
+
+-- | Width cap for modal and error message windows
+maxModalWindowWidth :: Int
```
Alexander-Block, line 306: What is the deeper reason to move `maxModalWindowWidth` from `View.hs` to `Model.hs`? It does not seem to interact with the rest of `Model.hs`.

src/Swarm/TUI/Controller.hs
```diff
@@ -149,6 +151,16 @@ toggleModal s modal = do
   -- the world event handler for the KChar 'p'.
   resetLastFrameTime curTime = uiState . lastFrameTime .~ curTime
 
+handleModalEvent :: AppState -> V.Event -> EventM Name (Next AppState)
+handleModalEvent s ev = case ev of
+  V.EvKey V.KEnter [] ->
+    case s ^? uiState . uiModal . _Just . modalDialog . to dialogSelectedIndex of
```
Alexander-Block, line 157: Using `dialogSelectedIndex` does not seem very type safe. Refering directly to a selected index leaks abstraction, since its semantics depends on the ordering of a list inside a subfunction of `generateModel` in `View.hs`. 
I propose replacing `dialogSelectedIndex` by `dialogSelection` which in this case is of type `Dialog Bool -> Maybe Bool` and gives us the state corresponding to the currently selected button. I would even propose going one step further and replacing `Bool` here by a custom `ButtonSelection` state because I imagine it to be an easy mistake to confuse whether `True` means cancel or quit.

src/Swarm/TUI/View.hs
```diff
@@ -167,31 +168,34 @@ chooseCursor s locs = case s ^. uiState . uiModal of
...
-      WinModal -> ("", txt "Congratulations!")
+  (title, widget, buttons, requiredWidth) =
+    case mt of
+      HelpModal -> ("Help", helpWidget, Nothing, maxModalWindowWidth)
+      WinModal -> ("", txt "Congratulations!", Nothing, maxModalWindowWidth)
+      QuitModal ->
+        let quitMsg = "Are you sure you want to quit?"
+         in ( ""
+            , padBottom (Pad 1) $ hCenter $ txt quitMsg
+            , Just (0, [("Cancel", False), ("Quit", True)])
```
Alexander-Block, line 196: As written in my review for `handleModalEvent` I would consider replacing `False` and `True` with values `ButtonCancel` and `ButtonQuit` of a `ButtonSelection` type in order to clarify the semantics.

Alexander-Block (changes requested):
I like this approach of embedding the quit dialog box in a general machinery for pop-up windows. However, I have some questions/remarks. The main one is that I find it problematic to use `dialogSelectedIndex` in `handleModalEvent`.

src/Swarm/TUI/Model.hs
byorgey: Oh, good catch, no reason. I was moving some things around and that got left behind.  I will move it back.

src/Swarm/TUI/Controller.hs
byorgey: Ah, you're absolutely right.  I remember seeing `dialogSelection` when I was looking at the brick documentation but then I guess I forgot/didn't notice it when I actually wrote the code.  And your idea of replacing `Bool` with a custom `ButtonSelection` state is good, I will do that.

src/Swarm/TUI/View.hs
byorgey: I just called them `Cancel` and `Confirm` to try to make it slightly more general in case we later want another type of dialog that asks you to confirm something.

byorgey:
@Alexander-Block , thanks so much for the review and the great suggestions.  I think I've implemented all of them now.

Alexander-Block (approved):
@byorgey: Thank you. This looks great now; I have no futher comments.

PR #338, opened 2022-05-22
Make only base display as Omega

### LINKED ISSUE #332: `turn down` should not make a robot look like the base
For expedience, the default display for any robot is to look like `<`, `>`, `^`, or `v` depending on its direction, and like `Ω` if the direction is (0,0).  But it is actually possible for a robot to execute `turn down` to set its direction to (0,0), which makes it suddenly look like the base.  `turn down` might seem useless at the moment but I was contemplating it as a mechanism for accessing sub-worlds in #144 .  In any case, we should probably pick something else for the display for a generic robot with direction (0,0), and set the display of the base specially.

xsebek:
- make robots look down as `'X'`

With the base not having any direction-specific characters, this just finishes the job and makes it fully distinct.

Closes #332

### DIFF
--- a/src/Swarm/Game/Display.hs
+++ b/src/Swarm/Game/Display.hs
@@ -133,11 +133,14 @@ defaultEntityDisplay c =
     }
 
 -- | Construct a default robot display, with display characters
---   @"Ω^>v<"@, the default robot attribute, and priority 10.
+--   @"X^>v<"@, the default robot attribute, and priority 10.
+--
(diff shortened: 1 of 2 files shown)


xsebek:
The motivation behind X is that it looks like four robots meeting. :slightly_smiling_face: 
```
  v
              v
>   <        > <        X
              ^
  ^
```

byorgey (approved):
Looks good to me!

byorgey:
Thanks @xsebek , nice to see you again. =)

PR #339, opened 2022-05-22
Find robots

xsebek:
- add `robotNamed : string -> cmd robot`
- add chess horse challenge :horse_racing:

In challenges, it's impossible to know beforehand what the robot ID will be.
Seems like we should be able to find it by its name though.


xsebek:
The challenge is a bit hard but not impossible:
Screenshot from 2022-05-22 11-42-54

```bash
move; move; move; turn right; move; turn left; move;
```
Screenshot from 2022-05-22 11-43-07

I like that we can customize all this in current challenges. :+1: 
Being able to query the position of an arbitrary robot makes it even sweeter.

data/challenges/chess_horse.yaml
```diff
@@ -0,0 +1,65 @@
+name: Moving, part 1
```
byorgey, line 1: Should the name be changed?

byorgey (approved):
Makes sense, I like it.

Hmm, maybe there should be a "compass" device which you need to use directions like `east`, `west` etc.  Then we could prevent cheesing this challenge by not starting the robot with a compass...

data/challenges/chess_horse.yaml
xsebek: Done :heavy_check_mark:

PR #352, opened 2022-06-03
Update Github Actions and fix Doctest

xsebek:
- fix doctests in code
- configure Haskell-CI to use `cabal-docspec` (compiled doctest)
- update cabal `tested-with` field
- update `.mergify.yaml` to new compilers
  - restore note about mergify in haskell-ci


cabal.haskell-ci
```diff
@@ -0,0 +1,3 @@
+no-tests-no-benchmarks: False
+unconstrained:          False
+docspec:                True
```
xsebek, line 3: I did not read up yet, on what those other options do, I just wanted to include doctest.

We could also look for inspiration in [lens](https://github.com/ekmett/lens/blob/master/cabal.haskell-ci).

xsebek:
@byorgey should we also test newer GHC versions? :thinking:

.github/workflows/haskell-ci.yml
```diff
@@ -36,10 +39,10 @@ on:
       - 'cabal.project*'
 jobs:
   linux:
-    # when changing the name or adding ghc version, please don't forget to update
-    # the .mergify.yml rules at the root of the project.
```
byorgey, line 40: Did you delete this comment on purpose because it's no longer relevant?  Or did you just miss it?  Assuming it's still relevant, we should leave it there.

byorgey (changes requested):
Looks good, I just have the one question about a comment that got removed.

I like doctests.  I don't know how much opportunity we'll have to use them but might as well!

.github/workflows/haskell-ci.yml
xsebek: I mistook it for autogenerated comment. I will move it up so it's visible - that way we don't have to edit the file too much.

cabal.haskell-ci
xsebek: I added documentation to the fields. :heavy_check_mark:

byorgey (approved):
LGTM!

Issue #25, opened 2021-09-21
Tutorial via challenge mode


byorgey:
Once we implement #15 , I think it would be nice to have a simple tutorial that advances through a series of simple challenges, guiding the user through doing a few of the kinds of things they will need to do at the start of the game (build some robots, make some definitions, craft some devices, etc.)

byorgey:
To make this work well might also require some kind of facility for stepping automatically through a specific list of challenges, with an option to continue to the next challenge after completing each one.

xsebek:
@byorgey would you mind if I also pinned this?

There can be at most three pinned issues, so we should choose where to focus. :slightly_smiling_face:

PR #361, opened 2022-06-04
Start base with finite supply of basic robot parts

### LINKED ISSUE #35: Starting base with a large but finite supply of basic robot parts

byorgey:
There is no longer a list of "default devices" which can be generated out of thin air.

Closes #35.


xsebek:
@byorgey I hope you won't mind that I updated the description with #395 :slightly_smiling_face:

byorgey:
@xsebek perfect, thanks!

byorgey:
A recipe for a `plasma cutter` is the only thing left!  Any ideas?

byorgey:
Technically we should probably update `TUTORIAL.md` to match,  but I'm getting sick of updating it, and would rather focus on #25 .

xsebek (approved):
:+1:

byorgey:
@xsebek what do you think?

xsebek:
The toolkit recipe seems fine to me :slightly_smiling_face: I checked the recipe graph and it's about the same difficulty as the other starting devices.

TristanCacqueray (approved):
:+1:

byorgey:
I called it a `toolkit` instead of a `repair kit` since if you are taking robots apart you are obviously not repairing them. :grin:  Hopefully we can find other uses for a `toolkit` later.

byorgey:
@mergifyio refresh

Issue #336, opened 2022-05-18
Add sand entity


byorgey:
Sand can be used to make glass in a furnace, which could in turn be required to make a solar panel, needed for a calculator.  See the discussion at #26 .

My idea is that sand can be found on some edges of water, perhaps at the boundary between two specific biomes.

byorgey:
Maybe sand could also be made from rocks somehow.

xsebek:
> from rocks

We do have drills that robots use to get through mountains, maybe they could be used to drill down rocks to sand?

But having sand on some of the edges of water could be visually nice too - there does not need to be much of it if it's a growing entity like water. :slightly_smiling_face:

PR #362, opened 2022-06-04
add sand, glass, and calculator

byorgey:
- Sand is now found near water.  Sand regenerates immediately (like water).
- Sand can be used in a furnace to make glass.
- Glass + copper wires + 3D printer can be used to make solar panels.
    - Note that with #361 I intend to start the base with a small number of 3D printers, so other robots could carry this out.  But also, the base will start with a bunch of solar panels too.
- Solar panel + counter can be used to make a calculator.


byorgey:
Once people are happy with we should also think about whether the tutorial needs to be updated.

byorgey:
At #336 we also discussed the possibility of being able to create sand by drilling rocks; we could easily add that later.

data/entities.yaml
```diff
@@ -378,7 +396,7 @@
 - name: plasma cutter
   display:
     attr: device
-    char: '+'
+    char: '*'
```
xsebek, line 399: This is also a flower, detonator and gear? `P` would make sense as a downward pointing plasma gun :thinking:

xsebek (approved):
> Once people are happy with we should also think about whether the tutorial needs to be updated.

With what? Otherwise, this looks good to me. :+1:

data/entities.yaml
```diff
@@ -216,6 +216,24 @@
   properties: [portable]
   capabilities: [float]
 
+- name: sand
+  display:
+    attr: copper
+    char: '░'
```
xsebek, line 222: Actually, I can not see the `░` character very well. Maybe its the dirt background and copper attribute?

xsebek: Oh, the name of the character is "light shade" and my font takes that literally:

image

data/entities.yaml
byorgey: They are all distinguished by color.  But P for plasma cutter sounds good.

data/entities.yaml
byorgey: It's the same character used for dirt, grass, and stone, but should be the same color as copper.  I do want to make sure it's distinguishable in different terminals which don't all handle color the same...

byorgey: Hmm, in my terminal the copper and dirt colors are easily distinguishable.  But I tried it in gnome-terminal and the two colors are very very hard to tell apart.  Any suggestions for a different color and/or character to use?

xsebek: I would suggest `IceT` background and the `sand` attribute which I completely forgot about. :smile:

byorgey: Hah, didn't know we had a `sand` attribute!  I updated it, how does it look now?

xsebek: Looks good enough, in light colour schemes it very much looks like sand for me, in dark ones it's a bit muddy but still distinguishable. :+1:

PR #367, opened 2022-06-05
Upgrade dependencies

### LINKED ISSUE #366: Add support for GHC 9.2
We should bump our dependencies. Namely, [dependent-sum](https://github.com/obsidiansystems/dependent-sum/commit/683ff2d1c649a4c8ec87f7be4c96e235dde518b2) and [th-extras](https://github.com/mokus0/th-extras/commit/fa8a2855883220480b80ca0d8224e9c3a8cd2c4a) have cut releases and we can get them from Hackage.

We can also check what other packages have been up to. :slightly_smiling_face: 

Finally, add [GHC 9.2.3](https://www.haskell.org/ghc/blog/20220527-ghc-9.2.3-released.html) to CI.

Note that Stack is not run in CI and needs to be tested manually so check that those packages are on stackage.

byorgey:
- bump upper bounds on dependencies
  - fixes breaking type changes in Aeson (`KeyMap`) and LSP (`UInt`)
- remove special `th-extras` and `depenent-sum` deps
- adds GHC 9.2 to CI
- closes #366

### DIFF
--- a/src/Swarm/Game/Entity.hs
+++ b/src/Swarm/Game/Entity.hs
@@ -484,6 +484,10 @@ instance Hashable Inventory where
   hash = inventoryHash
   hashWithSalt s = hashWithSalt s . inventoryHash
 
+-- | Inventories are compared by hash for efficiency.
+instance Eq Inventory where
+  (==) = (==) `on` hash
+
 -- | Look up an entity in an inventory, returning the number of copies
 --   contained.
 lookup :: Entity -> Inventory -> Count
--- a/.github/workflows/haskell-ci.yml
+++ b/.github/workflows/haskell-ci.yml
@@ -53,6 +53,11 @@ jobs:
     strategy:
       matrix:
         include:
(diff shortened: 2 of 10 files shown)


src/Swarm/Game/Entity.hs
```diff
@@ -487,6 +487,10 @@ instance Hashable Inventory where
   hash = inventoryHash
   hashWithSalt s = hashWithSalt s . inventoryHash
 
+-- | Entities are compared by hash for efficiency.
```
xsebek, line 490: <kbd>Ctrl</kbd><kbd>C</kbd> ...  :facepalm:

xsebek: Fixed :heavy_check_mark:

xsebek:
@byorgey Also could we have a <kbd>Rebase and merge</kbd> allowed for maintainers? :innocent:

PR #368, opened 2022-06-06
require `--cheat` flag to allow toggling creative mode with `m` key

### LINKED ISSUE #360: Get rid of `m` key to switch modes unless explicitly enabled via "cheat mode"
Now that we can start a new game in either classic or creative mode, there probably no longer should be a way to switch between the modes arbitrarily (as we currently have with the `m` key on the world).  However, being able to switch between modes is still useful for testing (e.g. to quickly generate a certain device and then test how it works in classic mode).  So I propose:

- Add a command-line flag `--cheat` which enables "cheat mode" (off by default)
- Normally, when cheat mode is off, there is no way to toggle creative mode.
- However, when cheat mode is enabled, then we can toggle creative mode with the `m` key as we can now.

byorgey:
Closes #360.


byorgey:
Good idea, done.  `Ctrl`+`M` is not possible because terminals apparently interpret it as a newline character or something weird like that.  So I changed it to `Ctrl`+`K`.  You know, `K` for "Kreative".

xsebek:
> Good idea, done. `Ctrl`+`M` is not possible because terminals apparently interpret it as a newline character or something weird like that. So I changed it to `Ctrl`+`K`. You know, `K` for "Kreative".

[Germans will be delighted](https://en.wiktionary.org/wiki/kreativ#German) :+1:

byorgey:
> @byorgey we should eventually note that in `TUTORIAL.md` when we get to update it 😅

Ah, whoops, I forgot to update the tutorial!  I don't think we need to mention cheat mode in the tutorial, we can just mention creative mode games.

PR #372, opened 2022-06-08
add recipe weights and choose randomly among matching recipes

byorgey:
Recipes can now optionally have a `weight` (with a default weight of
1).  Any time multiple recipes match the criteria by which we are
selecting a recipe, one of the recipes will be chosen randomly, with
probability proportional to its weight.

As a simple example, drilling a boulder now produces 3 rocks with
probability 3/4 and 4 rocks with probability 1/4.

The ultimate purpose of this is to support some things I would like to propose in relation to mines, iron, etc. but it seems like a useful/interesting feature on its own.


src/Swarm/Game/Step.hs
```diff
@@ -846,9 +865,11 @@ execConst c vs s k = do
         not (null recipes)
           `holdsOrFail` ["There is no known recipe for making", indefinite name, "."]
 
-        -- Try recipes and remeber the first one that we have ingredients for.
+        -- Try recipes and make a weighted random choice among the
+        -- ones we have ingredients for.
+        chosenRecipe <- weightedChoice (\(_, _, r) -> r ^. recipeWeight) (rights (map (make (inv, ins)) recipes))
```
xsebek, line 870:
```suggestion
        chosenRecipe <- weightedChoice (^. _3 . recipeWeight) (rights (map (make (inv, ins)) recipes))
```

xsebek (approved):
Nice, I might even be lucky in #373 and narrowly avoid merge conflict :sweat_smile:

Issue #201, opened 2021-10-08
`require` commands to artificially require devices or supplies


byorgey:
[**Edited to add**: _I have realized that this issue is even more general, and is not specifically about devices per se, but about requiring any kind of needed supplies.  See the comments later in the issue._]

In certain situations, it's possible you might want a robot to have a certain device installed even though its program does not require the capabilities provided by that device.  For example, this can happen with `logger`s: if the robot might crash and you want it to be able to log its failure.  Or you might be planning to `reprogram` the robot later and know the later program will require a certain capability that it does not require at first.  We should add a command `require ` which artificially creates a demand for a specific device, so the `build` command will install it.  For example,
```
build "x" {require "logger"; move; grab; move}
```
will build a robot `x` with a `logger` installed.

Of course `require` is a no-op at runtime.

I think there is actually a question here whether the argument to `require` should name a *capability* or a *device*.  In general I think we need to clarify the relationship between capabilities and devices.

byorgey:
Yes, `build_with` works up to a point, but:
1. If you need *multiple* devices, you would have to make a version of `build_with` that takes a list, or make `build_with2` and `build_with3`...
2. It is easy to get wrong.  In fact I think your `build_with` may have a subtle race condition: depending on the order in which the robots execute and what command `program` executes first, the newly built robot may drive off before the `install` command is complete.  It would be safer to say `x <- build name {wait 1; program}`.
3. It is not compositional.  In all other cases, what devices a program needs is determined simply by what commands it uses.  But in the case of a few devices like `boat` and `logger`, you must *globally* remember to build them with the right invocation of `build_with`.  Imagine that you were designing a robot to carry out some complex task which you broke down into pieces.  One of those pieces involves crossing a lake.  You want to be able to say `require "boat"` *right in the definition* of that piece; the property of requiring a boat is thus expressed as a *local* property of that definition, and will compositionally carry over to any other bigger definitions which reuse that one.

I am all for having users define things themselves, but in this case I don't think they have the tools to actually build the abstraction they would really want.

byorgey:
Tried starting on implementing this and it seems to be a bit trickier than I thought, so writing down what I've learned so far and some current thoughts.

- I feel pretty strongly that the `require` command should take *the name of a device* as an argument, since that is what users know about.  Users shouldn't need to know or care about the capability system.
- Currently, we figure out capability requirements by calling the `requiredCaps` function, which returns a set of capabilities.  In order to properly handle `require`, we would need to:
[...]

PR #373, opened 2022-06-08
Iron

### LINKED ISSUE #93: Add iron to allow more advanced recipes
Since copper is already implemented in the game, I'd suggest adding iron in order to allow more advanced recipes. 

When it comes to the world gen, I'd suggest an `F` for the symbol (after the chemical element `Fe`) and choose either an orange color (then change copper to blue, in order to stay as close as possible to real world) or blue to stay close to what most games use those days.

We could add iron veins close to the mountains (and/or make terrains like suggested in #18). That would mean that we could get iron/copper in the wild (like it is now for copper) then have those terrains/veins to farm more for later use.

Finally, this would give options for the potential `robot chassis` suggested in #35 and also add/change other recipe (like the `drill` for example). Here's a few recipe suggestions to start:

- 1 iron ore -> 2 iron plates (1:2 or 1:1)
- 2 iron plates -> 1 iron gear (that would mean to rename the current `Gear` to this or differentiate `Wooden Gear` and `Iron Gear`)
- 2 iron plates + 1 iron gear + 2 copper pipes -> 1 robot chassis
- 1 robot chassis + 4 iron plates + 8 copper wires -> 1 robot

Those are obviously open for discussion.

### LINKED ISSUE #341: Compass device
In some challenges it is undesirable to be able to turn to cardinal directions.

For example in some algorithmic challenges the robot should only turn `left` or `right`, to teach thinking relative to the robot.

The chess horse challenge sets the robot direction to `[2,1]` so it moves like the knight piece. Turning north cheeses the challenge. :sweat_smile: 

## Solution
Create a compass device, some cardinal orientation ability which will be required for cardinal directions.

The recipe could be something really simple like:
```
1 water ─┬───── compass
1 copper ┘
   wire
```

## Alternatives

We could limit abilities directly in challenges, but it would still require creating some compass ability.

---

- credit for the idea goes to @byorgey in #339

### LINKED ISSUE #342: Suggest required entities
Currently, the exception thrown when the robot is missing some capability looks like this:
```
│ ───────────────────Log────────────────── │
│                                          │
│ missing device(s) needed to execute      │
│   command 'blocked'                      │
└──────────────────────────────────────────┘
```
Some commands are obvious, like `drill` requiring `drill`, others like `blocked` requiring `scanner` should be probably mentioned in the exception message.

In my mind there are three cases:
1. The player knows the necessary entity but the base/robot does not have it.
2. The player does not know the required entity.
3. There is no entity that the player could obtain to gain the capability (e.g. the `CGod` capability).

Giving a nice message to each of those cases could be slightly harder, but would IMO help beginners a lot.

xsebek:
- add iron ore, iron mine and iron vein (closes #93)
  - split gear into iron/wooden gear
  - add metal drill
  - add faster recipes with the metal drill
- add compass (closes #341)
- handle multiple entities providing the same capability
  - try to find if the robot has at least one entity providing the capability
  - when no entity could provide the capability rejects it too
- list required devices in the `Incapable` error (closes #342)


data/entities.yaml
```diff
@@ -137,6 +137,50 @@
     water or steam.
   properties: [portable]
 
+- name: iron plate
+  display:
+    attr: iron
+    char: '■'
+  description:
+  - Raw iron ore. Used to create more resilient tools.
+  - It can only be mined by drilling in the mountains.
```
byorgey, line 146: Seems like this description got copied from `iron ore` and never edited?

data/entities.yaml
```diff
@@ -137,6 +137,50 @@
...
+  display:
+    attr: iron
+    char: '*'
+  description:
+  - An iron gear.
+  properties: [portable]
+
+- name: iron ore
+  display:
+    attr: copper
```
byorgey, line 159:
```suggestion
    attr: iron
```

data/entities.yaml
```diff
@@ -137,6 +137,50 @@
...
+    char: 'C'
```
byorgey, line 160:
```suggestion
    char: 'F'
```

data/entities.yaml
```diff
@@ -268,14 +323,23 @@
...
   display:
     attr: wood
     char: '*'
   description:
   - A wooden gear.
   properties: [portable]
 
+- name: iron gear
+  display:
+    attr: wood
```
byorgey, line 336:
```suggestion
    attr: iron
```

data/entities.yaml
```diff
@@ -137,6 +137,50 @@
...
+  display:
+    attr: iron'
+    char: 'A'
+  description:
+  - A place in the mountains where raw iron ore can be mined.
+    As it is hidden inside a mountain, a tunnel needs to be
+    first drilled through, so that the vein becomes accessible.
+  properties: [unwalkable]
+
+- name: iron mine
```
byorgey, line 176: So one idea I had was to get rid of the distinction between `copper vein`/`iron vein` and `copper mine`/`iron mine` and just have `vein` and `mine`.  Given #372, the idea is that drilling in a mine would give you copper with some probability and iron with another (smaller) probability. We can also easily do things like a wooden drill might get you copper but never iron, whereas an iron drill could get you both, etc. --- we just have to specify the right recipes. While this scenario of getting everything from a single mine may not be very accurate in the "real world", I like it for a few reasons:

- It keeps things simpler in terms of the number of different entities in the world.
- It leaves the door open for adding other mine-able things (gold? diamonds? etc.) with much lower probabilities, without having to add yet more separate types of veins and mines.
- With multiple different types of mines, the difficulty for a player lies in locating all the different types and building separate drilling operations at each. With a single mine, the difficulty for the player lies in building a drilling and processing/sorting operation.  The latter sounds much more fun and interesting from a programming perspective than the former.

What do you think?

src/Swarm/Game/Step.hs
```diff
@@ -877,8 +878,12 @@ execConst c vs s k = do
...
 
         let nextLoc = loc ^+^ applyTurn d (rDir ? V2 0 0)
-        em <- use entityMap
-        drill <- lookupEntityName "drill" em `isJustOr` Fatal "Drill does not exist?!"
+
+        let toyDrill = lookupByName "drill" ins
+            metalDrill = lookupByName "metal drill" ins
+            insDrill = listToMaybe $ metalDrill <> toyDrill
+
+        drill <- insDrill `isJustOr` Fatal "Drill is required but not installed?!"
```
byorgey, line 886: This should not be a `Fatal` error.  The  `Fatal` error from before was for the situation that looking up `drill` in *the entity map* failed, i.e. the system as a whole did not know about the `drill` entity at all.  But the new code is just checking what kind of drill the robot has installed.  It could certainly happen that the robot is trying to execute a `drill` command even though it has no drill installed.  That is a runtime error but not a `Fatal` one.

byorgey (commented):
Some initial comments.  I haven't finished looking carefully through the changes regarding how we check for required devices etc.

data/entities.yaml
xsebek: I actually like both ideas - maybe we can split it into different scenarios :thinking: 

It does not require any extra logic in Swarm so the only overhead is extra recipes in our main `data/recipes.yaml`.
I think moving it to some *adventure mode* scenario would both clean up and leave our options open.

My only concern is how to order the changes - I will have to check if scenarios support all I need now or if it is easier to move it over later. :slightly_smiling_face:

src/Swarm/Game/Step.hs
xsebek: I suspected as much, but what example would escape our capability checking ahead of drilling?

xsebek:
Thanks for taking a look @byorgey, the copy-pasting is definitely a problem :sweat_smile: 

I spent most type debugging the capability checking, so now that it (hopefully?) works I will come back and streamline it.

data/entities.yaml
xsebek: Alternatively, we could parametrize entities by their underlying material to cut down on boilerplate.

I am not sure how it would work, but maybe it will become more evident once we add gold and other late-game stuff.

byorgey: You said you like both ideas.  What is it you like about the idea of having separate `copper mine`, `iron mine`, etc.?

I don't mind supporting both, and I am not so much concerned about boilerplate as I am about the gameplay experience.  I don't mind having extra entities and recipes in `recipes.yaml` etc.  So we could have both `mine` that works as I described with weighted random results as well as specific `iron mine`, `copper mine`, etc.  The biggest issues then would be (1) which version do you get in the "default" mode?  Even if we support both we still have to decide which is the default game.  And (2) in order to support both we have to allow multiple "world functions" and provide a mechanism for scenarios to specify which world function they will use.  I think this was discussed in an issue somewhere but I can't find it at the moment.

src/Swarm/Game/Step.hs
byorgey: Hmm, good question.  I guess you're right, if this happens it means something has gone wrong with capability checking, so perhaps that should actually be a `Fatal` error.

byorgey:
By the way, I very much like all the new recipes, the way metal drill does things faster than a regular drill, etc.

data/entities.yaml
xsebek: Fixed :+1:

xsebek:
That Restyled action is a bit annoying since I by mistake pushed to my repo which confused Restyled and it wants to merge to `main`. :facepalm:

data/entities.yaml
xsebek: I like that it is visual, easier to debug and requires sending robots a bit further. :slightly_smiling_face: 

We could actually do with only mines and have them randomly drilled from `mountain`. That would address the number of entities in the world and make it simpler to add new mines, but it would be even harder to establish mining operations.

byorgey: Hmm, I like the idea of mines appearing from a drilled mountain with a certain probability (in all cases, whether there is only one kind of mine or multiple).  It requires more exploration.  It also feels weird to me that right now the mountains are different colors depending on whether there is a vein --- not that Swarm needs to reflect the real world necessarily, but if you have to drill for something, you should not be able to tell just by looking at the surface what you will find when you drill.

bench/Benchmark.hs
```diff
@@ -60,13 +60,7 @@ circlerProgram =
...
     in forever (
       move;
-      turn east;
-      move;
-      turn south;
-      move;
-      turn west;
-      move;
-      turn north
+      turn right;
```
byorgey, line 63: Let's change this to `turn right; move; turn right; move; ...` i.e. with the same number of turns and moves as before.  Otherwise the benchmark won't really be comparable to before, since the number of straight-line instructions the robot executes vs the number of recursive calls to `forever` can definitely make a performance difference.

data/recipes.yaml
```diff
@@ -220,6 +220,13 @@
   out:
   - [2, iron gear]
 
+- in:
+  - [1, iron plate]
+  - [1, water]
+  - [1, box]
+  out:
+  - [1, compass]
```
byorgey, line 228: I like it.

src/Swarm/Game/Step.hs
```diff
@@ -374,10 +374,19 @@ stepCESK cesk = case cesk of
...
   In (TBool b) _ s k -> return $ Out (VBool b) s k
+  -- Direction is simple too unless it is absolute (e.g. north)
+  -- and needs orient capability (provided by a compass).
+  -- This may be redundant just because of base escaping
+  -- capability checking (see #231).
+  In (TDir d) _ s k -> do
+    orient <- hasCapability COrient
+    if isCardinal d && not orient
+      then return $ Up (Incapable (S.singleton COrient) (TDir d)) s []
+      else return $ Out (VDir d) s k
```
byorgey, line 388: I am not sure about this.  Should we really need the `orient` capability just to *evaluate* a cardinal direction constant?

xsebek: Well, `base` would escape the check if I did not do it, but it still seems wrong:
```Haskell
> turn north -- compass is needed (but the error does not mention it)
> let d = north in build {turn d} -- once base has compass, nobody else needs it
```
I probably just overthought it and checking it when turning and drilling would work just fine.

- related #201 for the case when a robot is sent in some direction and it turns out it is cardinal at runtime (I should write a test for it...)
- related error message improvement #342

bench/Benchmark.hs
xsebek: Indeed it does seem to be 20% slower, I should have checked :+1:

data/entities.yaml
xsebek: I think I found a good balance:
- mountains can be drilled with an 8/1/1 probability to tunnel/copper mine/iron mine
- in scenarios, we can use the deterministic vein ➔ mine custom recipes and entities

bench/Benchmark.hs
xsebek: Fixed :heavy_check_mark:

src/Swarm/Game/Step.hs
xsebek: Reverted and moved the check to commands using the direction.

src/Swarm/Language/Syntax.hs
```diff
@@ -120,12 +120,12 @@ dirInfo d = case d of
   DLeft -> relative (\(V2 x y) -> V2 (- y) x)
   DRight -> relative (\(V2 x y) -> V2 y (- x))
   DBack -> relative (\(V2 x y) -> V2 (- x) (- y))
+  DDown -> relative (const down)
```
xsebek, line 123: I think I might have at some point advocated for `down` being a cardinal direction, but now I need it to be allowed without a compass :sweat_smile: I do not think it's worth coming up with another name for the set of directions that are cardinal but not `down`. Writing `/= down` everywhere is also an option, but not a good one. :shrug:

xsebek:
Hey @byorgey, could you please take a look and test if this works?

I made a testing playground with the old `copper vein`, `iron vein` and both drills and `compass` in inventory:
```bash
cabal run swarm:swarm -- --scenario 03Challenges/02-drill_test --cheat
```
Screenshot from 2022-06-11 02-22-12

I did not get around to automating the tests yet, but I would be happy to let someone else test my fragile code if they felt like it :wink:

src/Swarm/Game/Step.hs
xsebek: Given that even build inside build checks ahaed, I am going to leave it as Fatal and see if it comes up.

```
> build {r <- build {move; drill down; log ""}; log ""};
<r3> : robot
> salvage
```
```
[busy_williamson] build: You do not have required devices:
  - metal drill or drill
  - logger
```

data/entities.yaml
byorgey: Ah, OK, I like that.

byorgey:
The test scenario seems to work.  However, when I started creative mode and tried sending a robot to drill some mountains, it did not work.  Maybe it is failing due to having no `drill` installed even though we are in creative mode?

src/Swarm/Game/Step.hs
```diff
@@ -1412,6 +1383,68 @@ execConst c vs s k = do
...
+        -- capabilities.
+        (caps, _capCtx) = requiredCaps currentContext cmd
+
+        -- list of possible devices per capability
+        capDevices = map (`deviceForCap` em) . S.toList $ caps
+
+        -- device is ok if it is available in the inventory of parent
+        -- when building or installed in target robot when reprogramming
+        deviceOK d = inventory `E.contains` d
+        ignoreOK ([], miss) = ([], miss)
```
byorgey, line 1427: Took me a little while to figure out what this is for.  Maybe add a comment, something like: ignoreOK is given two lists of devices which all provide a given capability, a list of OK devices and a list of the other devices. If *at least one* device providing a capability is OK (i.e. the first list is not empty), then we can ignore all the alternative devices that are not installed.

src/Swarm/Util/Yaml.hs
```diff
@@ -61,6 +62,9 @@ liftE = E . const
 withE :: Semigroup e => e -> With e f a -> With e f a
 withE e (E f) = E (f . (<> e))
 
+fromE :: (Monad f) => With e f e
```
byorgey, line 65: I would call this `getE`.  `fromE` sounds like you're converting `e` to some other type.

src/Swarm/Game/Step.hs
```diff
@@ -1412,6 +1383,68 @@ execConst c vs s k = do
...
+
+    -- check if robot has all devices to execute new command
+    (creative || all null missingDeviceSets)
+      `holdsOrFail` ( singularSubjectVerb subject "do" :
+                      "not have required devices:" :
+                      (("\n  - " <>) . formatDevices <$> filter (not . null) missingDeviceSets)
+                    )
+    -- check that there are in fact devices to provide every required capability
+    (creative || not (any null deviceSets))
+      `holdsOrFail` [singularSubjectVerb subject "can", "not perform an impossible task"]
```
byorgey, line 1444: To make sure I'm understanding correctly, if this "impossible task" error is triggered, it means that there do not exist any devices to provide the required capability, right?  As in, we have some capability but no actual devices yet in `entities.yaml` which provide it?

src/Swarm/Language/Syntax.hs
byorgey: Yeah, making `down` a relative direction makes sense, given the existence of `compass`.  You should not need a `compass` to find the ground. :laughing:

byorgey (commented):
Everything looks good to me, but there's still something I don't understand going on with drilling in creative mode.

src/Swarm/Game/Step.hs
byorgey: Hmm, I actually got this error while playing in classic mode and it's very mysterious.  How hard would it be to output the name of the capability for which there is no device, with a message apologizing and a link to the issue tracker?

xsebek: I expected this to only come up for `CGod` tier tasks, like `create "life"`, where the message would make sense.

Maybe I could check if it is `CGod` and otherwise link to the issue. Something like:
```Haskell
[ singularSubjectVerb subject "do", "not have", indefiniteQ capName
, "capability, because no device provides it."
, "\nSee https://github.com/swarm-game/swarm/issues/26."
]
```

However, I feel like it should be `Incapable` exceptions job to figure this out. :smirk:

src/Swarm/Util/Yaml.hs
xsebek: I was considering `askE`, to make it look like `ReaderT`, but `getE` sounds good too.

byorgey: Yeah, `askE` is probably better!

xsebek: Too late, I used your suggestion :laughing:

byorgey: Hah! Either one is fine. 😄

src/Swarm/Game/Step.hs
xsebek: Some of it was common, but building and reprogramming can give a more specialised error message, so I keep some duplicity code. Oh well, at least I get to close #342 :innocent:

src/Swarm/Game/Step.hs
xsebek: Sure, I added some comments :+1:

data/scenarios/03Challenges/02-drill_test.yaml
```diff
@@ -0,0 +1,140 @@
...
+    display:
+      attr: water
+      char: ' '
+    description:
+    - An infinite ocean of water.
+    properties: [known, portable, growable, liquid]
+    growth: [0,0]
+    yields: water
+
+  - name: knownwavywater
```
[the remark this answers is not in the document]
xsebek: I suggest we solve it later, I made an Issue so we don't forget:
- #387

xsebek:
> Everything looks good to me, but there's still something I don't understand going on with drilling in creative mode.

@byorgey it was actually more than drilling, by mistake I did not install any required entities if they were not present in inventory.

The commit [9e24293](https://github.com/swarm-game/swarm/pull/373/commits/9e2429394850d9f51a9b857abd46924798512663) should fix that :slightly_smiling_face:

data/scenarios/03Challenges/02-drill_test.yaml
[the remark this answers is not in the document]
byorgey: 👍

src/Swarm/Game/Exception.hs
```diff
@@ -43,18 +75,84 @@ data Exn
...
+-- >>> incapableError (S.singleton CCreate) (TConst Create)
+-- Missing the create capability for:
+--   'create'
+--   but no device yet provides it. See
+--   https://github.com/swarm-game/swarm/issues/26
+formatIncapable :: EntityMap -> IncapableFix -> Set Capability -> Term -> Text
+formatIncapable em f caps tm
+  | CGod `S.member` caps =
+    unlinesExText
+      [ "Can not perform an impossible task:"
```
byorgey, line 128: I think this error message could still be improved.  I don't think it is going to make much sense if a player stumbles across it.  *e.g.* imagine that they type `as` accidentally and notice that it has a type, then they wonder what it does so they try it out. They are going to wonder *why* the task is impossible. Somehow we should convey that this is a command that is not available during normal play.  We can definitely make it funny though.  Some ideas:
- "This command is not available to mere mortals" / "This command is forbidden"
- "Thou shalt not utter such blasphemy"
- If we wanted to be more informative and less funny, "This command is only available to system robots and scenario definitions".

byorgey (approved):

byorgey:
This is looking great! :rocket:

src/Swarm/Game/Step.hs
byorgey: Also, just an update: the reason I got this error while playing in classic mode was actually due to #389 .  So that was making it more mysterious than necessary.

byorgey:
But up to you whether we want to go ahead and merge this and file a separate issue re the problems I mentioned above!  At this point I would actually lean towards doing it that way rather than hold this up longer.

byorgey:
A capability error reporting bug that I am likewise happy to split out into a separate issue after merging this, just want to write it down while it's fresh in my head:

- Start a classic mode game with `--cheat`
- Switch into creative mode and `create "branch predictor"`
- Type `build {move; if (1 > 2) {} {}}`
    - As expected we get an error saying no device provides the `compare` capability.
- Now type `build {move; turn left; if (1 > 2) {} {}}`
    - All we have done is add `turn left`; we would expect to get the same error.  Instead we get an error like `You do not have the devices required for: 'move; turn left; if (1 > 2) {noop} {noop}'  please obtain:   - branch predictor`
    - But we do have a branch predictor, so this error is bogus.  Not sure yet whether the capability checking or error reporting is to blame.

src/Swarm/Game/Exception.hs
xsebek: I guess the Elden ring is now popular, so blasphemy it is:
```Haskell
incapableError (S.singleton CGod) (TConst As)
-- Thee shalt not utter such blasphemy:
--   'as'
--   If't be true thee wanteth to playeth god, then tryeth Creative game.
```

xsebek:
@byorgey I think your first bug is present on main too, the Drill capability is not reported by `requiredCaps`.
The difference is, that I call that a `Fatal` error because `build` should have given it to the robot, while `main` drives the robots to rocks and there they stop without a drill.

See the debug logs I added:

### Main

### Iron
I was also missing standard devices, so I added them back:
Screenshot from 2022-06-14 16-48-55

---

I have no idea what the second bug is, but I hope it is not really my fault :smile:

PR #379, opened 2022-06-10
resolve seed every time a scenario is selected to be played

### LINKED ISSUE #369: Seed is not reset between multiple new games
**Describe the bug**
Within a given run of the app, selecting e.g. "classic game" twice yields the same world.

**To Reproduce**
1. Start the app
2. Select "New Game > Classic Game"
3. Quit back to the menu.
4. Select "New Game > Classic Game" again.

**Expected behavior**
A different randomly chosen world should appear the second time, but instead, we get exactly the same world both times.  

It seems to be an issue with not resetting the seed, since all other game state is in fact regenerated fresh.

byorgey:
Seed resolution used to happen in `loadScenarioFile`, but that was the
wrong place.  Resolution now happens in `playScenario`.  This means that, e.g. if you
select "New Game > Classic" from the menu, then quit back to the menu, then start
a second new game, you will get a different random world each time.

Fixes #369.

### DIFF
--- a/src/Swarm/Game/Scenario.hs
+++ b/src/Swarm/Game/Scenario.hs
@@ -43,7 +43,6 @@ module Swarm.Game.Scenario (
   loadScenarios,
 ) where
 
-import Control.Applicative ((<|>))
 import Control.Arrow ((***))
 import Control.Lens hiding (from, (<.>))
 import Control.Monad (filterM)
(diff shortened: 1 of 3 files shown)


byorgey:
The automatic restyler fix obscures the actual changes a bit.  Take a look at the original commit https://github.com/swarm-game/swarm/pull/379/commits/a7b0643b9e3aa5602e922324744d19347d11a25b for an easier view of what actually changed.

xsebek (approved):
Works for me :+1: 

Btw. if you restructure a function and Fourmolu changes indentation, you can ignore whitespace for review:


|Screenshot from 2022-06-10 10-36-15|
|--|

byorgey:
Cool, didn't know that, thanks!

PR #381, opened 2022-06-10
Add note about create

xsebek:
- add a note about 'create' to the make command
- also fix spacing before punctuation in error messages

### DIFF
--- a/src/Swarm/Game/Step.hs
+++ b/src/Swarm/Game/Step.hs
@@ -763,7 +763,7 @@ execConst c vs s k = do
         -- Make sure the other robot exists
         other <-
           robotWithID otherID
-            >>= (`isJustOrFail` ["There is no robot with ID", from (show otherID), "."])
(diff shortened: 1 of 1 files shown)


src/Swarm/Game/Step.hs
```diff
@@ -852,25 +852,28 @@ execConst c vs s k = do
         em <- use entityMap
         e <-
           lookupEntityName name em
-            `isJustOrFail` ["I've never heard of", indefiniteQ name, "."]
+            `isJustOrFail` ["I've never heard of", indefiniteQ name <> "."]
 
         outRs <- use recipesOut
 
+        creative <- use creativeMode
+        let create l = l <> ["You can use 'create" <> name <> "' instead." | creative]
```
byorgey, line 860:
```suggestion
        let create l = l <> ["You can use 'create \"" <> name <> "\"' instead." | creative]
```

byorgey (approved):

Issue #26, opened 2021-09-21
Design and implement more capabilities and devices


byorgey:
There are quite a few capabilities that are missing; see the comments in [Swarm.Language.Capability](https://github.com/byorgey/swarm/blob/main/src/Swarm/Language/Capability.hs).  This will require some creative thinking and careful planning and testing.

---

Here is a list of commands and related issues (TODO):

- [x] **View** this should also require something. #17 
- [x] **Not**, **And**, **Or** some kind of boolean logic cap? #562
- [x] **Inl**, **Inr**, **Case** should require a capability for sums #563
- [x] **Fst**, **Snd** should require a cap for pairs #563
- [x] **Try** definitely needs to require something #132 
- [x] **Wait** could require **time** capability #402 

Here is the list of capabilities that are missing devices:

- [x] **senseloc** (Whereami)
- [x] **appear** (Appear) #1027 
- [x] **setname** (Setname) #1028 
- [ ] #1041
- [x] **random** (Random) #497
- [x] ~~**create**~~ (Create) make restricted instead in #408 
- [x] **compare** (==,!=,<, <=,>,>=) added with comparator in #395
- [x] **arith** (+,-,*,/) added with calculator in #362

Here is a list of devices that provide a capability but are missing a recipe (see also #361):

- [ ] #1017
- [x] **dictionary** #507 
- [x] **tool kit** #361 
- [x] **grabber** #507 
- [x] **3D printer** #496
- [x] **flash memory** #498
- [x] **treads** #434 

In addition, the following commands require **god** capability and are currently only intended for scenario development and not available to players (except for creative mode):
- **RobotNamed**, **RobotNumbered**
- **As**
- **Create**

---

Since #373 the exception users will see if there are no devices to provide required capability looks like this:
```
Missing the random capability for:
  'random'
 but no device yet provides it. See
 https://github.com/swarm-game/swarm/issues/26
```
Restricted commands give different message:
```
Thee shalt not utter such blasphemy:
  'as'
  If't be true thee wanteth to playeth god, then tryeth Creative game.
```

xsebek:
I would really like to link the “counter” to a small solar panel and have a pocket calculator with arithmetic capability. 🧮

byorgey:
I like that idea! =)  I had some vague thoughts about making an "ALU" but I like this better.  What should be the recipe for a solar panel?

xsebek:
[...]

Issue #133, opened 2021-10-01
callcc


byorgey:
With CEK machines, it is extremely easy to add `callcc` and `throw`, so we might as well.  Perhaps this capability should be provided by some kind of `time machine` device (ingredients: `clock`, `phone booth`, ...?)

Concretely:
* Add a new type: if `t` is a type, `cont t` is the type of continuations expecting a value of type `t`.
* Add constants `callcc : forall t. (cont t -> t) -> t` and `throw : forall t1 t2. t1 -> cont t1 -> t2`.
* To evaluate `callcc`, just package up the current continuation in the CEK machine and pass it to the function (i.e. push an `FApp` frame and produce the current continuation as a value).
* To evaluate `throw v k`, produce the output `v` but *replace* the current continuation with `k`.

polux:
https://www.brics.dk/RS/06/15/BRICS-RS-06-15.pdf also describes a tweaked  version of the CEK machine that supports delimited continuations, which would be even more fun :)

byorgey:
It seems there are many different sets of control operators one can have for working with delimited continuations, but `shift/reset` seems the simplest and most standard?  https://okmij.org/ftp/continuations/index.html#tutorial is a relevant tutorial.

https://brics.dk/RS/03/41/BRICS-RS-03-41.pdf seems like a simpler / more directly relevant reference for implementing shift + reset in a CE(S)K machine context.

byorgey:
I really want the device providing continuations to be a `time machine`.  More thoughts on a potential recipe tree:

- `clock`
    - `quartz`
    - 32 `iron gear`s
    - `glass`
- 256 `solar panel`s (a time machine requires a lot of power!)
- a bunch (16? 32? 64?) of `circuit`s
- `phone booth`
    - 32 `board`s
    - some other silly things?

byorgey:
So the essential idea of how to update the CESK machine to support delimited continuations is just to store, instead of a single continuation, a *stack* of continuations (though this is perhaps a bit confusing since we must remember that a continuation is itself represented as a stack of frames).  Most of the time, things interact with the continuation on the top of the stack just as they did when we had a single continuation.  But:
- To evaluate `reset t`, we focus on evaluating `t` with a new empty continuation pushed on top of the continuation stack.
- To evaluate `shift k. t`, we focus on evaluating `t`, with a new binding `k -> C` in the environment, where `C` is the current continuation; and that current continuation is *replaced* by an empty continuation on top of the continuation stack.
[...]

PR #395, opened 2022-06-14
Some new devices and recipes

byorgey:
- Add quartz and quartz mines
- Add `clock`, `silicon`, and `circuit`
- Add `comparator` device
- Metal drills can now drill through boulders too
- Add pixels to the world, and recipes for `camera` and `scanner`

See discussion at https://github.com/swarm-game/swarm/issues/26#issuecomment-1152743741 and https://github.com/swarm-game/swarm/pull/361 .


xsebek:
> Metal drills can now drill through boulders too.

Hah, thanks :+1:

data/recipes.yaml
```diff
@@ -44,6 +44,18 @@
   out:
   - [1, wooden gear]
 
+- in:
+  - [1, board]
+  - [2, wooden gear]
+  out:
+  - [1, seesaw]
```
xsebek, line 51: Wouldn't it be better to use iron for a seesaw? It feels strange that we won't use it for cutting boards and stuff. :confused: 

I would not mind having a scale just from rocks and a branch, but some joke entity would be fun. :thinking:

byorgey: A seesaw is not for cutting: https://en.wikipedia.org/wiki/Seesaw  It is supposed to be a joke... :smile:   Maybe `seesaw` is a US-specific term?  We could call it a `teeter-totter` instead, maybe that sounds funnier anyway.

byorgey: Also, in terms of gameplay, I think a `comparator` should not be too hard to make (in particular I don't think it should require any iron).  They come in very handy.  You need a comparator if you want to do basic things like recurse on an `int` (so you can tell when it reaches 0).  I also do things like set up a robot to accept harvested trees, and when it notices that the number of trees is over a certain amount, then it crafts them into various products and sends them along.

TristanCacqueray: `teeter-totter` sounds better, I also thought seesaw was something for cutting stuff :)

byorgey: OK, done!

byorgey:
The `clock` will probably end up being an ingredient for other things (e.g. a `time machine`, see #133), but I'd like to make it a device in its own right.  We could make it provide the ability to `wait`.  We could also create an instruction `time` that gets the absolute number of ticks since the beginning of the game, perhaps.  But I think I might do that in a separate PR.

PR #398, opened 2022-06-15
copy over known (count 0) items when salvaging

### LINKED ISSUE #391: Salvaging a robot should salvage its knowledge as well as its actual contents
Suppose you make a robot to do some `scan`ning, planning to have it `upload` to the base, but something goes wrong.  If you send another robot out to `salvage` it, it seems like the knowledge it obtained never makes it back to the base.

byorgey:
Fixes #391.

### DIFF
--- a/src/Swarm/Game/Step.hs
+++ b/src/Swarm/Game/Step.hs
@@ -1294,6 +1294,11 @@ execConst c vs s k = do
                 `isJustOr` Fatal "While executing 'salvage': there's no such thing as a logger!?"
             when (creative || inst `E.contains` logger) $ robotLog <>= target ^. robotLog
 
+            -- Immediately copy over any items the robot knows about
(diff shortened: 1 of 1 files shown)


xsebek (approved):
Works for me :+1: 

I wonder if we can determine if some robot knows an entity in Swarm language?
Say I wanted to write a test/challenge checking that:
```
has "tree" // false
count "tree" // 0

build {move; scan left; turn back; move} // World 0
salvage

has "tree" // false
count "tree" // 0
```

byorgey:
> I wonder if we can determine if some robot knows an entity in Swarm language?

Ah, interesting question.  Right now, we cannot.  `count "tree"` will always evaluate to 0 whether the robot knows about that entity or not.  The easy way would be to add a command `knows : string -> bool`.

Issue #394, opened 2022-06-14
Inner build does not correctly require Drill


xsebek:
From #373 by @byorgey:

> So I ran into one additional issue while testing out some things. In creative mode I tried something like this:
> ```
> def forever = \c. c ; forever c end
> def unblock = try {drill forward} {} end
> def push = unblock; move end
> 
> build {forever (push; build {turn right; forever push})}
> ```
> ...
> However, it doesn't work. The initial robot works fine, drilling along and building another robot at each step. However, all those secondary robots don't end up with a drill device and crash with a fatal error when they try to execute the drill command.

The problem was present before, but not visible until the robots stopped without a drill.

In this new iron age, it is immediately a fatal bug once the robot attempts to drill.

## Screenshots

A bronze age screenshot from the main branch:

![Screenshot from 2022-06-14 17-01-12](https://user-images.githubusercontent.com/44544735/173629200-67c3adf3-3af3-4f2b-9f41-3940b1d4a359.png)
<sup>It's almost the same now too, so I won't include another screenshot here.</sup>

Log of the robot missing a drill:

![Screenshot from 2022-06-14 17-19-14](https://user-images.githubusercontent.com/44544735/173629709-3bdb3bce-fb4d-420f-aca5-a912561010b9.png)

## Additional context

I traced the issue to the `requiredCaps` function which did not require `CDrill`.

byorgey:
Oh, I like your use of `log` and `robotNamed` to be able to find and view the log for the inner robot. =D

xsebek:
Surprisingly this program in Classic mode does not run into a Fatal error:
```
def forever = \c. c ; forever c end;
def unblock = try {drill forward} {} end;
def push = unblock; move end;
log "Hi, I am base";
r <- build {
  wait 2;
  log "Hi, I am builder";
  forever (build {log "Hi, I am pusher"; turn forward; forever push}; log "- robot built")
};
give r "logger";
give r "drill";
```
These are all the robots logs:
```
Hi, I am base
Hi, I am builder
- robot built
build: You do not have required devices, please obtain: 
  - logger
Hi, I am pusher
You do not have the devices required for:
  'drill'
  please install:
   - metal drill or drill
```

xsebek:
Once #400 is merged, you can test this with an included scenario:
```bash
$ cabal run test:swarm-integration -- -p '/build with drill (#394)/'
Up to date
Tests
  Test scenario solutions
    Regression tests
[...]

PR #400, opened 2022-06-15
Test scenarios

### LINKED ISSUE #388: Test scenarios
Currently, we check that scenarios can be parsed.

I propose we test that they can be solved:

- add a `solution` field to scenarios, which will contain the winning program
- load this program to `base`
- run CESK machine
- verify the `win` condition

This would help us make sure that scenarios are functional, which will be harder once we have many of them for the tutorial.

It would also create an easy way to write tests without knowing any Haskell!

xsebek:
- add winning solutions to scenarios
- test the winning solutions
  - fail if any robot has a `Fatal error` in its log
  - timeout in X seconds (default 1)
- replace startup logging to `stdout` with marking errors
- fix creative mode ignoring scenario inventory
- fix integration test discovery (challenges are in a subfolder, so recurse)
- add a *Testing* scenario collection hidden without the `--cheat` flag
- closes #388


xsebek:
@byorgey I wrote a test case for #394, but then I needed `expectFail`, so I put the automatic testing of all scenarios to rest.

I also allow `testSolution'` to check the final `GameState` so automating that is probably not that useful.

xsebek:
I really like [tasty-expected-failure](https://hackage.haskell.org/package/tasty-expected-failure-0.12.3/docs/Test-Tasty-ExpectedFailure.html):
|image|
|-|

xsebek:
Looking at the image above, I noticed we do not actually test all scenarios. :sweat_smile: 

Now it is fixed. :heavy_check_mark:

src/Swarm/Game/State.hs
```diff
@@ -515,7 +522,7 @@ playScenario em scenario userSeed toRun g = do
       & ix 0 . robotInventory
         %~ case scenario ^. scenarioCreative of
           False -> id
-          True -> const (fromElems (map (0,) things))
+          True -> union (fromElems (map (0,) things))
```
byorgey, line 525: What is the purpose of this change?

src/Swarm/TUI/Controller.hs
```diff
@@ -145,7 +145,7 @@ handleNewGameMenuEvent scenarioStack@(curMenu :| rest) s = \case
             & gameState .~ gs'
       Just (SICollection _ c) ->
         continue $
-          s & uiState . uiMenu .~ NewGameMenu (NE.cons (mkScenarioList c) scenarioStack)
+          s & uiState . uiMenu .~ NewGameMenu (NE.cons (mkScenarioList (s ^. gameState . creativeMode) c) scenarioStack)
```
byorgey, line 148: I don't see how the `creativeMode` flag in the `gameState` is relevant here.  It doesn't really tell us much unless we're currently playing a game.  In particular it does *not* tell us whether Swarm was started with the `--cheat` flag enabled.

test/integration/Main.hs
```diff
@@ -0,0 +1,162 @@
...
+  children <- mapM (\path -> (,) path <$> readFile path) filePaths
+  -- recurse
+  sub <- filterM doesDirectoryExist paths
+  transChildren <- concat <$> mapM (`acquire` ext) sub
+  return $ children <> transChildren
+ where
+  hasExt path = takeExtension path == ("." ++ ext)
+
+data Time
+  = -- | One second should be anough to run most programs.
```
byorgey, line 87:
```suggestion
  = -- | One second should be enough to run most programs.
```

src/Swarm/Game/State.hs
xsebek: [Fix creative mode ignoring scenario inventory](https://github.com/swarm-game/swarm/pull/400/commits/602cefa70739883a076109495df622b17cbb9383) :slightly_smiling_face:

If I add inventory to the base in a scenario, originally it would overwrite it with zero of everything.
I could workaround by adding `program: "create X; create Y; ..."`, but I thought it was a bug, so I fixed it.

src/Swarm/TUI/Controller.hs
xsebek: Well spotted! :+1: Weird that I did not notice, but I guess I did not test it too much. :sweat_smile:

test/integration/Main.hs
xsebek: Thanks, fixed :+1:

src/Swarm/TUI/Controller.hs
xsebek: Now it is `(s ^. uiState . uiCheatMode)` as it should have been. :heavy_check_mark:

src/Swarm/Game/State.hs
xsebek: The reason why I looked for it was that the new test case would not work without inventory in creative mode. 

(I was also trying modified versions of it where you start with a logger and give it to the robot.)

I expect being able to specify inventory even in creative mode will come in handy. :slightly_smiling_face:

xsebek:
I rebased, cleaned up the history and added the new portal challenge solution to the tests :innocent: 

@byorgey given that there are some changes outside tests, I would like to merge the commits as they are now.
Unless you have more suggestions of course :slightly_smiling_face:

src/Swarm/Game/State.hs
byorgey: Aha, I see now! 👍

byorgey (approved):
Looks good!

Issue #343, opened 2022-05-28
Robot to name or id and back


xsebek:
This is a meta issue for discussion of the way we convert between robot, its name and id.

```
      self
       │
       ▼
   ┌─────────────┐◄───┐
   │current robot│    │setname
   └──────┬──────┴────┘
    whoami│
          └───►┌──────────┐
               │          │
┌─────────────►│robot name│◄───────────┐
│as r {whoami} │          │           ?│
│          ┌───┴──────────┴───┐        │
│          │find             ?│        │
│          ▼                  ▼        │
│ ┌──────────┐◄─────────────┬────────┐ │
└─┤          │      ?       │        ├─┘
  │robot type│              │robot ID│
┌─┤          │      ?       │        │
│ └──────────┴─────────────►└────────┘
│  ▲     ▲   ▲
└──┘     │   │
parent  base └─build
```

Why the robot ID? Because it is unique, stored in `VRobot Int` and printed in REPL:
```
> build {move; move}
<r1> : robot
```
The problem is that typing `<r1>` in REPL does not return `robot` value. What shall we do?

xsebek:
Also, since #303 each robot gets a random name, like Fermi or Gagarin. :smile:  This funny detail is sadly not shown to users. 

To kick off the discussion, my idea would be to add two functions:
```c
robotNamed : string -> cmd robot // current 'find'
robot      : int    -> cmd robot // currently missing
```

Then the `VRobot 1` could be shown as `robot 1`.

xsebek:
@byorgey what do you think? I can still rename `find` in #339. :slightly_smiling_face:

xsebek:
Btw. when `view`ing the robot, we do show its name in the inventory panel:
```
┌──────────────────────────────────┐
│    trusting_euclid  (2, 0)  >    │
│                                  │
│ ────────Installed devices─────── │
```

So I would argue for just printing `robotNamed "trusting_euclid" : robot` for consistency and fun.
<sub>I almost want to have that power in a regular game, could `robot registry` be a thing? :smile: </sub>

If we wanted to make it really opaque, we could print `undefined : robot` or make `unnamed` constant so it reads better and provides some nice error message.

byorgey:
One problem is that names are not guaranteed to be unique.  So printing out `robotNamed "foo"` as the value of the robot doesn't seem like a good option to me, since it might not actually evaluate to the same robot.

byorgey:
[...]

PR #405, opened 2022-06-16
Add robotNumbered command for debugging

xsebek:
Sometimes I need a dirty and direct way to get the n-th robot in the world quickly.

The name of the command is longer by design so that it does not get confused with the proposed `child` command or custom user definitions.

- part of #343


byorgey (approved):
Looks good to me.

Issue #392, opened 2022-06-13
`halt` command to cancel a robot's running program


byorgey:
**Is your feature request related to a problem? Please describe.**
Robots can be reprogrammed or salvaged when they are done; but currently there is no way to halt a robot's running program in the middle.  This is annoying especially in the case of robots set up to do something forever in an infinite loop.  If you messed up the program, or you just decide that you want to do something else now, there's currently no way to stop a robot in an infinite loop.

**Describe the solution you'd like**
There should be some way to explicitly tell a robot to stop its program.  Probably you should (1) need an explicit reference to the robot and (2) be physically close the robot. (If you want to be able to tell robots to halt from far away, well, you can program that yourself once we have #94 .)  For example, maybe it can be as simple as a new primitive `halt : robot -> cmd ()` which tells the given robot to halt if it is on the same cell as (or maybe within Manhattan distance 1 of) the robot executing the `halt` instruction.

**Describe alternatives you've considered**
I suppose an alternative would be to just say "tough luck, if you wanted to be able to cancel a robot program then you should have programmed that in from the beginning", with some kind of `cancellable_forever` function that has some built-in polling to see whether it should quit or keep going. But that would be pretty annoying.  No one can plan perfectly for the future.

byorgey:
Idea: have `halt` provided by a `halting oracle` device.  Proposed description:

> A device to solve the halting problem.  When asked if a particular robot program will halt, it always answers YES.  And it is always correct. Or else.

TristanCacqueray:
I like it. Perhaps we could also introduce a `chaitin constant` entity to make the oracle recipe :)

PR #406, opened 2022-06-16
Add teleport command

xsebek:
- add command `teleport :: robot -> (int * int) -> cmd ()` requiring god capability

In challenges, it is useful to be able to check the state of some remote position.
This gets you there.


src/Swarm/Language/Syntax.hs
```diff
@@ -489,6 +492,7 @@ constInfo c = case c of
   Format -> functionLow 1
   Concat -> binaryOp "++" 6 R
   AppF -> binaryOp "$" 0 R
+  Teleport -> commandLow 2
```
byorgey, line 495: Shouldn't this be `commandLow 1`?  It only takes one argument.

byorgey (changes requested):
This all looks good to me.  However, I might suggest creating a new `teleport` capability and using that instead of `god`.  The reason is that I can imagine it would be fun to make a (very difficult to construct) teleportation device which actually provides this capability in the game.  We could of course use `god` for now and just add a new capability later, but using its own capability now will serve as a reminder that we want to eventually create a device for it.

src/Swarm/Language/Syntax.hs
xsebek: It should, I am surprised that it worked. :open_mouth:

xsebek:
OK, let's commit to adding a teleportation device :smile:

As a side note, I am hoping that `halt` (#392) would allow us to stop base and reprogram it to teleport, so in-world portals operated by system robots become possible :crossed_fingers:

byorgey:
Ah, I see, yes, having `teleport : robot -> int * int -> cmd ()` does seem strictly more useful/interesting.

byorgey (approved):

src/Swarm/Game/Step.hs
```diff
@@ -1346,7 +1346,7 @@ execConst c vs s k = do
       [VBool b] -> return $ Out (VBool (not b)) s k
       _ -> badConst
     Neg -> case vs of
-      [VInt n] -> return $ Out (VInt (-n)) s k
+      [VInt n] -> return $ Out (VInt (- n)) s k
```
xsebek, line 1349: OMG, I really have to find out which version of Fourmolu Restyled uses :expressionless:

byorgey (approved):

PR #407, opened 2022-06-16
Retire raise command

xsebek:
It is less powerful then error:
```
let raised: string -> cmd a = error in raised "ha"
```


byorgey (commented):
Hmm, interesting.  I hadn't realized that `error` and `raise` were basically the same.  I think maybe the reason we ended up with both is that I initially thought that `raise` might eventually get a more sophisticated type, so you could raise exceptions with some custom type, instead of having to use strings.

- I agree `raise`/`error` should not require any capability to use.  A capability will be needed for `try` of course.
- I agree it seems silly to keep both just because we might eventually generalize the type.

I want to bikeshed a little bit about the name if we are only going to have one.  To me at least, `error` seems to imply an unrecoverable error, whereas `raise` or `throw` imply an exception which can potentially be caught. So I would prefer for this to be called `raise` or `throw` instead of `error`. I originally went with `raise` because it was the terminology used by the textbook I was referencing (PFPL), and because we don't have anything named `catch` to go along with `throw`.  But I could be convinced to use `throw` instead.

xsebek:
Well, "error" does make more sense to me because it can be used in pure contexts like `if (error "bottom") {} {}`.

But maybe it's just what Haskell made me expect and beginners will not appreciate the difference. :smile:

byorgey:
Oh! I didn't notice `error` and `raise` had different types.  Let me think about this some more.

xsebek:
@byorgey from your list I kind of like `fail` or `bad`.

Is there some recipe for `spit`? I do not see why `spit` would throw an exception, it sounds like a Llama attack from Minecraft.

byorgey:
Hah, ok, not `spit` then. :laughing:   I think I like `fail`.  It goes well with `try`, and seems reasonable to me that it could be used in pure code.

byorgey (approved):
Looks good, thanks!

PR #412, opened 2022-06-16
put base in a 32x32 patch with all necessary starting resources

### LINKED ISSUE #385: Ensure copper is not too hard to get initially
A small amount of copper is necessary in the early game - it's needed to make motors, and hence drills, so you need to get some surface copper to bootstrap your way into drilling more.  Copper is also needed for strange loops, which is not strictly necessary but is very helpful early on to be able to use functions like `forever`.

However, in my experience, in many worlds copper ends up being rather far away, and in particular there might not even be any visible on the screen depending on terminal size.  It seems a bit too difficult to get, especially considering that new players won't even necessarily know that they need it.

I am not quite sure what to do about this.   I don't just want to cover the world in lots of copper.  A few ideas, but I'm open to more:

- Change the function that looks for a good starting spot for the base so that it finds somewhere near both trees and copper.
- Start out the base with a bit of copper ore in its inventory.

byorgey:
Fixes #385 .  Note, we will need to update `TUTORIAL.md` since this changes what world 0 looks like, but I wanted to get some feedback before launching into updating that.


src/Swarm/Game/WorldGen.hs
```diff
@@ -113,17 +113,47 @@ testWorld2FromArray arr seed co@(Coords (r, c))
...
+  patchCoords = [(r, c) | r <- [-16 .. 16], c <- [-16 .. 16]]
+  isGoodPatch f = all (`S.member` es) reqs
+   where
+    es = S.fromList . mapMaybe (snd . f . Coords) $ patchCoords
+
 -- | Offset the world so the base starts on empty spot next to tree and grass.
-findGoodOrigin :: WorldFun t Text -> WorldFun t Text
-findGoodOrigin f = \(Coords (r, c)) -> f (Coords (r + rOffset, c + cOffset))
+findTreeOffset :: WorldFun t Text -> WorldFun t Text
+findTreeOffset = findOffset 1 isGoodPlace
```
xsebek, line 145: Which direction will this search?

I am concerned if it is guaranteed that the good place will stay in the good patch.
If it spirals out, then it should be fine, but if it goes in line, then it could escape. :thinking:

src/Swarm/Game/WorldGen.hs
```diff
@@ -17,9 +16,10 @@ import Data.Enumeration
 import Data.Hash.Murmur
 import Data.Int (Int64)
 import Data.List (find)
-import Data.Maybe (fromMaybe)
+import Data.Maybe (fromMaybe, mapMaybe)
+import Data.Set qualified as S
 import Data.Text (Text)
-import qualified Data.Text as T
+import Data.Text qualified as T
```
xsebek, line 22: Hah, post-qualified import is nice :grinning: 

Maybe we could have it in default extensions (along with the rest of [GHC2021](https://github.com/ghc-proposals/ghc-proposals/blob/ghc2021/proposals/0000-ghc2021.rst#proposed-change)) and add [-Wprepositive-qualified-module](https://ghc.gitlab.haskell.org/ghc/doc/users_guide/using-warnings.html#ghc-flag--Wprepositive-qualified-module).

xsebek (approved):
Works for me :+1: I'm surprised that the original starting zone is nowhere in sight :open_mouth:

src/Swarm/Game/WorldGen.hs
xsebek: This is not very related to this PR so I made #419, let us discuss it there. :slightly_smiling_face:

src/Swarm/Game/WorldGen.hs
byorgey: Good question.  The order in which it searches comes from the way the `simple-enumeration` package works.  `Int` values are enumerated in the order `0, 1, -1, 2, -2, 3, -3 ...`, and then pairs of ints are enumerated using a balanced diagonal sort of scheme.  It's not exactly a spiral, but generally speaking closer locations are checked before farther ones:

```
λ> take 20 . enumerate $ int >< int
[(0,0),(0,1),(1,0),(0,-1),(1,1),(-1,0),(0,2),(1,-1),(-1,1),(2,0),(0,-2),(1,2),(-1,-1),(2,1),(-2,0),(0,3),(1,-2),(-1,2),(2,-1),(-2,1)]
```

byorgey:
Drat, messed up the screenshots...

PR #416, opened 2022-06-17
Add recipe for `mirror` and make it required for `self`

byorgey:
The `whoami` command used to be more interesting back when robot names were used to uniquely identify them.  It seemed fitting to make `mirror` required for `self`, which now plays the same role.  I also added a recipe for `mirror` which we lacked (`glass` + `silver`) and made a new type of `deep mine` where we can mine `gold`, `silver`, and `mithril`.


byorgey:
Oh, good idea, I like it!  We can have mithril drop very occasionally from a deep mine. =)  I will change it.

byorgey:
I didn't add lava but we could definitely add that later.

src/Swarm/Language/Capability.hs
```diff
@@ -304,7 +305,6 @@ constCaps =
     Appear -> [CAppear] -- paint?
     Whereami -> [CSenseloc] -- GPS?
     Random -> [CRandom] -- randomness device (with bitcoins)?
-    Whoami -> [CWhoami] -- mirror, needs a recipe
```
xsebek, line 307: I think we might have had a similar idea, so I caused you a merge conflict. Sorry about that :sweat_smile:

xsebek (approved):
Amazing, the entity descriptions are perfect. :mage_man:

byorgey:
I was thinking that in some kind of difficult mode we add later, drilling in a deep mine can very occasionally cause a balrog to spawn, who then wanders around setting everything on fire. 😀 🔥

src/Swarm/Language/Capability.hs
byorgey: Hah, no worries.

PR #426, opened 2022-06-18
Add info command

TristanCacqueray:
This change adds a new command to get information about
constant, similarly to the GHCi `:info` command.

### DIFF
--- a/editors/emacs/swarm-mode.el
+++ b/editors/emacs/swarm-mode.el
@@ -34,7 +34,7 @@
                 "install" "make" "build" "salvage" "reprogram"
                 "say" "log" "view" "appear" "create" "getx" "gety"
                 "blocked" "scan" "upload" "ishere" "whoami"
-                "random" "not"
+                "random" "not" "info"
                 "left" "right" "back" "forward" "north" "south" "east" "west" "down"
                 ))
              (x-types '("int" "string" "dir" "bool" "cmd"))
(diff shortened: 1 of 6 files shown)


TristanCacqueray:
I'm not sure this is the right approach, but I think it would be nice to query information through the REPL. So here is a proof of concept, and if that fits the project, then could we extend the `ConstInfo` structure to contains a short description of the commands?

byorgey:
Hi @TristanCacqueray , thanks for suggesting this.  I am not opposed to the idea in general, but I am a little skeptical.  I'm willing to be convinced though.

- Currently, if you want to know the type of a constant, you can just write the constant (much easier than typing `info` and putting the name of the constant in quotes) and the type will pop up in the upper right of the REPL window.
- Showing the required capabilities is not very useful because we otherwise nowhere refer to capabilities by name in the UI, only to devices.

So as currently implemented in this PR, the `info` command is not all that useful.  Showing a description and a list of devices that could potentially supply the needed capabilities (instead of the capabilities themselves) could be more useful.  However, it's annoying and distracting having to put the name of the constant in quotes, and then getting the response also in quotes with `: string` printed after it.

As an alternative, elsewhere we have discussed putting this kind of information in the help popup, and potentially having that be dynamically populated depending on what the player has "discovered" so far.

TristanCacqueray:
Oops, I thought I removed the required capability print, as it is not useful indeed. I agree `info` is not useful as it is now, but if it returns the needed devices and a bit of documentation, then it might be good to use in a repl session. And that same information could be presented in the help popup too.

xsebek:
I like the idea of something like the Pythons `help` function.

Maybe it would be more useful to make the signature `info : a -> cmd string` and based on what we get we do our best?
So for `VApp c []` show documentation for the constant `c`, its type signature and required capabilities.
I think we would need to add `(operator)` syntax if we wanted to view documentation for them this way though.

TristanCacqueray:
I think #458 is a better solution.

PR #429, opened 2022-06-18
Destroy robot when needed

### LINKED ISSUE #428: Destroying robots doesn't work
**Describe the bug**
Destroying robots --- either via the `selfdestruct` command, or by drowning --- no longer seems to have any effect.

**To Reproduce**
- Start swarm with `-s 0`
- `build {turn right; move; move; move}`

**Expected behavior**
The robot should drown and disappear, but it does not.

**Additional context**
I think this may have been caused by some incorrect refactoring in https://github.com/swarm-game/swarm/commit/6d2550bb0f4ca3843a9631eaaf90b2506a950508 .  It seems that `destroyIfNotBase` no longer actually sets the `selfDestruct` flag.

TristanCacqueray:
This change fixes #428


TristanCacqueray:
Well I don't know if this is correct, but that seems to fix the bug described in #428. @xsebek should know if that is a good solution.

byorgey:
Hmm, I don't think that's quite right, because setting `selfDestruct .= True` will destroy the current robot.  But I think in the context of the `teleport` command we might need to destroy a *different* robot, because `teleport` takes a robot as an argument and teleports it, and it is destroyed if it tries to teleport somewhere bad like on top of a mountain or in the middle of a lake.  I think that's what the special `.==` operator was supposed to be doing somehow.

byorgey:
In other words, we need to make sure that any potential fix both fixes the bug described in #428 and also makes the teleport command continue to work.  @xsebek , maybe we can/should add automated tests for these things?

TristanCacqueray:
Oops indeed, the first commit was not correct, I mixed `.==` and `.=`.

xsebek (approved):
Oh sorry about that, thanks for fixing it and also naming it better :+1:

TristanCacqueray:
Not sure why, but the test fails with `Timed out`. I tried adding `log` statements, but they don't seems to be collected. How can I debug the test scenario?

xsebek:
@TristanCacqueray run with `--cheat` and you will see it. Let me take a look...

data/scenarios/04Testing/428-drowning-destroy.yaml
```diff
@@ -0,0 +1,46 @@
...
+  try {
+    _ <- robotNamed "bot";
+    return false
+  } { return true }
+
+solution: |
+  log "start";
+  wait 5;
+robots:
+  - name: bot
```
xsebek, line 15: I would suggest making the map 2x2 and having both base and bot attempt to go into the water.
Then check that base still lives and the bot does not.

xsebek:
As a side note, I am definitely not attached to the `.==` operator, I just did not have a good name for what it does:

"Give me a setter and a value and I will use them to modify the right robot - either this one or the one in map."

Maybe there is even some standard name or function for it, but I did not find any.

TristanCacqueray:
Ha, pretty cool to run the test scenario using the game :-) Thanks, @xsebek, the scenario needed a base to get the id 0.

data/scenarios/04Testing/428-drowning-destroy.yaml
```diff
@@ -0,0 +1,60 @@
+name: Drowning results in destruction
+description: |
+  Destroying robots doesn't work.
+  https://github.com/swarm-game/swarm/issues/428
+win: |
+  def isAlive = \n. try {
+    _ <- robotNamed n;
+    log ("Is alive: " ++ n);
```
xsebek, line 8: I don't think those logs will be considered (even if they contained "Fatal error") because there is a new robot with id 0 created for this purpose every tick and it is not in the robot map. Can you see them anywhere?

But I guess they are useful when trying this manually.

TristanCacqueray: Indeed the winning log is not visible.

TristanCacqueray:
@byorgey if that's ok with you, we should merge this fix.

byorgey:
I trust @xsebek , if he has approved it that's good enough for me!

PR #411, opened 2022-06-16
Add a few tutorial challenges

xsebek:
Start designing the tutorial (via challenges) so that we can figure out which new features are necessary.

Issue #410, opened 2022-06-16
Known entities do not include installed devices


xsebek:
## To Reproduce

1. Open Creative game
2. See that base has an installed drill in the inventory panel
3. Type `give base "drill"`

```
give: What is a drill ?
```

## Expected behaviour

I would expect installed devices to count as known - the same as `drill  0`.

I expect the same is true for devices `install`ed to other robots or built in.

byorgey:
- Looks like we have an opportunity to refactor out some common code between `give` and `install`.
- A nice opportunity might be to notice whether we are in creative mode, and if so, add a hint to this error message (or the related one saying you don't have X) that you could make one with `create`.

PR #433, opened 2022-06-18
Add knows command

### LINKED ISSUE #409: Add `knows` command
## Problem
When we get to the tutorial testing `scan` and `upload`, we will need to check that the base has obtained the knowledge.

## Alternatives

Make existing command (`has` or `count`) return an optional value.

TristanCacqueray:
Fixes #409

### DIFF
--- a/src/Swarm/Game/Step.hs
+++ b/src/Swarm/Game/Step.hs
@@ -968,6 +968,14 @@ execConst c vs s k = do
 
         return $ Out res s k
       _ -> badConst
+    Knows -> case vs of
+      [VString name] -> do
+        inv <- use robotInventory
+        let knows = case E.lookupByName name inv of
+              [] -> False
+              _ -> True
+        return $ Out (VBool knows) s k
+      _ -> badConst
     Upload -> case vs of
       [VRobot otherID] -> do
         -- Make sure the other robot exists and is close
--- a/src/Swarm/Game/Step.hs
+++ b/src/Swarm/Game/Step.hs
@@ -971,7 +971,9 @@ execConst c vs s k = do
     Knows -> case vs of
       [VString name] -> do
         inv <- use robotInventory
-        let knows = case E.lookupByName name inv of
+        ins <- use installedDevices
+        let allKnown = inv `E.union` ins
+        let knows = case E.lookupByName name allKnown of
               [] -> False
               _ -> True
         return $ Out (VBool knows) s k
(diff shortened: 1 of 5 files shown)


src/Swarm/Game/Step.hs
```diff
@@ -968,6 +968,14 @@ execConst c vs s k = do
 
         return $ Out res s k
       _ -> badConst
+    Knows -> case vs of
+      [VString name] -> do
+        inv <- use robotInventory
```
byorgey, line 973: We should also check in `installedDevices`, since they are certainly known but there may not be any in the inventory.  (I am coming to regret separating them into two different inventories.)

byorgey (changes requested):
Looks good, other than my one comment.

src/Swarm/Game/Step.hs
TristanCacqueray: Do you think it would be complicated to move the installed devices to the entity inventory? For example, could we just change the counts map to include a boolean value for the isInstalled status?

byorgey: I don't think it would be complicated. But we would have to be careful keeping track of counts, and carefully scrutinize all current uses of the inventory to see whether they need to be updated. It would mean a big refactoring, touching a lot of code, without the type checker's help.  I don't know if it would really be worth it.

src/Swarm/Game/Step.hs
```diff
@@ -968,6 +968,14 @@ execConst c vs s k = do
...
+        let knows = case E.lookupByName name inv of
```
xsebek, line 974:
```suggestion
        ins <- use installedDevices
        let allKnown = inv `E.union` ins
        let knows = case E.lookupByName name allKnown of
```
The same thing is needed in `place`, `give` and `install`. #410 

My take is that it would be best to install via a lens that adds to inventory a zero so that an invariant is maintained.

But that should not hold back this PR, with this change it should be good to go. :+1:

xsebek (approved):

src/Swarm/Game/Step.hs
byorgey: > My take is that it would be best to install via a lens that adds to inventory a zero so that an invariant is maintained.

Ah, that's a good idea.

xsebek:
Hey @TristanCacqueray, I am going to merge this so I can rebase #411 on it.

byorgey (approved):

Issue #334, opened 2022-05-13
Prevent base from being able to move too early


byorgey:
See the discussion at #92 .  Right now, the base can install anything it wants on itself, including treads, which let it drive around.

xsebek:
In #373 I am about to add a new stronger motor for a `metal drill`, so maybe we could use it to power the base moving around?

Maybe we could add a "heavy" property to the base robot, which would be checked on moving.

A possible alternative would be to limit the base to one square with life support, which would lead to a different playthrough more centred on programming robots.

Generally, I would like to make this configurable so that we could experiment with scenarios and then choose whichever approach works best as default.

byorgey:
It doesn't make sense to me that a certain *square* would have "life support" that you could drive away from.  Wouldn't the life support systems be built into the base itself?  Also I simply don't like the idea of *preventing* the player from doing things.  Making them difficult, sure, but making them arbitrarily impossible doesn't seem like good design to me.

I really do like the idea of having to construct some special "tank treads" or "heavy treads" using a stronger motor, some iron plates, etc. in order to move robots with a certain "heavy" property.

As an aside, I also had the idea of putting some harvestable entities around the world called `LaTeX`, represented by a `$` character, that you can use to make `rubber`, which would also be required for making treads (regular or heavy). =)  This has nothing in particular to do with the base moving around but just thought I would mention it.

PR #434, opened 2022-06-19
recipes for treads + tank treads: LaTeX -> rubber + motors + iron

byorgey:
This doesn't yet prevent normal treads from being installed on the base (see #334).  That will be a separate PR.

### DIFF
--- a/data/entities.yaml
+++ b/data/entities.yaml
@@ -346,6 +346,24 @@
   - A pane of a brittle, clear substance, made from melting sand in a furnace.
   properties: [portable]
 
+- name: LaTeX
+  display:
+    attr: flower
+    char: '$'
+  description:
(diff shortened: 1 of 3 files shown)


xsebek (approved):
I like the LaTeX entity growing in forests :+1: 

Technically it might be good to check that it is in starting tile, but I see one nearby and the chances are in its favour, so it's not worth remaking every World 0 picture. I also like the current starting area :smile: 

image
<sup>The LaTeX can be seen int he forests west and east of the base</sup>

byorgey:
> Technically it might be good to check that it is in starting tile,

I did think about that, but I think it's not really necessary since making rubber will not be needed right away.  And in any case yes, it's unlikely to be too far away. In fact of all the worlds I've tried, world 0 seems to be an outlier in terms of the distance to the closest `$`.

PR #437, opened 2022-06-19
Handle mouse click event 

### LINKED ISSUE #203: Cursor mode in world panel
One very annoying aspect of the early game is the need to precisely measure distances to various objects so you can send robots to fetch them: without any kind of conditionals or repetition, the only thing you can do is specify an exact distance to travel.  Of course there are clever things you can do besides squinting at your screen and trying to carefully count, such as sending robots to measure distances for you etc. But I wonder if it would be nice to add a "cursor" mode to the world view, where you can move a cursor around the world and see the coordinates of the current location of the cursor (and maybe even e.g. paste the current coordinates into the REPL when you hit Enter).

TristanCacqueray:
This change adds support for mouse event:
- Change panel focus
- Display world location distance to the center

Fixes #203

### DIFF
--- a/src/Swarm/Game/State.hs
+++ b/src/Swarm/Game/State.hs
@@ -48,6 +48,7 @@ module Swarm.Game.State (
   recipesIn,
   scenarios,
   world,
+  worldCursor,
   viewCenterRule,
   viewCenter,
   needsRedraw,
@@ -195,6 +196,7 @@ data GameState = GameState
   , _recipesIn :: IntMap [Recipe Entity]
   , _scenarios :: ScenarioCollection
   , _world :: W.World Int Entity
+  , _worldCursor :: Maybe W.Coords
   , _viewCenterRule :: ViewCenterRule
   , _viewCenter :: V2 Int64
   , _needsRedraw :: Bool
@@ -285,6 +287,9 @@ scenarios :: Lens' GameState ScenarioCollection
 -- | The current state of the world (terrain and entities only; robots
 --   are stored in the 'robotMap').
 world :: Lens' GameState (W.World Int Entity)
+
+-- | The last clicked position on the world view.
+worldCursor :: Lens' GameState (Maybe W.Coords)
 
 -- | The current rule for determining the center of the world view.
 --   It updates also, viewCenter and focusedRobotName to keep
@@ -451,6 +456,7 @@ initGameState cmdlineSeed scenarioToLoad toRun = do
           , _recipesIn = inRecipeMap recipes
           , _scenarios = loadedScenarios
           , _world = W.emptyWorld 0
+          , _worldCursor = Nothing
           , _viewCenterRule = VCRobot 0
           , _viewCenter = V2 0 0
           , _needsRedraw = False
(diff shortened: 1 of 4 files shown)


TristanCacqueray:
As noted in the source comment, this need more works to adjust the distance when the base is not centered. But I think that is already pretty useful for the starting condition.

src/Swarm/Game/State.hs
```diff
@@ -195,6 +196,7 @@ data GameState = GameState
   , _recipesIn :: IntMap [Recipe Entity]
   , _scenarios :: ScenarioCollection
   , _world :: W.World Int Entity
+  , _worldCursor :: Maybe W.Coords
```
TristanCacqueray, line 199: I guess this should go in the UIState, but the TUI.Model module does not import World, so maybe it's ok to keep this here?

byorgey: I don't see any problem with adding a `World` import to `TUI.Model`.

src/Swarm/TUI/View.hs
```diff
@@ -205,29 +205,43 @@ drawGameUI s =
...
+                      . padLeftRight 1
+                      $ drawREPL s
+                  )
             ]
         ]
   ]
  where
+  addCursorPos = case s ^. gameState . worldCursor of
+    Just coord -> topLabels . rightLabel ?~ drawWorldCursorInfo (s ^. gameState) coord
+    Nothing -> id
```
byorgey, line 230: I think it would look nice to add an extra space on either side of the cursor info, e.g. by calling `padLeftRight 1` on the output of `drawWorldCursorInfo`.

byorgey (commented):
This is looking nice so far.  Is the goal to show the absolute coordinates of the thing clicked on, or to show the coordinates relative to the focused robot?  Either could make sense, but I think

- If the former, then it seems like it's working great.
- If the latter, then I think we need to indicate that somehow.  The word "at" (as in, `$  at  14 2`) to me indicates absolute location.  If we are trying to show relative location, maybe we could say something like `$ offset: 14 2` ?  I am not 100% sure what the best wording would be.

src/Swarm/TUI/View.hs
TristanCacqueray: Indeed, that's better :)

TristanCacqueray:
The goal is to help figure out how to reach a location, thus I meant to provide the relative coordinate from the focused robot, but perhaps that's too helpful, and the absolute coordinate works fine.

byorgey:
Yeah, I'm inclined to agree.  I think it's great as is.

src/Swarm/TUI/Controller.hs
```diff
@@ -204,6 +204,22 @@ handleMainEvent s = \case
...
+        mouseCoordsM <- mouseLocToWorldCoords (s ^. gameState) mouseLoc
+        continue (s & uiState . uiWorldCursor .~ mouseCoordsM)
+      REPLPanel ->
+        -- Do not clear the world cursor when going back to the REPL
+        continueWithoutRedraw s
+      _ -> continueWithoutRedraw (s & uiState . uiWorldCursor .~ Nothing)
+  MouseUp n _ _mouseLoc -> do
+    setFocus s $ case n of
+      -- Adapt click event origin to their right panel.
+      -- TODO: figure how to force that using `clickable`
```
byorgey, line 219: I'm not sure what this comment means, can you elaborate? 

It would be cool if clicking in the inventory list caused the clicked item to be highlighted, but we can put that off to another PR.

byorgey (approved):

src/Swarm/TUI/Controller.hs
TristanCacqueray: I was hoping this part would not be necessary when using 'clickable', but I couldn't make it work. I updated the comment to make it more clear.

Issue #85, opened 2021-09-26
Add reverse-i-search  to REPL


lsmor:
**Summary**
It could be usefull to have reverse-i-search in the REPL as in regular unix shell. 

**Describe the solution you'd like**
Pressing `Crtl+r` should prompt the user with the word to find and it should look into repl history returning the first command containing that word

lsmor:
@byorgey  I am trying to complete this, and it is quite of an achivement. I don't know if I'm missing something about the code.

So AFAIK, the repl is represented as just `Text`, whereas input and outputs are properly represented using a ADT. In order to give the repl a behaviour other than just typing comands we should create an ADT representing the "kind" of prompt you want to use. Example

```haskell
-- | The ADT representing REPL functionality.
--   Typing a command, search in history, multiline edition, etc ...
--   The Text field is the text prompted to the user
data REPLPrompt = PromptCmd Text | PromptSearch Text | PromptEdit Text

data UIState = UIState
  { ...
  , _uiReplPrompt :: REPLPrompt
  , _uiReplForm   :: Form REPLPrompt AppEvent Name
  }

```
The idea is that it you type something, the behaviour is defferent whether you are in a search prompt or in a comand prompt. With respec to visualization, the REPL would become kind of a `Form REPLPrompt AppEvent Name` and all functions dealing with `uiReplForm` lens, should adapt to the new type. 

I willing to make this work, but it seems like a big change and I don't know if there is a better way to do it. Let me know If you think of a different solution or you just prefer to keep the repl as it is

byorgey:
Hi @lsmor , yes, it does seem tricky.  It's a shame we can't just reuse something like `haskeline` which already implements all of this, but that's not really possible because of the way the REPL is embedded inside a larger UI, so we have to implement all of these features ourselves. (Down the road it might be a nice idea to spin out some of the REPL stuff info an independent package, e.g. `brick-repl`, but that's quite a ways off at this point.)

I am not sure I understand 100% what the `REPLPrompt` type represents.  But it definitely makes sense that we will need to introduce a sum type somewhere to keep track of what mode the REPL is in.  I don't think there's any simpler way to do it.

lsmor:
> Perhaps implementing `history-search-backward` might be easier to begin with, that is when the user press pgup, look in the history for a line starting with the current prompt.

Yes, actually something like that would be much easier. But I think, sonner or later we will need a sum type for the repl.

[...]

PR #439, opened 2022-06-19
Find in repl

lsmor:
This PR solves #85. It's still a little bit rough since it only shows the last entry in history with the given text.

The solution if found is making the prompt a sum type. This allow to handle reple events differently depending on which mode the repl is. I've found this more difficult to implement than expected. The main problems I've found are 

- `UIState` is a huge data type. Maybe putting al repl-related stuff within the same record could improve things a little bit
- I think we can define a more complex Lenses, like `promptUpdateL` which can handle many parts of the logic. I think this is better than having a huge `handleREPLEvent` function.

 All of this would imply a big refactor tough.


src/Swarm/TUI/Controller.hs
```diff
@@ -515,13 +515,44 @@ handleREPLEvent s (ControlKey 'c') =
...
 handleREPLEvent s (Key V.KEnter) =
   if not $ s ^. gameState . replWorking
-    then case processTerm' topTypeCtx topCapCtx entry of
-      Right mt -> do
-        let s' = advanceREPL s
-            s'' = maybe id startBaseProgram mt s'
-        continue s''
-      Left err ->
-        continue $ s & uiState . uiError ?~ err
+    then case entry of
```
TristanCacqueray, line 518: nit, perhaps the case expression could return the state transformation so that we can factor out the continue like this:

```suggestion
    then continue $ s & case entry of
```

src/Swarm/TUI/Model.hs
```diff
@@ -555,16 +636,21 @@ focusedEntity =
 initFocusRing :: FocusRing Name
 initFocusRing = focusRing [REPLPanel, InfoPanel, RobotPanel, WorldPanel]
 
--- | The default REPL prompt.
-replPrompt :: Text
-replPrompt = "> "
+{-
```
TristanCacqueray, line 639: nit: leftover debug comment?

src/Swarm/TUI/Model.hs
```diff
@@ -298,6 +307,61 @@ getCurrentItemText history = replItemText <$> Seq.lookup (history ^. replIndex)
...
+
+-- | Turn the repl prompt into a decorator for the form
+replPromptAsWidget :: REPLPrompt -> Widget Name
+replPromptAsWidget (CmdPrompt _) = txt "> "
+replPromptAsWidget (SearchPrompt t rh) =
+  case lastEntry t rh of
+    Nothing -> txt $ "[find command " <> t <> " nothing found] "
+    Just lastentry
+      | T.null t -> txt $ "[find command " <> t <> "] "
+      | otherwise -> txt $ "[find command " <> t <> ": \"" <> lastentry <> "\" found] "
```
byorgey, line 359: These prompts seem a bit too verbose.  I think we could get rid of the word `command` and also the word `found` at the end, and it would save some horizontal space and still make perfect sense.

byorgey (changes requested):
Great work, I'm excited to have this feature.  A few changes I'd like to see:

- `Esc` and/or `^G` (ideally both) should cancel the search and return to the regular prompt.  I could not figure out any way to cancel the search other than deleting what I had typed and then hitting `Enter`.  Since then, I figured out that `^R` actually cancels the search, but I would expect hitting `^R` to find the next matching item in the history, not cancel.
- Making the prompt less verbose, as explained in my other comment.
- After finding something with a search and hitting `Enter`, the text of that history entry is placed in the REPL prompt, but it is not checked.  In other words, if you select something from the history that has an error, it will appear in white letters, and will not turn red until you type something else or move the cursor.  Ideally, the prompt entry would be checked immediately.

Making `REPLPrompt` a sum type seems like exactly the right way to handle this.  The refactoring you talked about sounds great, and we should definitely be looking for ways to clean up and simplify the code, but I don't think we need to do that big refactoring before merging this.

src/Swarm/TUI/Controller.hs
lsmor: Absolutely! will commit along with other changes

lsmor:
> * `Esc` and/or `^G` (ideally both) should cancel the search and return to the regular prompt.  I could not figure out any way to cancel the search other than deleting what I had typed and then hitting `Enter`.  Since then, I figured out that `^R` actually cancels the search, but I would expect hitting `^R` to find the next matching item in the history, not cancel.

Yes, I think all can be done easily except for the "next matching item" thing. I'd tried that before, but maybe I can workaround it... Probably some nasty code would be required...

> * Making the prompt less verbose, as explained in my other comment.

Agree

> * After finding something with a search and hitting `Enter`, the text of that history entry is placed in the REPL prompt, but it is not checked.  In other words, if you select something from the history that has an error, it will appear in white letters, and will not turn red until you type something else or move the cursor.  Ideally, the prompt entry would be checked immediately.

Oh! didn't spot that out. I'll change it

byorgey:
> Yes, I think all can be done easily except for the "next matching item" thing. I'd tried that before, but maybe I can workaround it... Probably some nasty code would be required...

Oh, to be clear, I was not saying you have to do that for this PR.  I was just saying that's what I would expect ^R to do, if it does anything at all.  But for this PR let's not worry about it.

byorgey:
Also, I just went and made #450 which uses ^G to display a dialog with the scenario goal.  So forget what I said about ^G, just `Esc` can cancel the search. :smile:   Or else we could think of a different shortcut to display the goal.

lsmor:
> Oh, to be clear, I was not saying you have to do that for this PR. I was just saying that's what I would expect ^R to do, if it does anything at all. But for this PR let's not worry about it.

Well It turned out to be easier than expected

> Also, I just went and made #450 which uses ^G to display a dialog with the scenario goal. So forget what I said about ^G, just `Esc` can cancel the search. smile Or else we could think of a different shortcut to display the goal.

I read that 10 minutes later hahaha also added ^G as a command. I'm removing it

lsmor:
The new commits:
- implements the "keep searching as you press ^R" 
- reduce verbosity of search prompt
- Introduce Escape key to reset the prompt. Notice this works in Search prompt by canceling the search, and also in regular prompt by erasing it

src/Swarm/TUI/Model.hs
```diff
@@ -302,6 +312,59 @@ getCurrentItemText history = replItemText <$> Seq.lookup (history ^. replIndex)
...
+replPromptAsWidget (SearchPrompt t rh) =
+  case lastEntry t rh of
+    Nothing -> txt $ "[find " <> t <> " nothing found] "
+    Just lastentry
+      | T.null t -> txt $ "[find " <> t <> "] "
+      | otherwise -> txt $ "[find " <> t <> ": \"" <> lastentry <> "\"] "
+
+-- | Creates the repl form as a decorated form.
+mkReplForm :: REPLPrompt -> Form REPLPrompt AppEvent Name
+mkReplForm r = newForm [(replPromptAsWidget r <+>) @@= editTextField promptTextL REPLInput (Just 1)] r
```
TristanCacqueray, line 366: Perhaps we should not repeat the search input, it is already displayed in the `[find XX]`. For example, in bash:

- `$` <kbd>CTRL-R</kdb>
- `(reverse-i-search)``':` <kbd>swarm</kbd>
- `(reverse-i-search``swarm': ./dist-newstyle/build/...`

src/Swarm/TUI/Controller.hs
```diff
@@ -590,25 +612,49 @@ handleREPLEvent s (Key V.KUp) =
   continue $ s & adjReplHistIndex Older
 handleREPLEvent s (Key V.KDown) =
   continue $ s & adjReplHistIndex Newer
+handleREPLEvent s (ControlKey 'r') = continue $
```
TristanCacqueray, line 615: pressing <kbd>CTRL-R</kbd> should scroll back in the history, to access previous matching entry.

TristanCacqueray (changes requested):

src/Swarm/TUI/Model.hs
```diff
@@ -302,6 +312,59 @@ getCurrentItemText history = replItemText <$> Seq.lookup (history ^. replIndex)
...
+  = -- | Interpret the given text as a regular command
+    CmdPrompt Text
+  | -- | Interpret the given text as "search this text in history"
+    SearchPrompt Text REPLHistory
+
+-- | Get the last Entry in REPLHistory matching the given text
+lastEntry :: Text -> REPLHistory -> Maybe Text
+lastEntry t h = case filter (liftA2 (&&) matchesText isREPLEntry) $ toList $ h ^. replSeq of
+  [] -> Nothing
+  rh : _ -> Just $ replItemText rh
```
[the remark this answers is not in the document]
TristanCacqueray: Ha yes, that is it :-)

src/Swarm/TUI/Model.hs
```diff
@@ -302,6 +312,59 @@ getCurrentItemText history = replItemText <$> Seq.lookup (history ^. replIndex)
...
+  s (SearchPrompt _ h) t = SearchPrompt t h
+
+-- | Turn the repl prompt into a decorator for the form
+replPromptAsWidget :: REPLPrompt -> Widget Name
+replPromptAsWidget (CmdPrompt _) = txt "> "
+replPromptAsWidget (SearchPrompt t rh) =
+  case lastEntry t rh of
+    Nothing -> txt $ "[find " <> t <> " nothing found] "
+    Just lastentry
+      | T.null t -> txt $ "[find] "
```
byorgey, line 361:
```suggestion
      | T.null t -> txt "[find] "
```

This will fix the `hlint` suggestion currently causing CI to fail.

src/Swarm/TUI/Controller.hs
```diff
@@ -561,13 +565,36 @@ handleREPLEvent s (ControlKey 'c') =
...
+              (uiState . uiReplForm .~ mkReplForm (CmdPrompt ""))
+                . (uiState . uiReplType .~ Nothing)
+                . (uiState . uiError .~ Nothing)
+                $ s
+            | otherwise ->
+              validateREPLForm
+                . (uiState . uiReplForm .~ mkReplForm (CmdPrompt found))
+                . (uiState . uiReplType .~ Nothing)
+                . (uiState . uiError .~ Nothing)
+                $ s
```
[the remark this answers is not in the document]
lsmor: Indeed

src/Swarm/TUI/Controller.hs
```diff
@@ -590,25 +612,49 @@ handleREPLEvent s (Key V.KUp) =
...
 handleREPLEvent s (Key V.KDown) =
   continue $ s & adjReplHistIndex Newer
+handleREPLEvent s (ControlKey 'r') = continue $
+  case s ^. uiState . uiReplForm . to formState of
+    CmdPrompt uinput ->
+      let newform = mkReplForm $ SearchPrompt uinput (s ^. uiState . uiReplHistory)
+       in s & uiState . uiReplForm .~ newform
+    SearchPrompt ftext rh -> case toList $ rh ^. replSeq of
+      [] -> s
+      _ : rhis ->
```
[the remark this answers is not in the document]
lsmor: Actually, I think the problem is that if you type many times the same command, It will appear many times in the history and you al filtering out once at time. 

Using `Data.Seq` directly is by far a better solution... you still have the problem of repeated elements in history I'll work that around

src/Swarm/TUI/Controller.hs
```diff
@@ -590,25 +612,49 @@ handleREPLEvent s (Key V.KUp) =
...
+        let newform = mkReplForm $ SearchPrompt ftext (newREPLHistory rhis)
+         in s & uiState . uiReplForm .~ newform
+handleREPLEvent s EscapeKey =
+  continue $
+    (uiState . uiReplForm .~ mkReplForm (CmdPrompt ""))
```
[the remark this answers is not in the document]
lsmor: Oh that's a good point!

src/Swarm/TUI/Model.hs
lsmor: Yes, the thing happening is that the form display the text, and also does the decorator. Should be only one of both

byorgey (approved):
Thanks for the hard work getting this into shape!  I am excited to use this feature.  The code looks great now and everything seems to work as expected.

TristanCacqueray (approved):

PR #441, opened 2022-06-19
Add available recipes help

### LINKED ISSUE #436: Available recipes and capabilities view
Perhaps I'm doing it wrong, but so far, each time i acquire a new item, either through `grab`/`give` or `make`, then i have to go through the entire inventory to see if there is anything new i can make with it.

I would like to have a "Available" view list, ordered by my the most recently unlocked one, along with a popup message or a notification icon when the list changed. I think that would be more user friendly than having to remember all the different recipes, in particular if we are going to add many more of them.

A similar mechanism could be used for capabilities, when building an installable device, then a similar view can be used to display the new available commands.

A more traditional tech tree view could be useful too, but that does not help much when the available recipes are not usable right away.

**Additional context**
Here is a concrete example:
- Get `tree`, list contains `log` recipe.
- Make `log`, list contains `board` recipe.
- Make `board`, list contains `box` recipe.
- Get `bit 0` and `bit 1`, list contains `drill bit`

TristanCacqueray:
This change adds a new help panel with the list of recently acquired recipes.

Fixes #436

### DIFF
--- a/src/Swarm/Game/State.hs
+++ b/src/Swarm/Game/State.hs
@@ -39,6 +39,7 @@ module Swarm.Game.State (
   robotMap,
   robotsByLocation,
   activeRobots,
+  availableRecipes,
   gensym,
   randGen,
   adjList,
@@ -186,6 +187,7 @@ data GameState = GameState
     -- append to a list than to a Set.
     _waitingRobots :: Map Integer [RID]
   , _robotsByLocation :: Map (V2 Int64) IntSet
+  , _availableRecipes :: (Int, [Recipe Entity])
   , _gensym :: Int
   , _randGen :: StdGen
   , _adjList :: Array Int Text
@@ -247,6 +249,9 @@ robotMap :: Lens' GameState (IntMap Robot)
 --   Fortunately, there are relatively few ways for these things to
 --   happen.
 robotsByLocation :: Lens' GameState (Map (V2 Int64) IntSet)
+
+-- | The list of available recipes, with the position of the last known item.
+availableRecipes :: Lens' GameState (Int, [Recipe Entity])
 
 -- | The names of the robots that are currently not sleeping.
 activeRobots :: Getter GameState IntSet
@@ -440,6 +445,7 @@ initGameState cmdlineSeed scenarioToLoad toRun = do
           , _runStatus = Running
           , _robotMap = IM.empty
           , _robotsByLocation = M.empty
+          , _availableRecipes = (0, mempty)
           , _activeRobots = IS.empty
           , _waitingRobots = M.empty
           , _gensym = 0
--- a/src/Swarm/Game/Step.hs
+++ b/src/Swarm/Game/Step.hs
@@ -25,6 +25,7 @@ import Control.Monad (forM_, guard, msum, unless, when)
 import Data.Array (bounds, (!))
 import Data.Bool (bool)
 import Data.Either (rights)
+import Data.Foldable (traverse_)
 import qualified Data.Functor.Const as F
 import Data.Int (Int64)
 import qualified Data.IntMap as IM
@@ -367,6 +368,7 @@ stepCESK cesk = case cesk of
       Left exn -> return $ Up exn s k
       Right wo -> do
         robotInventory %= robotUpdateInventory rf
+        updateSelfAllAvailableRecipes
         world .= wo
         needsRedraw .= True
         stepCESK (Out v s k)
@@ -752,6 +754,7 @@ execConst c vs s k = do
         Just n -> fromMaybe e <$> uses entityMap (lookupEntityName n)
 
       robotInventory %= insert e'
+      updateSelfAvailableRecipes e'
 
       -- Return the name of the item grabbed.
       return $ Out (VString (e ^. entityName)) s k
@@ -813,6 +816,12 @@ execConst c vs s k = do
           robotMap . at otherID . _Just . robotInventory %= insert item
           robotInventory %= delete item
 
+          otherRobotm <- use $ robotMap . at otherID
+          case otherRobotm of
+            Just otherRobot ->
+              updateAvailableRecipes (otherRobot ^. robotInventory, otherRobot ^. installedDevices) item
+            Nothing -> pure ()
+
           -- Flag the UI for a redraw if we are currently showing either robot's inventory
           when (focusedID == myID || focusedID == otherID) flagRedraw
 
@@ -844,6 +853,7 @@ execConst c vs s k = do
               installedDevices %= insert item
               robotInventory %= delete item
 
+              updateSelfAllAvailableRecipes
               -- Flag the UI for a redraw if we are currently showing our inventory
               when (focusedID == myID) flagRedraw
           False -> do
@@ -891,6 +901,7 @@ execConst c vs s k = do
 
         -- take recipe inputs from inventory and add outputs after recipeTime
         robotInventory .= invTaken
+
         finishCookingRecipe recipe (WorldUpdate Right) (RobotUpdate changeInv)
       _ -> badConst
     Has -> case vs of
@@ -1075,6 +1086,8 @@ execConst c vs s k = do
             `isJustOrFail` ["I've never heard of", indefiniteQ name <> "."]
 
         robotInventory %= insert e
+        updateSelfAvailableRecipes e
+
         return $ Out VUnit s k
       _ -> badConst
     Ishere -> case vs of
@@ -1354,7 +1367,7 @@ execConst c vs s k = do
       [VBool b] -> return $ Out (VBool (not b)) s k
       _ -> badConst
     Neg -> case vs of
-      [VInt n] -> return $ Out (VInt (- n)) s k
+      [VInt n] -> return $ Out (VInt (-n)) s k
       _ -> badConst
     Eq -> returnEvalCmp
     Neq -> returnEvalCmp
@@ -1617,3 +1630,32 @@ safeExp :: Has (Throw Exn) sig m => Integer -> Integer -> m Integer
 safeExp a b
   | b < 0 = throwError $ CmdFailed Exp "Negative exponent"
   | otherwise = return $ a ^ b
+
+-- | Check if any new recipes is available from the given entity
+updateSelfAvailableRecipes :: (Has (State GameState) sig m, Has (State Robot) sig m) => Entity -> m ()
+updateSelfAvailableRecipes e = do
+  inventory <- use robotInventory
+  devices <- use installedDevices
+  updateAvailableRecipes (inventory, devices) e
+
+-- | Check if any new recipes is available from all the inventory
+updateSelfAllAvailableRecipes :: (Has (State GameState) sig m, Has (State Robot) sig m) => m ()
+updateSelfAllAvailableRecipes = do
+  inventory <- use robotInventory
+  devices <- use installedDevices
+  traverse_ (updateAvailableRecipes (inventory, devices) . snd) (filter (\(c, _) -> c > 0) $ E.elems inventory)
+
+updateAvailableRecipes :: Has (State GameState) sig m => (Inventory, Inventory) -> Entity -> m ()
+updateAvailableRecipes invs e = do
+  allInRecipes <- use recipesIn
+  let entityRecipes = recipesFor allInRecipes e
+  let canMake recipe = case missingIngredientsFor invs recipe of
+        [] -> True
+        _ -> False
+  let usableRecipes = filter canMake entityRecipes
+  (knownLength, knownRecipes) <- use availableRecipes
+  let newRecipes = filter (`notElem` knownRecipes) usableRecipes
+      newLength = case newRecipes of
+        [] -> knownLength
+        _ -> length newRecipes
+  availableRecipes .= (newLength, newRecipes <> knownRecipes)
--- a/src/Swarm/TUI/Controller.hs
+++ b/src/Swarm/TUI/Controller.hs
@@ -204,6 +204,9 @@ handleMainEvent s = \case
   ControlKey 'k'
     | s ^. uiState . uiCheatMode -> continue (s & gameState . creativeMode %~ not)
   FKey 1 -> toggleModal s HelpModal >>= continue
+  FKey 2 -> do
+    s' <- toggleModal s RecipesModal
+    continue (s' & gameState . availableRecipes %~ markAsKnown)
   -- dispatch any other events to the focused panel handler
   ev ->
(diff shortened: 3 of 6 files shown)


TristanCacqueray:
I think the implementation is not very efficient, and I'd like to investigate a better way to implement this feature, but that seems to be working so this is proposed as a draft for now.

Also it seems like the `toggleModal` helper does not automatically unpause the game as expected, e.g. pressing <kbd>F1</kbd> then <kbd>q</kbd> presently leaves the game in a pause state, and it's not clear how to unpause it (or why the REPL appears stuck after printing '...')

src/Swarm/Game/Step.hs
```diff
@@ -752,6 +753,7 @@ execConst c vs s k = do
         Just n -> fromMaybe e <$> uses entityMap (lookupEntityName n)
 
       robotInventory %= insert e'
+      updateDiscoveredEntities e'
```
byorgey, line 756: I worry that in the future when anyone writes code to adjust the `robotInventory` or installed devices, they will forget to call the proper `update` function, or not even realize that they are supposed to.  (And indeed, how can we be confident that you have found all the right places to insert calls to these functions?)  Could we somehow refactor into *e.g.* a single function that both updates the robot inventory and also updates the discovered entity set appropriately?

src/Swarm/Game/Step.hs
```diff
@@ -1617,3 +1623,43 @@ safeExp :: Has (Throw Exn) sig m => Integer -> Integer -> m Integer
...
   | otherwise = return $ a ^ b
+
+-- | Update the global list of discovered entities, and check for new recipes.
+updateDiscoveredEntities :: (Has (State GameState) sig m, Has (State Robot) sig m) => Entity -> m ()
+updateDiscoveredEntities e = do
+  allDiscovered <- use allDiscoveredEntities
+  if E.contains0plus e allDiscovered
+    then pure ()
+    else do
+      let newAllDiscovered = E.insertCount 9001 e allDiscovered
```
byorgey, line 1634: `9001`?  I assume the idea is that this is supposed to be sufficiently large so that any recipe will use at most 9001 of something?  If so, I definitely don't like this on principle, though I think I understand why you did it this way.  We want to say a recipe is "available" even if you don't actually yet have *enough* of some entity, just as long as you *know about* all the inputs to the recipe, right?

How about we add a new function to the `Recipes` module, like `missingIngredientsFor`, but which just returns a `Bool` saying whether you know about all the ingredients?

src/Swarm/Game/State.hs
```diff
@@ -248,6 +254,15 @@ robotMap :: Lens' GameState (IntMap Robot)
...
 robotsByLocation :: Lens' GameState (Map (V2 Int64) IntSet)
 
+-- | The list of entities that have been discovered.
+allDiscoveredEntities :: Lens' GameState Inventory
+
+-- | The list of devices that have been installed once.
+allInstalledDevices :: Lens' GameState Inventory
+
+-- | The list of available recipes, with the position of the last known item.
+availableRecipes :: Lens' GameState (Int, [Recipe Entity])
```
byorgey, line 264: I don't think I understand what the `Int` is for.

src/Swarm/Game/Step.hs
```diff
@@ -1617,3 +1623,43 @@ safeExp :: Has (Throw Exn) sig m => Integer -> Integer -> m Integer
...
+  let canMake recipe = case missingIngredientsFor invs recipe of
+        [] -> True
+        _ -> False
+  let usableRecipes = filter canMake entityRecipes
+  (knownLength, knownRecipes) <- use availableRecipes
+  let newRecipes = filter (`notElem` knownRecipes) usableRecipes
+      newLength = case newRecipes of
+        [] -> knownLength
+        _ -> length newRecipes
+  availableRecipes .= (newLength, newRecipes <> knownRecipes)
```
byorgey, line 1665: I see two potential sources of inefficiency here:
- Every time we discover a new entity, we iterate through the entire list of recipes to see which ones we can make.
    - I am really not too concerned about this. Trying to do something more clever seems like it would definitely be a case of premature optimization.  One doesn't discover new entities all that often, and I can't imagine the number of recipes will ever grow more than a few thousand.
- For each usable recipe, we do a linear search through the list of known recipes to see if we already know it.
    - This is a little more troubling, since it's quadratic in the number of recipes.  But again, it probably doesn't really make that much difference until we get up to thousands of recipes.

src/Swarm/Game/State.hs
byorgey: Oh, wait, I get it, it's to keep track of which available recipes are "read" and which are "unread", right?

src/Swarm/TUI/Controller.hs
```diff
@@ -189,7 +189,7 @@ handleMainEvent s = \case
     continue s
   Key V.KEsc
     | isJust (s ^. uiState . uiError) -> continue $ s & uiState . uiError .~ Nothing
-    | isJust (s ^. uiState . uiModal) -> continue $ s & uiState . uiModal .~ Nothing
+    | isJust (s ^. uiState . uiModal) -> continue $ s & maybeUnpause s . (uiState . uiModal .~ Nothing)
```
byorgey, line 192: We need to also call `resetLastFrameTime` here; see the comment below next to the other use of `maybeUnpause`.  Probably it would be better to refactor them both into the same function, so we can't call only one without the other.

byorgey (commented):
This looks really cool so far!  A few general comments:

- At the start of the game, when there are no recipes available, pressing F2 seems to open a tiny empty dialog box in the middle of the screen.
- I think recipes should become available as soon as we *know* about an entity, not just as soon as we actually have one.  When I started a new classic game and sent a robot to scan a tree, no recipes showed up.  But when I sent a robot to actually fetch a tree, then the recipe showed up.
- The recipes don't seem to be updated when I obtain an entity via the `create` command. (See, this is exactly what I meant about forgetting to call `updateAvailableRecipes` in all the right places.)
- I think the recipe dialog should put a blank line between recipes to make it easier to tell them apart.
- Recipes seem to be stretched across the entire available horizontal space which makes them difficult to read.
- What will happen when the available recipes dialog gets bigger than the screen?
- I'd like some way in the available recipes dialog to tell which recipes are new/unread.  Maybe a horizontal red line separating the new ones from the old ones?

src/Swarm/TUI/Controller.hs
TristanCacqueray: Good catch. I've proposed a fix separately in #452

src/Swarm/Game/Step.hs
TristanCacqueray: I wanted to keep the overhead of this nice-to-have feature as low as possible, and I figured we only need to update the lists on Make, Create and Grab, any other inventory update could go without updating the new lists (because the entity would already be known). On the other hand, having a common function might be useful for calculating achievements in the future, but unless the overhead is minimal, I think this should be only used by Make, Create and Grab.

src/Swarm/Game/Step.hs
TristanCacqueray: That's exactly it! I can add a new ingredients check that does not look for quantities.

src/Swarm/Game/State.hs
TristanCacqueray: Yes, and I'd like to add a separator so that the new recipes are separated from the known one.

src/Swarm/Game/Step.hs
TristanCacqueray: Oh yes, that is worrisome, that's the reason why I added two new Inventories to avoid doing this update unless necessary. I'll copy this comment for future reference, when it will become a bottleneck.

src/Swarm/Game/Step.hs
byorgey: But what about `give`?  i.e. suppose you program a robot to go grab a thing that you don't know about and then give it to you.  Shouldn't that be able to trigger the availability of a new recipe?  What about if you `salvage` that robot instead of having it `give` you something? There are many, many ways a robot's inventory can gain new entities (I'm quite sure I'm forgetting some).

TristanCacqueray:
Thank you for the detailed review @byorgey , really appreciate it!

>     * At the start of the game, when there are no recipes available, pressing F2 seems to open a tiny empty dialog box in the middle of the screen.
> 

The dialog is disabled when no recipes are available: 368f2c3.

>     * I think recipes should become available as soon as we _know_ about an entity, not just as soon as we actually have one.  When I started a new classic game and sent a robot to scan a tree, no recipes showed up.  But when I sent a robot to actually fetch a tree, then the recipe showed up.
> 

The list is updated on scan too, but I wonder if that is desirable: scan may be used in bulk, and I think it would be better to wait for an actual entity acquisition. 66574c4

>     * The recipes don't seem to be updated when I obtain an entity via the `create` command. (See, this is exactly what I meant about forgetting to call `updateAvailableRecipes` in all the right places.)
> 

That's odd, it should have worked... Perhaps you were missing an installed devices?

>     * I think the recipe dialog should put a blank line between recipes to make it easier to tell them apart.
> 

Done in e919ef0

>     * Recipes seem to be stretched across the entire available horizontal space which makes them difficult to read.
> 

Fixed in 44576ac

>     * What will happen when the available recipes dialog gets bigger than the screen?
> 

143d8e0 adds a viewport to enable scrolling.


>     * I'd like some way in the available recipes dialog to tell which recipes are new/unread.  Maybe a horizontal red line separating the new ones from the old ones?

Done in e919ef0, though I wasn't able to change the line color.

src/Swarm/Game/Step.hs
TristanCacqueray: I initially implemented something along those lines, but it sounded like a lot of work (to check the current or receiving robot inventory). Instead the current solution only keeps track of a single global list of known entities, so that the available recipes list is still updated, even if the current robot does not have all the ingredients. The reason being that if the player acquired an entity by any mean, then the player would know how to gather the ingredients for making the new recipe.

byorgey: Ah, I see now, that makes sense.

src/Swarm/Game/Step.hs
TristanCacqueray: Fixed in e660c2d

byorgey:
> wonder if that is desirable: scan may be used in bulk, and I think it would be better to wait for an actual entity acquisition

Hmm, I think I disagree. It's a bit difficult to scan all that many things at once in the beginning of the game, and knowing there are things I can make or do with the entities I just learned about is great incentive to go get some.

src/Swarm/Game/Step.hs
TristanCacqueray: Though I agree that the present solution is not great either. But perhaps if we miss a case, then that's not the end of the world, the player may still go through the inventory to check the full list.

TristanCacqueray:
Alright, that works for me. Then I think the PR is now ready for review again, I believe it's working as expected.

byorgey:
Thanks for addressing all those things, this is coming along nicely.

Unfortunately it still doesn't seem to be working quite right for me.  I got a tree and it told me about a recipe to make branch + log; I made branches and it told me about the recipe to make a branch predictor.  So far so good.... except that since I also obtained a `log` from that `make "branch"`, I would have expected it to also tell me about new available recipes to make a `logger` or `board`s out of a `log`.  I guess it is only looking at the explicit argument to `make`, instead of all the products of the recipe?  I later did `create "log"` and *then* the log recipes showed up.

I then switched into creative mode and did `create "furnace"`, which I expected to make several more recipes available (at that point I had `LaTeX` and `sand`, so it should have enabled recipes for `glass` and `rubber`), but I didn't get any.

I still find the recipes are hard to read because they are very wide.  Can we add some blank space on either side of the recipes (i.e. with `padLeftRight 8` or something like that)?

TristanCacqueray:
I see, the update only considered the given Make entity, not the recipe by-products. This should now be fixed with 155c64d (along with the padding).

For the recipes enabled by install-able entities such as the `furnace`, they only appears after the device is installed. I don't mind changing the logic, and it would be simpler (e.g. we could consider every discovered entities to be already installed), but perhaps it would be better to wait for the player to actually installed the device, presumably after reading its description.

byorgey:
> I see, the update only considered the given Make entity, not the recipe by-products. This should now be fixed with [155c64d](https://github.com/swarm-game/swarm/commit/155c64dab6f15b896f4faa0921c5f0d222952181) (along with the padding).

Works great now!  And looks much better with the padding.  I personally would also prefer a blank line between the top of the dialog box and the first recipe, but now we're definitely just getting into the territory of personal preferences and I'm happy to merge it either way. :smile: 

> For the recipes enabled by install-able entities such as the `furnace`, they only appears after the device is installed.

Ah, I see the confusion.  Furnaces don't actually need to be installed, nor do any entities listed as a required "catalyst" (*i.e.* centered at the top of a recipe), at least not as far as the recipe is concerned.  You certainly can `install` a `furnace`, and it works either way, but it's not necessary to make those recipes.  Things that need to be installed are things that provide one or more capabilities and enable commands.  Perhaps this is part of the confusion: we abuse the recipe system a little bit in the case of, say, drills.  The reason they need to be installed is that they provide the "drill" capability, which enables you to execute the `drill` command.  But the `drill` command also interprets recipes in a special way, and the drill is listed as a catalyst both to help the player understand what the drill will do and also to conveniently record the results of using different types of drills.

byorgey:
One other thing: I think it would be nice if there was a slightly more prominent way to alert the player that they have new recipes available.  A single asterisk showing up is pretty easy to miss if your eyes are looking somewhere else.  I like the idea (I think @xsebek mentioned this?) of highlighting `[F2] Recipes` in a different color.

TristanCacqueray:
That makes sense, today I learned furnaces do not have to be installed :-) In the last change I've also updated the styling as you suggested, though only the `[F2]` button is now highlighted. I don't mind extending to the whole `[F2] Recipes` if you prefer, but I find it less intrusive that way. Moreover, If the menu gets too crowded, we might want to remove the name altogether.

Thank you for helping me figure all that out!

byorgey (approved):
All looks great!  I like highlighting just the `[F2]`, that seems noticeable without being too in-your-face.  Thanks for your work on this great feature, I think it represents a huge jump in the playability/discoverability of the game.

As for the menu getting too crowded, I think @xsebek suggested moving all the `[Fn]` key shortcuts to the top left of the world, which is something we might consider.  But that can wait for another PR!

byorgey:
@TristanCacqueray feel free to merge this unless there are any last-minute tweaks you want to make!

PR #445, opened 2022-06-20
reset some UI state fields when beginning a new game

### LINKED ISSUE #444: Inventory position persists across games
When I start the `build` challenge after completing the previous one, `compass` is highlighted instead of `Goal`.  I think there is some state relating to the inventory list that is being persisted across games that really shouldn't be.

_Originally posted by @byorgey in https://github.com/swarm-game/swarm/issues/411#issuecomment-1159882208_

You can also see this by e.g. starting any challenge, highlighting the second item in the inventory list, then quitting to the menu and starting a creative mode game.  The second item in the inventory will be highlighted.

byorgey:
Reset `uiFocusRing`, `uiInventory`, `uiShowFPS`, and `uiShowZero`.

Fixes #444.

### DIFF
--- a/src/Swarm/TUI/Controller.hs
+++ b/src/Swarm/TUI/Controller.hs
@@ -169,6 +169,7 @@ resetUIState =
     . (uiInventory .~ Nothing)
     . (uiShowFPS .~ False)
     . (uiShowZero .~ True)
+    . (lgTicksPerSecond .~ initLgTicksPerSecond)
 
 mkScenarioList :: Bool -> ScenarioCollection -> BL.List Name ScenarioItem
 mkScenarioList cheat = flip (BL.list ScenarioList) 1 . V.fromList . filterTest . scenarioCollectionToList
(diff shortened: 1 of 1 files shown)


byorgey:
I think an argument could be made for leaving `uiShowFPS` and `uiShowZero` alone instead of resetting them.  Curious what others think.

TristanCacqueray (approved):
That's much better for the tutorials.

src/Swarm/TUI/Controller.hs
```diff
@@ -161,6 +162,14 @@ handleNewGameMenuEvent scenarioStack@(curMenu :| rest) s = \case
     continue $ s & uiState . uiMenu .~ NewGameMenu (menu' :| rest)
   _ -> continueWithoutRedraw s
 
+-- | Reset the UI state when beginning a new game.
+resetUIState :: UIState -> UIState
+resetUIState =
+  (uiFocusRing .~ initFocusRing)
+    . (uiInventory .~ Nothing)
+    . (uiShowFPS .~ False)
+    . (uiShowZero .~ True)
```
TristanCacqueray, line 171: Perhaps we could reset the tps to the default value, e.g. `lgTicksPerSecond .~ initLgTicksPerSecond`

byorgey: Oh, good catch!

PR #448, opened 2022-06-20
Handle mouse click event for robot inventory

### LINKED ISSUE #446: Select inventory items via click
After #437, clicking on a panel now shifts the focus to that panel.  It would be nice if clicking on the inventory panel in particular not only switched the focus to the panel, but actually caused the clicked inventory item to be highlighted.  This would obviously require a bit of work to translate a screen click location into an inventory list index.

See https://github.com/swarm-game/swarm/pull/437#discussion_r901157888 .

TristanCacqueray:
This change makes inventory click select the inventory item.

Fixes #446


src/Swarm/TUI/Controller.hs
```diff
@@ -224,12 +224,17 @@ handleMainEvent s = \case
         continueWithoutRedraw s
       _ -> continueWithoutRedraw (s & uiState . uiWorldCursor .~ Nothing)
   MouseUp n _ _mouseLoc -> do
-    setFocus s $ case n of
+    let s' =
+          s & case n of
+            InventoryListItem pos -> uiState . uiInventory %~ maybe Nothing (\(h, l) -> Just (h, BL.listMoveTo pos l))
```
TristanCacqueray, line 229: There is surely a better lense to update the second element of a maybe tuple, but I couldn't find it :)

byorgey (approved):
Works great!

src/Swarm/TUI/Controller.hs
byorgey: How's that?  `traverse` is to a lens as `fmap` is to applying a function.

TristanCacqueray: :exploding_head:

Issue #196, opened 2021-10-07
Guidance for Unicode and `--ascii` mode


byorgey:
As we were discussing in #186 it became clear that we need a more consistent approach to how we pick characters to use.  I think ideally, we should do the following:

1. Come up with a clear standard regarding which ranges of Unicode characters we will or won't use in the UI, ideally based on some data about sets of codepoints that are widely supported by many terminal fonts.
2. Come up with some terminal font recommendation(s) which do a good job displaying the characters we use.
3. Add an `--ascii` command-line parameter which forces the UI to use only ASCII characters.
    * This would require adding some kind of optional attributes for a `Display` specifying fallback ASCII characters for things that normally use non-ASCII.

byorgey:
Relevant, though perhaps somewhat outdated: https://www.johndcook.com/blog/2013/04/11/which-unicode-characters-can-you-depend-on/

byorgey:
I think a good approach would be to figure out (1) what are some of the most popular/widely used fixed-width fonts, that people might be likely to use in their terminal, (2) what subset of Unicode is supported by all of them.  That would give us both a clear standard in terms of what characters we can use, and also allow us to recommend any of those fonts.

byorgey:
Writing down a few recommendations for fixed-width fonts covering a large number of glyphs that I've found in various places around the web.

- DejaVu Sans Mono
- Fixedsys Excelsior
- Everson Mono
- Roboto Mono
- https://en.wikipedia.org/wiki/Monospace_%28Unicode%29

xsebek:
I would also consider terminal - it seems that Gnome terminal renders `░` as "light shade" and does not leave it to the font.

I am not 100% sure about that, but I first noticed the `░` in #362.

Either way, there are probably a few terminal settings that could improve the way characters are shown on screen. :slightly_smiling_face:

xsebek:
@byorgey could we add a note to the README.md and maybe CONTRIBUTING.md before the alpha release and close this Issue? 🙂 

AFAIK we stay clear of wide characters that cause trouble and only use basic Latin, Greek and Cyrillic characters.

So any popular monospace font should work. We could check what is the default on popular distributions like Ubuntu or what is used in Windows with WSL and on macOS. But I hope they all work with at least those characters in the old list you linked.

byorgey:
[...]

PR #449, opened 2022-06-20
switch from U+2591 Light Shade to U+2592 Medium Shade

byorgey:
Some terminals (*e.g.* `gnome-terminal`) use the character to decide
how dark or light the color should be, rather than showing the
character itself (like *e.g.* `rxvt-unicode`).  Light Shade looks OK
in terminals that display the actual character, but too dark in
terminals that use it as a brightness hint.  Medium Shade seems to
look good (at least, IMO) in both.

See https://github.com/swarm-game/swarm/issues/196#issuecomment-1147298159 .


xsebek (approved):
It does look a bit more colourful :+1: I will have to try `urxvt` someday. :slightly_smiling_face:

Issue #356, opened 2022-06-04
Improvements to challenge scenario UI


byorgey:
Some ideas:

- [x] Instead of having a special item denoting the goal, there should be a special `goal` field in the scenario description file; the goal can then be shown in a pop-up dialog box when the scenario is first loaded, with a key shortcut to pop up the goal dialog again any time the user wants. #450 
- [x] When completing a scenario, the "congratulations" dialog box should give the options to continue or return to the menu.

TristanCacqueray:
- [ ] keep track of the completed scenarios and adds a mark in the menu for the one that are already done.

byorgey:
@TristanCacqueray see #357 .

byorgey:
Closing now since the only thing not done is #357.

PR #450, opened 2022-06-20
add `goal` field to scenarios

byorgey:
If there is a goal, immediately pop up a dialog with the goal, then
allow showing it any time afterwards with ^G.

See #356.


byorgey:
Note this suffers from a bug where popping up the scenario goal dialog autopauses the game but then it is not unpaused when closing the dialog.  I think that bug already existed before, and is also being addressed in #441 .  Not quite sure what the best way forward is.

TristanCacqueray:
I've proposed the modal fix separately in #452

src/Swarm/Game/State.hs
```diff
@@ -162,11 +166,34 @@ data RunStatus
...
+markGoalRead :: GoalStatus -> GoalStatus
+markGoalRead (UnreadGoal g) = ReadGoal g
+markGoalRead gs = gs
+
 -- | The main record holding the state for the game itself (as
 --   distinct from the UI).  See the lenses below for access to its
 --   fields.
 data GameState = GameState
   { _creativeMode :: Bool
+  , _gameGoal :: GoalStatus
```
xsebek, line 196: Out of curiosity, why do we keep this information in GameState and not the UI state?

It does not seem to interact with the game logic at all.

My guess is it's because of the YAML --> Scenario --> GameState loading, but surely that should be possible to tweak.

byorgey: Yes, that's the simplest reason why. But yes, maybe the `playScenario` function should update the entire `AppState`, not just the `GameState`.  Come to think of it that might be a better place to put the code from #445 .  I will look into that.

byorgey:
Also, note that I intentionally did not update any of the scenarios in `data/scenario` to have a `goal`, to avoid causing merge conflicts with #411 .  If you want to test this out, you can just temporarily edit one of the scenarios to have a `goal` field containing a list (similar to `description`).

byorgey:
@xsebek , thanks for the feedback, I think this is now ready again.

src/Swarm/Game/State.hs
```diff
@@ -414,10 +433,63 @@ addRobot r = do
     %= M.insertWith IS.union (r ^. robotLocation) (IS.singleton rid)
   internalActiveRobots %= IS.insert rid
 
--- | Create an initial game state record for a particular game type,
---   first loading entities and recipies from disk.
-initGameState :: Maybe Seed -> Maybe String -> Maybe String -> ExceptT Text IO GameState
-initGameState cmdlineSeed scenarioToLoad toRun = do
+maxMessageQueueSize :: Int
```
byorgey, line 436: I didn't change anything from here down to `Initialization`, I just moved it to try to organize the code a bit better.

xsebek:
> Also, note that I intentionally did not update any of the scenarios in data/scenario to have a goal, to avoid causing merge conflicts with https://github.com/swarm-game/swarm/pull/411

Thanks, @byorgey, but I did not touch any *Challenges* so those can get the new goal right now :wink: 
Feel free to merge my PR #456 to this branch :innocent:

src/Swarm/TUI/Model.hs
```diff
@@ -678,6 +721,35 @@ populateInventoryList (Just r) = do
...
+  let skipMenu = isJust scenarioName || isJust toRun || isJust userSeed
+  gs <- initGameState
+  ui <- initUIState (not skipMenu) cheatMode
+  case skipMenu of
+    False -> return $ AppState gs ui
+    True -> do
+      scenario <- loadScenario (fromMaybe "00-classic" scenarioName) (gs ^. entityMap)
+      liftIO $ scenarioToAppState scenario userSeed toRun $ AppState gs ui
+
+-- XXX do we need to keep an old entity map around???
```
xsebek, line 734: I do not think so, we can always reload it if we wanted to say compare it before saving the game (or just save the whole map). Do you have some use cases in mind?

xsebek (approved):

src/Swarm/TUI/Model.hs
byorgey: I'm worried that when we load a scenario with custom entities, they get *permanently* added to the entity map...

byorgey: Well, that's unrelated to this PR though, and it would have already been an issue before.  I'll think about it later.

PR #458, opened 2022-06-21
Add available commands help

TristanCacqueray:
This change adds a new help panel with the list of recently acquired commands. The panels currently shows `command name :: command type signature`, for each newly available command.

Fixes https://github.com/swarm-game/swarm/issues/436

### DIFF
--- a/src/Swarm/Game/State.hs
+++ b/src/Swarm/Game/State.hs
@@ -186,10 +186,13 @@ data Notifications a = Notifications
   }
   deriving (Eq, Show)
 
-makeLenses ''Notifications
+instance Semigroup (Notifications a) where
+  Notifications count1 xs1 <> Notifications count2 xs2 = Notifications (count1 + count2) (xs1 <> xs2)
 
-emptyNotifications :: Notifications a
-emptyNotifications = Notifications 0 []
+instance Monoid (Notifications a) where
+  mempty = Notifications 0 []
+
+makeLenses ''Notifications
 
 ------------------------------------------------------------
 -- The main GameState record type
@@ -546,8 +549,8 @@ initGameState = do
       , _runStatus = Running
       , _robotMap = IM.empty
       , _robotsByLocation = M.empty
-      , _availableRecipes = emptyNotifications
-      , _availableCommands = emptyNotifications
+      , _availableRecipes = mempty
+      , _availableCommands = mempty
       , _allDiscoveredEntities = empty
       , _activeRobots = IS.empty
       , _waitingRobots = M.empty
--- a/src/Swarm/Game/State.hs
+++ b/src/Swarm/Game/State.hs
@@ -37,7 +37,7 @@ module Swarm.Game.State (
   robotsByLocation,
   activeRobots,
   availableRecipes,
-  availableRecipesNewCount,
+  availableCommands,
   allDiscoveredEntities,
   gensym,
   randGen,
@@ -56,6 +56,11 @@ module Swarm.Game.State (
   messageQueue,
   focusedRobotID,
   ticks,
+
+  -- ** Notifications
+  Notifications (..),
+  notificationsCount,
+  notificationsContent,
 
   -- ** GameState initialization
   initGameState,
@@ -82,7 +87,7 @@ module Swarm.Game.State (
 
 import Control.Applicative ((<|>))
 import Control.Arrow (Arrow ((&&&)))
-import Control.Lens hiding (use, uses, view, (%=), (+=), (.=), (<+=), (<<.=))
+import Control.Lens hiding (Const, use, uses, view, (%=), (+=), (.=), (<+=), (<<.=))
 import Control.Monad.Except
 import Data.Array (Array, listArray)
 import Data.Int (Int64)
@@ -118,7 +123,7 @@ import Swarm.Game.WorldGen (Seed)
 import qualified Swarm.Language.Context as Ctx
 import Swarm.Language.Pipeline (ProcessedTerm)
 import Swarm.Language.Pipeline.QQ (tmQ)
-import Swarm.Language.Syntax (Term (TString))
+import Swarm.Language.Syntax (Const, Term (TString))
 import Swarm.Language.Types
 import Swarm.Util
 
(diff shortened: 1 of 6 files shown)


src/Swarm/Game/Step.hs
```diff
@@ -1663,3 +1664,20 @@ updateAvailableRecipes invs e = do
...
   availableRecipesNewCount += newCount
+  updateAvailableCommands e
+
+updateAvailableCommands :: Has (State GameState) sig m => Entity -> m ()
+updateAvailableCommands e = do
+  let newCaps = S.fromList (e ^. entityCapabilities)
+      keepConsts s
+        | S.null s = False
+        -- This presently only works for single capability command.
+        -- If a command needs more than one capability, then we need combine all the known capabilities
```
byorgey, line 1675: I think we currently don't have any commands that require multiple capabilities, though in theory we could.

src/Swarm/TUI/View.hs
```diff
@@ -406,13 +432,14 @@ drawKeyMenu s =
   cheat = s ^. uiState . uiCheatMode
   goal = (s ^. uiState . uiGoal) /= NoGoal
 
-  availRecipes
-    | null (s ^. gameState . availableRecipes) = []
+  notificationKey :: Lens' GameState (Notifications a) -> Text -> Text -> [(KeyHighlight, Text, Text)]
```
TristanCacqueray, line 435: Not sure why, but without this annotation, ghc fails with `Couldn't match type ‘[a]’ with ‘Int’`. And this needs RankNType to work. Thus I'm not sure it's worth the trouble to use a generic Notifications type, and perhaps it would be better to just duplicate the code. On the other hand, we might want to use this logic for an events log menu too, e.g. when a monster appears, or when a robot gets destroyed.

src/Swarm/Game/Step.hs
TristanCacqueray: Until then, I guess it would be safer to change `constCap` to provides a `Maybe Capability`

TristanCacqueray:
I think this is starting to work as expected. Though I may need to change the widget to a Table if we want to add more information, such as a help text or the required entity.

src/Swarm/Game/Step.hs
byorgey: Sure, that's a good idea.  I can't think of any compelling reason we really *need* commands that require multiple capabilities.  In fact, if we ever wanted a command that required capabilities A and B, we could just make a new capability C, and the device that provides capability C could be constructed out of the devices that provide capabilities A and B.  So I would be totally fine with enshrining that restriction in the types.

byorgey:
This is looking good, but what about the commands the base can already execute from the beginning because of the devices it has installed?  It would be nice if those were already populated in the Commands modal from the start.

src/Swarm/Game/Step.hs
```diff
@@ -1675,8 +1678,20 @@ updateAvailableRecipes invs e = do
...
   let entityRecipes = recipesFor allInRecipes e
       usableRecipes = filter (knowsIngredientsFor invs) entityRecipes
-  knownRecipes <- use availableRecipes
+  Notifications prevCount knownRecipes <- use availableRecipes
   let newRecipes = filter (`notElem` knownRecipes) usableRecipes
       newCount = length newRecipes
-  availableRecipes .= newRecipes <> knownRecipes
-  availableRecipesNewCount += newCount
+  availableRecipes .= Notifications (newCount + prevCount) (newRecipes <> knownRecipes)
+  updateAvailableCommands e
```
byorgey, line 1685: This looks to me like `Notifications` really wants to be a `Monoid`. 

:thought_balloon:   `availableRecipes <>= Notifications newCount newRecipes`

src/Swarm/TUI/View.hs
byorgey: I'm not sure I understand what's going on here in particular, but I do like the generic `Notifications` type.  It does feel like something we might reuse for other purposes later.

byorgey: Anything that takes a lens as an argument is going to require `RankNTypes`, that's not a problem.  The fact that there's a type failure without the annotation makes sense to me (even though I would be hard-pressed to explain the error itself), since GHC will never infer a higher-rank type.

src/Swarm/Game/State.hs
```diff
@@ -626,6 +629,12 @@ scenarioToGameState scenario userSeed toRun g = do
           False -> id
           True -> const (fromList devices)
 
+  initialCaps = mconcat $ map (^. robotCapabilities) robotList
```
TristanCacqueray, line 632: Perhaps, when creativeMode is on, the initial list should be empty, so that there is still a sense of discovery?

byorgey (approved):
Looks great!

byorgey:
I think this also raises the issue of what we want to do with the `F1` Help modal.  Some of the information there is redundant now.  I don't think that should be addressed in this PR, just mentioning it while I'm thinking of it.

src/Swarm/Game/State.hs
byorgey: Hmm, no, when I'm playing creative mode I just want the game to stay out of my way and let me do whatever the heck I want (which includes showing me documentation for whatever the heck I want).

byorgey:
I think there is also more info we can imagine adding to the commands dialog but that also can wait for other PRs.

src/Swarm/Game/Step.hs
TristanCacqueray: That's tempting, but I think we need another operator to flips the mappend arguments so that the newRecipes are added to the front.

src/Swarm/Game/State.hs
TristanCacqueray: Makes sense.

src/Swarm/Game/Step.hs
TristanCacqueray: Otherwise we can reverse the list when doing the rendering.

byorgey: Oh, right.  I mean, it would still work with something like `deriving Monoid via (Sum Int, Dual [a])` but it's probably not worth it, the current code is fine.  I just really like monoids. 😁

TristanCacqueray: In bc73d6f I used `%= mappend` which is a good alternative. I agree, monoids looks much better, thanks!

src/Swarm/TUI/View.hs
TristanCacqueray: I see, thank you so much for the explanations.

Issue #387, opened 2022-06-12
Add a simpler way to make default entites known


xsebek:
There is a hacky way to make some entity known with a `known` property, but it needs an extra entity:
```yaml
  - name: knownwater
    display:
      attr: water
      char: ' '
    description:
    - An infinite ocean of water.
    properties: [known, portable, growable, liquid]
    growth: [0,0]
```

This results in a lot of copy-pasting and would eventually be pain to keep up to date.

### A simple field

Add a `known` field to scenario taking a list of entities.

### Hiding 0 (or less) entities

Just add the entity to inventory with a zero count and hide it.

Problem is, this is useful for letting the player know what to look for ("find lambda, you currently have 0") and adding other entites would be confusing.

We could hack it with -1 number of entities and always hide <0, but that sounds awful.

### DSL

Allow modifying default entities based on parameters.

```yaml
modify:
  - op: add
    properties: [known]
    entities: [water, wavywater]
```

Possible other use cases:
- add some description to entities in that scenario
- add unwalkable property and change attr to fire to make the floor lava

byorgey:
I agree adding 0 of something to the inventory is not a great way to solve the problem.  I think I definitely prefer the first solution (adding a `known` field with a list of entities) since it seems simple and easy to implement.

A DSL to modify default entities sounds cool and useful, but also like a lot of work. =)  I propose that we (1) solve this issue in a simple way, (2) file a separate issue with the idea for a DSL to modify default entities (& recipes?), (3) maybe implement that later if we find some other concrete use cases.

PR #461, opened 2022-06-22
Return the obtained entity in grab

xsebek:
When grabbing entities that have `yield` property, I want to use what I obtained.
But `t <- grab` would give me "wavy water" instead of "water".

I noticed this because of our hack with known entities, which also use yield (#387).


byorgey (approved):
Nice catch!

PR #464, opened 2022-06-22
Fix reprogram requiring wrong inventory

xsebek:
- fix the mistake introduces in #373


xsebek:
I was tearing my hair over this in #411 - the secret robot could not reprogram the target to log a message because it was looking for the logger device in secret inventory. :sweat_smile:

byorgey (approved):

PR #472, opened 2022-06-23
Tweaks to mines

byorgey:
- Give all mines priority 11 so they display on top of robots
- Make drilled mines known by adding them to the inventory with count 0

### DIFF
--- a/src/Swarm/Game/Step.hs
+++ b/src/Swarm/Game/Step.hs
@@ -947,7 +947,9 @@ execConst c vs s k = do
             `isJustOrFail` ["You don't have the ingredients to drill", indefinite (nextE ^. entityName) <> "."]
 
         let (out, down) = L.partition ((`hasProperty` Portable) . snd) outs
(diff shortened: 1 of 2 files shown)


src/Swarm/Game/Step.hs
```diff
@@ -947,7 +947,9 @@ execConst c vs s k = do
             `isJustOrFail` ["You don't have the ingredients to drill", indefinite (nextE ^. entityName) <> "."]
 
         let (out, down) = L.partition ((`hasProperty` Portable) . snd) outs
-            changeInv inv' = L.foldl' (flip $ uncurry insertCount) inv' out
+            changeInv =
+              flip (L.foldl' (flip $ uncurry insertCount)) out
+                . flip (L.foldl' (flip $ insertCount 0)) (map snd down)
```
xsebek, line 952: I can not read those flips and folds, but I trust it works :smile:

xsebek (approved):
Thanks, no idea why I did not add the priority when I created the first mine. :+1:

Issue #344, opened 2022-05-28
Create command line tools to automate documentation


xsebek:
It would be nice to have either a separate executable or a new parameter to generate documentation.

Some ideas:

- [x] We have nice cheatsheets on the Wiki that are hard to keep up to date. Generate those markdown tables from code.
  - [x] **Commands** (name, type, capability, description) #531
  - [x] **Capabilities** (name, install, commands, description)
  - [x] **Entities** (character, name, capabilities, properties, portable)
  - [x] **Recipes** (in, out, required, time)
- [x] Recipe dependencies graph #476
- [x] Editor keywords  #526

xsebek:
Also, there are the identifier lists in `editors/` that I did not update in a while :sweat_smile: 

- VS Code: [swarm.tmLanguage.json](https://github.com/swarm-game/swarm/blob/a6715e4bbde41a1d82ece7e77d3a17fe06a650ad/editors/vscode/syntaxes/swarm.tmLanguage.json#L59)
  ```json
  {
  "name": "keyword.operator",
  "match": "(<|>|==|<=|>=|!=|=|;|<-|-|\\+|\\*|\\^|\\$|/(?![/|*]))"
  },
  {
  "name": "keyword.other",
  "match": "\\b(?i)(noop|wait|selfdestruct|move|turn|grab|place|give|install|make|has|count|drill|build|salvage|reprogram|say|log|view|appear|create|whereami|blocked|scan|upload|ishere|whoami|random|run|if|inl|inr|case|fst|snd|force|return|try|raise|not)\\b"
  }
  ```
- Emacs: [swarm-mode.el](https://github.com/swarm-game/swarm/blob/a6715e4bbde41a1d82ece7e77d3a17fe06a650ad/editors/emacs/swarm-mode.el#L39)
  ```lisp
  (x-builtins '("if" "run" "return" "try" "raise" "force" "fst" "snd"))
  (x-commands
  '("noop" "wait" "selfdestruct" "move" "turn" "grab" "place" "give"
    "install" "make" "build" "salvage" "reprogram"
    "say" "log" "view" "appear" "create" "getx" "gety"
    "blocked" "scan" "upload" "ishere" "whoami"
    "random" "not"
    "left" "right" "back" "forward" "north" "south" "east" "west" "down"
    ))
  ```
  There is even a nice note: :spiral_notepad: 
  ```lisp
  ;; We should figure out how to autogenerate these, so we don't have
[...]

PR #476, opened 2022-06-24
Add generator for recipe dependencies

xsebek:
- add `generate` subparser to the executable
- create a generator for a Graphviz entity dependencies graph
- part of #344 

You can interactively test this with:
```bash
$ cabal run swarm:swarm -- generate recipes \
 | sed -n '/^digraph/,$p' > docs/recipes.dot
$ xdot docs/recipes.dot
```

### DIFF
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -31,7 +31,7 @@ cliParser =
   subparser
     ( command "format" (info (format <**> helper) (progDesc "Format a file"))
         <> command "lsp" (info (pure LSP) (progDesc "Start the LSP"))
-        <> command "generate" (info (docgen <**> helper) (progDesc "Generate docs"))
+        <> command "generate" (info (DocGen <$> docgen <**> helper) (progDesc "Generate docs"))
     )
     <|> Run <$> seed <*> scenario <*> run <*> cheat
  where
@@ -39,8 +39,10 @@ cliParser =
   format =
     (Format Stdin <$ switch (long "stdin" <> help "Read code from stdin"))
       <|> (Format . File <$> strArgument (metavar "FILE"))
-  docgen :: Parser CLI
-  docgen = DocGen RecipeGraph <$ switch (long "recipe-graph" <> help "Output graphviz dotfile of entity dependencies based on recipes")
+  docgen :: Parser GenerateDocs
+  docgen =
+    subparser
+      (command "recipes" (info (pure RecipeGraph) $ progDesc "Output graphviz dotfile of entity dependencies based on recipes"))
   seed :: Parser (Maybe Int)
   seed = optional $ option auto (long "seed" <> short 's' <> metavar "INT" <> help "Seed to use for world generation")
   scenario :: Parser (Maybe String)
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -14,6 +14,7 @@ import Swarm.App (appMain)
 import Swarm.Language.LSP (lspMain)
 import Swarm.Language.Pipeline (processTerm)
 import System.Exit
+import Swarm.DocGen (GenerateDocs (..), generateDocs)
 
 data CLI
   = Run
@@ -22,6 +23,7 @@ data CLI
       (Maybe FilePath) -- file to run
       Bool -- cheat mode
   | Format Input
+  | DocGen GenerateDocs
   | LSP
 
 cliParser :: Parser CLI
@@ -29,6 +31,7 @@ cliParser =
   subparser
     ( command "format" (info (format <**> helper) (progDesc "Format a file"))
         <> command "lsp" (info (pure LSP) (progDesc "Start the LSP"))
+        <> command "generate" (info (docgen <**> helper) (progDesc "Generate docs"))
     )
     <|> Run <$> seed <*> scenario <*> run <*> cheat
  where
@@ -36,6 +39,8 @@ cliParser =
   format =
     (Format Stdin <$ switch (long "stdin" <> help "Read code from stdin"))
       <|> (Format . File <$> strArgument (metavar "FILE"))
+  docgen :: Parser CLI
+  docgen = DocGen RecipeGraph <$ switch (long "recipe-graph" <> help "Output graphviz dotfile of entity dependencies based on recipes")
   seed :: Parser (Maybe Int)
   seed = optional $ option auto (long "seed" <> short 's' <> metavar "INT" <> help "Seed to use for world generation")
   scenario :: Parser (Maybe String)
@@ -87,4 +92,5 @@ main = do
   case cli of
     Run seed scenario toRun cheat -> appMain seed scenario toRun cheat
     Format fo -> formatFile fo
+    DocGen g -> generateDocs g
     LSP -> lspMain
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -11,10 +11,10 @@ import qualified Data.Text.IO as Text
 import GitHash
 import Options.Applicative
 import Swarm.App (appMain)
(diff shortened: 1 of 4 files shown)


xsebek:
This is how the graph looks:
Screenshot from 2022-06-24 15-24-21
<sup>The "test world 2 entites" are in green, base inventory in grey and each level of dependency is in yellow.</sup>

app/Main.hs
```diff
@@ -22,20 +23,24 @@ data CLI
...
   | Format Input
+  | DocGen GenerateDocs
   | LSP
 
 cliParser :: Parser CLI
 cliParser =
   subparser
     ( command "format" (info (format <**> helper) (progDesc "Format a file"))
         <> command "lsp" (info (pure LSP) (progDesc "Start the LSP"))
+        <> command "generate" (info (docgen <**> helper) (progDesc "Generate docs"))
```
xsebek, line 34: I wonder if it is possible to have a subparser in a subparser, kind of like `docker image ps -a`, except in our case it would be `swarm generate recipes --graph`.

xsebek: After inspecting the types a bit, I found it. :grinning:

TristanCacqueray:
That's interesting, though the output is a bit crowded. Would it be possible to use this layout instead: https://graphviz.org/Gallery/directed/ninja.html ?

src/Swarm/DocGen.hs
```diff
@@ -0,0 +1,247 @@
...
+-- These are the exported functions used by the executable.
+--
+-- ----------------------------------------------------------------------------
+
+data GenerateDocs where
+  -- | Entity dependencies by recipes.
+  RecipeGraph :: GenerateDocs
+  deriving (Eq, Show)
+
+generateDocs :: GenerateDocs -> IO ()
```
TristanCacqueray, line 50: Perhaps we only need one command to generate the documentation files directly. I guess we could also render a markdown reference file, and it might be easier to just add `swarm generate-doc` in the ci, and fail if the output changed.
For the final dot render, I think it is not deterministic, so updating the image could be a manual step, though we could commit the dot source file.

app/Main.hs
```diff
@@ -22,20 +23,26 @@ data CLI
...
   | Format Input
+  | DocGen GenerateDocs
   | LSP
 
 cliParser :: Parser CLI
 cliParser =
   subparser
     ( command "format" (info (format <**> helper) (progDesc "Format a file"))
         <> command "lsp" (info (pure LSP) (progDesc "Start the LSP"))
+        <> command "generate" (info (DocGen <$> docgen <**> helper) (progDesc "Generate docs"))
```
TristanCacqueray, line 34: We might want to add the `hidden` mod.

xsebek:
@TristanCacqueray changing the `rankdir` to `LR` like in the ninja example does not improve things that much:
image

I could add an extra node per each recipe and see if that would clear up things or become too big mess. :thinking: 

But for my usecase I can highlight the edges and nodes in `xdot` see where they lead. :shrug:

app/Main.hs
xsebek: Do you mean for the whole subparser? All of the commands show just as COMMAND, so it does not clutter much:
```
Swarm game - pre-alpha version (generate-docs@58cd7c00c3)

Usage: swarm [COMMAND | [-s|--seed INT] [-c|--scenario FILE] [-r|--run FILE] 
               [-x|--cheat]]

  To play the game simply run without any command.

Available options:
  -s,--seed INT            Seed to use for world generation
  -c,--scenario FILE       Name of a scenario to load
  -r,--run FILE            Run the commands in a file at startup
  -x,--cheat               Enable cheat mode
  -h,--help                Show this help text

Available commands:
  format                   Format a file
  lsp                      Start the LSP
  generate                 Generate docs
```

src/Swarm/DocGen.hs
xsebek: We could generate all the documentation and put it in appropriate files.
I can imagine a `--dir` option to `generate` that would create the files in the chosen directory.

I like the default `stdout` output for debugging and it also does not require the user to be in the swarm repo.

---

Eventually, I would like to create the render in CI and add it to GitHub Webpage.

The editor updates suggested in #344 seem more useful to track in this repository to me.

But we can set up the CI to push to GitHub Wiki or GitHub Webpage and track the generated stuff there.

TristanCacqueray:
Adding `ranksep="2"` seems to help unclutter. What is adding all the `pos` attributes to the dot source, could we let graphviz do the layout instead?

app/Main.hs
xsebek: But developers will need the `--help` to show something and I do not think that *one line* is worth adding a `--help-all` or conditionally hiding it based on for example Git branch or build flag.

xsebek:
@TristanCacqueray thanks for the `ranksep` tip, that seems to help. :+1:

I am not sure what you mean by `pos` in dot source, I do not see any there.
The layout is forced to be somewhat linear by invisible nodes and edges, as otherwise, it is a bit messy:

image

app/Main.hs
TristanCacqueray: I remember having the same discussion for the `lsp` sub command, where I suggested a more minimalist usage where only the relevant option are presented to the user. Though I don't mind the other way, and let's not bother adding an extra `--help-all` :)

TristanCacqueray:
Oh I see, I was looking at the output of `generate recipes | dot`. 

What about the many edges between rocks and drills, could we remove the duplicates?

xsebek:
@TristanCacqueray here is the funny thing - there are no duplicates, I already removed them.

What you see are the many versions of recipes that yield a different number of entities noted on each end of the arrow.

But for a dependency view that is a huge overkill, so I will remove it:

graphviz

app/Main.hs
xsebek: As long as it fits on a small terminal screen I think it's fine.

Besides, players can now use the menu screen. :smile:

xsebek:
One thing that we could do to simplify the graph significantly would be to note that some entities are upgrades over others.

Then we could add one edge from the original to upgrade and remove all that they share from the upgraded one.

This would hopefully allow Graphviz to place e.g. metal drill more to the right where it belongs.

I imagine this would be done in the `entites.yaml` file via a new `upgrades:` field. But maybe that should be discussed in a separate issue since doing it right could be complicated. :sweat_smile:

src/Swarm/Game/WorldGen.hs
```diff
@@ -77,7 +99,7 @@ testWorld2 baseSeed (Coords ix@(r, c)) =
     | h `mod` 10 == 0 = (GrassT, Just (T.concat ["bit (", from (show ((r + c) `mod` 2)), ")"]))
     | otherwise = (GrassT, Nothing)
   genBiome Big Soft Artificial
-    | h `mod` 5000 == 0 = (DirtT, Just "linux")
+    | h `mod` 5000 == 0 = (DirtT, Just "Linux")
```
byorgey, line 102: Hah, I had been wondering why I have never, ever seen a `Linux`.

byorgey (approved):
Looks good to me!  I didn't realize we were already at 13 levels. :astonished:

src/Swarm/Game/WorldGen.hs
xsebek: Weird that it did not throw an exception. :confused:

xsebek:
Thanks again @TristanCacqueray for your tips and careful scrutiny! :+1: 

I will try to merge this now and we can improve it with #480 and maybe some automatic CI export in the future. :slightly_smiling_face:

PR #482, opened 2022-06-24
Make ctrl-q works inside modal

TristanCacqueray:
This change enables using <kbd>CTRL</kbd><kbd>q</kbd> to close a visible modal.


byorgey (approved):
To be clear, it seems what this does is that hitting Ctrl-Q while a modal is open will just close the modal.  That seems like a reasonable behavior to me, and assuming that's what you intended this looks fine.

TristanCacqueray:
Yes thanks, it seems like a pretty natural thing to expect.

PR #483, opened 2022-06-24
Add robots list view

TristanCacqueray:
This change adds a new dialog to list the robots and their status.

### DIFF
--- a/src/Swarm/TUI/Controller.hs
+++ b/src/Swarm/TUI/Controller.hs
@@ -193,10 +193,11 @@ handleMainEvent s = \case
     | isJust (s ^. uiState . uiError) -> continue $ s & uiState . uiError .~ Nothing
     | isJust (s ^. uiState . uiModal) -> maybeUnpause s >>= (continue . (uiState . uiModal .~ Nothing))
   FKey 1 -> toggleModal s HelpModal >>= continue
-  FKey 2 | not (null (s ^. gameState . availableRecipes . notificationsContent)) -> do
+  FKey 2 -> toggleModal s RobotsModal >>= continue
+  FKey 3 | not (null (s ^. gameState . availableRecipes . notificationsContent)) -> do
     s' <- toggleModal s RecipesModal
     continue (s' & gameState . availableRecipes . notificationsCount .~ 0)
-  FKey 3 | not (null (s ^. gameState . availableCommands . notificationsContent)) -> do
+  FKey 4 | not (null (s ^. gameState . availableCommands . notificationsContent)) -> do
     s' <- toggleModal s CommandsModal
     continue (s' & gameState . availableCommands . notificationsCount .~ 0)
   ControlKey 'g' -> case s ^. uiState . uiGoal of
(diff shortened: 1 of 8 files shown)


TristanCacqueray:
I think we might want a multi column strategy when the list becomes too long to better utilize the available viewport space.

Also, I'm not sure what to put in the status column, for example how to check from the CESK machine when a robot is idle?

src/Swarm/TUI/Controller.hs
```diff
@@ -279,10 +280,14 @@ toggleModal s mt = case s ^. uiState . uiModal of
   Nothing -> pure $ s & (uiState . uiModal ?~ generateModal s mt) . ensurePause
   Just _ -> maybeUnpause s <&> uiState . uiModal .~ Nothing
  where
+  -- these modals do not pause the game
+  runningModals = [RobotsModal]
```
TristanCacqueray, line 284: even though the game is not paused, the modal is somehow not updated.

xsebek:
It should be idle when there is an `Out {}` value I think. :thinking:

TristanCacqueray:
> > how to check from the CESK machine when a robot is idle?
> 
> It should be idle when there is an `Out {}` value I think. thinking `not . isActive` should do the trick.

Thanks, that works! How about getting the age of a robot, is this something we can already know, or should i add a new `createdAt :: TimeSpec` robot (or entity) attribute?

TristanCacqueray:
And I found a way to make the modal screen render update. The issue was that the widget was pre-generated when the modal was created; I guess this was done like that because it is easier to do a single pattern match. Thus this PR changes the rendering logic a bit so that the modal are updated when the game state change.

byorgey:
> How about getting the age of a robot, is this something we can already know, or should i add a new `createdAt :: TimeSpec` robot (or entity) attribute?

I don't think that's something we can already know, but it sounds like a useful addition.  A `createdAt :: TimeSpec` field would be able to tell us the actual wall clock age of a robot.  It might also be nice to know the age of a robot in game ticks.

src/Swarm/TUI/View.hs
```diff
@@ -288,53 +291,98 @@ renderErrorDialog err = renderDialog (dialog (Just "Error") Nothing (maxModalWin
...
             )
-      GoalModal g -> (" Goal ", padLeftRight 1 (displayParagraphs g), Nothing, 80)
+      GoalModal _ -> (" Goal ", Nothing, 80)
+
+robotsListWidget :: GameState -> Widget Name
+robotsListWidget g = viewport RobotsViewport Vertical (hCenter table)
+ where
+  table =
+    BT.renderTable
+      . BT.setDefaultColAlignment BT.AlignCenter
```
byorgey, line 348: In ASCII art, I don't like the default of having lines drawn on all the borders.  It looks visually distracting and takes up a lot of space, especially vertical space.  How would it look to turn off the cell borders?  Can we do that + add enough space between things so that it looks OK?

byorgey (approved):
This is really nice!

A few ideas for improvement (either now or later):
- Include a little picture of the robot somewhere in each row, like we see at the top of the robot panel in the upper left.  May be especially useful once we get the `appear` command.
- Show the terrain and entity (if any) each robot is currently standing on.

src/Swarm/TUI/View.hs
TristanCacqueray: Agreed, in b2221fd the vertical border are removed, what do you think?

byorgey: That looks nice!

byorgey:
Another idea: let's make the words "idle", "waiting", "working" different colors.  I noticed it is particularly hard to distinguish between "waiting" and "working" when there are a bunch of robots in the list.

byorgey:
Looks great!  This is so extremely useful.  Goes a long way towards debugging and figuring out what's going on.  It also makes a nice impetus to use the `setname` command more often.

TristanCacqueray:
That's great to hear! This will probably needs more work to be usable with many robots, e.g. a search box would be nice. Being able to move the dialog position would also be helpful, to have the robot list on the side.

byorgey:
Agree that some kind of search would be helpful.  For moving the dialog we should probably just make some generic system for moving any dialog.

PR #485, opened 2022-06-24
Let robots wait 1 without going to queue

### LINKED ISSUE #475: `yield` primitive
Consider the following code:
```
def while : cmd bool -> {cmd a} -> cmd () = \test. \body.
end
def waitFor : string -> cmd () = \thing.
  while (b <- ishere thing; return (not b)) {}
end
```
`waitFor "tree"`, for example, simply does nothing until a `tree` is present.

This all seems reasonable and it works fine.  However, there is one annoying problem: when a robot takes its turn to execute and it is running `waitFor`, if the item is not present, `waitFor` will **execute repeatedly** until it has used up all the robot's allotted steps for the current tick, even though we can see that the status of the test cannot possibly change until some other robots have had a chance to run.

I propose adding a primitive `yield : cmd ()`, which has the same semantics as `noop`, but indicates that the robot is done running for this tick and does not wish to run again until the next.  Given `yield`, we could rewrite `waitFor` like so:
```
def waitFor : string -> cmd () = \thing.
  while (b <- ishere thing; return (not b)) { yield }
end
```
Now if the looked-for thing is not present, the robot will immediately yield until the next tick, when the status of the test could possibly have changed.

Like `noop`, I think `yield` should not require any capabilities.

Using `wait 1` could in theory have a similar effect.  However, that would mean repeatedly putting the robot in the waiting robots queue and then immediately taking it back out again.  Also, I imagine that `wait` may eventually require a device like a `clock`.

xsebek:
- closes #475

### DIFF
--- a/src/Swarm/Game/Step.hs
+++ b/src/Swarm/Game/Step.hs
@@ -113,12 +113,14 @@ gameTick = do
             when (newLoc /= oldLoc) $ do
               robotsByLocation . at oldLoc %= deleteOne rn
               robotsByLocation . at newLoc . non Empty %= IS.insert rn
+            time <- use ticks
             case waitingUntil curRobot' of
-              Just wakeUpTime ->
-                sleepUntil rn wakeUpTime
+              Just wakeUpTime
+                -- if w=2 t=1 then we do not needlessly put robot to waiting queue
+                | wakeUpTime - 2 <= time -> return ()
+                | otherwise -> sleepUntil rn wakeUpTime
               Nothing ->
-                unless (isActive curRobot') do
-                  sleepForever rn
+                unless (isActive curRobot') (sleepForever rn)
 
   -- See if the base is finished with a computation, and if so, record
   -- the result in the game state so it can be displayed by the REPL;
@@ -373,7 +375,7 @@ stepCESK cesk = case cesk of
   -- then stepCESK is a no-op.
   Waiting wakeupTime cesk' -> do
     time <- use ticks
-    if wakeupTime == time
+    if wakeupTime <= time
       then stepCESK cesk'
       else return cesk
   Out v s (FImmediate wf rf : k) -> do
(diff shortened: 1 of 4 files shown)


data/scenarios/04Testing/475-wait-one.yaml
```diff
@@ -0,0 +1,43 @@
...
+      - grabber
+    inventory:
+      - [1, lambda]
+  - name: sleeper
+    loc: [0,-1]
+    dir: [1,0]
+    devices:
+      - logger
+    program: |
+      log "I shall sleep"; wait 1; log "I have awoken"
```
xsebek, line 25: @byorgey does this test your use case enough? I was thinking about one robot `wait 3` and another `wait 1; log "Is it time?"; wait 1` or we could use something from #475 description.

I guess this could also be a normal test, but I wanted to make sure we use the same `gameTick`.

byorgey: At first I was confused what this is testing but then I also saw the extra test code in `integration/Main.hs`.  Yes, I think this is sufficient.

src/Swarm/Game/Step.hs
```diff
@@ -111,12 +111,14 @@ gameTick = do
             when (newLoc /= oldLoc) $ do
               robotsByLocation . at oldLoc %= deleteOne rn
               robotsByLocation . at newLoc . non Empty %= IS.insert rn
+            time <- use ticks
             case waitingUntil curRobot' of
-              Just wakeUpTime ->
-                sleepUntil rn wakeUpTime
+              Just wakeUpTime
+                -- if w=2 t=1 then we do not needlessly put robot to waiting queue
+                | wakeUpTime - 2 <= time -> return ()
```
byorgey, line 118: It doesn't really matter much, but I am curious why you chose `2` here instead of `1`.

byorgey (approved):

TristanCacqueray (approved):

src/Swarm/Game/Step.hs
xsebek: Because I am very bad with off-by-one errors and 2 seemed safer. It still took me several tries to get it right.

byorgey: That's totally fair. 😄

PR #488, opened 2022-06-25
Add branch filter to reduce CI cost

TristanCacqueray:
The action presently runs twice because both `on` conditions matche
every contributions. With this change, the push action only triggers
for the final push to the main branch, and the pull_requeste action
only triggers for PR made for the main branch.

That also means restyled PR should not trigger CI as they are made
for non main branches.

### DIFF
--- a/.github/workflows/haskell-ci.yml
+++ b/.github/workflows/haskell-ci.yml
@@ -23,7 +23,7 @@
 #
 name: Haskell-CI
 on:
-  push: 
+  push:
     paths:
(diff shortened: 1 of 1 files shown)


byorgey:
The CI didn't seem to run at all for this PR, is that expected?

TristanCacqueray:
Yes that was expected, the workflow file was not matched.

byorgey (approved):
OK, looks good to me.

PR #491, opened 2022-06-25
Repl mouse event

lsmor:
This draft explores the idea of handling the Mouse event and manually set the cursor position. It is clearly not working. If anyone has a better idea feel free to comment :)


byorgey:
I don't think this approach is going to work.  The problem is that even though you can set the visual position of the cursor, the REPL form has its own internal idea of where the cursor is, and those won't match.

byorgey:
I'm going to close this now that #750 is merged.

PR #494, opened 2022-06-26
Add a few more JSON instances

### LINKED ISSUE #347: Add missing ToJSON instances
I would like to have [`ToJSON`](https://hackage.haskell.org/package/aeson-2.0.3.0/docs/Data-Aeson.html#t:ToJSON) to complement the current [`FromJSON`](https://hackage.haskell.org/package/aeson-2.0.3.0/docs/Data-Aeson.html#t:FromJSON). 

- **With #337 adding a `ToJSON Scenario` would solve 99% of #50!** But for purposes of this issue adding most of the necessary instances would already be great.
- If designing `ToJSONE`, consider making the class generate JSON Schema for free and solve #346.

TristanCacqueray:
- part of #347


src/Swarm/Game/CESK.hs
```diff
@@ -373,3 +378,8 @@ instance Show RobotUpdate where show _ = "RobotUpdate {???}"
 instance Eq WorldUpdate where _ == _ = True
 
 instance Eq RobotUpdate where _ == _ = True
+
+instance FromJSON WorldUpdate where parseJSON _ = pure $ WorldUpdate $ \w -> Right w
```
TristanCacqueray, line 382: I guess this needs to be updated with the actual update function after decoding.

xsebek: I think `WorldUpdate` is my mistake. It is too generic when all we use it for is updating one robot or cell.

We should add more concrete constructors that would be possible to serialize and also easier to debug.

But it does not have to be in this PR.

xsebek (approved):
Wow, `Generic` is great! :+1: 

Please just change the description to "- part of #347" as I think this does not add all missing `ToJSON` instances.

My main interest is the `World` which has only a parser in `Scenario` but is missing a serializer.

TristanCacqueray:
I think it's worth using `Generic`, however it seems to double, or maybe triple, the build time. Let me try adding instances for World.

TristanCacqueray:
I guess the alternative to `Generic` would be to use `Data.Aeson.TH`, though I'm not familiar with it.

src/Swarm/Game/Robot.hs
```diff
@@ -150,8 +155,14 @@ data RobotR f = RobotR
   , _tickSteps :: Int
   , _robotCreatedAt :: TimeSpec
   }
+  deriving (Generic)
 
 deriving instance Show (f RID) => Show (RobotR f)
+deriving instance FromJSON (f RID) => FromJSON (RobotR f)
+deriving instance ToJSON (f RID) => ToJSON (RobotR f)
```
xsebek, line 162: This is a bit surprising since we have the `FromJSONE` instance above.

That one requires the entity map and also does not allow specifying robot ID nor some other fields like the log.

I am not sure what this `ToJSON` instance produces, but I would like it to match the `FromJSONE`.
Though if the generated instance is reasonable maybe it could be useful and we could align them later.

TristanCacqueray:
Here are the missing instances to be able to serialize the GameState:

instance FromJSONKey (V2 Int64)
instance ToJSONKey (V2 Int64)

instance FromJSON (Array Int Text)
instance ToJSON (Array Int Text)

instance FromJSON (W.World Int Entity)
instance ToJSON (W.World Int Entity)

instance FromJSON ScenarioCollection
instance ToJSON ScenarioCollection

instance FromJSON StdGen
instance ToJSON StdGen

src/Swarm/Game/CESK.hs
TristanCacqueray: Ok, I've added a TODO

src/Swarm/Game/Robot.hs
TristanCacqueray: To be honest, I have no idea, I just typed on the keyboard until GHC could link the binary :-)

TristanCacqueray:
Oh well, build time are back to the 5min range (they went up to 15min in the previous checks)

xsebek:
- `V2 Int64` is just a pair, should be easy
- `Array Int Text` is just an efficient `Map` in my mind
- `World Int Entity` this one requires a bit of work
  - you need to serialize chunks of the world and the entity palette (in the scenarios we only do one chunk)
  - we would need to express our `WorldFunction` as a data type and use that as the generator - but maybe we can pretend that it is always either test world 2 or one entity everywhere for now :slightly_smiling_face: 
- `ScenarioCollection` IMHO this should be in UI state, it is a bit annoying that a broken scenario does not let me run any test
- `StdGen` its current state can be shown, but it does not have a read instance :confused:

src/Swarm/Game/Robot.hs
byorgey: Note, when I made the `FromJSONE` instances I was just interested in putting in enough functionality to be able to describe robots in the ways that I wanted in scenario description files.  But if there are fields that are still missing we can definitely add them.

byorgey:
> * `ScenarioCollection` IMHO this should be in UI state, it is a bit annoying that a broken scenario does not let me run any test

Agreed, let's move it.

byorgey:
> * `StdGen` its current state can be shown, but it does not have a read instance confused

Relevant: https://www.reddit.com/r/haskell/comments/ponow7/saving_and_loading_a_stdgen/hcxvrcp/?utm_source=reddit&utm_medium=web2x&context=3

TristanCacqueray:
Actually we may not want to serialize the `StdGen`. What matters is to be able to recreate the world, so we should record the initial seed, and use it to regenerate the same world, but how random number are produced could be random across reload.

src/Swarm/Game/Scenario.hs
```diff
@@ -1,5 +1,7 @@
 {-# LANGUAGE DefaultSignatures #-}
+{-# LANGUAGE DeriveAnyClass #-}
 {-# LANGUAGE DeriveFunctor #-}
+{-# LANGUAGE DeriveGeneric #-}
```
TristanCacqueray, line 4: It's odd hlint does not complain about unused extensions, but I think that's fine because we'll probably need them for the scenarios.

xsebek:
> Actually we may not want to serialize the StdGen.

Counterexample: A player submits a bug report with a saved game and it depends on a random number somewhere.
If you want to randomize it again, you can just call `random` once more or we could ignore it at loading if the player wants to.

We already allow a custom seed, so keeping the current random generator in saves seems natural. :man_shrugging:

byorgey:
I was going to say the same thing re: saving the `StdGen`.  I agree it's not strictly necessary as far as gameplay goes but it would help with reproducibility, and we might as well.

TristanCacqueray:
@byorgey if that's ok with you, i'd like to merge this PR for #537

byorgey (approved):
Sounds good to me!

PR #499, opened 2022-06-27
Give the option to go immediately to the next challenge after completing one

### LINKED ISSUE #355: "Meta-scenarios" consisting of sequences of other scenarios
For the in-game tutorial (#25) we are going to want some way to have a chained sequence of scenarios, where completing one scenario causes the next one in the sequence to load automatically (well, probably mediated by a dialog that lets you choose between continuing to play the current scenario, loading the next, or quitting).  This could also be useful for creating other curated/sequenced sets of scenarios.

My idea for this (as mentioned in #296) is to simply have this be yet another type of scenario, described in a `.yaml` file, with a name, description, etc. but then having a list of references to other scenarios instead of a description of a world etc.

byorgey:
- Always have `uiMenu` store the current menu; while playing, it represents the menu we should return to if we quit
- Get rid of `uiPrevMenu` which is thus no longer needed
- `NoMenu` is now only used to represent the scenario when we started via command-line options and thus should not return to a menu when we quit
- Add `uiPlaying` to denote whether we are playing the game or displaying a menu (`NoMenu` used to be used for this purpose)
- Only advance to the next scenario in the menu when actually completing the previous one.  For example if we start a scenario then quit before completing it, the menu will still be highlighting the same scenario.  But if we complete a scenario and then quit to the menu, the next one will be highlighted.
- Closes #355.  As [explained in a comment there](https://github.com/swarm-game/swarm/issues/355#issuecomment-1166343849), I think true "meta-scenarios" as originally described are needlessly complex. This PR gives us something that should be good enough: scenarios that have win conditions, placed in the same folder, can now be played sequentially without ever returning to the menu in between.

### DIFF
--- a/src/Swarm/Game/Scenario.hs
+++ b/src/Swarm/Game/Scenario.hs
@@ -41,7 +41,7 @@ module Swarm.Game.Scenario (
   loadScenario,
   ScenarioCollection (..),
   scenarioCollectionToList,
-  ScenarioItem (..),
+  ScenarioItem (..), _SISingle, _SICollection,
   scenarioItemName,
   loadScenarios,
 ) where
@@ -354,3 +354,10 @@ loadScenarioFile em fileName = do
   case res of
     Left parseExn -> throwError @Text (from @String (prettyPrintParseException parseExn))
     Right c -> return c
+
+------------------------------------------------------------
+-- Some lenses + prisms
+------------------------------------------------------------
+
+makePrisms ''ScenarioItem
+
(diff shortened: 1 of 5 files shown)


src/Swarm/Game/Scenario.hs
```diff
@@ -42,6 +42,8 @@ module Swarm.Game.Scenario (
   ScenarioCollection (..),
   scenarioCollectionToList,
   ScenarioItem (..),
+  _SISingle,
+  _SICollection,
```
TristanCacqueray, line 46: `_SICollection` does not seems to be used. Also isn't it odd to export symbol prefixed with `_`, I thought this was for reserved for internal name.

PR #501, opened 2022-06-27
Tab completion

### LINKED ISSUE #207: Tab completion
It would be very nice for the REPL to support tab completion. At a minimum we should be able to complete names of built-in commands, but ideally also any currently defined name.

I haven't thought a lot about what would be required to make this work well.

### LINKED ISSUE #316: Do away with Tab to cycle panels
Cycling between panels using Tab is annoying and error-prone, and clashes with the desire to have tab completion for the REPL (#207).  A few thoughts on what we might do instead:

- Allow switching panels by clicking on the panel to focus.
- Of course we also have shortcuts for switching panels.  Maybe allow users to customize their panel shortcuts via a `.swarm` file or something like that.

byorgey:
- Tab no longer cycles between panels (closes #316).  Now that we have keyboard shortcuts *and* you can click to move the panel focus, that seems sufficient.
- Very basic REPL tab completion (closes #207).  Currently tries to complete against all reserved words and all in-scope names.

Known limitations/things that should be improved:
- Doesn't do anything when there are multiple matches.  So if you type a partial word and hitting Tab does nothing, there is no way to know whether it's because it matched nothing, or it matched more than one thing.
- Doesn't work if the word to complete is preceded by punctuation instead of whitespace.

### DIFF
--- a/src/Swarm/TUI/Controller.hs
+++ b/src/Swarm/TUI/Controller.hs
@@ -81,6 +81,7 @@ import Swarm.Game.Value (Value (VUnit), prettyValue)
 import Swarm.Game.World qualified as W
 import Swarm.Language.Capability
 import Swarm.Language.Context
+import Swarm.Language.Parse (reservedWords)
 import Swarm.Language.Pipeline
 import Swarm.Language.Pretty
 import Swarm.Language.Syntax
@@ -612,6 +613,12 @@ handleREPLEvent s = \case
         Just found ->
           let newform = mkReplForm $ SearchPrompt ftext (removeEntry found rh)
            in s & uiState . uiReplForm .~ newform
+  CharKey '\t' -> do
+    let formSt = s ^. uiState . uiReplForm . to formState
+        newform = mkReplForm $ formSt & promptTextL %~ tabComplete s
+    continue $
+      s & uiState . uiReplForm .~ newform
+        & validateREPLForm
   EscapeKey ->
     case s ^. uiState . uiReplForm . to formState of
       CmdPrompt _ -> continueWithoutRedraw s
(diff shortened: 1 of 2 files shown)


src/Swarm/Util.hs
```diff
@@ -165,6 +170,31 @@ readAppData = do
...
+-- >>> isIdentChar '$' || isIdentChar '.' || isIdentChar ' '
+-- False
+isIdentChar :: Char -> Bool
+isIdentChar c = isAlphaNum c || c == '_' || c == '\''
+
+-- | @replaceLast r t@ replaces the last word of @t@ with @r@.
+--
+-- >>> :set -XOverloadedStrings
+-- >>> replaceLast "foo" "bar baz quux"
+-- "bar baz foo"
```
xsebek, line 192: Glad to see the doctest put to good use! :+1:

src/Swarm/TUI/Controller.hs
```diff
@@ -587,76 +587,96 @@ stripCmd pty = pty
...
+tabComplete :: AppState -> REPLPrompt -> REPLPrompt
+tabComplete _ p@(SearchPrompt {}) = p
+tabComplete _ (CmdPrompt "" []) = CmdPrompt "" []
+tabComplete s (CmdPrompt t []) = case matches of
+  [] -> CmdPrompt t []
+  [m] -> CmdPrompt (completeWith m) []
+  (m : ms) -> CmdPrompt (completeWith m) (ms ++ [m])
+ where
+  completeWith m = T.append t (T.drop (T.length lastWord) m)
+  lastWord = T.takeWhileEnd isIdentChar t
```
xsebek, line 668: This does not work for completing operators, e.g. `=` get completed as `=noop`.

Maybe there should be a check that before the "last word", there is either nothing or a whitespace character?

xsebek (approved):
It is not perfect, but now I won't be able to live without this feature, so please merge soon. :+1:  

We should also add support for autocompletion to LSP, the common code should be then easy to refactor out. :nerd_face:

TristanCacqueray (approved):

src/Swarm/TUI/Controller.hs
byorgey: Thanks, filed as #518 .

PR #511, opened 2022-06-30
Add scenario entites to game entity map

### LINKED ISSUE #509: Scenario entites are not considered for install
When a scenario-only entity (like `Tardis`) provides a capability (like `teleport`) it should be installed when building a robot that requires it. But it does not happen.

```Yaml
robots:
  - name: base
    loc: [0,0]
    dir: [1,0]
    devices:
      - logger
      - 3D printer
    inventory:
      - [1, mirror]
      - [1, Tardis]
      - [1, logger]
entities:
  - name: Tardis
    display:
      attr: water
      char: '█'
    description:
    - Bigger on the inside.
    properties: [known, portable]
    capabilities: [senseloc, teleport]
```
```
build {log "Hi, I can teleport!"; teleport self (5,0)}
```

Screenshot from 2022-06-30 15-45-32

xsebek:
- fixes #509

### DIFF
--- a/src/Swarm/Game/State.hs
+++ b/src/Swarm/Game/State.hs
@@ -609,6 +609,7 @@ scenarioToGameState scenario userSeed toRun g = do
       , _waitingRobots = M.empty
       , _gensym = initGensym
       , _randGen = mkStdGen seed
+      , _entityMap = g ^. entityMap <> scenario ^. scenarioEntities
       , _recipesOut = addRecipesWith outRecipeMap recipesOut
       , _recipesIn = addRecipesWith inRecipeMap recipesIn
       , _world = theWorld seed
--- a/test/integration/Main.hs
+++ b/test/integration/Main.hs
@@ -160,7 +160,7 @@ testScenarioSolution _ci _em =
             assertBool "Robot 1 should be still active" active
             assertBool "Robot 1 should not be in waiting set" $ not waiting
         , testSolution Default "Testing/504-teleport-self"
-        , expectFailBecause "Awaiting fix (#509, #508?)" $
+        , expectFailBecause "Awaiting fix (#508)" $
             testSolution Default "Testing/508-capability-subset.yaml"
         ]
     ]


src/Swarm/Game/State.hs
```diff
@@ -609,6 +609,7 @@ scenarioToGameState scenario userSeed toRun g = do
       , _waitingRobots = M.empty
       , _gensym = initGensym
       , _randGen = mkStdGen seed
+      , _entityMap = g ^. entityMap <> scenario ^. scenarioEntities
```
byorgey, line 612: I agree that #509 is a bug and this is the change needed to fix it. However, we now have another problem which is what my comment `do we need to keep an old entity map around???` was talking about: changes to the entity map are permanent.  For example, to reproduce:
1. Start Swarm in cheat mode
2. Load the new testing scenario containing `Tardis`
3. Quit to the menu
4. Start a creative mode game
5. Execute `create "Tardis"`: it *should* say "I've never heard of a Tardis", but instead, it works.  That's because the `Tardis` was added to the `entityMap` when loading the test scenario, and it was never removed.

We will have the same problem with recipes.

The problem is that `scenarioToGameState` *modifies* the existing `GameState` instead of creating a brand new one from scratch.  I am actually not sure right now whether there are good reasons for this.  But I think the "right" solution, assuming it is not too difficult for some reason I can't think of right now, is to create a new `GameState` from scratch every time we load a new scenario.  But if we don't want to actually reread the entities and recipes from disk every time, we will have to save copies of the "pristine" versions somewhere, which is what my comment `do we need to keep an old entity map around???` was supposed to mean.

I don't care whether we try to fix the problem in this PR, or split out another issue and fix it later.

xsebek:
I figured making a new branch rebased on the main branch will be simpler. :slightly_smiling_face:

byorgey:
Hah, I had just finished writing a long comment when you closed it. :laughing: No worries, I can copy it to the new PR.

PR #520, opened 2022-06-30
Make GHC2021 extensions default

### LINKED ISSUE #419: Consider using GHC2021 extensions by default
Let us discuss the `default-extensions` field in `swarm.cabal` - I suggest we put there all the extensions that we use a lot and that are harmless and not very interesting.

AFAIK we can not use [`GHC2021`](https://github.com/ghc-proposals/ghc-proposals/blob/ghc2021/proposals/0000-ghc2021.rst) yet, but it is the current "standard" way of writing Haskell, based on some collected statistics, community votes and a few votes by committee.

### PROs

- cleans up the module header
- highlights the more interesting extensions (looking at TH and undecidable instances :raised_eyebrow:)
- makes it easier to use nice syntax sugar anywhere we like

### CONs

- some bikeshedding could take place
- ?

## Extensions

Here are the extensions that we use in Swarm along with their count with those in GHC2021 checked:

- [ ] 1 AllowAmbiguousTypes
- [x] 1 BangPatterns
- [ ] 1 BlockArguments
- [ ] 1 DerivingStrategies
- [ ] 1 GADTs (we could use GADTSyntax which is in GHC2021, related #331) 
- [x] 1 ScopedTypeVariables
- [ ] 1 TemplateHaskellQuotes
- [ ] 2 DefaultSignatures
- [ ] 2 DerivingVia
- [x] 2 ImportQualifiedPost (we could require it with [-Wprepositive-qualified-module](https://ghc.gitlab.haskell.org/ghc/doc/users_guide/using-warnings.html#ghc-flag--Wprepositive-qualified-module))
- [x] 2 NumericUnderscores
- [x] 3 DeriveFunctor
- [x] 3 DeriveTraversable
- [x] 3 TypeOperators
- [ ] 3 ViewPatterns
- [x] 4 StandaloneDeriving
- [ ] 4 TypeFamilies
- [ ] 4 UndecidableInstances
- [x] 5 DeriveDataTypeable
- [ ] 5 PatternSynonyms
- [ ] 5 QuasiQuotes
- [x] 6 DeriveAnyClass
- [x] 6 DeriveGeneric
- [x] 6 RankNTypes
- [x] 6 TupleSections
- [x] 8 FlexibleContexts
- [ ] 8 TemplateHaskell
- [ ] 9 LambdaCase (let's add this one too)
- [x] 10 MultiParamTypeClasses
- [x] 11 FlexibleInstances
- [x] 15 TypeApplications
- [ ] 20 OverloadedStrings

---

What do you think? :slightly_smiling_face:

xsebek:
- closes #419


src/Swarm/Game/Exception.hs
```diff
@@ -37,7 +36,7 @@ import Swarm.Util
 -- $setup
 -- >>> :set -XOverloadedStrings
 -- >>> import Control.Lens
--- >>> import qualified Data.Set as S
+-- >>> import Data.Set qualified as S
```
xsebek, line 39: OK, this was a mistake :sweat_smile:

xsebek:
I wonder why Fourmolu failed in restyled? :thinking: It should find the extension in the cabal file just fine.

byorgey:
No opinion on whether we should wait to merge this.  We are going to have merge conflicts one way or another.  I guess you mean if we wait to merge this, then maybe we will just be able to fix a bunch of merge conflicts here at once, instead of having to fix merge conflicts for every other pending PR.

On the other hand there are always going to be more PRs... :smile:

byorgey:
> I wonder why Fourmolu failed in restyled?  It should find the extension in the cabal file just fine.

Indeed, that is strange.  https://hackage.haskell.org/package/fourmolu-0.7.0.1#language-extensions-dependencies-and-fixities does seem to imply this should work.

xsebek:
Yes, it does, if you provide the `--cabal-default-extensions` flag. :unamused: 

I am looking for a way to enable that in CI...

xsebek:
No, but we should be able to add it to restyled.

swarm.cabal
```diff
@@ -130,7 +131,32 @@ library
                       yaml                          >= 0.11 && < 0.12,
     hs-source-dirs:   src
     default-language: Haskell2010
-    default-extensions: StrictData
+    default-extensions:
```
TristanCacqueray, line 134: perhaps we should move default-language and default-extensions to a new common used by every target. Like that tests and executables uses the same extensions as the lib.

xsebek:
Hey @TristanCacqueray, would you happen to know how to coerce restyled to pass a flag to fourmolu?

I tried something in #521, but it did not work. :disappointed:

PR #521, opened 2022-06-30
Restyled extensions

xsebek:
Just trying this out...


xsebek:
I looked at https://github.com/project-chip/connectedhomeip/blob/master/.restyled.yaml, this should work. :confused:

xsebek:
Here is the relevant wiki:
https://github.com/restyled-io/restyled.io/wiki/Configuring-Restyled#restylers-1

.hlint.yaml
```diff
@@ -9,7 +9,26 @@
 # Specify additional command line arguments
 #
 # - arguments: [--color, --cpp-simple, -XQuasiQuotes]
-
+- arguments:
+  - '-XBangPatterns'
```
TristanCacqueray, line 13: I think this should be: `--ghc-opt -XBangPatterns --ghc-opt -XDeriveAnyClass ...`

xsebek: You probably mean a different file, but thanks, it helped me find the right thing. :+1:

The syntax I needed was ['--ghc-opt=-X...','--ghc-opt=-X',...]

Issue #508, opened 2022-06-30
`build` does not always install a minimal set of devices


byorgey:
For example, a `grabber` and `harvester` might both be installed when the `grab` and `harvest` capabilities are needed, even though a `harvester` already provides both capabilities.  It would be nice if the analysis figured out that it could get away with just installing a `harvester` alone.

Note, this depends on merging #503 , where this problem was first noted.

To reproduce, make sure you have a `harvester` in your inventory, then execute `r <- build {grab; harvest}; view r`.  You can see that the newly built robot has both a `harvester` and a `grabber` installed.

...actually, I am not sure whether this is a problem.  Right now, *every* robot has a `grabber` installed whether it needs one or not.  We will have to see if this is still a problem after merging #361 .

byorgey:
The test we added (`508-capability-subset`) actually *succeeds* and I don't yet understand if it's because the bug is not a bug, or if it just so happens to succeed but another similar test would fail.

xsebek:
@byorgey it seems to me like the harvester commit fixed it. :smile:

xsebek:
I guess we could wait for #361 to be merged and if everything works then, we would close this issue.

But I suspect we just got lucky with the current order of entities and other capabilities/devices will not work.

byorgey:
Yeah, I suspect the same thing.  I want to at least spend some time studying the code to think about what it does, and maybe come up with a failing test case if I can.

byorgey:
In general, if the result depends on some arbitrary order, and can change when other unrelated entities are added, I would consider that a bug in and of itself!

byorgey:
Hah, OK, so I looked at the code and indeed it depends on the order of the entities for each capability.  But the entities are stored in a `Set Entity`, and hence ordered by *hash*.  😄   Obviously this is very unpredictable and could change if, say, someone edited the punctuation in the description of one of the entities.

byorgey:
My idea for solving this is to add a minimization post-processing step.  After figuring out which capabilities we need, and then finding which devices that we have could provide each capability, we end up with a collection of non-empty `Set Entity`, one per capability.  That is, for each capability, we have a set of devices which are in our inventory that would provide that capability.  Right now, we simply convert these sets to lists and choose the first device from each set, which guarantees that every capability will be covered.  Instead, we can do something like this:
- If there are any singleton sets, those devices must be chosen, so add them to a set $D$.
- Now go through each of the remaining sets.
[...]

PR #522, opened 2022-07-01
make issue #508 test fail (as expected)

byorgey:
Just tweak the description for the GPS device so its hash changes, and
hence the order of the `GPS` and `Tardis` entities in a `Set Entity`
of devices that provide the `senseloc` capability.  When `GPS` comes
first, it will be chosen to provide the capability instead of
`Tardis`.


xsebek (approved):
😅

xsebek:
No idea what confused Mergify here, but its merged. :shrug:

PR #533, opened 2022-07-02
Add `require` directive

### LINKED ISSUE #201: `require` commands to artificially require devices or supplies

### LINKED ISSUE #508: `build` does not always install a minimal set of devices

byorgey:
Add new special syntactic forms `require <string literal>` to require a device to be installed, and `require <int literal> <string literal>` to require a certain number of a given entity in the inventory.

- Replace `Set Capability` with a new `Requirements` type which records various types of requirement.
- Refactor requirements checking into a new function `checkRequirements` which also does device set minimization (#508).  
- `reprogram` will now install extra devices or transfer extra inventory as necessary.
- Add a bunch of tests.
- Add new `installed` command, similar to `has`, which checks whether a given device is installed.

Closes #201.
Closes #508.


src/Swarm/Game/Step.hs
```diff
@@ -531,10 +539,10 @@ stepCESK cesk = case cesk of
   -- environment, store, and type and capability contexts into the robot's
   -- top-level environment and contexts, so they will be available to
   -- future programs.
-  Out (VResult v e) s (FLoadEnv ctx cctx : k) -> do
+  Out (VResult v e) s (FLoadEnv ctx rctx : k) -> do
     robotContext . defVals %= (`union` e)
     robotContext . defTypes %= (`union` ctx)
-    robotContext . defCaps %= (`union` cctx)
+    robotContext . defReqs <>= rctx
```
byorgey, line 545: Does the composition happen in the right order?

src/Swarm/Game/Step.hs
```diff
@@ -1211,14 +1224,29 @@ execConst c vs s k = do
...
+            FixByObtain
 
         -- update other robot's CESK machine, environment and context
         -- the childRobot inherits the parent robot's environment
         -- and context which collectively mean all the variables
         -- declared in the parent robot
         robotMap . at childRobotID . _Just . machine .= In cmd e s [FExec]
         robotMap . at childRobotID . _Just . robotContext .= r ^. robotContext
+
+        -- Provision the target robot with any require devices and
```
byorgey, line 1245:
```suggestion
        -- Provision the target robot with any required devices and
```

src/Swarm/Game/Step.hs
```diff
@@ -1487,9 +1540,34 @@ execConst c vs s k = do
                           (("\n  - " <>) . formatDevices <$> filter (not . null) missingDeviceSets)
                         )
         -- check that there are in fact devices to provide every required capability
-        not (any null deviceSets) `holdsOr` Incapable fixI missingCaps cmd
-        -- give back the devices required per capability
-        return deviceSets
+        not (any null deviceSets) `holdsOr` Incapable fixI (R.Requirements missingCaps S.empty M.empty) cmd
+
+        let minimalDeviceSet = smallHittingSet deviceSets
+            minimalInstallSet = minimalDeviceSet `S.difference` alreadyInstalled
```
byorgey, line 1546: Take out already installed BEFORE finding hitting set.

byorgey:
Hmm, I should probably add a tutorial challenge or two about this.  I will definitely do that before merging.

byorgey:
Tutorials should be good to go now!

data/scenarios/Tutorials/requireinv.yaml
```diff
@@ -0,0 +1,78 @@
...
+    loc: [0,0]
+    dir: [1,0]
+    display:
+      char: Ω
+      attr: robot
+    devices:
+      - logger
+      - 3D printer
+      - dictionary
+    inventory:
```
xsebek, line 49: I would like a compass so that I do not have to test which direction the robots will go by default.
```suggestion
    inventory:
     - [10, compass]
```

data/scenarios/Tutorials/requireinv.yaml
```diff
@@ -0,0 +1,78 @@
...
+      - [10, solar panel]
+      - [10, logger]
+      - [10, treads]
+      - [10, boat]
+      - [10, grabber]
+      - [10, scanner]
```
xsebek, line 55: Let's make the number of devices 16 so that I can send a robot to each position. :smile:

editors/vscode/syntaxes/swarm.tmLanguage.json
```diff
@@ -56,7 +56,7 @@
 				},
 				{
 				"name": "keyword.other",
-				"match": "\\b(?i)(noop|wait|selfdestruct|move|turn|grab|harvest|place|give|install|make|has|count|drill|build|salvage|reprogram|say|log|view|appear|create|whereami|blocked|scan|upload|ishere|self|parent|base|whoami|setname|random|run|if|inl|inr|case|fst|snd|force|return|try|undefined|fail|not|format|teleport|as|robotnamed|robotnumbered|knows)\\b"
+				"match": "\\b(?i)(noop|wait|selfdestruct|move|turn|grab|harvest|place|give|install|make|has|installed|count|drill|build|salvage|reprogram|say|log|view|appear|create|whereami|blocked|scan|upload|ishere|self|parent|base|whoami|setname|random|run|if|inl|inr|case|fst|snd|force|return|try|undefined|fail|not|format|teleport|as|robotnamed|robotnumbered|knows)\\b"
```
xsebek, line 59: I will also need to add `require` as special syntax and then update the plugin.

I might not have time to do it soon, so could you please create a ticket and assign me, so that I do not forget?

src/Swarm/Language/Parse.hs
```diff
@@ -242,6 +243,10 @@ parseTermAtom =
         <|> TInt <$> integer
         <|> TString <$> stringLiteral
         <|> TBool <$> ((True <$ reserved "true") <|> (False <$ reserved "false"))
+        <|> reserved "require"
+          *> ( (TRequireDevice <$> stringLiteral)
+                <|> (TRequire <$> (fromIntegral <$> integer) <*> stringLiteral)
+             )
```
xsebek, line 249: Could you add a help text to the `stringLiteral` parser?

It is not easy to understand the error message when I write `\x. require 1 x` and get:
```
Error: unexpected 'x', expecting '"'
```
Something like expecting `"a literal string that is the name of an entity"` or specifically "device" in the first case, would be more helpful IMO. :slightly_smiling_face:

test/unit/Main.hs
```diff
@@ -627,8 +639,84 @@ inventory =
...
+prop_hittingSet ss = not (any S.null ss) ==> isHittingSet (smallHittingSet ss) ss
+
+isHittingSet :: (Foldable t, Ord a) => Set a -> t (Set a) -> Bool
+isHittingSet hs = not . any (S.null . S.intersection hs)
+
+-- This property does *not* hold (and should not, because the problem
+-- of producing a minimal hitting set is NP-hard), but we can use it
+-- to generate counterexamples.
+prop_hittingSetMinimal :: [Set El] -> Property
+prop_hittingSetMinimal ss = not (any S.null ss) ==> all ((S.size hs <=) . S.size) allHittingSets
```
xsebek, line 708: Nice property test! :+1:

Do you think Quickcheck would find the counterexample with real sets of capabilities that entities have, or would that take too long?

xsebek (approved):
Thanks, @byorgey, it will be great not to drown robots! :boat:

It was fun solving the tutorials, though I had a peek at the solution before solving the second one - definitely not intentionally! :+1:

test/unit/Main.hs
byorgey: It would take way too long, I think.  Also, I'm not sure that any counterexamples actually exist with the real sets of capabilities that entities currently have.

src/Swarm/Language/Parse.hs
byorgey: Good idea, thanks! :heavy_check_mark:

data/scenarios/Tutorials/require.yaml
```diff
@@ -0,0 +1,79 @@
...
+    description:
+    - A running stream of of water.
+    properties: [known, portable, growable, liquid]
+    growth: [0,0]
+win: |
+  try {
+    as base {has "periwinkle"}
+  } { return false }
+solution: |
+  build {require "boat"; move; move; move; move; move; f <- grab; turn back; move; move; move; move; move; give base f}
```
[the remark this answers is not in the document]
byorgey: Yes, I am thinking about adding a simple `def` tutorial somewhere in here as well.

data/scenarios/Tutorials/require.yaml
```diff
@@ -0,0 +1,79 @@
...
+      attr: robot
+    devices:
+      - logger
+      - 3D printer
+    inventory:
+      - [10, solar panel]
+      - [10, logger]
+      - [10, treads]
+      - [10, boat]
+      - [10, grabber]
```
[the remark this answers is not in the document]
byorgey: > There is no need to have so many devices in inventory because if the first robot grabs the flower and does not give it to the base you have failed.

Actually, that's not true: I made the flower `infinite` for exactly this reason.  However, I think we definitely should offer the ability to start over.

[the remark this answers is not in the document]
byorgey: Filed #549 to track this.

PR #535, opened 2022-07-03
when pressed ^c restore repl

lsmor:
This closes #65 . It is a very simple change but notice that ^C already cancels current execution. Is it convinient to have more than one action per keybinding?


byorgey:
This looks good to me.  As to whether we want ^C to do multiple things, I am not sure.  Would users expect this?  Or be surprised by it?  Note that you can already clear the input by doing ^A and then ^K.

byorgey:
Ah, I just tried it in my terminal and ^C does indeed work this way.  I guess I never knew that or used it before!

byorgey (approved):

xsebek:
@byorgey @lsmor I think it is consistent with cancel computation in that it cancels the REPL prompt and sets SIGINT status:
image

I could never use that in REPL because the status would annoy me, but it is a neat shortcut. :slightly_smiling_face:

PR #537, opened 2022-07-03
Add a web interface

### LINKED ISSUE #493: Web based debug interface
Being able to debug complex issue, such as the one described in #490 might become tricky to do with the ncurses based interface.

The game starts a wai application that can be used to query the current game state. Initially the interface provides a simple `/export` endpoint to dump the state. But the interface could also serve a little web ui to display the state through a web browser.

With #50, we might be able to inspect the save file for debug purpose too. Though the web service might be more convenient to debug live situations.

**Additional context**
The web service could also receive REPL commands so that the IDE could directly interact with the game, without having to switch focus and use the `run` command.

It also open the possibility of a new web base interface for the game. Though that is a subject for another discussion, the purpose of this issue is purely for adding a debug capability.

TristanCacqueray:
This change adds a new `--web` command line argument to start a web service to serve a new user interface.
The interface presently lists the robot and dump a single robot state.

TODO:

- [ ] Add a websocket handler to refresh the UI automatically, for example to update the robots list when it changes.
- [ ] Format the robot state using html elements (the representation is presently using the Show instance).
- [ ] Use [setBeforeMainLoop](https://hackage.haskell.org/package/warp-3.3.21/docs/Network-Wai-Handler-Warp.html#v:setBeforeMainLoop) to abort if the service can't be started.

Fixes #493


byorgey:
Do you want to mark the PR as draft while you're working on the TODOs?  Or are you OK with this being merged?

TristanCacqueray:
I'd rather close the PR because I don't have much time to make progress on it. Though it might already be useful, and I don't think it should be difficult to maintain, so I'd be happy to merge it.

byorgey:
Then I think we might as well merge it, and create an issue to track the TODOs that are left.  That way other people can potentially contribute as well, if you don't have time to work on it.

src/Swarm/App.hs
```diff
@@ -60,12 +67,34 @@ appMain seed scenario toRun cheat = do
           threadDelay 33_333 -- cap maximum framerate at 30 FPS
           writeBChan chan Frame
 
-      -- Run the app.
+      eventHandler <- case port of
```
TristanCacqueray, line 70: When the web interface is running, it is using a custom handleEvent handler to update a shared IORef. I made this optional to not interfere when the web interface is not activated, but perhaps the overhead is negligible, and in that case, I think we should always starts the web service. That way, there is no need to restart the game if the user forgot to use the `--web` argument, and the web view would be always available if needed.

src/Swarm/Web/Index.hs
```diff
@@ -0,0 +1,106 @@
...
+-- | Display a single robot information
+componentRobot :: Int -> GameState -> Html ()
+componentRobot rid g = do
+  with' div_ "grid p-4 place-content-center" do
+    with' span_ "font-semibold pb-3" do
+      "Robot " <> toHtml (show rid)
+    with' pre_ "whitespace-pre-wrap" do
+      case IM.lookup rid (g ^. robotMap) of
+        Nothing -> "Unknown robot"
+        Just r -> toHtml (pShowNoColor r)
```
TristanCacqueray, line 81: Perhaps it would be better to use the ToJSON instance from #494 with aeson-pretty. With #494, it will also be easy to handle `Accept: application/json` header and serve the json directly.

TristanCacqueray:
I'd be happy to fix minor usability issues, e.g. making the service starts by default, but I agree it would be better to land this initial work before addressing the TODOs.

TristanCacqueray:
Also I picked `tailwind` to create the initial html styles, and if we want to use something else, then it would be better to decide now and avoid adding the js blob (381kb). I think `bootstrap` would also work great, but it is less flexible.

byorgey:
> if we want to use something else, then it would be better to decide now and avoid adding the js blob (381kb)

I have no opinion on the matter.  If you think `tailwind` is a good choice, that's good enough for me.

byorgey:
Making the service start by default sounds like a good idea, unless it makes a noticeable performance difference (which I doubt).

app/Main.hs
```diff
@@ -54,6 +55,8 @@ cliParser =
       ]
   seed :: Parser (Maybe Int)
   seed = optional $ option auto (long "seed" <> short 's' <> metavar "INT" <> help "Seed to use for world generation")
+  webPort :: Parser (Maybe Int)
+  webPort = optional $ option auto (long "web" <> metavar "INT" <> help "Start the web interface")
```
byorgey, line 59: If I saw the words "web interface" I would assume it is an entire alternative UI for playing the game.  Of course if people think that, they are going to be disappointed.  Maybe call it the `web debugging interface` to make it more clear what it is?

byorgey (approved):
I got a chance to try it out this evening.  Other than the issue getting it to build with `stack` (see my PR against this branch), all works great.  I think we can merge this whenever you think it's ready.

swarm.cabal
```diff
@@ -154,6 +164,8 @@ library
                       unordered-containers          >= 0.2.14 && < 0.3,
                       vector                        >= 0.12 && < 0.13,
                       vty                           >= 5.33 && < 5.36,
+                      warp                          >= 3.2,
+                      wai                           >= 3.2,
```
xsebek, line 168: We might want to set some upper bounds so that the build does not unexpectedly fail when some breaking changes are released in the packages. :slightly_smiling_face:

xsebek (approved):
Nice work @TristanCacqueray, I am also happy to merge as is and fix the TODOs incrementally :slightly_smiling_face: 

I am personally not likely to use the web page to view the game state, but I am very much looking forward to adding JSON export paths once #494 lands. That would be a huge step toward a multiplayer game! :rocket:

TristanCacqueray:
It may be premature to pull in CSS and Javascript files into the project, and the choice of libraries should be defined based on an actual implementation. Moreover if we provide a web based user interface, then I think it should match the current esthetic of the game.

Thus I've re-purpose this PR into #623 which I think is more useful for now. I'm closing this one, but perhaps this initial work may be useful in the future.

byorgey:
@TristanCacqueray , makes sense!

TristanCacqueray:
@byorgey good :) Well I would be happy to rebase that effort in the future, but it seems more practical to first add the base API.

PR #541, opened 2022-07-03
Fix tab

noahyor:
a little thing related to tab completion. (#501)

### DIFF
--- a/src/Swarm/TUI/View.hs
+++ b/src/Swarm/TUI/View.hs
@@ -544,12 +544,12 @@ drawKeyMenu s =
     [(NoHighlight, "F1", "help"), (NoHighlight, "F2", "robots")]
       <> notificationKey availableRecipes "F3" "Recipes"
       <> notificationKey availableCommands "F4" "Commands"
-      <> [(NoHighlight, "Tab", "cycle")]
       <> [(NoHighlight, "^v", "creative") | cheat]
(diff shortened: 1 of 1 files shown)


src/Swarm/TUI/View.hs
```diff
@@ -544,12 +544,12 @@ drawKeyMenu s =
     [(NoHighlight, "F1", "help"), (NoHighlight, "F2", "robots")]
       <> notificationKey availableRecipes "F3" "Recipes"
       <> notificationKey availableCommands "F4" "Commands"
-      <> [(NoHighlight, "Tab", "cycle")]
       <> [(NoHighlight, "^v", "creative") | cheat]
       <> [(NoHighlight, "^g", "goal") | goal]
 
   keyCmdsFor (Just REPLPanel) =
     [ ("↓↑", "history")
+    , ("Tab" "Tab Completion")
```
byorgey, line 552: I think we probably don't need tab completion listed in the help bar for the REPL; tab completion is pretty standard and many users will probably just figure it out by trying it.  Also, this seems to be missing comma which is causing compilation to fail.
```suggestion
```

byorgey (changes requested):
Thanks!

byorgey (approved):
Looks good, thanks!  I definitely forgot about this when we merged #501 :smile:

PR #542, opened 2022-07-04
hide `-Wprepositive-qualified-module` behind `ci` flag

### LINKED ISSUE #528: Build fails on prepositive `qualified` in autogen module with `cabal` < 3.6
This is what happens when I do `stack build`:

```
10:40:36 starting   stack test --fast --flag=swarm:ci
Building all executables for `swarm' once. After a successful build of all of them, only specified executables will be rebuilt.
swarm> build (lib + exe + test)
Preprocessing library for swarm-0.1.0.0..
Building library for swarm-0.1.0.0..
[ 1 of 37] Compiling Paths_swarm
            
/home/brent/projects/swarm/.stack-work/dist/x86_64-linux-tinfo6/Cabal-3.4.1.0/build/autogen/Paths_swarm.hs:11:8: error: [-Wprepositive-qualified-module, -Werror=prepositive-qualified-module]
    Found ‘qualified’ in prepositive position
    Suggested fix: place  ‘qualified’ after the module name instead.
   |        
11 | import qualified Control.Exception as Exception
   |        ^^^^^^^^^
Progress 1/2

--  While building package swarm-0.1.0.0 (scroll up to its section to see the error) using:
      /home/brent/.stack/setup-exe-cache/x86_64-linux-tinfo6/Cabal-simple_mPHDZzAJ_3.4.1.0_ghc-9.0.2 --builddir=.stack-work/dist/x86_64-linux-tinfo6/Cabal-3.4.1.0 build lib:swarm exe:swarm test:swarm-integration test:swarm-unit --ghc-options " -fdiagnostics-color=always"
    Process exited with code: ExitFailure 1
10:40:36 exited:    failed: 1
10:40:36 took       0.55s
```

-  It works if I use `cabal-3.6.2.0` to build directly.  
- But I noticed that the above build is using `cabal-3.4.1.0`. If I downgrade to that version and try `cabal build -fci`, it fails with the same error as above.

So I think this is not really a `stack` vs `cabal` thing, it's just that only the latest version of `cabal` works, and `stack` has its own idea about which version of `cabal` it wants to use with specific versions of GHC.

Is there a way we can turn off the `-Wprepositive-qualified-module` only for the autogenerated `Paths_swarm` module?

byorgey:
Closes #528.

Hopefully this is a relatively simple solution.  We will still
warn (and hence fail) on prepositive `qualified` in CI, but it won't
trip up anyone building with `stack` and/or an older version of
`cabal` unless they explicitly use the `ci` flag.

### DIFF
--- a/feedback.yaml
+++ b/feedback.yaml
@@ -1,2 +1,4 @@
+# For use with https://github.com/NorfairKing/feedback
+
 loops:
   test: stack test --fast
--- a/feedback.yaml
+++ b/feedback.yaml
@@ -1,2 +1,2 @@
 loops:
-  test: stack test --fast --flag=swarm:ci
+  test: stack test --fast
--- a/swarm.cabal
+++ b/swarm.cabal
@@ -54,7 +54,8 @@ common stan-config
 
 -- Harmless extensions from GHC2021
 common ghc2021-extensions
-    ghc-options:  -Wprepositive-qualified-module
+    if flag(ci)
+      ghc-options:  -Wprepositive-qualified-module
     default-extensions:
       BangPatterns
       DeriveAnyClass


byorgey:
I don't think there's any way to add that kind of thing in `stack.yaml`.  But in any case I realized I am being silly: the only reason this matters is that I have been building locally with the ci flag just to get `-Werror`.  But it's really not necessary for me to do that.

PR #543, opened 2022-07-04
don't tab complete when there are no identifier chars typed

### LINKED ISSUE #518: Tab completion does not work for operators
This does not work for completing operators, e.g. `=` get completed as `=noop`.

Maybe there should be a check that before the "last word", there is either nothing or a whitespace character?

_Originally posted by @xsebek in https://github.com/swarm-game/swarm/pull/501#discussion_r911333192_

byorgey:
When the prompt was blank, it didn't tab complete; however, when there
were any characters typed at all, it found the "last word" (all
identifier characters at the end) and tried to match on it.  However,
if there were no identifier chars, the resulting empty "last word"
matched everything.

This fix now has a special case to check whether the `lastWord` is
blank, which also takes care of the original special case when the
prompt is completely empty.

Fixes #518.


xsebek (approved):
LGTM! :slightly_smiling_face:

byorgey:
Figured it out, I was just confused which last case you were referring to. :smile:

PR #546, opened 2022-07-04
List system robots when cheating

xsebek:
- show system robots when the `--cheat` flag was used and creative mode is on
- do not limit the distance to 32 when in creative mode

### DIFF
--- a/src/Swarm/TUI/View.hs
+++ b/src/Swarm/TUI/View.hs
@@ -419,11 +419,12 @@ robotsListWidget s = viewport RobotsViewport Vertical (hCenter table)
   isNear robot = creative || distance (realToFrac <$> robot ^. robotLocation) basePos < 32
   robots :: [Robot]
   robots =
-    filter (\robot -> cheat || (isRelevant robot && isNear robot))
+    filter (\robot -> debugging || (isRelevant robot && isNear robot))
       . IM.elems
       $ g ^. robotMap
   creative = g ^. creativeMode
   cheat = s ^. uiState . uiCheatMode
+  debugging = creative && cheat
   g = s ^. gameState
 
 helpWidget :: Widget Name
--- a/src/Swarm/TUI/View.hs
+++ b/src/Swarm/TUI/View.hs
@@ -413,13 +413,13 @@ robotsListWidget s = viewport RobotsViewport Vertical (hCenter table)
 
   basePos :: V2 Double
   basePos = realToFrac <$> fromMaybe (V2 0 0) (g ^? robotMap . ix 0 . robotLocation)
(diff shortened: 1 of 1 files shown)


byorgey (approved):

TristanCacqueray (approved):

src/Swarm/TUI/View.hs
```diff
@@ -415,13 +416,16 @@ robotsListWidget s = viewport RobotsViewport Vertical (hCenter table)
   -- Keep the base and non sytem robot (e.g. no seed)
   isRelevant robot = robot ^. robotID == 0 || not (robot ^. systemRobot)
   -- Keep the robot that are less than 32 unit away from the base
-  isNear robot = distance (realToFrac <$> robot ^. robotLocation) basePos < 32
+  isNear robot = creative || distance (realToFrac <$> robot ^. robotLocation) basePos < 32
```
TristanCacqueray, line 419:
```suggestion
  isCreativeOrNear robot = creative || distance (realToFrac <$> robot ^. robotLocation) basePos < 32
```

xsebek: I would not overcomplicate it - if you are in creative mode then the definition of near simply changes to "anywhere".

src/Swarm/TUI/View.hs
```diff
@@ -415,13 +416,16 @@ robotsListWidget s = viewport RobotsViewport Vertical (hCenter table)
...
   robots :: [Robot]
   robots =
-    filter (\robot -> isRelevant robot && isNear robot)
+    filter (\robot -> cheat || (isRelevant robot && isNear robot))
```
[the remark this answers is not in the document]
xsebek: Done :heavy_check_mark:

src/Swarm/TUI/View.hs
```diff
@@ -419,11 +419,12 @@ robotsListWidget s = viewport RobotsViewport Vertical (hCenter table)
   isNear robot = creative || distance (realToFrac <$> robot ^. robotLocation) basePos < 32
   robots :: [Robot]
   robots =
-    filter (\robot -> cheat || (isRelevant robot && isNear robot))
+    filter (\robot -> debugging || (isRelevant robot && isNear robot))
       . IM.elems
       $ g ^. robotMap
   creative = g ^. creativeMode
   cheat = s ^. uiState . uiCheatMode
+  debugging = creative && cheat
```
byorgey, line 427: I was going to say this should just be `creative`, but after thinking about it a minute I think I understand the point --- you don't necessarily want to see all the system robots whenever you are a normal player playing in creative mode.  `debugging = creative && cheat` makes sense to me.

PR #553, opened 2022-07-10
New `atomic` syntax for blocks to be executed atomically

### LINKED ISSUE #479: `grabif` primitive to provide a lock/semaphore mechanism

Consider the following code:
```
b <- ishere "tree"; if b { grab } {}
```
Seems straightforward: we check if a tree is present, and if so, we grab it.   Of course if we just executed a bare `grab`, it might throw an exception when there's no entity present; so that's why we included the `if`.  In contrast to a bare `grab` command, the above code can never crash... right?

Wrong!  There is a race condition here: if our robot's execution just so happens to be suspended in between evaluating `ishere` and executing the `grab`, another robot could `grab` the tree out from under us in the meantime.  The `grab` would then crash when execution resumes.  (Of course we could use `try` to catch the exception thrown by the `grab`, but that's a more advanced feature, and it doesn't really address the fundamental issue, it can only mitigate its impact.)

I am not sure whether I have encountered this in the wild, but I think I possibly may have.  It's very difficult to tell, of course.  But thinking about possible causes of some unexpected behavior led me to realize this issue.

I would like to have a primitive command `grabif : string -> cmd bool` which **atomically** grabs the named entity if one is present (and does nothing otherwise), returning a `bool` to tell us which happened.  In other words, the test for the entity and the grab happen as one atomic operation which cannot be interrupted.

In addition to resolving the above issue with `grab`, this primitive could be used to implement foolproof locks/semaphores; it is basically the Swarm equivalent of a [compare-and-swap primitive](https://en.wikipedia.org/wiki/Compare-and-swap).   For example, an entity at a specific location in the world can serve as a lock; when a robot needs to execute some critical code, or have exclusive access to some resource, it can first try to acquire the lock by executing `grabif`.  Even if multiple robots try to grab the lock at the same time, exactly one of them will succeed.  Once the robot is done it can simply `place` the item to release the lock.

As another example (this was my original motivating example), `grabif` can also be used to program safe resource acquisition from a "depot".  The depot is programmed to place a certain resource whenever its cell becomes empty.  Other robots wishing to acquire the resource can come to the depot, wait for the resource to be placed, and then grab it.  However, using separate `ishere` and `grab` instructions, the race condition is a problem: if multiple robots come to acquire the resource at the same time, there is a chance that one of them will crash when it tries to grab the resource that it thinks is there but which has actually just been grabbed by the other robot.  If the robots use an atomic `grabif` instruction instead, this problem cannot happen.

I suppose there are other atomic primitives we might consider that could be used to implement similar features, but I have not thought much about this area before.

We might also consider a corresponding `placeif` which atomically does a `place` only if the cell is empty; doing an `ishere` test followed by a `place` suffers from the same race condition as the version with `grab`.  On the other hand, it is possible to implement `placeif` using `grabif`: have an item at location 1 serve as a lock; if a robot wants to attempt to place an item at location 2, it must first acquire the lock using `grabif`, and only then can it attempt to place an item at location 2.

**Additional context**

The only additional question is what capability should be required, and what device(s) might provide it.  It seems natural to say it should require *both* the `sensehere` *and* `grab` capabilities, but I literally [just got through saying](https://github.com/swarm-game/swarm/pull/458/files#r906234923) that we should just make every command require at most one capability. :facepalm:   So I guess I'll put my money where my mouth is and suggest that we make a new device, called, um... a `smart grabber`?  (bikeshedding welcome), with the recipe `smart grabber = scanner + grabber`, that provides the `grabif` capability.

byorgey:
Provide a new `atomic` construct; when a robot is executing `atomic c` it is guaranteed that no other robots will execute any commands while the robot is executing `c`.  Maybe this isn't all that plausible physically, but I guess we can imagine that the base is in communication with all robots and implements some kind of global lock mechanism.

- Closes #479. `grabif`, `harvestif`, and `placeif` can all be implemented simply in terms of `atomic`.  For example:

    ```
    def grabif : string -> cmd bool = \thing.
      atomic (
        b <- ishere thing;
        if b {grab; return true} {return false}
      )
    end
    ```

    ```
    def placeif : string -> cmd bool = \thing.
      atomic (
        res <- scan down;
        if (res == inl ()) {place thing; return true} {return false}
      )
    end
    ```

- The argument to an `atomic` block must have type `cmd a` for some type `a`.
- `atomic` blocks have some restrictions to ensure they can't run for a long time:
    - No `def`, `let`, or lambdas are allowed
    - Variables can only be referenced if they are bound locally or have a "simple type" (= sums and products of base types; no `cmd`, `delay`, or function types)
    - Taken together, these ensure that computations can't be duplicated, and any variables referenced must contain simple data values which have already been evaluated, which means that an argument to `atomic` will have an execution time no more than linear in its length.
- `atomic` blocks also can't be nested, just because that seems weird and I don't know what the purpose would be
- `atomic` blocks may contain at most one "external" command which normally requires a tick to execute.  Allowing *e.g.* multiple `move` commands in an `atomic` block would be problematic: it would either mean that robots could use `atomic` blocks as a way to execute faster than normal, or that we would have to freeze other robots for multiple ticks while some robot executed its atomic block.  But typically, we want to do some kind of sensing command (which is an internal command that doesn't require a tick) + some external command.
- Relatedly, also refactored the notion of external/internal commands so that information is now stored in the `ConstInfo` record.  This means from now on we will be required to declare whether each new constant is internal or external; we hadn't been keeping up with the old definition of `takesTick` at all.
- Add the `rubber band` device which allows doing `atomic`, and make a recipe for it (`rubber` + `strange loop`).
    - This is a fairly simple recipe, but being able to do `atomic` early can be helpful for setting up automatic resource production pipelines.

### DIFF
--- a/data/entities.yaml
+++ b/data/entities.yaml
@@ -882,10 +882,9 @@
 
   - "This seems like a safe way to execute 'grab' only when there is a
     rock to grab.  However, it is actually possible for the 'grab' to
-    fail, if some robot B tries to grab the rock at the same time:
-    first, robot A executes 'ishere' which returns true; next, robot B
-    grabs the rock; finally, robot A tries to execute 'grab', which throws
-    an exception."
+    fail, if some other robot B snatches the rock right after robot A sensed
+    it and before robot A got around to grab it on the next game tick."
+  - This will make robot A very sad and it will crash.
   - "To prevent this situation, robot A can wrap the commands in 'atomic', like so:"
   - |
     atomic (b <- ishere "rock"; if b {grab} {})
--- a/data/entities.yaml
+++ b/data/entities.yaml
@@ -740,7 +740,7 @@
 - name: life support system
   display:
     attr: device
-    char: 'O'
+    char: 'Ж'
   description:
   - A state-of-the-art life support system which maintains the
     particular temperature and mixture of gases you need to survive.
@@ -868,3 +868,18 @@
   description:
   - An I/O cable can be used to communicate with an adjacent robot.
   properties: [portable]
+
+- name: rubber band
+  display:
+    attr: device
+    char: 'O'
+  description:
+  - "A rubber band can tie multiple commands together so that other robots can't execute
+    commands in between them.  It can be used via the 'atomic' command, for example:"
+  - |
+    atomic (b <- ishere "rock"; if b {grab} {})
+  - "Without the 'atomic' wrapper, it would be possible for some other robot to grab the
+    rock in between checking for its presence and grabbing it, causing the grab to fail.
+    'atomic' ensures this cannot happen."
+  properties: [portable]
+  capabilities: [atomic]
--- a/data/entities.yaml
+++ b/data/entities.yaml
@@ -875,11 +875,20 @@
     char: 'O'
   description:
   - "A rubber band can tie multiple commands together so that other robots can't execute
-    commands in between them.  It can be used via the 'atomic' command, for example:"
+    commands in between them.  It can be used via the 'atomic' command. For example, suppose
+    robot A executes the following code:"
+  - |
+    b <- ishere "rock"; if b {grab} {}
+
+  - "This seems like a safe way to execute 'grab' only when there is a
+    rock to grab.  However, it is actually possible for the 'grab' to
+    fail, if some robot B tries to grab the rock at the same time:
+    first, robot A executes 'ishere' which returns true; next, robot B
+    grabs the rock; finally, robot A tries to execute 'grab', which throws
+    an exception."
+  - "To prevent this situation, robot A can wrap the commands in 'atomic', like so:"
   - |
     atomic (b <- ishere "rock"; if b {grab} {})
-  - "Without the 'atomic' wrapper, it would be possible for some other robot to grab the
-    rock in between checking for its presence and grabbing it, causing the grab to fail.
-    'atomic' ensures this cannot happen."
+
   properties: [portable]
   capabilities: [atomic]
--- a/data/scenarios/Testing/479-atomic-race.yaml
+++ b/data/scenarios/Testing/479-atomic-race.yaml
@@ -1,0 +1,76 @@
+# This scenario is specially engineered to almost guarantee a crash
+# due to a race condition in short order.
+#
+#   - The steps per tick is set to 1, so robots are interrupted after
+#     every single step, making bugs due to bad interleaving much more
+#     probable.
+#   - A `weed` entity is used which grows very quickly
+#   - `harvest` is used, which causes more weeds to grow in the same spot
+#   - A crash happens when the planter robot tests for the presence of a weed
+#     and doesn't find one, but then is interrupted and a new weed grows;
+#     then the planter robot crashes when it tries to place a weed.
+#   - The harvesting robot waits a random number of ticks between 1-4 each
+#     time, to give the planting robot time to run into the race condition.
+#
+# This may work slightly less well once we implement #502, but I think it should
+# still work.
+
+name: Grab/harvest/place (with race)
+description: |
+  Make sure we can cause a robot to crash due to a race condition, in order to
+  know that atomic blocks are doing their job.
+  https://github.com/swarm-game/swarm/issues/479
+stepsPerTick: 1
+creative: true
+win: |
+  try {
+    p <- robotNamed "planter";
+    as p {n <- count "weed"; return (n == 0)}
+  } { return false }
+solution: |
+  def forever = \c. force c ; forever c end
+  def tryharvest =
+    b <- ishere "weed";
+    if b {harvest; n <- random 4; wait n} {}
+  end;
+  forever {tryharvest}
+robots:
+  - name: base
+    loc: [0,0]
+    dir: [1,0]
+    devices:
+      - logger
+      - harvester
+    inventory:
+      - [1, lambda]
+  - name: planter
+    loc: [0,0]
+    dir: [1,0]
+    devices:
+      - logger
+      - grabber
+    inventory:
+      - [100, weed]
+    program: |
+      def forever = \c. force c ; forever c end
+      def tryplant =
+        b <- ishere "weed";
+        if b {} {place "weed"}
+      end;
+      forever {tryplant}
+entities:
+  - name: weed
+    display:
+      attr: plant
+      char: 'W'
+    description:
+    - A fast-growing weed.
+    properties: [known, portable, growable]
+    growth: [1,2]
+world:
+  default: [blank, none]
+  palette:
+    '.': [grass, null]
+  upperleft: [0,0]
+  map: |
+    .
--- a/data/scenarios/Testing/479-atomic-race.yaml
+++ b/data/scenarios/Testing/479-atomic-race.yaml
@@ -25,7 +25,7 @@ creative: true
 win: |
   try {
     p <- robotNamed "planter";
-    as p {n <- count "weed"; return (n == 0)}
+    as p {has "Win"}
   } { return false }
 solution: |
   def forever = \c. force c ; forever c end
@@ -57,7 +57,7 @@ robots:
         b <- ishere "weed";
         if b {} {place "weed"}
       end;
-      forever {tryplant}
+      forever {try {tryplant} {create "Win"}}
 entities:
   - name: weed
     display:
@@ -67,6 +67,13 @@ entities:
     - A fast-growing weed.
     properties: [known, portable, growable]
     growth: [1,2]
+  - name: Win
+    display:
+      attr: device
+      char: 'W'
+    description:
+      - You win!
+    properties: [known, portable]
 world:
   default: [blank, none]
   palette:
--- a/src/Swarm/Language/Capability.hs
+++ b/src/Swarm/Language/Capability.hs
@@ -169,6 +169,7 @@ constCaps = \case
   Exp -> Just CArith
   Whoami -> Just CWhoami
   Self -> Just CWhoami
+  Atomic -> Just CAtomic
   -- Some God-like abilities.
   As -> Just CGod
   RobotNamed -> Just CGod
--- a/src/Swarm/Language/Capability.hs
+++ b/src/Swarm/Language/Capability.hs
@@ -16,6 +16,7 @@ module Swarm.Language.Capability (
   constCaps,
 ) where
 
+import Data.Aeson (FromJSONKey, ToJSONKey)
 import Data.Char (toLower)
 import Data.Hashable (Hashable)
 import Data.Text (Text)
@@ -108,7 +109,7 @@ data Capability
     --   checking challenge mode win conditions, and not for use by
     --   players.
     CGod
-  deriving (Eq, Ord, Show, Read, Enum, Bounded, Generic, Hashable, Data)
+  deriving (Eq, Ord, Show, Read, Enum, Bounded, Generic, Hashable, Data, FromJSONKey, ToJSONKey)
 
 capabilityName :: Capability -> Text
 capabilityName = from @String . map toLower . drop 1 . show
--- a/src/Swarm/Language/Capability.hs
+++ b/src/Swarm/Language/Capability.hs
@@ -102,6 +102,9 @@ data Capability
     CSetname
   | -- | Capability to move unrestricted to any place
     CTeleport
+  | -- | Capability to run commands atomically
+    -- XXX need to make device that provides this capability
+    CAtomic
   | -- | God-like capabilities.  For e.g. commands intended only for
     --   checking challenge mode win conditions, and not for use by
     --   players.
--- a/src/Swarm/Language/Capability.hs
+++ b/src/Swarm/Language/Capability.hs
@@ -103,7 +103,6 @@ data Capability
   | -- | Capability to move unrestricted to any place
     CTeleport
   | -- | Capability to run commands atomically
-    -- XXX need to make device that provides this capability
     CAtomic
   | -- | God-like capabilities.  For e.g. commands intended only for
     --   checking challenge mode win conditions, and not for use by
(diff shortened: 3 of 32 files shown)


src/Swarm/Language/Capability.hs
```diff
@@ -102,6 +102,9 @@ data Capability
     CSetname
   | -- | Capability to move unrestricted to any place
     CTeleport
+  | -- | Capability to run commands atomically
+    -- XXX need to make device that provides this capability
```
byorgey, line 106:
```suggestion
```

src/Swarm/Language/Parse.hs
```diff
@@ -98,6 +98,7 @@ reservedWords =
        , "false"
        , "forall"
        , "require"
+       , "atomic"
```
byorgey, line 101: Need to update the editor support/ syntax highlighting

byorgey:
Something I just thought of: I made `atomic` a special syntactic form, rather than a constant, because of the need to specially check its argument.  But actually, it's worth looking into whether we can make it a regular constant but still have a special case for it in the typechecker.  I think that could actually work in this case,  and it would simplify a lot of things.

src/Swarm/Language/Syntax.hs
```diff
@@ -403,6 +406,15 @@ data MUnAssoc
     S
   deriving (Eq, Ord, Show)
 
+-- | Whether a command is internal or external commands.  External
```
byorgey, line 409:
```suggestion
-- | Whether a command is internal or external.  External
```

TristanCacqueray:
Shouldn't the commit message and/or the documentation indicate what happens in a `atomic` block? Maybe I missed it, but I'm not sure what it does, e.g. does it prevent other robots from using external command on the same tile?

data/scenarios/Testing/479-atomic-race.yaml
```diff
@@ -0,0 +1,76 @@
...
+  Make sure we can cause a robot to crash due to a race condition, in order to
+  know that atomic blocks are doing their job.
+  https://github.com/swarm-game/swarm/issues/479
+stepsPerTick: 1
+creative: true
+win: |
+  try {
+    p <- robotNamed "planter";
+    as p {n <- count "weed"; return (n == 0)}
+  } { return false }
```
xsebek, line 29: I find testing this later with `expectedFail` confusing since we otherwise use it for bugs that need fixing.

I would suggest winning by creating the Win entity when harvesting fails.

Something like `s/tryharvest/try {raceharvest} {create "Win"}/g`.

src/Swarm/Language/Syntax.hs
```diff
@@ -404,6 +412,15 @@ data MUnAssoc
     S
   deriving (Eq, Ord, Show)
 
+-- | Whether a command is internal or external.  External
+--   commands have some kind of effect on the external world; at most
+--   one external command can be executed per tick.  Internal commands
+--   are things like sensing commands, or commands that solely modify
+--   a robot's internal state; multiple internal commands may be
+--   executed per tick.
+data Facing = Internal | External
```
xsebek, line 421: I would suggest _silent_ vs _observable_ visibility (found it on the [bisimulation Wiki page](https://en.wikipedia.org/wiki/Bisimulation#Variants_of_bisimulation)) so that it's not as easy to confuse with other things we keep internal like the `CGod` commands.

src/Swarm/Language/Parse.hs
xsebek: Seems like you already did that? :slightly_smiling_face:

data/entities.yaml
```diff
@@ -868,3 +868,27 @@
...
+    robot A executes the following code:"
+  - |
+    b <- ishere "rock"; if b {grab} {}
+
+  - "This seems like a safe way to execute 'grab' only when there is a
+    rock to grab.  However, it is actually possible for the 'grab' to
+    fail, if some robot B tries to grab the rock at the same time:
+    first, robot A executes 'ishere' which returns true; next, robot B
+    grabs the rock; finally, robot A tries to execute 'grab', which throws
+    an exception."
```
xsebek, line 888: I think this is overly detailed and something funny would be more appropriate.
```suggestion
    fail, if some other robot B snatches the rock right after robot A sensed
    it and before robot A got around to grab it on the next game tick."
  - This will make robot A very sad and it will crash.
```

data/entities.yaml
```diff
@@ -740,7 +740,7 @@
 - name: life support system
   display:
     attr: device
-    char: 'O'
+    char: 'Ж'
```
xsebek, line 743: What does that device look like? If it's some space suite, I might suggest [Be `Б`](https://en.wikipedia.org/wiki/Be_(Cyrillic)) or [Lje `Љ`](https://en.wikipedia.org/wiki/Lje).

Still, any Cyrillic letter is cool. :smile:

xsebek:
> atomic blocks also can't be nested, just because that seems weird and I don't know what the purpose would be

You could use it to make handcoded `register` function faster:
```
\i. atomic (
  if (i == n) {
    return rob
  } {
    atomic (if (i == n-1) ...) i
  }
) 
```
I think this would give you a sort of $O(1)$ access list, which sounds a bit like cheating. :sweat_smile:

src/Swarm/Language/Parse.hs
byorgey: Indeed!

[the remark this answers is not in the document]
byorgey: Good catch! :heavy_check_mark:

data/scenarios/Testing/479-atomic-race.yaml
byorgey: Oh, good point, and good suggestion!  It's actually the planter robot that is expected to crash but yes, a `try {...} {create "Win"}` does the trick just fine.

src/Swarm/Language/Syntax.hs
byorgey: Hmm, I agree internal/external is confusing.  Silent/observable don't really sound like opposites to me, so I consulted a thesaurus and went with "tangible/intangible" --- how does that sound to you?

data/entities.yaml
byorgey: I wasn't thinking of it as a space suit, more like an air circulation + filtration system that keeps the temperature + atmosphere inside your base within acceptable levels.

A `toolkit` is already Yat `Ѣ` which looks too similar to Be I guess.  Lje could be good.  Honestly my choice of Zhe `Ж` was pretty arbitrary.  I just recently realized that we are kind of running out of interesting ASCII characters and most fonts can probably  be counted on to have Greek and Cyrillic characters.  Maybe it looks kinda like a thing with a bunch of tubes coming out, for air + water + etc??

byorgey: Obviously the reason I changed it is that `O` is perfect for a rubber band.

src/Swarm/Language/Typecheck.hs
```diff
@@ -508,3 +539,117 @@ check t ty = do
...
+
+-- | Analyze an argument to @atomic@: ensure it contains no nested
+--   atomic blocks and no references to external variables, and count
+--   how many tangible commands it will execute.
+analyzeAtomic :: Set Var -> Syntax -> Infer Int
+analyzeAtomic locals (Syntax l t) = case t of
+  TUnit {} -> return 0
+  -- No nested 'atomic' allowed!
+  TConst Atomic -> throwError (InvalidAtomic l NestedAtomic t)
+  TConst c -> return $ if isTangible c then 1 else 0
```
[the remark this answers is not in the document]
byorgey: Oh, great catch! We should disallow `drill` too.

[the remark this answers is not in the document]
byorgey: Maybe it would be better to have a whitelist of allowed constants?  Otherwise I just know we are going to add a command in the future that takes multiple ticks and forget to check for it here.

src/Swarm/Language/Typecheck.hs
```diff
@@ -508,3 +541,195 @@ check t ty = do
...
+  -- We should never encounter a TRef since they do not show up in
+  -- surface syntax, only as values while evaluating (*after*
+  -- typechecking).
+  TRef {} -> throwError (CantInfer l t)
+
+-- | Constants which are allowed inside an @atomic@ block.  This is
+--   intentionally expressed as a whitelist, even though it's much
+--   longer this way, so that we can't create problems in the future
+--   by adding a constant that is problematic to use with @atomic@ but
+--   forgetting to disallow it.
```
xsebek, line 646: I am not sure if there is an ergonomic way to do that, but would it not be better to include this in const info?

This is how I understand it:
- functions take 0 ticks
- intangible commands take 1 tick normally but do not count in atomic
- tangible commands take 1 tick normally and there can be at most one in atomic (if branches are an exception)
  - some of those commands can take a higher number of ticks and should _not_ be in atomic at all

Sounds to me like the safest option would be to add a parameter to `Tangible` that would tell us if the command takes just one tick or can take more. A boolean would work, but maybe a `data Length = Short | Long` would be more descriptive, I am definitely open to other name ideas.

Sorry to nitpick on this so much @byorgey, in case you are tired of this atomic business, we can merge as is and improve the ergonomics later :wink:

src/Swarm/Language/Syntax.hs
```diff
@@ -562,28 +585,57 @@ constInfo c = case c of
...
-  As -> commandLow 2 "Hypothetically run a command as if you were another robot."
-  RobotNamed -> commandLow 1 "Find a robot by name."
-  RobotNumbered -> commandLow 1 "Find a robot by number."
-  Knows -> commandLow 1 "Check if the robot knows about an entity."
+  Atomic ->
+    command 1 Intangible . doc "Execute a block of commands atomically." $
+      [ "When executing `atomic c`, a robot will not be interrupted, that is, no other robots will execute any commands while the robot is executing @c@."
+      ]
+  Teleport -> command 2 Tangible "Teleport a robot to the given location."
+  As -> command 2 Tangible "Hypothetically run a command as if you were another robot."
```
xsebek, line 601: Should `As` be intangible too? It's only hypothetical.

src/Swarm/Language/Typecheck.hs
```diff
@@ -508,3 +541,195 @@ check t ty = do
...
+  TString {} -> return 0
+  TAntiString {} -> return 0
+  TBool {} -> return 0
+  TRobot {} -> return 0
+  TRequireDevice {} -> return 0
+  TRequire {} -> return 0
+  SPair s1 s2 -> (+) <$> analyzeAtomic locals s1 <*> analyzeAtomic locals s2
+  -- Special case for if: number of tangible commands is *max* of branches
+  -- instead of sum, since exactly one of them will be executed
+  SApp (STerm (SApp (STerm (SApp (STerm (TConst If)) tst)) thn)) els ->
```
xsebek, line 598: `As` should be another exception, it does not really take an extra game time and its arguments should not be checked.

E.g. it took me a minute to go 600 steps with the base until I reached a mountain, but `As` had it in less than a tick.

It is a big responsibility to not freeze the game, but god-like powers should not be limited by mortal concerns. :innocent: 

---

P.S. Do we not have some pattern for binary operators/functions to make this pattern matching shorter?

xsebek (approved):
LGTM! :+1: Just a few notes (that we could fix later) below.

src/Swarm/Language/Typecheck.hs
byorgey: @xsebek don't apologize, I'm glad for all the nitpicking!  And you're absolutely right, this should be included in the const info, I don't know why that didn't occur to me. :facepalm:   Just to clarify:

* All functions and binary operators are intangible.
* Intangible commands do not take up a tick.  That is, a robot can execute any number of intangible commands in one tick (up to the limit of how many CESK machine steps may be taken per tick).
* Tangible commands take up a tick, that is, a robot can execute at most one of them per tick.
* Since we implement `atomic` by suspending the normal mechanisms for limiting the number of steps per tick, we need to make sure that `atomic` still executes at most one tangible command, and moreover does not take multiple ticks, since otherwise it can cause the game to freeze and/or a robot to do more than it should be able to do in one tick. Hence:
    - An `atomic` block can contain at most one tangible comand
    - Commands that can take multiple ticks cannot be allowed in `atomic` at all

I like the idea of adding `data Length = Short | Long` as an argument to `Tangible`.

src/Swarm/Language/Syntax.hs
byorgey: That definitely seems to make sense, but if I make it intangible, several tests start failing.  See https://github.com/swarm-game/swarm/pull/553/commits/6afc232c6dddeb97a3b5b9f444227f062bca0c9a .  I should look into this more and at least try to understand and document why it needs to be tangible.

src/Swarm/Language/Typecheck.hs
byorgey: > As should be another exception, it does not really take an extra game time and its arguments should not be checked.

That's... true, I suppose, although it does take *computational* time even if it doesn't take ticks in the game.  And I am not sure why you would need to use `as` inside an `atomic` block.  I am more inclined to leave it as is and revisit this if someone has a compelling use case for a special case later.

byorgey: > Do we not have some pattern for binary operators/functions to make this pattern matching shorter?

We do now! :smile: :heavy_check_mark:

src/Swarm/Language/Syntax.hs
byorgey: @xsebek `as` is now intangible, as I agree it should be.  That caused two tests to fail which were apparently inadvertently depending on the fact that `as` took a tick.  I figured out one of them, but for `Tutorials/crash` I still have no idea.  I was able to "fix" it by adding a manual `wait 1` after the `as` command, but I don't understand why!  Since you wrote that scenario, do you think you could take a look?  It doesn't matter in some sense but I'd like to understand what was going on.

PR #554, opened 2022-07-11
Set `replStatus` to `REPLWorking` when running a file via `--run`

### LINKED ISSUE #333: Definitions loaded via `--run` have their memory cells trashed
**Describe the bug**
Any `def`s in a file loaded via the `--run` command-line argument (#312) will not work since their corresponding memory cell gets thrown away.

**To Reproduce**
1. In `test.sw`:
    ```
    def m = move end
    ```
2. Start the game via `swarm --run test.sw`.
3. Type `m` at the REPL.

This should generate an error about not having the right devices to move; instead, it generates `fatal error: Reference to unknown memory cell 0`.

**Additional context**

I think I already know why this is happening, but I am filing the issue just to be able to properly track and discuss it.  This bug apparently has existed ever since #312 was merged, which I guess was insufficiently tested.  It can ultimately be traced back to strangeness with typechecking `run`, and points toward the need to do something like https://github.com/swarm-game/swarm/issues/208#issuecomment-1072492695 .

When the `--run` flag is used, we call `initMachine` to create a CESK machine that will execute the relevant `run` command.  However, `initMachine` looks at the type *and generated environment* of the given term to decide what stack frames the CESK machine should start with.  Since `run(foo)` has type `cmd ()` but has an *empty environment*, `initMachine` starts the stack with only an `FExec` frame but no `FLoadEnv` frame.  That means any definitions in the file being run will have the corresponding store thrown away.  If the type of `run` properly took into account the definitions that will be generated when it is executed, this wouldn't be an issue.

byorgey:
Fixes #333.

### DIFF
--- a/src/Swarm/Game/State.hs
+++ b/src/Swarm/Game/State.hs
@@ -617,7 +617,9 @@ scenarioToGameState scenario userSeed toRun g = do
       , _viewCenterRule = VCRobot baseID
       , _viewCenter = V2 0 0
       , _needsRedraw = False
-      , _replStatus = REPLDone
+      , _replStatus = case toRun of
+          Nothing -> REPLDone
+          Just _ -> REPLWorking (Forall [] (TyCmd TyUnit)) Nothing
       , _messageQueue = []
       , _focusedRobotID = baseID
       , _ticks = 0
--- a/src/Swarm/Game/State.hs
+++ b/src/Swarm/Game/State.hs
@@ -617,7 +617,9 @@ scenarioToGameState scenario userSeed toRun g = do
       , _viewCenterRule = VCRobot baseID
       , _viewCenter = V2 0 0
       , _needsRedraw = False
-      , _replStatus = case toRun of
+      , -- When starting base with the run flag, REPL status must be set to working,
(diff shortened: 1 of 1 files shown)


src/Swarm/Game/State.hs
```diff
@@ -617,7 +617,9 @@ scenarioToGameState scenario userSeed toRun g = do
       , _viewCenterRule = VCRobot baseID
       , _viewCenter = V2 0 0
       , _needsRedraw = False
-      , _replStatus = REPLDone
+      , _replStatus = case toRun of
```
xsebek, line 620: Maybe a short explanation would help here? This is how I understand it:
```suggestion
      , -- When starting base with the run flag, REPL status must be set to working,
        -- otherwise the store of definition cells is not saved (see #333)
        _replStatus = case toRun of
```

xsebek (approved):
Works for me, good job debugging this @byorgey! :+1:

src/Swarm/Game/State.hs
byorgey: Nice idea, thanks!

Issue #540, opened 2022-07-03
Definition capabilities are not available later in the same term


byorgey:
**Describe the bug**
When making a definition, the definition can be used later in the same term (separated by semicolon), but it is not in the map that records the capabilities/requirements of definitions, so a `build` using it may not give the proper devices.

**To Reproduce**
Start a new classic mode game, and execute
```
def m = move end; build {m}
```
You will see that the robot is built but does not move.  To see why this happens, quit and start a new game with `--cheat`.  First, switch to creative and `create "logger"`, then switch back.  Now execute
```
def m = move end; r <- build {log "hi"; m}; view r
```
We can see that the built robot does not have any `treads`, and failed with a runtime capability check saying that it does not have the devices necessary to `move`.  The problem is that at the time the `build` is executed, the capability map does not have an entry for `m`, so capability analysis on the program `log "hi"; m` concludes that it needs only a `logger` and a `solar panel`.

Note that `let` seems to have the same issue.  For example, we get the same behavior using something like `let m = move in build {m}`.

**Expected behavior**
Anything defined with `def` should be fully available (including capability analysis) immediately after the definition.

**Additional context**
Discovered this while preparing #533.  This is especially a problem in test scenarios, where a solution program to be run on the base must consist of a single term (as opposed to interactive use, where many separate programs are executed in sequence).  Any test scenario which wishes to use a `def` as part of its solution program will likely not work at the moment.

Also, none of this will be an issue once we (eventually) implement some kind of effect type system with built-in capability tracking, as in #231 .

byorgey:
My current idea for fixing this, after thinking through #394, is simply to add capabilities to environments.  This is actually quite logical: an environment is a mapping from names to their values at runtime; since we (currently) need capability information *at runtime*, environments should carry that information too.  Practically speaking, this will fix the issue since we already have machinery in place for updating environments properly with the results of definitions.

PR #556, opened 2022-07-12
Refactor `Display` code and add invisibility

### LINKED ISSUE #421: Invisibility
Robots should have a flag specifying whether they are visible or invisible.  Potential applications include:

- It could be fun to have a (powerful, difficult to construct) "cloaking device" with which a robot can make itself invisible.  This would be merely a curiosity at this point, but could become more interesting/important if we ever add some kind of networked PvP mode in the future.
- More practically, invisible system robots could impart some sort of global dynamicism or interactivity to the world without having to appear in a specific location.  As a simple example, suppose we wanted an additional entity (say, a snake) to spawn in a specific location every 10000 ticks.  We could effect this by having an invisible system robot at the location, programmed to `build` a new snake every 10000 ticks.

Invisible robots obviously should not be rendered, but also should not be detectable by #306 .

### LINKED ISSUE #555: `teleport` does not update `robotsByLocation` map
The `robotsByLocation` map in the `GameState` is supposed to be kept up to date with the robots at each location.  However, when `teleport` is applied to a robot other than `self`, it does not get updated.  This was not apparent since we were not using the `robotsByLocation` map to draw the world.  However, other commands which do use `robotsByLocation`, such as `salvage`, will not work correctly if you try to apply them to a robot which has just been teleported.

I noticed this when I switched the world drawing to make use of `robotsByLocation`, and teleported robots did not get drawn correctly.  I am already fixing this as part of my work in another PR (and in such a way that it will be much harder to make the same mistake again), but just wanted to file this issue so we have a place to discuss it and refer to it in the future.

byorgey:
- Add invisibility (closes #421).  Every `Display` has an `invisible` flag, so entities can be invisible too.
- Refactor the world drawing code to simply collect and compose all the `Display`s for everything at each location.  The world drawing code is much simpler now.
- Along the way, I also found and fixed #555 : `teleport`, when applied to another robot, did not properly update the `robotsByLocation` map.  I've refactored things so it should be much harder to make this error in the future.

### DIFF
--- a/src/Swarm/Game/Display.hs
+++ b/src/Swarm/Game/Display.hs
@@ -23,6 +23,7 @@ module Swarm.Game.Display (
   curOrientation,
   displayAttr,
   displayPriority,
+  invisible,
 
   -- ** Rendering
   displayChar,
@@ -63,11 +64,15 @@ data Display = Display
   , _curOrientation :: Maybe Direction
   , _displayAttr :: AttrName
   , _displayPriority :: Priority
+  , _invisible :: Bool
   }
   deriving (Eq, Ord, Show, Generic, Hashable)
 
 instance Semigroup Display where
-  (<>) = maxOn _displayPriority
+  d1 <> d2
+    | _invisible d1 = d2
+    | _invisible d2 = d1
+    | otherwise = maxOn _displayPriority d1 d2
 
 makeLensesWith (lensRules & generateSignatures .~ False) ''Display
 
@@ -91,14 +96,18 @@ displayAttr :: Lens' Display AttrName
 --   on top of lower.
 displayPriority :: Lens' Display Priority
 
+-- | Whether the entity is currently invisible.
+invisible :: Lens' Display Bool
+
 instance FromJSON Display where
   parseJSON = withObject "Display" $ \v ->
     Display
-      <$> v .: "char"
+      <$> v .:? "char" .!= ' '
       <*> v .:? "orientationMap" .!= M.empty
       <*> v .:? "curOrientation"
       <*> v .:? "attr" .!= entityAttr
       <*> v .:? "priority" .!= 1
+      <*> v .:? "invisible" .!= False
 
 instance ToJSON Display where
   toJSON d =
@@ -108,6 +117,7 @@ instance ToJSON Display where
       , "priority" .= (d ^. displayPriority)
       ]
         ++ ["orientationMap" .= (d ^. orientationMap) | not (M.null (d ^. orientationMap))]
+        ++ ["invisible" .= (d ^. invisible) | d ^. invisible]
 
 -- | Look up the character that should be used for a display.
 displayChar :: Display -> Char
@@ -142,6 +152,7 @@ defaultEntityDisplay c =
     , _curOrientation = Nothing
     , _displayAttr = entityAttr
     , _displayPriority = 1
+    , _invisible = False
     }
 
 -- | Construct a default robot display for a given orientation, with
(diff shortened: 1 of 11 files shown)


src/Swarm/Game/Display.hs
```diff
@@ -76,20 +85,29 @@ defaultChar :: Lens' Display Char
 --   the 'defaultChar' will be used.
 orientationMap :: Lens' Display (Map Direction Char)
 
+-- | The display caches the current orientation of the entity, so we
+--   know which character to use from the orientation map.
+curOrientation :: Lens' Display (Maybe Direction)
```
xsebek, line 90: I am getting very lost here. Do all entities have an orientation map just because robots need it?

And now we need to put here the robot direction because we are at display level?

byorgey:
The tests aren't passing because of #540, and I can't think of a good workaround.  I think I will go fix #540 and then rebase this PR on top once we merge that fix.

xsebek:
@byorgey maybe we could move `meet` command ([f6eaad](https://github.com/swarm-game/swarm/pull/556/commits/f6eaad480d0463c5cb42aa897b5b0defea59f196)) into separate PR?

I would not mind merging the refactoring parts with tests marked as `xfail`. :slightly_smiling_face:

src/Swarm/Game/Display.hs
xsebek: @byorgey I wonder if that is a good responsibility division - perhaps there is a better abstraction/wrapper for displayable things that have orientation, and their drawing depends on it. :thinking: 

I am hoping things would get simpler if robots (or later other entities) set the drawn character from their direction map when they are themselves changing their orientation, keeping the source of truth in the robot (or entity).

Caching seems like overengineering to me - I am worried it would just cause bugs.

byorgey:
Sure, I will move `meet` and the invisible maze challenge into separate PRs.  Merging this with an expected fail test is probably best, especially since I spent some time thinking about #540 and it is not a simple fix.

src/Swarm/Game/Display.hs
byorgey: These are good concerns.  To be honest I'm not super happy with this either but it was the best I could come up with.  I'll try to give it some more thought.

If we have robots/entities set the drawn character from their direction map, that is still caching something --- it's just caching a character instead of a `Direction`.  I don't really see how that would be an improvement.  Notice how I changed the `robotDisplay` to be a special lens that automatically updates the cached direction when returning the `Display`.  So there shouldn't be any way to get a `Display` that is out of sync.  Well, I mean, there is always a way, but it seems like you would have to try pretty hard to turn it into a bug.

I think the direction map should remain part of the `Display` --- it is part of the description of how an entity should be drawn on the screen.

byorgey: Hmm, but maybe it would indeed be better to store the current character in the `Display` instead of a `Direction`.  Then robots and entities could do things like e.g. temporarily change their display by setting the current character to something else?

byorgey:
OK, I have taken `meet` and the invisible maze challenge out of this PR.  The invisible maze challenge solution was actually the one that ran into #540 so everything here is fine.  I still can't think of any better way to organize the `Display` + orientation stuff but I'm open to ideas.

xsebek (approved):
@byorgey If you want we can merge this* as is :slightly_smiling_face: 

<sup>*) As discussed above I think that `Display` should not know about directions, but changing that is beyond refactoring. :sweat_smile: </sup>

Also thanks for making the portals in the teleport room truly invisible. :+1:

byorgey:
Thanks.  I definitely agree `Display` could be improved, I just haven't been able to figure out something better than what we have.  Sure, let's merge as is, and I created #570 to keep track of the technical debt.

Issue #561, opened 2022-07-18
Update to new `brick` API


jtdaugherty:
I'm working on an overhaul of `brick`'s `EventM` API and I'd love to get your feedback on how it might impact `swarm`. Please see the details and post comments on https://github.com/jtdaugherty/brick/issues/379. Thanks!

xsebek:
Thanks, @jtdaugherty, that looks like it could simplify things a bit in our `Controller.hs`. :slightly_smiling_face: 

There we use the `S -> EventM Name (Next S)` functions a lot, so changing it to `EventM Name S ()` would hopefully make the code look more like the functions in `Step.hs` where we use lenses for state updates.

I can test your branch with swarm this week unless @byorgey is faster. :wink:

xsebek:
I started working on this in the [brick-event](https://github.com/swarm-game/swarm/tree/brick-event) branch. :construction: 

The controller code will require some shuffling of parameters, but I look forward to replacing `StateT AppState (EventM Name) ()` with `EventM Name AppState ()` :+1:

jtdaugherty:
Nice!

jtdaugherty:
Hi, folks - Brick 1.0 is out. Here are the details: https://github.com/jtdaugherty/brick/blob/master/CHANGELOG.md#10

Note that 1.0 also includes the custom keybinding API to help you with #523. Let me know how that turns out!

Issue #340, opened 2022-05-25
Update to latest version of brick


byorgey:
In particular, it looks like newer versions of brick support scrollbars in viewports, which could be very nice for us.  In addition, if we move towards adding mouse support, (e.g. see the discussion at #203 ) there are apparently some bug fixes related to mouse handling in recent releases.

Issue #94, opened 2021-09-27
Inter-robot communication


byorgey:
This may be still a ways off, but eventually we certainly want to be able to have inter-robot communication.  This would enable lots of cool things such as remotely controlling or reprogramming robots, creating teams of robots that can dynamically coordinate among one another, etc. There is some discussion of this in https://github.com/byorgey/swarm/issues/17#issuecomment-927669308 .

There is a lot of language design work to be done here.  My initial vague thought to look for inspiration to things like pi-calculus and session types.  Probably full-on session types are too heavy-handed, though.  Maybe something simple like typed broadcast channels: a value of type `chan t` is a channel over which values of type `t` can be broadcast, supporting operations like
```
newChannel : cmd (chan t)
broadcast : forall t. chan t -> t -> cmd ()
read : forall t. chan t -> cmd t
readNonBlocking : forall t. chan t -> cmd (maybe t)
```
Channels will be private in the sense that only robots with a reference to a channel can use it.  There would have to be some special support for `newChannel` similar to `def`, so a robot can declare a new channel and then have it available globally to all robots it built after that.

There are also game design issues to think about; e.g. what sort(s) of device(s) should be needed to use these features?

byorgey:
I'm upgrading this from `Nice to have` to `Moderate` --- it seems to keep coming up!  I think this is probably one of the first things we should tackle after the alpha release.

byorgey:
We should also put a bit of thought into propagation delay: the time it takes for a `broadcast` message to reach a listener can be proportional to the distance.

xsebek:
> I'm upgrading this from Nice to have to Moderate --- it seems to keep coming up! I think this is probably one of the first things we should tackle after the alpha release.

@byorgey shall I pin this then? :slightly_smiling_face:

xsebek:
> We should also put a bit of thought into propagation delay: the time it takes for a broadcast message to reach a listener can be proportional to the distance.

I just realized that if you send a robot to walk away, then we would have to store all messages that the robot could possibly have access to by the time they reach it. That could easily be more than 10000 if you have a lot of robots broadcasting a lot of messages.

byorgey:
Yeah, that gets complicated.  For the sake of argument, what if we just limit the broadcast radius but have instantaneous propagation within that radius?  And for I/O cables what if we just allow instantaneous transmission for any distance?  That is "un-physical", of course, but is there any way it would mess up the game or make things inconsistent?

xsebek:
[...]

PR #565, opened 2022-07-18
Improve say, add listen command and F5 messages

### LINKED ISSUE #512: Check errors in win condition
If an exception occurs in the win condition, we do not test it:
```Haskell
-- Execute the win condition check *hypothetically*: i.e. in a
-- fresh CESK machine, using a copy of the current game state.
v <- runThrow @Exn . evalState @GameState g $ evalPT t
case v of
  Left _exn -> return () -- TODO: Check errors in win condition
  Right (VBool True) -> winCondition .= Won False
  _ -> return ()
```
Possible solutions:
- keep the robot around and add the error message to its log
- put the error message to the message queue (used by `say`) and check that
- add an option for tests to cause this to throw an IO error

### LINKED ISSUE #513: Revive message queue
From #26:
> I was thinking `say` would become useful as a global God log (we could read `messageQueue` in some pop-up creative window) and check in the tests (no log required). Maybe other robots could hear you too if they are nearby and `listen`.

- add an <kbd>F7</kbd> window where the message queue would be shown in cheating/creative mode
- consider using `LogEntry` instead of raw `Text`, which could help with message formatting:
    ```Haskell
    Say -> case vs of
      [VString msg] -> do
        rn <- use robotName -- TODO: use robot name + ID
        emitMessage (T.concat [rn, ": ", msg])
    ```
- bonus: add a `listen: cmd (robot * string)` command that would allow you to hear other robots `say` something

### LINKED ISSUE #577: Add listen command
Once #565 is merged it would be nice to also allow robots to hear each other and react programmatically.

## Proposed solution
- add the command `listen: cmd string`
  - it will get the first message said by any robot within a radius of 32 cells
  - if there are more messages, take the closest one
  - if there are still more messages, take one at random or jumble them together if you are feeling evil :smiling_imp:
- when a message is obtained, add it to the robot log (and filter message notifications to unique)
  - maybe drop the `Bool` from message notifications and add a field if the message was logged or said to the `LogEntry`?
- create a hearing capability provided by some "listening pipe" (pipe + copper wire + paper?) or something better

## Alternatives
- the command could take an `int` timeout parameter or we could have separate `timeout` command
- the command could fold over messages
- the command could wait for only one tick and have return type `() + string`

xsebek:
- use `LogEntry` for said messages
- add fields to log entries for sorting
- add a messages panel that also shows logs (logger required)
  - colour messages (also in logger)
    - logs bold, message colour per robot and exceptions red
  - show message time
- test notifications for new messages
- add `listen` command and capability
  - add a hearing aid device that provides the capability
- move the F-key modal hints to the top left world corner
- close #513 
- close #512
- close #577

### DIFF
--- a/data/entities.yaml
+++ b/data/entities.yaml
@@ -388,7 +388,7 @@
 
 - name: rubber
   display:
-    attr: stone
+    attr: rubber
     char: '%'
   description:
   - A flexible, durable material made from LaTeX.
--- a/data/entities.yaml
+++ b/data/entities.yaml
@@ -816,6 +816,22 @@
   properties: [portable]
   capabilities: [log]
 
+- name: hearing aid
+  display:
+    attr: device
+    char: '@'
+  description:
+  - "Allows robot to hear what nearby robots are saying."
+  - "Simply having this device installed will automatically
+     add messages said by nearby robots to this robots log,
+     assuming it has a logger installed."
+  - "That way you can view any heard message later either in
+     the logger or the message window."
+  - "To wait for a message and get the string value, use:"
+  - "`l <- listen; log $ \"I have waited for someone to say \" ++ l`"
+  properties: [portable]
+  capabilities: [log]
+
 - name: counter
   display:
     attr: device
--- a/data/entities.yaml
+++ b/data/entities.yaml
@@ -830,7 +830,7 @@
   - "To wait for a message and get the string value, use:"
   - "`l <- listen; log $ \"I have waited for someone to say \" ++ l`"
   properties: [portable]
-  capabilities: [log]
+  capabilities: [listen]
 
 - name: counter
   display:
--- a/data/entities.yaml
+++ b/data/entities.yaml
@@ -821,9 +821,9 @@
     attr: device
     char: '@'
   description:
-  - "Allows robot to hear what nearby robots are saying."
+  - "Allows a robot to hear what nearby robots are saying."
   - "Simply having this device installed will automatically
-     add messages said by nearby robots to this robots log,
+     add messages said by nearby robots to this robot's log,
      assuming it has a logger installed."
   - "That way you can view any heard message later either in
      the logger or the message window."
--- a/data/recipes.yaml
+++ b/data/recipes.yaml
@@ -478,6 +478,15 @@
   out:
   - [1, strange loop]
 
+- in:
+  - [2, copper pipe]
+  - [4, copper wire]
+  - [1, paper]
+  - [1, lodestone]
+  out:
+  - [2, hearing aid]
+
+
 #########################################
 ##           QUARTZ + SILICON          ##
 #########################################
--- a/src/Swarm/Game/State.hs
+++ b/src/Swarm/Game/State.hs
@@ -488,16 +488,16 @@ messageNotifications = to getNotif
     -- classic players only get to see messages that they said and a one message that they just heard
     -- other they have to get from log
     latestMsg = messageIsRecent gs
-    closeMsg = messageIsFromNearby gs
+    closeMsg = messageIsFromNearby (gs ^. viewCenter)
     focusedOrLatestClose mq =
       (Seq.take 1 . Seq.reverse . Seq.filter closeMsg $ Seq.takeWhileR latestMsg mq)
         <> Seq.filter ((== gs ^. focusedRobotID) . view leRobotID) mq
 
 messageIsRecent :: GameState -> LogEntry -> Bool
-messageIsRecent gs e = e ^. leTime == (gs ^. ticks)
+messageIsRecent gs e = e ^. leTime >= gs ^. ticks - 1
 
-messageIsFromNearby :: GameState -> LogEntry -> Bool
-messageIsFromNearby gs e = maybe False (\r -> manhattan (r ^. robotLocation) (e ^. leLocation) <= hearingDistance) (focusedRobot gs)
+messageIsFromNearby :: V2 Int64 -> LogEntry -> Bool
+messageIsFromNearby l e = manhattan l (e ^. leLocation) <= hearingDistance
 
 -- | Given a current mapping from robot names to robots, apply a
 --   'ViewCenterRule' to derive the location it refers to.  The result
--- a/src/Swarm/Game/State.hs
+++ b/src/Swarm/Game/State.hs
@@ -36,10 +36,12 @@ module Swarm.Game.State (
   robotMap,
   robotsByLocation,
   robotsAtLocation,
+  robotsInArea,
   activeRobots,
   waitingRobots,
   availableRecipes,
   availableCommands,
+  messageNotifications,
   allDiscoveredEntities,
   gensym,
   randGen,
@@ -73,6 +75,7 @@ module Swarm.Game.State (
   classicGame0,
 
   -- * Utilities
+  manhattan,
   applyViewCenterRule,
   recalcViewCenter,
   modifyViewCenter,
@@ -88,7 +91,8 @@ module Swarm.Game.State (
   deleteRobot,
   activateRobot,
   toggleRunStatus,
-  messageNotifications,
+  messageIsRecent,
+  messageIsFromNearby,
 ) where
 
 import Control.Algebra (Has)
@@ -144,10 +148,14 @@ import Swarm.Language.Pipeline (ProcessedTerm)
 import Swarm.Language.Pipeline.QQ (tmQ)
 import Swarm.Language.Syntax (Const, Term (TString), allConst)
 import Swarm.Language.Types
-import Swarm.Util (isRightOr, (<+=), (<<.=), (?))
+import Swarm.Util (isRightOr, uniq, (<+=), (<<.=), (?))
 import System.Clock qualified as Clock
 import System.Random (StdGen, mkStdGen, randomRIO)
 import Witch (into)
+
+-- $setup
+-- >>> import qualified Data.Map as M
+-- >>> import Linear.V2
 
 ------------------------------------------------------------
 -- Subsidiary data types
@@ -338,6 +346,51 @@ robotsAtLocation loc gs =
     . view robotsByLocation
     $ gs
 
+-- | Manhattan distance between world locations.
+manhattan :: V2 Int64 -> V2 Int64 -> Int64
+manhattan (V2 x1 y1) (V2 x2 y2) = abs (x1 - x2) + abs (y1 - y2)
+
+-- | Get elements that are in manhattan distance from location.
+--
+-- >>> v2s i = [(v, manhattan (V2 0 0) v) | x <- [-i..i], y <- [-i..i], let v = V2 x y]
+-- >>> v2s 0
+-- [(V2 0 0,0)]
+-- >>> map (\i -> length (getElemsInArea (V2 0 0) i (M.fromList $ v2s i))) [0..8]
+-- [1,5,13,25,41,61,85,113,145]
+--
+-- The last test is the sequence "Centered square numbers":
+-- https://oeis.org/A001844
+getElemsInArea :: V2 Int64 -> Int64 -> Map (V2 Int64) e -> [e]
+getElemsInArea o@(V2 x y) d m = M.elems sm'
+ where
+  -- to be more efficient we basically split on first coordinate
+  -- (which is logarithmic) and then we have to linearly filter
+  -- the second coordinate to get a square - this is how it looks:
+  --         ▲▲▲▲
+  --         ││││    the arrows mark points that are greater then A
+  --         ││s│                                 and lesser then B
+  --         │sssB (2,1)
+  --         ssoss   <-- o=(x=0,y=0) with d=2
+  -- (-2,-1) Asss│
+  --          │s││   the point o and all s are in manhattan
+  --          ││││                  distance 2 from point o
+  --          ▼▼▼▼
+  sm =
+    m
+      & M.split (V2 (x - d) (y - 1)) -- A
+      & snd -- A<
+      & M.split (V2 (x + d) (y + 1)) -- B
+      & fst -- B>
+  sm' = M.filterWithKey (const . (<= d) . manhattan o) sm
+
+-- | Get robots in manhattan distastance from location.
+robotsInArea :: V2 Int64 -> Int64 -> GameState -> [Robot]
+robotsInArea o d gs = map (rm IM.!) rids
+ where
+  rm = gs ^. robotMap
+  rl = gs ^. robotsByLocation
+  rids = concatMap IS.elems $ getElemsInArea o d rl
+
 -- | The list of entities that have been discovered.
 allDiscoveredEntities :: Lens' GameState Inventory
 
@@ -462,17 +515,31 @@ replWorking = to (\s -> matchesWorking $ s ^. replStatus)
   matchesWorking (REPLWorking _ _) = True
 
 -- | Get the notification list of messages from the point of view of focused robot.
-messageNotifications :: Getter GameState (Notifications (LogEntry, Bool))
+messageNotifications :: Getter GameState (Notifications LogEntry)
 messageNotifications = to getNotif
  where
-  getNotif gs = Notifications {_notificationsCount = newCount, _notificationsContent = toList allMessages}
+  getNotif gs = Notifications {_notificationsCount = length new, _notificationsContent = allUniq}
    where
+    allUniq = uniq $ toList allMessages
+    new = takeWhile (\l -> l ^. leTime > gs ^. lastSeenMessageTime) $ reverse allUniq
+    -- creative players and system robots just see all messages (and focused robots logs)
+    unchecked = gs ^. creativeMode || fromMaybe False (focusedRobot gs ^? _Just . systemRobot)
+    messages = (if unchecked then id else focusedOrLatestClose) (gs ^. messageQueue)
     allMessages = Seq.sort $ focusedLogs <> messages
-    focusedLogs = (,True) <$> maybe Empty (view robotLog) (focusedRobot gs)
-    messages = (,False) <$> (if gs ^. creativeMode then id else Seq.filter noOther) (gs ^. messageQueue)
-    noOther e = gs ^. focusedRobotID == e ^. leRobotID
-    newCount = max 0 $ Seq.length allMessages - fromMaybe 0 oldestSeen
-    oldestSeen = succ <$> Seq.findIndexR (\(l, _) -> l ^. leTime <= gs ^. lastSeenMessageTime) allMessages
+    focusedLogs = maybe Empty (view robotLog) (focusedRobot gs)
+    -- classic players only get to see messages that they said and a one message that they just heard
+    -- other they have to get from log
+    latestMsg = messageIsRecent gs
+    closeMsg = messageIsFromNearby gs
+    focusedOrLatestClose mq =
+      (Seq.take 1 . Seq.reverse . Seq.filter closeMsg $ Seq.takeWhileR latestMsg mq)
+        <> Seq.filter ((== gs ^. focusedRobotID) . view leRobotID) mq
+
+messageIsRecent :: GameState -> LogEntry -> Bool
+messageIsRecent gs e = e ^. leTime == (gs ^. ticks)
+
+messageIsFromNearby :: GameState -> LogEntry -> Bool
+messageIsFromNearby gs e = maybe False (\r -> manhattan (r ^. robotLocation) (e ^. leLocation) <= hearingDistance) (focusedRobot gs)
 
 -- | Given a current mapping from robot names to robots, apply a
 --   'ViewCenterRule' to derive the location it refers to.  The result
--- a/src/Swarm/Game/State.hs
+++ b/src/Swarm/Game/State.hs
@@ -147,7 +147,7 @@ import Swarm.Language.Pipeline (ProcessedTerm)
 import Swarm.Language.Pipeline.QQ (tmQ)
 import Swarm.Language.Syntax (Const, Term (TString), allConst)
 import Swarm.Language.Types
-import Swarm.Util (isRightOr, uniq, (<+=), (<<.=), (?), getElemsInArea, manhattan)
+import Swarm.Util (getElemsInArea, isRightOr, manhattan, uniq, (<+=), (<<.=), (?))
 import System.Clock qualified as Clock
 import System.Random (StdGen, mkStdGen, randomRIO)
 import Witch (into)
--- a/src/Swarm/Game/State.hs
+++ b/src/Swarm/Game/State.hs
@@ -56,6 +56,7 @@ module Swarm.Game.State (
   replStatus,
   replWorking,
   messageQueue,
+  lastSeenMessageTime,
   focusedRobotID,
   ticks,
   robotStepsPerTick,
@@ -86,6 +87,8 @@ module Swarm.Game.State (
   wakeUpRobotsDoneSleeping,
   deleteRobot,
   activateRobot,
+  toggleRunStatus,
+  messageNotifications,
 ) where
 
 import Control.Algebra (Has)
@@ -98,6 +101,7 @@ import Control.Monad.Except
 import Data.Aeson (FromJSON, ToJSON)
 import Data.Array (Array, listArray)
 import Data.Bifunctor (first)
+import Data.Foldable (toList)
 import Data.Int (Int64)
 import Data.IntMap (IntMap)
 import Data.IntMap qualified as IM
@@ -110,12 +114,14 @@ import Data.List.NonEmpty qualified as NE
 import Data.Map (Map)
 import Data.Map qualified as M
 import Data.Maybe (fromMaybe, isJust, mapMaybe)
+import Data.Sequence (Seq ((:<|)))
+import Data.Sequence qualified as Seq
 import Data.Set qualified as S
 import Data.Text (Text)
 import Data.Text qualified as T (lines)
 import Data.Text.IO qualified as T (readFile)
 import GHC.Generics (Generic)
-import Linear (V2 (..))
+import Linear
 import Paths_swarm (getDataFileName)
 import Swarm.Game.CESK (emptyStore, initMachine)
 import Swarm.Game.Entity
@@ -193,6 +199,13 @@ data RunStatus
     AutoPause
   deriving (Eq, Show, Generic, FromJSON, ToJSON)
 
+-- | Switch (auto or manually) paused game to running and running to manually paused.
+--
+--   Note that this function is not safe to use in the app directly, because the UI
+--   also tracks time between ticks - use 'Swarm.TUI.Controller.safeTogglePause' instead.
+toggleRunStatus :: RunStatus -> RunStatus
+toggleRunStatus s = if s == Running then ManualPause else Running
+
 -- | A data type to keep track of discovered recipes and commands
 data Notifications a = Notifications
   { _notificationsCount :: Int
@@ -257,7 +270,8 @@ data GameState = GameState
   , _viewCenter :: V2 Int64
   , _needsRedraw :: Bool
   , _replStatus :: REPLStatus
-  , _messageQueue :: [Text]
+  , _messageQueue :: Seq LogEntry
+  , _lastSeenMessageTime :: Integer
   , _focusedRobotID :: RID
   , _ticks :: Integer
   , _robotStepsPerTick :: Int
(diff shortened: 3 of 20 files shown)


xsebek:
There is a problem with the panel, it does not scroll past the first page and a little:
```
def for = \i.\c. if (i > 0) {c i; for (i-1) c} {} end
for 10 (\i. log ("index" ++ format i))
for 50 (\i. build {say ("index" ++ format i)})
```
Screenshot from 2022-07-18 20-02-31

byorgey (approved):
Looks great!  I am not sure what is causing the scroll limit on the messages dialog, I will give that some more thought.

src/Swarm/TUI/View.hs
```diff
@@ -511,6 +514,25 @@ descriptionTitle e = " " ++ from @Text (e ^. entityName) ++ " "
 descriptionWidget :: AppState -> Entity -> Widget Name
 descriptionWidget s e = padLeftRight 1 (explainEntry s e)
 
+messagesWidget :: GameState -> Widget Name
+messagesWidget gs = viewport MessageViewport Vertical (vBox widgetList)
+ where
+  widgetList = focusNewest . map drawLogEntry' . toList . Seq.sort $ focusedLogs <> messages
```
TristanCacqueray, line 520: This seems to fix the scroll issue, not sure why though :)
```suggestion
  widgetList = map drawLogEntry' . toList . Seq.sort $ focusedLogs <> messages
```

src/Swarm/TUI/View.hs
```diff
@@ -551,11 +573,16 @@ drawKeyMenu s =
...
-      <> notificationKey availableRecipes "F3" "Recipes"
-      <> notificationKey availableCommands "F4" "Commands"
-      <> [(NoHighlight, "^v", "creative") | cheat]
-      <> [(NoHighlight, "^g", "goal") | goal]
+    catMaybes
+      [ Just (NoHighlight, "F1", "help")
+      , Just (NoHighlight, "F2", "robots")
+      , notificationKey availableRecipes "F3" "Recipes"
+      , notificationKey availableCommands "F4" "Commands"
+      , Just (NoHighlight, "F4", "Commands")
```
TristanCacqueray, line 581: This seems to cause the F4 command help to appear twice.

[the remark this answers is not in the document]
TristanCacqueray: Perhaps we should use the same notification logic so that the button is highlighted when there are new messages?

src/Swarm/TUI/View.hs
```diff
@@ -511,6 +514,25 @@ descriptionTitle e = " " ++ from @Text (e ^. entityName) ++ " "
 descriptionWidget :: AppState -> Entity -> Widget Name
 descriptionWidget s e = padLeftRight 1 (explainEntry s e)
 
+messagesWidget :: GameState -> Widget Name
+messagesWidget gs = viewport MessageViewport Vertical (vBox widgetList)
+ where
+  widgetList = focusNewest . map drawLogEntry' . toList . Seq.sort $ focusedLogs <> messages
+  focusNewest = over (element 1) visible
```
byorgey, line 521: Explicitly setting `visible` on the newest log message is what is causing the scrolling not to work.  `visible` specifically demands that the given element is visible, so you cannot scroll it out of the window.  If I remove the `focusNewest` call then scrolling works as expected.

byorgey: Also, isn't `element 1` the *second* newest log message?  That's why you can scroll the window by one line but no more.

xsebek: Hmm, we use it in the other log list, how is it unset then?

I wanted to go back to that lens and try `_1` or head or something, never did I expect it would behave so weirdly. 🤦

src/Swarm/TUI/View.hs
```diff
@@ -511,6 +514,25 @@ descriptionTitle e = " " ++ from @Text (e ^. entityName) ++ " "
...
+      hBox
+        [ txt $ "[" <> view leRobotName e <> "] "
+        , txtWrapWith indent2 (e ^. leText)
+        ]
+  focusedLogs = (,True) <$> gs ^. robotMap . ix (gs ^. focusedRobotID) . robotLog
+  messages = (,False) <$> (if gs ^. creativeMode then id else Seq.filter noOther) (gs ^. messageQueue)
+  noOther e = gs ^. focusedRobotID == e ^. leRobotID
+  -- color each robot message with different color of the world (except those with background)
+  robotColor rid = fgCols !! (rid `mod` fgColLen)
+  fgCols = filter (`notElem` [waterAttr, iceAttr]) worldAttributes
```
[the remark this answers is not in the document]
xsebek: Yeah, I should filter it properly and check that background is default.

src/Swarm/TUI/View.hs
```diff
@@ -511,6 +514,25 @@ descriptionTitle e = " " ++ from @Text (e ^. entityName) ++ " "
...
+  focusNewest = over (element 1) visible
+  drawLogEntry' (e, isLog) =
+    withAttr (if isLog then notifAttr else robotColor (e ^. leRobotID)) $
+      hBox
+        [ txt $ "[" <> view leRobotName e <> "] "
+        , txtWrapWith indent2 (e ^. leText)
+        ]
+  focusedLogs = (,True) <$> gs ^. robotMap . ix (gs ^. focusedRobotID) . robotLog
+  messages = (,False) <$> (if gs ^. creativeMode then id else Seq.filter noOther) (gs ^. messageQueue)
+  noOther e = gs ^. focusedRobotID == e ^. leRobotID
```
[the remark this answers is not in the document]
xsebek: Exactly, though its hard to get right given that robots move.

I have some ideas about how listen or always-on-listen could work, but I will have to think it through.

src/Swarm/TUI/View.hs
TristanCacqueray: It doesn't like the scroll is working properly yet.

xsebek: Fixed :white_check_mark:

src/Swarm/TUI/View.hs
xsebek: Fixed :white_check_mark:

xsebek:
@TristanCacqueray  I tried adding a button to allow (un)pausing the game from the message view, but it runs into some problem mentioned in `maybeUnpause`:
```Haskell
  -- When unpausing, it is critical to ensure the next frame doesn't
  -- catch up from the time spent in pause.
  -- TODO: manage unpause more safely to also cover
  -- the world event handler for the KChar 'p'.
```
Should I just make `maybeUnpause` toggle the run status to also set the time in the world event handler and my message queue?

xsebek:
Scrolling now works somewhat intuitively (unpaused autoscroll, paused unobstructed scroll), please give it a try! :slightly_smiling_face: 
Screenshot from 2022-07-19 02-27-04

xsebek:
I squashed and rebased the previous commits.

The last commit should fix the unpausing, please check it out as it's easy to break. I think I got it right on the 3rd try though. :sweat_smile: 

I will try to use the notification logic and then it should be good to merge.

Thanks, @byorgey  and @TristanCacqueray, for helping me with this UI work. :slightly_smiling_face: 

Finally, I can't wait to try out #340, having `AppState` as the monad state just makes sense and I will not miss `continue`.

src/Swarm/TUI/Controller.hs
```diff
@@ -319,6 +319,7 @@ handleModalEvent s = \case
     case s ^? uiState . uiModal . _Just . modalDialog . to dialogSelection of
       Just (Just QuitButton) -> quitGame s'
       Just (Just (NextButton scene)) -> startGame scene s'
+      Just (Just PauseButton) -> continue $ s & gameState . runStatus %~ toggleRunStatus
```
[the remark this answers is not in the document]
xsebek: Should work now! :heavy_check_mark:

src/Swarm/TUI/Attr.hs
```diff
@@ -132,6 +132,31 @@ sepAttr = "sep"
 infoAttr = "info"
 defAttr = "def"
 
+worldAttributes :: [AttrName]
```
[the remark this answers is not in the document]
xsebek: One way to do it is to not allow accessing world attribute names and create a `[(AttrName, Attr)]` list of world attributes.

Those I could filter and pass directly to the attribute map. :smiling_imp:

byorgey (approved):
This looks good to me, and seems to work properly except for exiting a modal via the `Esc` key (see my other comment).

byorgey:
@xsebek #561 will be great, go ahead and give it a shot!

src/Swarm/TUI/Attr.hs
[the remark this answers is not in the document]
byorgey: Yeah, we really don't need Haskell constants, I think we should get rid of them.  We can easily look up attributes by name from some kind of map or association list.

src/Swarm/Game/Step.hs
```diff
@@ -110,7 +110,7 @@ gameTick = do
 
             when (newLoc /= oldLoc) $ do
               robotsByLocation . at oldLoc %= deleteOne rn
-              robotsByLocation . at newLoc . non Empty %= IS.insert rn
+              robotsByLocation . at newLoc . non Empty %= IS.insert rn -- TODO: should this be above?
```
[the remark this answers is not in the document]
byorgey: I'm not sure what you mean by "should be above", do you mean the `non Empty` should be on the previous line instead, with `oldLoc`?  I don't think that makes sense.  `at` yields a `Maybe Set`; `deleteOne` explicitly takes a `Maybe Set` because we want it to explicitly remove the mapping entirely if the set becomes empty.  On the other hand, the `non Empty` on the second line turns the `Maybe Set` into a `Set`, substituting an empty set for `Nothing`, so that if no mapping existed for the new location previously, it creates an empty set for us to insert `rn` into.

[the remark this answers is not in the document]
xsebek: Oh, I thought it would remove the empty set, so I had it backwards. Thanks for the explanation :slightly_smiling_face:

src/Swarm/TUI/Controller.hs
```diff
@@ -214,7 +214,7 @@ handleMainEvent s = \case
     continue s
   Key V.KEsc
     | isJust (s ^. uiState . uiError) -> continue $ s & uiState . uiError .~ Nothing
-    | isJust (s ^. uiState . uiModal) -> maybeUnpause s >>= (continue . (uiState . uiModal .~ Nothing))
+    | isJust (s ^. uiState . uiModal) -> safeTogglePause s >>= (continue . (uiState . uiModal .~ Nothing))
```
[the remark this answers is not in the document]
xsebek: Should be fixed now :+1:

src/Swarm/TUI/Controller.hs
```diff
@@ -223,7 +223,15 @@ handleMainEvent s = \case
...
-  FKey 5 -> toggleModal s MessagesModal >>= continue
+  FKey 5 -> do
+    let t = s ^. gameState . ticks
+    s' <- toggleModal s MessagesModal
+    liftIO $ appendFile "/tmp/debug.log" $ "\n\nOLD LEN - " <> show (s ^. gameState . messageNotifications . notificationsCount)
+    liftIO $ appendFile "/tmp/debug.log" $ "\nOLD TIME - " <> show (s' ^. gameState . lastSeenMessageTime)
+    liftIO $ appendFile "/tmp/debug.log" $ "\nNEW TIME - " <> show t
+    liftIO $ appendFile "/tmp/debug.log" $ "\nNEW LEN - " <> show (s' ^. gameState . messageNotifications . notificationsCount)
+    liftIO $ appendFile "/tmp/debug.log" $ "\nMESSAGES - " <> show (view leTime <$> s' ^. gameState . messageQueue)
+    continue (s' & gameState . lastSeenMessageTime .~ t)
```
xsebek, line 234: Somehow, I am not counting the number of new messages right and it's always zero! :facepalm: 

The flow in Controller should be correct, so it has to be some stupid mistake in State.

xsebek: Should work now! :heavy_check_mark:

src/Swarm/TUI/View.hs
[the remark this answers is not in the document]
xsebek: Done! :white_check_mark:

src/Swarm/TUI/View.hs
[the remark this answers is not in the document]
xsebek: Let's do it in separate PR.

src/Swarm/TUI/Attr.hs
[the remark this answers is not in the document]
xsebek: Well, I meant keeping constants _only_ for those we use in the Haskell code and _**not** those used in the world map*_.

<sup>*) With the exception of the few attributes that we use for drawing the logo.</sup>

It would not be great to have outdated attribute strings in code, so I think it is a good idea to keep those. :slightly_smiling_face:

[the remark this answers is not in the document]
xsebek: @byorgey I just pushed my solution, what do you think? :slightly_smiling_face:

src/Swarm/TUI/View.hs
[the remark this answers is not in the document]
xsebek: The code should now prevent this, all world attributes have only foreground colour and water is the only exception.

src/Swarm/Game/State.hs
```diff
@@ -440,7 +445,7 @@ viewCenterRule = lens getter setter
     case rule of
       VCLocation v2 -> g {_viewCenterRule = rule, _viewCenter = v2}
       VCRobot rid ->
-        let robotcenter = g ^? robotMap . ix rid <&> view robotLocation
+        let robotcenter = g ^? robotMap . ix rid . robotLocation
          in -- retrieve the loc of the robot if it exists, Nothing otherwise.
             -- sometimes, lenses are amazing...
             case robotcenter of
```
xsebek, line 451: _Sometimes, lenses are amazing..._ and sometimes they give you a headache, trying to find out why is the special robot that you added (later) to the game state not focused. :smile:

src/Swarm/TUI/View.hs
```diff
@@ -84,6 +80,10 @@ import Swarm.TUI.Border
 import Swarm.TUI.Model
 import Swarm.TUI.Panel
 import Swarm.Util
+import System.Clock (TimeSpec (..))
+import Text.Printf
+import Text.Wrap
+import Witch (from)
```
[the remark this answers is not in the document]
xsebek: Done in #579! :+1:

byorgey (commented):
What is the difference between `log` and `say`?  I'm sure there is a difference, but I don't think I am clear on exactly what the difference is.

src/Swarm/Game/State.hs
xsebek: The lenses I used were ordered with `.` not `&`, so when I added a robot on line X and called this lens on line X+1 it did not do anything.

It did not really matter, because there were other issues, so it was more of a funny discovery 😄

src/Swarm/TUI/Controller.hs
```diff
@@ -223,7 +223,10 @@ handleMainEvent s = \case
   FKey 4 | not (null (s ^. gameState . availableCommands . notificationsContent)) -> do
     s' <- toggleModal s CommandsModal
     continue (s' & gameState . availableCommands . notificationsCount .~ 0)
-  FKey 5 -> toggleModal s MessagesModal >>= continue
+  FKey 5 -> do
+    let t = s ^. gameState . ticks
+    s' <- toggleModal s MessagesModal
+    continue (s' & gameState . lastSeenMessageTime .~ t)
```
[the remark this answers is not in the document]
xsebek: Oh good point, I will fix that. 👍

src/Swarm/Language/Syntax.hs
```diff
@@ -555,9 +555,12 @@ constInfo c = case c of
     command 0 long . doc "Deconstruct an old robot." $
       ["Salvaging a robot will give you its inventory, installed devices and log."]
   Say ->
-    command 1 short . doc "Emit a message." $ -- TODO: #513
-      [ "The message will be in a global log, which you can not currently view."
-      , "https://github.com/swarm-game/swarm/issues/513"
+    command 1 short . doc "Emit a message." $
+      [ "The messages can be viewed in the message window."
+      , "However, there is a limit to how distant messages you can hear."
```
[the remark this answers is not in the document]
xsebek: The problem with showing all messages in that place is that it might give the wrong impression that the robot was able to hear them. But it could have just moved to the radius of that message or conversely not see messages it would have been able to hear!

My plan was to be strict and just include those that the robot heard using the `listen` command. So just removing the duplicity in Creative mode and colouring those logs that are saying "I heard robot X say Y".

But that is not very user-friendly and in hindsight, I should show at least some messages.

A truly ~evil~ realistic thing to do would be to show the messages whose distance equals their age. Meaning that sound travels one cell each tick.
Litlle less ~evil~ realistic would be to have some interval like one real-world second to make this more user-friendly.
I am starting to like this idea - it can populate the message queue history and relativity is just cool.
Bonus points for checking that the message has a path to the robot that is not blocked! :rofl: 

Easier to implement would be to just show the messages in the radius and limit them to last real-world second - so 16 ticks.
But that is a bit annoying and does not make much sense.

Whichever way we choose, I feel it should be how `listen` finds messages, so that the players can use the message queue as a guide to how 

---

**TL;DR:** I propose I will implement one of the following options, which will form the logic of `listen`:
- show say messages of other robots in radius _and time limit of one real-world second_
- show say messages that would reach the robot travelling at one cell per tick
- show say messages that would reach the robot travelling at one cell per tick _with a tolerance of one real-world second_
- _show all messages_ as we are not pedantic relativity nerds and #577 is not implemented

What do you think @byorgey?

xsebek:
> @byorgey What is the difference between log and say? I'm sure there is a difference, but I don't think I am clear on exactly what the difference is.

Easy - log writes to the logger, otherwise, it fails. Said messages go to the global message queue and do not require anything.

---

The current logic for showing things in the panel is:
- show focused robots logs and messages - it would know what it logged and said
- in creative mode show all messages (but only focused robot logs)

As you pointed out, that may be too pedantic, see [my ideas](https://github.com/swarm-game/swarm/pull/565#discussion_r928314003) about how to improve the situation.

---

P.S. The reason why I implemented it so was that I just wanted a way to debug scenarios and win conditions without the hassle of inspecting the installed robot's logger. But making it more useful outside creative mode should be doable in this PR.

src/Swarm/Language/Syntax.hs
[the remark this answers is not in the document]
byorgey: Relativity is fun but doesn't seem relevant here, since even audio signals typically travel fast enough that the delay is not really noticeable within the range of the sensor.  When we get to #94 we should definitely come back to that idea.  

How about this: 
- Messages from other robots only show up in the Messages dialog *if you have a listening device installed*
- Then we just show say messages of other robots within the radius and a certain time limit.
    - Am I understanding correctly that by "a time limit of one real-world second" you mean that messages would flash on the Messages dialog for one second and then disappear again?
    - One second seems way too short to see and read any messages.
    - Why one real-world second?  That would be more complex to implement and I'm not sure what the benefit would be.  Why not just say "16 ticks" or whatever?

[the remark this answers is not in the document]
xsebek: > Am I understanding correctly that by "a time limit of one real-world second" you mean that messages would flash on the Messages dialog for one second and then disappear again?

Exactly, so it's not great. But maybe there is a better option that I forgot about:
- say will put the message to the global queue and to all robot logs of robots that:
  - are within the radius
  - have logger
  - have some listening device

Suggestion for "hearing aid" recipe? :smile: Maybe we can do #577 here too.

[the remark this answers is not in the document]
byorgey: Ah, right, if we put it in the logger then they will remember it so it doesn't matter if they move later. I like it!

For a hearing aid, maybe something like a copper pipe, some paper, a lodestone, and some wires?

src/Swarm/TUI/Controller.hs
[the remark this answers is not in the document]
xsebek: Should be fixed now :slightly_smiling_face:

xsebek:
Say and listen should now work:
Screenshot from 2022-08-01 00-30-41

xsebek:
__I think this is ready!__ :smile: 

Say should work pretty well with the new messages window. There are a few tests, but I also tested it manually and that `[F5] Messages` getting highlighted with new messages feels very satisfying. Big thanks to @TristanCacqueray for laying the groundwork for that! :+1: 

Overall the messages window should be easier to read and scroll through than the logger and I hope to improve it with e.g. filters in the future. It reuses a lot of the logic for the other windows and also the time widget. :framed_picture: 

Finally the `listen` command is a bit of new addition and may have some rough edges. But it works well enough in the starting tutorial scenario, so let's hope there are no nasty bugs. :crossed_fingers:

data/entities.yaml
```diff
@@ -816,6 +816,22 @@
   properties: [portable]
   capabilities: [log]
 
+- name: hearing aid
+  display:
+    attr: device
+    char: '@'
+  description:
+  - "Allows robot to hear what nearby robots are saying."
```
byorgey, line 824:
```suggestion
  - "Allows a robot to hear what nearby robots are saying."
```

data/entities.yaml
```diff
@@ -816,6 +816,22 @@
...
+  - "Simply having this device installed will automatically
+     add messages said by nearby robots to this robots log,
```
byorgey, line 826:
```suggestion
     add messages said by nearby robots to this robot's log,
```

data/recipes.yaml
```diff
@@ -478,6 +478,15 @@
   out:
   - [1, strange loop]
 
+- in:
+  - [2, copper pipe]
+  - [4, copper wire]
+  - [1, paper]
+  - [1, lodestone]
+  out:
+  - [2, hearing aid]
```
byorgey, line 487: Haha, I like that the recipe produces one for each ear :smile:

src/Swarm/Game/State.hs
```diff
@@ -395,40 +458,46 @@ viewCenterRule = lens getter setter
...
+    -- classic players only get to see messages that they said and a one message that they just heard
+    -- other they have to get from log
+    latestMsg = messageIsRecent gs
+    closeMsg = messageIsFromNearby (gs ^. viewCenter)
+    focusedOrLatestClose mq =
+      (Seq.take 1 . Seq.reverse . Seq.filter closeMsg $ Seq.takeWhileR latestMsg mq)
+        <> Seq.filter ((== gs ^. focusedRobotID) . view leRobotID) mq
+
+messageIsRecent :: GameState -> LogEntry -> Bool
+messageIsRecent gs e = e ^. leTime >= gs ^. ticks - 1
```
byorgey, line 497: Does this mean it will only show messages that happened more recently than 1/16 of a second ago?

src/Swarm/Game/Step.hs
```diff
@@ -249,16 +259,28 @@ randomName = do
...
+createLogEntry :: (Has (State GameState) sig m, Has (State Robot) sig m) => Bool -> Text -> m LogEntry
+createLogEntry isSaid msg = do
+  rid <- use robotID
   rn <- use robotName
   time <- use ticks
-  robotLog %= (Seq.|> LogEntry msg rn time)
+  loc <- use robotLocation
+  pure $ LogEntry time isSaid rn rid loc msg
+
+-- | Print some text via the robot's log.
```
byorgey, line 273: Why do we need the Bool argument? Can we add a comment explaining what it is for?

data/entities.yaml
```diff
@@ -816,6 +816,22 @@
...
+     assuming it has a logger installed."
+  - "That way you can view any heard message later either in
+     the logger or the message window."
+  - "To wait for a message and get the string value, use:"
+  - "`l <- listen; log $ \"I have waited for someone to say \" ++ l`"
+  properties: [portable]
+  capabilities: [log]
```
byorgey, line 833: Should this be `listen`?

src/Swarm/TUI/Controller.hs
```diff
@@ -341,31 +348,45 @@ toggleModal mt = do
...
+  let isRunning = maybe False isRunningModal mt
+  case ev of
+    V.EvKey V.KEnter [] -> do
+      mdialog <- preuse $ uiState . uiModal . _Just . modalDialog
+      toggleModal QuitModal
+      case dialogSelection <$> mdialog of
+        Just (Just QuitButton) -> quitGame
+        Just (Just (NextButton scene)) -> startGame scene
+        _ -> return ()
+    -- TODO: copy pasted here from 'handleMainEvent'
```
byorgey, line 375: What needs to be done here? Per our new policy, we should either do it or turn it into an issue. :smile:

src/Swarm/Util.hs
```diff
@@ -134,6 +142,58 @@ cycleEnum e
...
+--
+-- >>> uniq []
+-- []
+-- >>> uniq [1..5]
+-- [1,2,3,4,5]
+-- >>> uniq (replicate 10 'a')
+-- "a"
+-- >>> uniq "abbbccd"
+-- "abcd"
+uniq :: Eq a => [a] -> [a]
```
byorgey, line 155: Why do we need a quadratic-time `uniq` rather than an n lg n -time version using `Ord`?

byorgey (approved):

src/Swarm/Util.hs
xsebek: This should be linear time unix style uniq, unless I made a mistake. 🙂 

It is used on already sorted list.

src/Swarm/TUI/Controller.hs
xsebek: I would like to avoid duplication, but I am not sure if it is feasible.

The modal handler is plugged in higher than the time handlers, so it overrides them.
But maybe they could be conditionally disabled there and moved up? :thinking:

src/Swarm/Game/Step.hs
xsebek: Thanks for a reminder, I think a small enumeration `data MessageSource = Said | Logged | ErrorTrace` would work best here and I could colour the errors red. :fire:

src/Swarm/Game/State.hs
xsebek: Oh yes, without a logger it is very weak and only suited for stepping through.

I expect listening robots will likely either have a logger (base, better-equipped relay robots,...) or they will be system robots.

data/recipes.yaml
xsebek: Thanks, that was the idea. :ear_with_hearing_aid:

src/Swarm/Util.hs
byorgey: Oh, whoops, that's what I get for reading too quickly. :facepalm:

data/entities.yaml
xsebek: Right... :sweat_smile: 
```suggestion
  capabilities: [listen]
```

src/Swarm/Game/Step.hs
xsebek: Looks pretty! :art: 
image

xsebek:
@byorgey is it fine if I move the modal help up? The line above REPL was already full.
Screenshot from 2022-08-03 23-08-26

src/Swarm/TUI/Controller.hs
xsebek: Turns out that was pretty easy and it seems to work well. :+1:

byorgey:
Yeah, I think that's a good idea.

src/Swarm/Game/Step.hs
```diff
@@ -1067,15 +1089,51 @@ execConst c vs s k = do
...
+                | manhattan rl (e ^. leLocation) > manhattan rl (m ^. leLocation) -> es |> m
+                | otherwise -> es |> e
+        let addToRobotLog :: Has (State GameState) sgn m => Robot -> m ()
+            addToRobotLog r = do
+              r' <- execState r $ do
+                hasLog <- hasCapability CLog
+                hasListen <- hasCapability CListen
+                loc' <- use robotLocation
+                when (hasLog && hasListen) (robotLog %= addLatestClosest loc')
+              addRobot r'
```
kostmo, line 1110: For reference: original definition of `addToRobotLog`

PR #567, opened 2022-07-18
Add `time` command and UI clock

### LINKED ISSUE #402: Make `clock` provide `time` capability
The `clock` will probably end up being an ingredient for other things (e.g. a `time machine`, see #133), but I'd like to make it a device in its own right.  We could make it provide the ability to `wait`.  We could also create an instruction `time` that gets the absolute number of ticks since the beginning of the game, perhaps.

_Originally posted by @byorgey in https://github.com/swarm-game/swarm/issues/395#issuecomment-1156492715_

### LINKED ISSUE #559: Time display in world panel
It would be fun to display the time (elapsed number of ticks) in the world panel.  I'm thinking something like xxxx:yy:zz where everything  is in hex, zz is the number of 8-tick "seconds" that have elapsed, and it goes from there.  So for example `3:8e:0a` would mean `0x38e0a * 8` ticks.

byorgey:
Add `time` capability and related things:
- Add a new constant `time : cmd int`
- Make `clock` device provide `time` capability
- If the focused robot has a `clock` device installed (`base` has one by default in classic mode), display hexadecimal time in bottom left of the world panel.
- Add binary, octal, and hexadecimal integer literal syntax to the language.
- Make 16 ticks/s the default speed
    - The old default, 8 ticks/s, was annoyingly slow
    - Now you can simply type in `0x` followed by all the digits in the clock display, followed by a single trailing 0, to get the `time` in ticks

#402 discussed the possibility of having the `wait` command also require the `time` capability.  This breaks a number of tests which rely on the `wait` command.  I can fix the tests of course, but I'd like some feedback on whether requiring a `clock` in order to `wait` is a good idea.

Closes #402.
Closes #559.

### DIFF
--- a/src/Swarm/Language/Parse.hs
+++ b/src/Swarm/Language/Parse.hs
@@ -143,10 +143,20 @@ identifier = (lexeme . try) (p >>= check) <?> "variable name"
 stringLiteral :: Parser Text
 stringLiteral = into <$> lexeme (char '"' >> manyTill L.charLiteral (char '"'))
 
--- | Parse a positive integer literal token.  Note that negation is
---   handled as a separate operator.
+-- | Parse a positive integer literal token, in decimal, binary,
+--   octal, or hexadecimal notation.  Note that negation is handled as
+--   a separate operator.
 integer :: Parser Integer
-integer = lexeme L.decimal
+integer =
+  label "integer literal" $
+    lexeme $ do
+      n <-
+        try (string "0b" *> L.binary)
+          <|> try (string "0o" *> L.octal)
+          <|> try (string "0x" *> L.hexadecimal)
+          <|> L.decimal
+      notFollowedBy alphaNumChar
+      return n
 
 braces :: Parser a -> Parser a
 braces = between (symbol "{") (symbol "}")
--- a/src/Swarm/Language/Parse.hs
+++ b/src/Swarm/Language/Parse.hs
@@ -151,9 +151,9 @@ integer =
   label "integer literal" $
     lexeme $ do
       n <-
-        try (string "0b" *> L.binary)
-          <|> try (string "0o" *> L.octal)
-          <|> try (string "0x" *> L.hexadecimal)
+        string "0b" *> L.binary
+          <|> string "0o" *> L.octal
+          <|> string "0x" *> L.hexadecimal
           <|> L.decimal
       notFollowedBy alphaNumChar
       return n
--- a/src/Swarm/Language/Parse.hs
+++ b/src/Swarm/Language/Parse.hs
@@ -161,9 +161,6 @@ integer =
 braces :: Parser a -> Parser a
 braces = between (symbol "{") (symbol "}")
 
--- dbraces :: Parser a -> Parser a
--- dbraces = between (symbol "{{") (symbol "}}")
-
 parens :: Parser a -> Parser a
 parens = between (symbol "(") (symbol ")")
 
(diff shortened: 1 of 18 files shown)


src/Swarm/TUI/View.hs
```diff
@@ -233,6 +238,14 @@ drawGameUI s =
   addCursorPos = case s ^. uiState . uiWorldCursor of
     Just coord -> topLabels . rightLabel ?~ padLeftRight 1 (drawWorldCursorInfo (s ^. gameState) coord)
     Nothing -> id
+  -- Add clock display in bottom left of the world view if focused robot
+  -- has a clock installed
+  addClock = case s ^. gameState . to focusedRobot of
```
TristanCacqueray, line 243: Could we also display `paused` when the game is not running, what ever the current focused robot is?

byorgey: Great idea, done! :heavy_check_mark:

byorgey: I also decided to show individual ticks in the clock display only when the game is paused, *i.e.* the decimal point and hex digit at the end of `0:00:03.a`.  That way you can see the tick count increase as you single-step the game.

xsebek:
> ... whether requiring a clock in order to wait is a good idea.

Technically getting something like `wait` manually requires arithmetic and recursion, so I would argue it should not be much harder than obtaining lambda and counter, but probably should not be free.

I think the clock requirement is sensible, though it will be annoying to add to challenges (like waiting for trees to grow). 

But hey, at least you can now do a speedrun of those challenges using the in-game clock! :smile:

xsebek:
Could the clock be on the top line instead? It's already crowded right above the REPL:
| image |
|---|

AFAIK some OS put the clock top middle (macOS, Gnome) and some bottom right (Windows) by default. I would not look for the clock in the middle of the screen, but maybe some OS or program put it there.

src/Swarm/Language/Parse.hs
```diff
@@ -143,10 +143,20 @@ identifier = (lexeme . try) (p >>= check) <?> "variable name"
...
+--   a separate operator.
 integer :: Parser Integer
-integer = lexeme L.decimal
+integer =
+  label "integer literal" $
+    lexeme $ do
+      n <-
+        try (string "0b" *> L.binary)
+          <|> try (string "0o" *> L.octal)
+          <|> try (string "0x" *> L.hexadecimal)
```
xsebek, line 156: I think that will also need to be added to the editors highlighting. :sweat_smile: 

Should be an easy change to the `constant.numeric` match: `([0-9]*|0b[01]*|0o[0-8]*|0x\x*)`

xsebek (approved):
LGTM! :+1: It's definitely more fun to use step with the clock. :slightly_smiling_face:

byorgey:
> Could the clock be on the top line instead? It's already crowded right above the REPL:

Makes sense.  I moved the clock to the upper right, which seems like a natural place for a clock.  The cursor info used to be there, so I moved that down to the lower left: which actually seems more useful to me since I'm often copying numbers from the cursor info into a command I type at the REPL, so having it right next to the REPL helps.

byorgey:
> Technically getting something like `wait` manually requires arithmetic and recursion, so I would argue it should not be much harder than obtaining lambda and counter, but probably should not be free.

Yeah, that makes sense, but I see now what you mean about challenges --- the challenges themselves do not require `wait` (because a human can just wait long enough manually) but expressing a working `solution` requires `wait`.  It seems annoying and wrong to add a `clock` to the challenge just so we can write down a solution.

byorgey:
So I ended up adding a `clock` only in one case.  In a couple other cases I did some creative things to wait without requiring a `clock`.

PR #571, opened 2022-07-20
Ability to "paint" robot locations in world descriptions

### LINKED ISSUE #557: Drawing robot starting locations in scenario descriptions
It's frustrating, when creating a challenge scenario, having to do a lot of manual counting to specify the starting locations of the robots relative to the map and its upper left corner.

I'd like to be able to DRAW the locations of the starting robots on the map somehow.  Not quite sure exactly how this will work.  We would need to be able to specify the starting location of the robot AND the entity and terrain at that cell.
- The `loc` field for a robot should be optional.
- One possibility would be to (optionally) use TWO characters per cell when drawing the map, one for the terrain+entity and one for a robot.
- Another interesting possibility is that it will simply take the terrain+entity under a robot to be the majority value from the surrounding cells.
- Or the palette could simply specify that a specific character corresponds to a specific robot+entity+terrain.
- I can't think of any good way to draw more than one robot on the same cell, but that's OK.  In that case you can still specify their locations manually.

byorgey:
When describing a scenario, it is now possible to use characters to draw the location of a particular robot.  This can also be used to easily make many copies of the same robot.

- Robots in the `robots` list with a concrete `loc` will actually be created in the world.  Those without a `loc` are only templates.
- A `palette` entry can now optionally have a 3-tuple, indicating terrain, entity, and robot name.
- Add a Game of Life example making use of the new functionality.
- Did a lot of code refactoring along the way.

Closes #557 .


src/Swarm/Game/Robot.hs
```diff
@@ -191,16 +215,19 @@ type Robot = RobotR Identity
...
 
--- | The name of a robot.  Note that unlike entities, robot names are
---   expected to be globally unique
+-- | The name of a robot.
 robotName :: Lens' Robot Text
 robotName = robotEntity . entityName
 
+-- | The name of a robot template.
+trobotName :: Lens' TRobot Text
+trobotName = robotEntity . entityName
```
xsebek, line 229: Why do we need a separate lens for the name? It should not depend on phase. :thinking:

src/Swarm/Game/Robot.hs
```diff
@@ -133,20 +136,40 @@ makeLenses ''LogEntry
...
+-- | Robot templates have no ID; concrete robots definitely do.
+type family RobotID (phase :: RobotPhase) :: * where
+  RobotID 'TemplateRobot = ()
+  RobotID 'ConcreteRobot = RID
+
 -- | A value of type 'RobotR' is a record representing the state of a
 --   single robot.  The @f@ parameter is for tracking whether or not
 --   the robot has been assigned a unique ID.
-data RobotR f = RobotR
+data RobotR (phase :: RobotPhase) = RobotR
```
xsebek, line 162: Cool use of `DataKinds`, I like that it's now confined to just two options. :+1:

src/Swarm/Game/Robot.hs
```diff
@@ -243,12 +275,16 @@ robotContext :: Lens' Robot RobotContext
...
-setRobotID i r = r {_robotID = Identity i}
+-- | Set the ID number of a robot, changing it from template to
+--   concrete.  If it didn't have a location already, just set the
+--   location to (0,0) by default.
+setRobotID :: RID -> TRobot -> Robot
+setRobotID i r =
+  r
+    { _robotID = i
+    , _robotLocation = fromMaybe (V2 0 0) (_robotLocation r)
+    }
```
xsebek, line 287: I would not expect `setRobotID` to set the location, maybe change the name?

src/Swarm/Game/WorldGen.hs
```diff
@@ -66,46 +70,58 @@ testWorld2Entites =
...
+testWorld2 :: EntityMap -> Seed -> WorldFun TerrainType Entity
+testWorld2 em baseSeed = second (readEntity em) . WF $ \(Coords ix) ->
   genBiome
+    ix
     (bool Small Big (sample ix pn0 > 0))
     (bool Soft Hard (sample ix pn1 > 0))
     (bool Natural Artificial (sample ix pn2 > 0))
  where
-  h = murmur3 0 (into (show ix))
+  h ix = murmur3 0 (into (show ix))
```
xsebek, line 93: Why is it necessary to put `ix` everywhere? `H` never seems to use anything else. :raised_eyebrow:

xsebek (approved):
LGTH! :+1: I look forward to not having to count offset by hand on the small map.

src/Swarm/Game/Robot.hs
xsebek: Maybe you do not want to expose lenses with the `RobotR phase` type?

If the idea is to hide it, then why export it?

data/scenarios/Fun/GoL.yaml
```diff
@@ -0,0 +1,71 @@
...
+  palette:
+    'o': [ice, rock, cell]
+    '.': [ice, null, cell]
+  upperleft: [1,-1]
+  map: |
+    ..............................
+    ..............................
+    ......o.......................
+    .......o......................
+    .....ooo......................
```
xsebek, line 60: This game of life is amazing! I totally overlooked that `.` is another robot.

I wonder if you could make it infinite by building robots in the other cells. :thinking:

data/scenarios/Fun/GoL.yaml
```diff
@@ -0,0 +1,71 @@
...
+        h <- scan down;
+        alive <- return (h != inl ());
+        n1 <- count3;
+        turn left; move; turn right; n2 <- count3;
+        turn right; move; move; turn left; n3 <- count3;
+        turn left; move; turn right;
+        total <- return (n1 + n2 + n3 - if alive {1} {0});
+        if (alive && (total < 2 || total > 3))
+          {grab; return ()}
+          { if (not alive && total == 3) { place "rock" } {}
```
xsebek, line 46: Should there be a `wait 1` in the else branch? How do the robots avoid going out of sync?

data/scenarios/Fun/GoL.yaml
```diff
@@ -0,0 +1,71 @@
...
+        b <- scan back;
+        return (cnt h + cnt f + cnt b)
+      end;
+      forever (
+        h <- scan down;
+        alive <- return (h != inl ());
+        n1 <- count3;
+        turn left; move; turn right; n2 <- count3;
+        turn right; move; move; turn left; n3 <- count3;
+        turn left; move; turn right;
```
xsebek, line 42: @byorgey you know, if `as` was special in `atomic`, you could check all of this hypothetically in one tick _and_ place/grab the rock. :wink:

src/Swarm/Game/Robot.hs
byorgey: We don't really need it of course, but if we make it polymorphic then type inference starts failing at use sites with horrible error messages like
```
 Could not deduce (Control.Effect.Sum.Member
                          (State (RobotR phase5)) sig)
        arising from a use of ‘.=’
```
Of course we could fix this by putting type annotations or explicit type applications whenever we use `robotName`, but I thought this seemed nicer.  I will add a comment explaining this.

src/Swarm/Game/Robot.hs
byorgey: Yeah, I think this is way cleaner than what we had before, even though it uses a bunch more `LANGUAGE` extensions.

src/Swarm/Game/Robot.hs
byorgey: Good idea. :heavy_check_mark:

src/Swarm/Game/WorldGen.hs
byorgey: Since I made `WorldFun` into a newtype, `ix` is no longer in scope.  We used to be able to write  something like `testWorld2 baseSeed ix@(Coords ...) = ...` but now we have to write  `testWorld2 baseSeed = WF $ \ix@(Coords ...) -> ...` .   I think it's worth it because newtypes are good documentation, and we get a convenient `Bifunctor` instance, but it is kind of annoying.  Unless there's a better way to refactor the code now that I haven't thought of.

data/scenarios/Fun/GoL.yaml
byorgey: Yes, very good point.  The glider example seems to work fine but while playing around with some other examples I witnessed it going out of sync.  Adding a `wait 1` in the else branch is the first thing I tried but it didn't fix it.  So now I have all the robots wait until the next tick which is a multiple of 32 before starting the next loop.  That seems to work great with all the examples I have tried.

data/scenarios/Fun/GoL.yaml
byorgey: :laughing:  Haha, that definitely seems like an abuse of `atomic`!

src/Swarm/Game/WorldGen.hs
byorgey: OK, never mind, I fixed it with some nested `where` blocks.

data/scenarios/Fun/GoL.yaml
byorgey: Glad you like it, this is the example I've had in mind all along when working on #421 and this PR. :smile:

PR #578, opened 2022-07-22
Split unit test file

xsebek:
Split the unit test file into a separate file for each test tree (except miscellaneous).

It was already a thousand lines long and hard to read.


xsebek:
I wanted to avoid conflicts, so I made the PR on top of `say-revival` branch.

After that is merged this should be changed to point to the main branch.

test/unit/TestParser.hs
```diff
@@ -0,0 +1,254 @@
...
+                    ]
+                )
+            )
+        ]
+    , testGroup
+        "json encoding"
+        [ testCase "simple expr" (roundTrip "42 + 43")
+        , testCase "module def" (roundTrip "def x = 41 end; def y = 42 end")
+        ]
+    , testGroup
```
byorgey, line 154: Not your fault, but I just realized that this test group isn't about the parser at all, but rather about type checking.  But it's nice that the same framework can deal with both.  I don't know if its worth moving them, or renaming the module?

byorgey (approved):

test/unit/TestParser.hs
xsebek: The whole test tree is named "Language - pipeline", what should I name the module and the exported test tree? :thinking: 

Ideas:
- `TestLanguagePipeline`
- `TestLanguage`
- `TestTypecheck`

byorgey: Ah, sure, any of those would be good.  Personally I'd lean towards `TestLanguagePipeline`, although it's kinda long.

xsebek: `TestLanguagePipeline` it is! :heavy_check_mark:

PR #581, opened 2022-07-22
Update editors

### LINKED ISSUE #548: Update editor plugins with `require` as special syntax
I will also need to add `require` as special syntax and then update the plugin.

I might not have time to do it soon, so could you please create a ticket and assign me, so that I do not forget?

_Originally posted by @xsebek in https://github.com/swarm-game/swarm/pull/533#discussion_r912895528_

xsebek:
- update the highlighting for `require` in editors
- bump VSCode plugin version to 0.0.5
- update VSCode plugin changelog
- update the VSCode plugin image link
- copy LICENSE to the VSCode folder (`vsce` was complaining about it)
- rename the team in LICENSE file to Swarm
- closes #548

### DIFF
--- a/editors/vscode/LICENSE
+++ b/editors/vscode/LICENSE
@@ -14,7 +14,7 @@ modification, are permitted provided that the following conditions are met:
       disclaimer in the documentation and/or other materials provided
       with the distribution.
 
-    * Neither the name of Disco team nor the names of other
+    * Neither the name of Swarm team nor the names of other
       contributors may be used to endorse or promote products derived
       from this software without specific prior written permission.
 
--- a/editors/vscode/LICENSE
+++ b/editors/vscode/LICENSE
@@ -1,0 +1,31 @@
+Copyright Brent Yorgey 2021-2022
+SPDX-License-Identifier: BSD-3-Clause
+
+All rights reserved.
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions are met:
+
+    * Redistributions of source code must retain the above copyright
+      notice, this list of conditions and the following disclaimer.
+
+    * Redistributions in binary form must reproduce the above
+      copyright notice, this list of conditions and the following
+      disclaimer in the documentation and/or other materials provided
+      with the distribution.
+
+    * Neither the name of Disco team nor the names of other
+      contributors may be used to endorse or promote products derived
+      from this software without specific prior written permission.
+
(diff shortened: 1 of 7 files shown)


xsebek:
@byorgey how should I push tags? :thinking: Should they go to this branch or should I push them directly to `main`?

Or maybe it would be best to use GitHub UI in releases since I am such a noob with git tags. :smile:

byorgey:
@xsebek tags are like branches that never move. That is, a tag is associated with a particular commit.  So if you tagged a commit in this branch it would always stat on that commit, it would never end up in main when this branch is merged.

editors/vscode/LICENSE
```diff
@@ -0,0 +1,31 @@
...
+
+    * Redistributions of source code must retain the above copyright
+      notice, this list of conditions and the following disclaimer.
+
+    * Redistributions in binary form must reproduce the above
+      copyright notice, this list of conditions and the following
+      disclaimer in the documentation and/or other materials provided
+      with the distribution.
+
+    * Neither the name of Disco team nor the names of other
```
byorgey, line 17: Swarm?

editors/vscode/LICENSE
```diff
@@ -0,0 +1,31 @@
+Copyright Brent Yorgey 2021-2022
```
byorgey, line 1: Makes more sense to put your name, I think.

byorgey (approved):

editors/vscode/LICENSE
xsebek: It's what's in the LICENSE file that I just copied. :man_shrugging:

editors/vscode/LICENSE
xsebek: It's just a copy-paste of the Microsoft template with generated TextMate grammar and keywords generated from the swarm binary, so I don't think it matters. :smile:

If we ever want to change the license for some reason, keeping it simple and not making it look like you have to get my signed permission could save us some headache. Licenses are complicated and I dread lawyers. :fearful:

byorgey: OK, fair enough!

byorgey:
@xsebek can we merge this (after updating `Disco` to `Swarm` of course)?  Or was there still more you wanted to do?

xsebek:
@TristanCacqueray I think the current rule for Mergify might prevent PRs not changing Haskell from merging?

It seems like it is waiting for CI to run but it never will. :confused:

PR #582, opened 2022-07-22
Add negation and lodestone

### LINKED ISSUE #481: Make `counter` easier to construct
`counter` devices are useful on their own, as they let you count how many of an item you have (though there are ways around this), and they also can be used to construct a `calculator` which is extremely useful.  However, a `counter` requires a lot of bits (8 of each) and bits grow slowly.  For comparison, here are the current [min,max] growth rates for various entities:

```
flower: [30, 50]
lambda: [100, 200]
tree: [500, 600]
bit: [1000, 1500]
pixel: [1000, 1500]
LaTeX: [2000, 3000]
```

To me it feels a little unbalanced --- to construct counters at the rate I would like, I would need a really massive bit harvesting operation.

There are a couple ways I can think of to make things more balanced:
- Require fewer bits for a `counter`.  (Though I kind of like the number 8 here; it feels like it makes sense that a `counter` can hold 1 byte.)
- Make bits grow faster.  I'm thinking something like `[200, 400]`---slower than lambda but faster than trees.

Open to other suggestions as well.

### LINKED ISSUE #562: Add boolean logic capability
There should be a device(s?) to provide a boolean logic capability (the **Not**, **And**, **Or** functions).

Ideas:
- `lodestone` that will give a new _Negation_ capability (and have recipes for `bit (0) <--> bit (1)`)
- reuse `branch predictor` and _Cond_ capability for **And** and **Or**, since its just syntactic sugar

xsebek:
- add **lodestone** (+ recipe to flip bits)
- add **negation** capability (and use **cond** for _and_, _or_)
- cleanup capability requirement list
- closes #562 
- closes #481


xsebek:
I have added a lodestone challenge but did not test it yet... :sweat_smile:

data/entities.yaml
```diff
@@ -611,7 +621,7 @@
     of the original entity.
   - For entities which do not grow, 'harvest' behaves exactly the same
     as 'grab'.
-  capabilities: [grab, harvest]
+  capabilities: [grab, harvest, give, place, install]
```
xsebek, line 624: @byorgey was this intended? I thought it was a strict upgrade so that I could use a harvester instead of a grabber.

byorgey: This was definitely intended.  It is not supposed to be a strict upgrade: you start with no harvesters but they are supposed to be easy to build (you can make one with three trees); whereas you start with a bunch of grabbers but they are more difficult to build.  You can use a harvester instead of a grabber to do `grab`, but not to do the other things a grabber can do.

data/recipes.yaml
```diff
@@ -173,6 +198,17 @@
...
+  - [1, mountain]
+  out:
+  - [16, lodestone]
+  - [8, iron ore]
+  - [1, mountain tunnel]
+  required:
+  - [1, metal drill]
+  time: 9
+  weight: 10
+
```
byorgey, line 211: Note that this means lodestones are a non-renewable resource.  I think that is perfectly OK (there should be plenty of them available), but it's worth noting.

byorgey (changes requested):
I like this a lot, but I think we should
- Get rid of the extra `harvester` capabilities (happy to discuss further if you don't agree with my reasoning)
- Make `lodestone` part of the recipe for `compass`.  We can probably just replace the `iron plate` with a `lodestone`?  That would make it much easier to construct, though.  Or maybe we can just keep the `iron plate` and *add* a `lodestone` to the recipe?

data/recipes.yaml
xsebek: Based on what I read on [Wikipedia](https://en.wikipedia.org/wiki/Compass#Magnetic_needle), I planned to make the compass recipe take lodestone as a catalyst.

The idea is that you rub the iron needle with a lodestone and it becomes magnetized.

So you really only need one. :slightly_smiling_face:

data/entities.yaml
xsebek: I would at least add **place**, so it can be used to grab&fetch without salvaging. :thinking:

That sounds realistic to me and leaves **give** and **install** to the grabber.

data/recipes.yaml
byorgey: Ah, OK, makes sense!

data/entities.yaml
byorgey: OK, sure, that sounds good to me.

byorgey (approved):
LGTM! :rocket:

PR #583, opened 2022-07-23
Update to new brick EventM API

### LINKED ISSUE #561: Update to new `brick` API

xsebek:
- closes #561


xsebek:
Btw. this will absolutely run in a few (hopefully small) conflicts with #565, so I would like to merge that first. :sweat_smile:

xsebek:
But it looks like the CI did not use the brick version set in `cabal.project`:
```
src/Swarm/TUI/List.hs:29:3: error:
    • Expected kind ‘* -> *’,
        but ‘EventM n (BL.GenericList n t e)’ has kind ‘*’   <---- This does not look like the new API
   |
29 |   EventM n (BL.GenericList n t e) ()
   |   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
Maybe it is caused by the build cache?

xsebek:
To reiterate the comment from d86d283 here:

This PR needs to be **carefully** checked for mistakes!

The order of lenses matters - the original code was using (.) and (&) a lot - where more of them are used together a subtle bug could be introduced. :bug: 

So places with complicated compositions should be double-checked. Thinks like:
- mixing `(.)` and `(&)`
- `<&>` which I am not that familiar with
- any weird lens like `%%~` that I am not proficient with (so all lenses except `^.` and `.~`)
- probably any block of `EventM` code that is longer than one line

These things should be visible in the diff, but I am looking for places where more actions on the state could be in the wrong order.

---

_This should not be taken as a critique of current code_ - it is written more smartly than I could have written it if at all.

But precisely that makes it unlikely that I got everything right or did not mess up code that depended on the order. :sweat_smile:

cabal.project
```diff
@@ -3,4 +3,9 @@ source-repository-package
   location: https://github.com/colinhect/hsnoise
   tag: 4ccff11dea7e8d94e6a5fcaf8f43857bd65bd72d
 
+source-repository-package
+  type: git
+  location: https://github.com/jtdaugherty/brick.git
```
byorgey, line 8: Just based on comparing with the existing `source-repository-package`, maybe the `.git` should not be there?

byorgey (commented):
Haven't looked at this carefully yet, but I will later.

cabal.project
xsebek: Did not work, could it be that the `cabal.project` file is not even there?

Otherwise, I do not see why the build did not download the commit. :confused:

byorgey: Hmm, I don't know.  I guess if CI was building against the version of `hsnoise` on Hackage instead of getting it from github, we would never have noticed before, since it just affects the way randomly-generated worlds look.

xsebek: Well, `rm -f cabal.project cabal.project.local` in **generate cabal.project** is kind of the obvious villain. :facepalm:

src/Swarm/TUI/Controller.hs
```diff
@@ -102,238 +102,245 @@ pattern FKey :: Int -> BrickEvent n e
...
+    MetaKey 'e' -> setFocus RobotPanel
+    MetaKey 'r' -> setFocus REPLPanel
+    MetaKey 't' -> setFocus InfoPanel
+    -- toggle creative mode if in "cheat mode"
+    ControlKey 'v'
+      | s ^. uiState . uiCheatMode -> gameState . creativeMode %= not
+    MouseDown n _ _ mouseLoc ->
+      case n of
+        WorldPanel -> do
+          mouseCoordsM <- Brick.withLens gameState (mouseLocToWorldCoords mouseLoc)
```
byorgey: This needs to be `zoom` now, right?

src/Swarm/TUI/Controller.hs
```diff
@@ -616,55 +620,58 @@ handleREPLEvent s = \case
...
-          let newform = set promptUpdateL t (s ^. uiState)
-           in s & uiState . uiReplForm .~ newform
+    replForm <- use $ uiState . uiReplForm
+    f' <- nestEventM' replForm (handleFormEvent ev)
+    case formState f' of
+      CmdPrompt {} -> do
+        uiState . uiReplForm .= f'
+        modify validateREPLForm
+      SearchPrompt t _ -> do
+        -- TODO: why does promptUpdateL not update the uiState?
```
byorgey, line 672: Yes, `promptUpdateL` has a strange type, I am not sure why it is that way.

src/Swarm/TUI/Controller.hs
```diff
@@ -808,27 +816,30 @@ adjustTPS (+/-) = uiState . lgTicksPerSecond %~ (+/- 1)
...
-handleRobotPanelEvent s (CharKey 'm') =
-  maybe (continueWithoutRedraw s) (makeEntity s) (focusedEntity s)
-handleRobotPanelEvent s (CharKey '0') = do
-  continue $ s & (uiState . uiShowZero %~ not) . (uiState . uiInventoryShouldUpdate .~ True)
-handleRobotPanelEvent s (VtyEvent ev) = do
-  let mList = s ^? uiState . uiInventory . _Just . _2
+handleRobotPanelEvent :: BrickEvent Name AppEvent -> EventM Name AppState ()
+handleRobotPanelEvent (Key V.KEnter) = do
+  focused <- gets focusedEntity
+  maybe continueWithoutRedraw descriptionModal focused
```
byorgey, line 822: Would `gets focusedEntity >>= maybe continueWithoutRedraw descriptionModal` be simpler?

src/Swarm/TUI/Controller.hs
```diff
@@ -808,27 +816,30 @@ adjustTPS (+/-) = uiState . lgTicksPerSecond %~ (+/- 1)
...
   case mList of
-    Nothing -> continueWithoutRedraw s
+    Nothing -> continueWithoutRedraw
     Just l -> do
-      l' <- handleListEventWithSeparators ev (is _Separator) l
-      let s' = s & uiState . uiInventory . _Just . _2 .~ l'
-      continue s'
-handleRobotPanelEvent s _ = continueWithoutRedraw s
+      l' <- nestEventM' l (handleListEventWithSeparators ev (is _Separator))
+      uiState . uiInventory . _Just . _2 .= l'
```
byorgey, line 835: Could this whole thing somehow be done more simply with `zoom (uiState . uiInventory . _Just . _2)`?

src/Swarm/TUI/Controller.hs
```diff
@@ -837,32 +848,31 @@ makeEntity s e = do
...
-      continue $
-        s
-          & gameState . replStatus .~ REPLWorking mkTy Nothing
-          & gameState . robotMap . ix 0 . machine .~ initMachine mkPT empty topStore
-          & gameState %~ execState (activateRobot 0)
-    _ -> continueWithoutRedraw s
+    Just False -> do
+      gameState . replStatus .= REPLWorking mkTy Nothing
+      gameState . robotMap . ix 0 . machine .= initMachine mkPT empty topStore
+      gameState %= execState (activateRobot 0)
```
byorgey, line 854: Would this be better as `zoom gameState $ activateRobot 0`?  (That would ultimately have the same effect, right?)

src/Swarm/TUI/Controller.hs
```diff
@@ -837,32 +848,31 @@ makeEntity s e = do
...
+    _ -> continueWithoutRedraw
 
 -- | Display a modal window with the description of an entity.
-descriptionModal :: AppState -> Entity -> EventM Name (Next AppState)
-descriptionModal s e =
-  continue $ s & uiState . uiModal ?~ generateModal s (DescriptionModal e)
+descriptionModal :: Entity -> EventM Name AppState ()
+descriptionModal e = do
+  s <- get
+  uiState . uiModal ?= generateModal s (DescriptionModal e)
```
byorgey, line 861: There are many places where we have to do `s <- get` and then call `generateModal s`.  Could we somehow rewrite `generateModal` to have a type like `(MonadReader AppState m) => generateModal :: ModalType -> m Modal` and then use some `zoom`-like thing to adapt it to `EventM Name AppState ()` ?

byorgey (commented):
This is wonderful!  The code is so much simpler and cleaner.

src/Swarm/TUI/Controller.hs
```diff
@@ -102,238 +102,245 @@ pattern FKey :: Int -> BrickEvent n e
...
-    case s ^? uiState . uiModal . _Just . modalDialog . to dialogSelection of
-      Just (Just QuitButton) -> quitGame s'
-      Just (Just (NextButton scene)) -> startGame scene s'
-      _ -> continue s'
+    toggleModal QuitModal
+    modal <- use $ uiState . uiModal
+    case modal ^? _Just . modalDialog . to dialogSelection of
+      Just (Just QuitButton) -> quitGame
+      Just (Just (NextButton scene)) -> startGame scene
+      _ -> return ()
```
xsebek, line 323: Found it! The original code got the dialogue selection on the original state `s` not `s'`.

src/Swarm/TUI/Controller.hs
xsebek: Already done :grinning:

src/Swarm/TUI/Controller.hs
byorgey: Aha, right!

byorgey:
Interestingly, optimization now seems to matter a lot.  I compiled it first with `--ghc-options='-O0'` just so it would compile faster, but the app was incredibly sluggish.  After rebuilding without that option, so it used the default `-O1`, it felt (I think) just as responsive as before.  Though we should probably do some benchmarking comparisons.

xsebek:
Interesting, maybe some monad transformer stack inlining is taking place?

I never tried `-O0`, does it improve the compilation speed a lot for Swarm?

Benchmarking without `-O1` optimizations does not make much sense to me and it would not benchmark the UI anyway. :shrug:

src/Swarm/TUI/Controller.hs
xsebek: Well, the only other place where its called is this:
```Haskell
uiState %= resetWithREPLForm (set promptUpdateL "" (s ^. uiState))
```
Honestly, I do not want to touch that `promptUpdateL` lens in this PR anymore.
It could cause more problems and I would not be sure what caused them. :fearful:

byorgey:
> Curious, maybe some monad transformer stack inlining is taking place?

Yes, that seems plausible.

> I never tried `-O0`, does it improve the compilation speed a lot for Swarm?

Yes, it makes a big difference.  Doing some informal timings on this branch, for me it takes 105s to build with `-O1` and only 20s to build with `-O0`.

byorgey:
Also, to be clear, I didn't mean we should benchmark the difference between `-O0` and `-O1`.  I meant we should benchmark the difference between this branch and `main`.

src/Swarm/TUI/Controller.hs
byorgey: I agree, we should not try to change it here.  Just something to think about for the future.

xsebek:
Yeah, I considered that you might have meant it that way. But as I said, would we even be able to benchmark it besides manually with stopwatches in hand?

This only changes the controller code which we do not benchmark now.

We could maybe profile it while running, but I have never played with that before in Haskell.

src/Swarm/TUI/Controller.hs
xsebek: I think not, we want to continue without redrawing the screen in the other case.

If it was just `return ()` then `zoom` would work.

src/Swarm/TUI/Controller.hs
xsebek: To me, 2 places are not that many. :wink: It might be worth it to rewrite all functions in View to use `MonadState AppState`, but they look kind of nicely simple now.

At the moment I am leaning to leave such refactoring to other PR, it is not obvious to me how to make it nice and keep it simple.

src/Swarm/TUI/Controller.hs
xsebek: That does not work, because the `fused-effect` signature does not find the right instance.

It is not that much boiler plate, so I can probably live with it for now.

xsebek:
@byorgey btw. I intend to merge this without waiting for the brick release.

It seems like a strict improvement to me and I do not want to go back to passing `AppState` through functions to `continue`. :sweat_smile:

src/Swarm/TUI/Controller.hs
byorgey: It looked like a lot more than 2 places to me.  In some places we do `flip generateModal`.  Anyway, leaving it for now is fine, it was just an offhanded suggestion.

byorgey (approved):

jtdaugherty:
FYI, I don't know when the Brick changes will get released. At this point I don't know of any show-stopping reasons why the changes should not go out, though. And it's helpful for me to know that they have been a win for swarm!

jtdaugherty:
W.r.t. your performance observations, I would be curious to know whether switching EventM to a lazy state monad makes an obvious difference.

byorgey:
Timing of the release is not a big deal, at this point we can easily just depend on a particular commit rather than depending on getting brick from Hackage.

src/Swarm/TUI/Controller.hs
xsebek: It might have been another function or maybe you looked/remembered the main branch?

I was actually looking forward to changing that function and then I came to the View module and found every function there looked like that! :sweat_smile:

src/Swarm/TUI/Controller.hs
xsebek: Sure, I almost want a name for a "gets maybe or no redraw" helper function, but that can be done later. :+1:

PR #585, opened 2022-07-23
Sequence of multiple winning objectives for one scenario

### LINKED ISSUE #378: Generalize scenarios to have a sequence of objectives instead of a single win condition
The goal would be to enable challenge scenarios with a *sequence* of conditions that have to be fulfilled in order.  Once you complete the first objective, it enables the second objective, and so on.

At the same time, I'd like to generalize a win condition to come with some text which will be displayed in a dialog when the condition is fulfilled.  *i.e.* right now it always says "Congratulations!" but we should be able to specify that along with the win condition.  This could be used to display a specific congratulatory message, but also to explain what the next objective will be.  I think we can also have a `goal` along with each condition which can be displayed somehow (see #356).

Right now we just have a text field `win` which should contain the text of a swarm program.  I propose generalizing this to something like the following:

```
objectives:
  - condition: |
      try {
        as base {has "lambda"}
      } { return false }
    goal: See if you can find a lambda.
    message: Congratulations, you found a lambda!  Now, see if you can find a tree.
  - condition: |
      try {
        as base {has "tree"}
      } { return false }
    goal: Find a tree.
    message: Good work, now you're ready for the next challenge!
```

Together with #355 this would give a very flexible framework for constructing multi-step challenges, which would be especially helpful for #25 .

byorgey:
Closes #378 .

- Old format of `goal` + `win` is still accepted, and is parsed as a single objective.
- Alternatively, an `objectives` list field can be specified, where each item consists of two fields, `goal` and `condition`.


src/Swarm/Game/State.hs
```diff
@@ -172,7 +174,7 @@ data WinCondition
     NoWinCondition
   | -- | The player has not won yet; this 'ProcessedTerm' of type @cmd
```
byorgey, line 175: Should update this comment.

data/scenarios/Tutorials/move.yaml
```diff
@@ -4,19 +4,20 @@ description: |
...
+      - To complete this challenge, move your robot two spaces to the right,
+        to the coordinates (2,0).
+      - Note that you can chain commands with ';'.
+      - You can open this popup window at any time to remind yourself of the goal
+        using Ctrl-G.
+    condition: |
+      try {
+        loc <- as base {whereami};
+        return (loc == (2,0))
+      } { return false }
```
byorgey, line 20: Put this back to the old style.

src/Swarm/TUI/Controller.hs
```diff
@@ -224,9 +222,8 @@ handleMainEvent s = \case
     s' <- toggleModal s CommandsModal
     continue (s' & gameState . availableCommands . notificationsCount .~ 0)
   ControlKey 'g' -> case s ^. uiState . uiGoal of
```
byorgey, line 224: Can we just call `toggleModal` instead of having a `case`?

byorgey: Ah, no, that would be different: the point is that `Ctrl-G` does nothing if there is no current goal.

src/Swarm/Game/Scenario.hs
```diff
@@ -224,16 +257,28 @@ instance FromJSONE EntityMap Scenario where
...
+    parseObjectives :: Object -> Parser [Objective]
+    parseObjectives v =
+      fromMaybe []
+        <$> liftA2
+          (<|>)
+          (v .:? "objectives")
+          ( do
+              g <- v .:? "goal" .!= []
+              (fmap . fmap) ((: []) . Objective g) (v .:? "win")
+          )
```
xsebek, line 281: If I did not see the type, I could not tell what this function does.

Do we need to support the old-style global goal&win syntax at all, since we do not save much on typing?

For new scenario writers, it would be better to have the YAML syntax simple with a minimum of exclusive alternatives.

data/scenarios/Testing/378-objectives.yaml
```diff
@@ -0,0 +1,64 @@
...
+      - grabber
+      - life support system
+      - logger
+      - toolkit
+      - solar panel
+      - workbench
+      - clock
+    inventory:
+      - [5, 3D printer]
+      - [100, treads]
```
xsebek, line 43: Why do we not check that base has to have `tank treads`? I refuse to build robots unless I have to! :smile:

data/scenarios/Testing/378-objectives.yaml
```diff
@@ -0,0 +1,64 @@
+name: Complete two objectives in succession.
+description: |
+  First, get some trees; then, use them to build a harvester.
```
xsebek, line 3: I would not mind including this as part of the Place tutorial. :wink:

data/scenarios/Testing/378-objectives.yaml
```diff
@@ -0,0 +1,64 @@
...
+      - solar panel
+      - workbench
+      - clock
+    inventory:
+      - [5, 3D printer]
+      - [100, treads]
+      - [70, grabber]
+      - [100, solar panel]
+      - [50, scanner]
+      - [5, toolkit]
```
xsebek, line 47: I prefer keeping the numbers the same, like 10, to make it easier to see what is important.
Also, `[0, harvester]` would be nice.

But for testing it of course does not matter much.

xsebek (approved):
LMCM - Looks like Merge Conflicts to Me! :sweat_smile:

src/Swarm/Game/Scenario.hs
byorgey: No, we don't need to support the old-style syntax, it is just a bunch of work to update them all :sweat_smile: .  But I think you're right that it would be better to just support one simple syntax.  I will update them and simplify this code.

data/scenarios/Testing/378-objectives.yaml
byorgey: I am including it as part of https://github.com/swarm-game/swarm/tree/multistep-tutorial .

data/scenarios/Testing/378-objectives.yaml
byorgey: Ah, right, I keep forgetting about this!  I've added #334 to the board for the alpha release project.

src/Swarm/Game/Scenario.hs
byorgey: 15 tedious minutes and many emacs macros later, I have updated all the existing test scenarios. :sweat_smile:

PR #590, opened 2022-07-25
Add heavy robots that require tank treads to move

### LINKED ISSUE #334: Prevent base from being able to move too early

byorgey:
Closes #334.

### DIFF
--- a/src/Swarm/Language/Capability.hs
+++ b/src/Swarm/Language/Capability.hs
@@ -37,6 +37,8 @@ data Capability
     CPower
   | -- | Execute the 'Move' command
     CMove
+  | -- | Execute the 'Move' command for a heavy robot
+    CMoveheavy
   | -- | Execute the 'Turn' command
     --
     -- NOTE: using cardinal directions is separate 'COrient' capability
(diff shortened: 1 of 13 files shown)


src/Swarm/Language/Capability.hs
```diff
@@ -37,6 +37,8 @@ data Capability
     CPower
   | -- | Execute the 'Move' command
     CMove
+  | -- | Execute the 'Move' command for a heavy robot
+    CMoveheavy
```
xsebek, line 41: Looking into the thesaurus, "locomote" or "mobile" sound good to me, but "move-heavy" is at least literal. :slightly_smiling_face:

xsebek (approved):
LGTM :+1:

src/Swarm/Language/Capability.hs
byorgey: Yeah, since users will never see these names at all, I'd prefer to make it more literal.

PR #591, opened 2022-07-25
Fix robot display

byorgey:
Use `renderDisplay (r ^. robotDisplay)` instead of `displayEntity (r ^. robotEntity)`, since the latter no longer takes into account the robot's direction.  Also get rid of the `displayEntity` function since it's unnecessary and tempting to use in this incorrect way.


xsebek (approved):
Nice, I just noticed yesterday that robots showed as `X` in the list. :+1:

PR #594, opened 2022-07-26
small improvement to capability checking for heavy robots

byorgey:
Previously, trying to execute `move` on the base generated an error
saying you needed `treads` or `tank treads`; but of course, installing
`treads` will not work.  This fixes the problem by moving the heavy
check from the specific runtime code for `Move` to the more general capability check before executing any command. 
Now executing `move` on the base will correctly tell you that you need to install `tank
treads`.

### DIFF
--- a/src/Swarm/Game/Step.hs
+++ b/src/Swarm/Game/Step.hs
@@ -264,19 +264,30 @@ traceLogShow = traceLog . from . show
 -- Exceptions and validation
 ------------------------------------------------------------
 
+-- | Capabilities needed for a specific robot to evaluate or execute a
+--   constant.  Right now, the only difference is whether the robot is
+--   heavy or not when executing the 'Move' command, but there might
+--   be other exceptions added in the future.
+constCapsFor :: Robot -> Const -> Maybe Capability
+constCapsFor r Move
+  | r ^. robotHeavy = Just CMoveheavy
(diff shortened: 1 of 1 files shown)


src/Swarm/Game/Step.hs
```diff
@@ -264,19 +264,30 @@ traceLogShow = traceLog . from . show
...
-  Just cap -> do
-    creative <- use creativeMode
-    sys <- use systemRobot
-    robotCaps <- use robotCapabilities
-    let hasCaps = cap `S.member` robotCaps
-    (sys || creative || hasCaps)
-      `holdsOr` Incapable FixByInstall (R.singletonCap cap) (TConst c)
+ensureCanExecute c = do
+  r <- get @Robot
+  case constCapsFor r c of
```
xsebek, line 282: If you flipped `constCapsFor` arguments, this could be simplified to `gets (constCapsFor c) >>= \case` :slightly_smiling_face: 

But it does not change anything except keeping the diff smaller.

xsebek (approved):

src/Swarm/Game/Step.hs
byorgey: Good idea, although the diff still doesn't end up being smaller since fourmolu's layout rules mean that `gets ...` still ends up on its own line with all the branches indented two more spaces.

PR #598, opened 2022-07-27
Add dependency config

xsebek:
- whitelist a few packages
- disable no new release fail
- disable single author fail


byorgey:
Looks good to me.

byorgey (approved):

PR #600, opened 2022-07-27
Make goal dialog scrollable; all dialogs now shrink/expand vertically

### LINKED ISSUE #592: Make goal dialog scrollable
When the terminal window is too small and there is a long goal description, the end of the goal simply gets cut off, with no indication that there is any additional text and no way to access it (other than making the terminal bigger).

The goal dialog should be scrollable, just like e.g. the commands dialog.

byorgey:
Closes #592 .

Also, refactor to use the same code path for all dialogs instead of having separate viewports, `Name` values, etc. for each different one.  All dialogs now use the slightly fancy `maybeScroll` function which adds a viewport only if the content of the dialog would be bigger than the available vertical space.  This way, the dialog will always take up as much vertical space as the content, and no more.


src/Swarm/TUI/View.hs
```diff
@@ -326,9 +326,26 @@ renderErrorDialog err = renderDialog (dialog (Just "Error") Nothing (maxModalWin
...
 drawDialog :: AppState -> Widget Name
 drawDialog s = case s ^. uiState . uiModal of
-  Just (Modal mt d) -> renderDialog d (withVScrollBars OnRight $ drawModal s mt)
+  Just (Modal mt d) -> renderDialog d (maybeScroll ModalViewport $ drawModal s mt)
   Nothing -> maybe emptyWidget renderErrorDialog (s ^. uiState . uiError)
 
+-- | Make a widget scrolling if it is bigger than the available
+--   vertical space.  Thanks to jtdaugherty for this code.
+maybeScroll :: (Ord n, Show n) => n -> Widget n -> Widget n
+maybeScroll vpName contents =
```
xsebek, line 335: Sumptuous! 🧐 (I like it 😉)

xsebek (approved):

src/Swarm/TUI/View.hs
byorgey: Haha, thanks! Though I can't claim much credit. :smile:

PR #601, opened 2022-07-28
make initial list of commands more complete

### LINKED ISSUE #588: Initial list of available commands should take into account devices in the inventory
When starting classic mode, the base has a bunch of devices installed (3D printer, workbench, etc.) but also a bunch of devices in its inventory which it can use to build more robots (treads, scanner, etc.).    Currently, the only commands that show up in the F4 Commands window is commands corresponding to *installed* devices, but we should also populate the initially available commands from devices in the inventory.  For example, the player should be able to see the `move` command immediately available --- although the base itself cannot `move`, the player can program robots to `move`.

Simply populate the initial set of commands from devices in the `robotInventory` as well as `installedDevices`.

byorgey:
Initial list of commands in the commands dialog now includes any commands which
- require a capability provided by an installed device
- require a capability provided by a device in the inventory, or
- do not require any capability.

Right now there are too many things in the list because we need to finish #26 , but that's a separate issue of course.

Closes #588 .


xsebek (approved):
Nice, this motivates me to finish #26. 😄

byorgey:
Yes, it's definitely extra motivation. :smile:

PR #603, opened 2022-07-28
A couple maze challenges

byorgey:


data/scenarios/Challenges/Mazes/loopy_maze.yaml
```diff
@@ -0,0 +1,87 @@
...
+      - [50, solar panel]
+      - [50, treads]
+      - [50, compass]
+      - [50, scanner]
+      - [50, lambda]
+      - [50, branch predictor]
+      - [50, strange loop]
+      - [50, logger]
+      - [50, grabber]
+      - [1000, rock]
```
xsebek, line 41: A cool upgrade to the challenge would be adding a workbench and requiring the shortest path of e.g. furnaces from start to finish. :fire:

xsebek (approved):

data/scenarios/Challenges/Mazes/invisible_maze.yaml
```diff
@@ -0,0 +1,97 @@
...
+      - [50, solar panel]
+      - [50, treads]
+      - [50, compass]
+      - [50, scanner]
+      - [50, lambda]
+      - [50, branch predictor]
+      - [50, strange loop]
+      - [50, logger]
+      - [50, grabber]
+      - [0, boulder]
```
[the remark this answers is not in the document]
xsebek: nvm, I missed that the boulders are on the edges.

data/scenarios/Challenges/Mazes/loopy_maze.yaml
byorgey: Ah, that does sound like fun.  How would you do it, though?  I thought a little bit about how to do BFS and I am not sure how to do it.  Hmm, maybe a sort of iterative deepening thing, where we explore everywhere we can get in one step, then everywhere we can get in at most two steps, etc...

PR #608, opened 2022-07-29
add a shortest path challenge

byorgey:
No solution yet... checking the solution will also be somewhat nontrivial.  I'm thinking that if we precompute the length of a shortest path, then we can write code to follow a marked path, make sure it doesn't branch, leads to the goal, and has the correct length.


xsebek:
@byorgey do you want to try BFS solution with infinite items? :slightly_smiling_face:

byorgey:
Nice, that should simplify things a good deal.  I will try to get back to this at some point soon.

PR #622, opened 2022-08-05
Use lazy event state

xsebek:
This is the lazy version of the state in `brick`'s new `EventM`.

Let's compare how it fares with `-O0` against:
- 0150b8a9 (refactor to `EventM` with strict state)
- 35ff8352 (just before refactor but with latest `brick`)

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -5,7 +5,7 @@ source-repository-package
 
 source-repository-package
   type: git
-  location: https://github.com/jtdaugherty/brick
-  tag: dee90efa8cde1755f435c1f7dd9b6828b1bd0089
+  location: https://github.com/xsebek/brick.git
+  tag: 757123f5b89c4d07a72851fb24f499bcf2efd979
 
 packages: .


byorgey:
- With `cabal run swarm --ghc-options='-O0'`, all three are extremely sluggish for me.  Starting a new classic game takes several seconds between hitting Enter and having the new world pop up.  After hitting Ctrl-Q I have to wait a second or so for the quit modal to pop up.
- With `cabal run swarm` (which seems to use `-O1` by default), all three feel nice and snappy at runtime (but they take much longer to compile).  I don't really notice any difference in terms of responsiveness of the UI.
- Strangely, with `stack build --fast` (which the docs say implies `-O0`), all are fast to compile and *also* feel snappy at runtime!  I do not understand how `stack build --fast` ends up getting the best of both worlds.

byorgey (commented):
- With `cabal run swarm --ghc-options='-O0'`, all three are extremely sluggish for me.  Starting a new classic game takes several seconds between hitting Enter and having the new world pop up.  After hitting Ctrl-Q I have to wait a second or so for the quit modal to pop up.
- With `cabal run swarm` (which seems to use `-O1` by default), all three feel nice and snappy at runtime (but they take _much_ longer to compile).  I don't really notice any difference in terms of responsiveness of the UI.
- Strangely, with `stack build --fast` (which the docs say implies `-O0`), all are fast to compile and *also* feel snappy at runtime!  I do not understand how `stack build --fast` ends up getting the best of both worlds.

byorgey:
So it turns out there is a big difference between passing `-O0` and `--ghc-options='-O0'` to `cabal run`.
- `cabal run --ghc-options='-O0'` seems to compile *all dependencies* using `-O0` as well as `swarm`. This leads to the long compile times and sluggish runtime performance I was seeing.
- `cabal run -O0` only compiles `swarm` itself with optimizations turned off.  It compiles quickly and runtime UI performance seems good.

PR #623, opened 2022-08-06
Add debug web service

TristanCacqueray:
This change introduce a new web API for the game.
The initial goal is to expose the internal state using the
`/robots` and `/robot/$id` endpoint to query the json representation
of the robots.

A new command line argument is added:
- When `--port 0` is set, the web service is disabled.
- When `--web port` is set, the game won't start if the web service fail to bind the port.
- Otherwise, when `--web` is not set, the service may start on 5357 if the port is available.

Example usage:

```
 swarm --web 5357
 curl localhost:5357/robots | jq
```

Future improvement may include:

- Ability to load program to the base (to send command from external text editor)
- Serve a html interface for alternative visualization

### DIFF
--- a/src/Swarm/App.hs
+++ b/src/Swarm/App.hs
@@ -70,7 +70,7 @@ appMain port seed scenario toRun cheat = do
         Just port' -> do
           -- Share a reference to the state with the web ui
           gsRef <- newIORef (s ^. gameState)
-          void $ forkIO $ webMain port' gsRef
+          Swarm.Web.startWebThread port' gsRef
           pure $ \e -> do
             s' <- get
             liftIO $ writeIORef gsRef (s' ^. gameState)
@@ -91,7 +91,7 @@ demoWeb = do
     Left errMsg -> T.putStrLn errMsg
     Right s -> do
       gsRef <- newIORef (s ^. gameState)
-      webMain 8080 gsRef
+      webMain Nothing 8080 gsRef
  where
   demoScenario = Just "./data/scenarios/Testing/475-wait-one.yaml"
 
--- a/src/Swarm/App.hs
+++ b/src/Swarm/App.hs
@@ -65,13 +65,12 @@ appMain port seed scenario toRun cheat = do
           threadDelay 33_333 -- cap maximum framerate at 30 FPS
           writeBChan chan Frame
 
-      eventHandler <- case port of
-        Nothing -> pure handleEvent
-        Just port' -> do
-          -- Share a reference to the state with the web ui
-          gsRef <- newIORef (s ^. gameState)
-          Swarm.Web.startWebThread port' gsRef
-          pure $ \e -> do
+      -- Start the web service with a reference to the game state
+      gsRef <- newIORef (s ^. gameState)
+      Swarm.Web.startWebThread port gsRef
+
+      -- Update the reference for every event
+      let eventHandler e = do
             s' <- get
             liftIO $ writeIORef gsRef (s' ^. gameState)
             handleEvent e
--- a/src/Swarm/App.hs
+++ b/src/Swarm/App.hs
@@ -11,29 +11,36 @@ module Swarm.App where
 import Brick
 import Brick.BChan
 import Control.Concurrent (forkIO, threadDelay)
+import Control.Lens ((^.))
 import Control.Monad.Except
+import Data.IORef (newIORef, writeIORef)
 import Data.Text.IO qualified as T
 import Graphics.Vty qualified as V
+import Network.Wai.Handler.Warp (Port)
+import Swarm.Game.State
 import Swarm.TUI.Attr
 import Swarm.TUI.Controller
 import Swarm.TUI.Model
 import Swarm.TUI.View
+import Swarm.Web
+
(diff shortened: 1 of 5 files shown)


src/Swarm/Web.hs
```diff
@@ -0,0 +1,41 @@
...
+mkApp gsRef =
+  robotsHandler
+    :<|> robotHandler
+ where
+  robotsHandler = do
+    g <- liftIO (readIORef gsRef)
+    pure $ IM.elems $ g ^. robotMap
+  robotHandler rid = do
+    g <- liftIO (readIORef gsRef)
+    pure $ fromMaybe (error "Unknown robot") (IM.lookup rid (g ^. robotMap))
```
byorgey, line 33: Currently, requesting a robot that doesn't exist generates a "connection was reset" error in the web browser, and spews a bunch of errors in the terminal.  How hard would it be to instead serve a proper error message via the web server?  That seems like it would be more reasonable, and wouldn't mess up the terminal display.

byorgey (approved):
Looks good to me, other than a couple suggestions for improvement.

src/Swarm/Web.hs
TristanCacqueray: Of course, using Maybe does the trick, and the handler is even shorter :)

src/Swarm/App.hs
```diff
@@ -58,17 +65,41 @@ appMain seed scenario toRun cheat = do
...
+      eventHandler <- case port of
+        Nothing -> pure handleEvent
+        Just port' -> do
+          -- Share a reference to the state with the web ui
+          gsRef <- newIORef (s ^. gameState)
+          void $ forkIO $ webMain port' gsRef
+          pure $ \e -> do
+            s' <- get
+            liftIO $ writeIORef gsRef (s' ^. gameState)
+            handleEvent e
```
TristanCacqueray, line 77: I think it would also make sense to always start the service by default and use this handler.

xsebek: I am not against it, but what will happen if 8080 is already taken?

src/Swarm/Web.hs
```diff
@@ -0,0 +1,55 @@
...
+-- |
+-- A web service for Swarm.
+--
+-- The service can be started using the `--port 8080` command line argument,
+-- or through the REPL by calling `Swarm.App.demoWeb`.
+--
+-- Once running, here are the available endpoints:
+--
+--   * /robots : return the list of robots
+--   * /robot/ID : return a single robot identified by its id
```
xsebek, line 14: Running `curl localhost:8080/robots/1` or even `curl localhost:8080/robots?id=1` does not work for me. :confused:

xsebek:
Hi @TristanCacqueray, thanks for working on this, I would love to get some REST API for Swarm. :+1:

The autogenerated ToJSON instances will need some tweaking, but it's great having easy access to data.
For reference, this is how to get the robot IDs:
```bash
curl --silent localhost:8080/robots | jq '.[]._robotID'
```

src/Swarm/Web.hs
TristanCacqueray: @xsebek the path should be `/robot/1`, without the plural s

src/Swarm/App.hs
TristanCacqueray: Depending on how important the service is, we could:
- fail to start the game
- pick the next available port (and add the info to the help page or something)
- don't start the api

Or perhaps, we could make the game fail to start only if the user explicitly provide a port number, otherwise we simply continue without the api.

Also we could use a less common port, for example 5357 (which is SW in hex)

byorgey: Failing to start just because the default port is already taken sounds like a terrible option.  Otherwise I don't have a strong opinion.

src/Swarm/Web.hs
xsebek: Oh, I am not a web dev so this feels a bit surprising (also `/robot` silently gives nothing), but I guess it is easier this way.

xsebek (approved):

src/Swarm/App.hs
byorgey: I agree re: always starting the service by default, perhaps with a way to explicitly disable it?

xsebek: I also vote to continue without API, but I was thinking more about discoverability from the player's perspective.
Basically logging anything to stderr or stdout does not make sense - you will only see that after closing the game.

A simple `txt` widget in <kbd>F1</kbd> Help modal would work better IMO, but that one seems buggy atm. We can work on that later though, a (new?) config overview modal could be a lot of work.

TristanCacqueray: Alright, I think the last commit and the updated description should resolve this conversation.

TristanCacqueray: @xsebek about user notification, in #627 I suggest we re-use the existing message queue.

byorgey (approved):
Looks great!

xsebek (approved):
LGTM! :partying_face: :rocket:

TristanCacqueray:
Thank you for the prompt feedback!

PR #628, opened 2022-08-06
Lower the maximum modal height

### LINKED ISSUE #626: The F1 Help modal is too greedy
The F1 Help modal loads slowly and then has seemingly infinite empty space under the two tables:
Screenshot from 2022-08-06 20-03-38
<sup>I took the picture on the web branch, but it is the same on main.</sup>

Scrolling is terribly slow too. What is going on?!

---

After looking at the code, the culprit is the new `maybeScroll` function (da4db6e61) which lets Help greedily take too much space.

xsebek:
- lower the maximum height for scrolling to 10 000
  - one million was too slow to scroll and load when filled by greedy widgets
- make the Help widget less greedy
  - `<+> fill ' '` literally fills the rest with space, thanks for spotting that @byorgey :)
- closes #626

### DIFF
--- a/src/Swarm/TUI/View.hs
+++ b/src/Swarm/TUI/View.hs
@@ -321,7 +321,7 @@ maybeScroll :: (Ord n, Show n) => n -> Widget n -> Widget n
 maybeScroll vpName contents =
   Widget Greedy Greedy $ do
     ctx <- getContext
-    result <- withReaderT (availHeightL .~ 1000000) (render contents)
+    result <- withReaderT (availHeightL .~ 10000) (render contents)
     if V.imageHeight (result ^. imageL) <= ctx ^. availHeightL
       then return result
       else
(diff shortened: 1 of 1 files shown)


src/Swarm/TUI/View.hs
```diff
@@ -321,7 +321,7 @@ maybeScroll :: (Ord n, Show n) => n -> Widget n -> Widget n
 maybeScroll vpName contents =
   Widget Greedy Greedy $ do
     ctx <- getContext
-    result <- withReaderT (availHeightL .~ 1000000) (render contents)
+    result <- withReaderT (availHeightL .~ 10000) (render contents)
```
byorgey, line 324: What's the max message queue size?  Could it ever reach 10000 lines?

byorgey (approved):
Seems reasonable, though I don't remember how big the message queue can get.

src/Swarm/TUI/View.hs
xsebek: Hah, it could be 10 001 because of the padding maybe? Also, I guess line overflow and multiline messages can make it arbitrarily long. I will test it.

xsebek: It does not work anyway, so I filed #629, let's fix that one separately. :disappointed:

PR #620, opened 2022-08-05
Add scenario json schema

xsebek:
- added JSON schemas for scenarios, entities and recipes
- closes #346

PR #630, opened 2022-08-06
Allow non-Haskell PRs to be merged

xsebek:
Hopefully, this fixes the problems with the PRs like #620 being stuck in merge queues.

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -1,8 +1,18 @@
 queue_rules:
   - name: default
     conditions:
-      - check-success=Haskell-CI - Linux - ghc-9.0.2
-      - check-success=Haskell-CI - Linux - ghc-8.10.7
+      - or:
+        - and:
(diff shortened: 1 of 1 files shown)


.mergify.yml
```diff
@@ -1,8 +1,18 @@
...
+          - -files~=\.hs$
+          - -files~=\.sw$
+          - -files~=\.cabal$
+          - -files=stack.yaml
+          - -files~=cabal.project.*
+          - -files~=^[^/]*\.yaml$
+          - -files~=^data/.*\.yaml$
+        - and:
+          - check-success=Haskell-CI - Linux - ghc-9.0.2
+          - check-success=Haskell-CI - Linux - ghc-8.10.7
```
byorgey, line 15: Don't we check on GHC 9.2 now also?  Why isn't that listed here?

Issue #587, opened 2022-07-24
Reject unknown fields when parsing `.yaml` files


byorgey:
**Is your feature request related to a problem? Please describe.**
I just spent a frustrating 20 minutes trying to figure out why my scenario didn't work the way I wanted, until I finally realized that the `seed:` field should be at the top level, rather than indented under `world:`.  No error was generated because (1) the `seed` is optional, and (2) unknown fields are simply ignored.

**Describe the solution you'd like**
We should have a standard way to check for, and reject, unknown/misplaced/misspelled fields.  For example, see this StackOverflow question: https://stackoverflow.com/questions/67375099/prevent-unknown-field-names-in-aeson-parsejson

xsebek:
We could show a Warning on it and ask the player to confirm:
```
$ swarm --scenario MyScenario.yaml
WARNING: unrecognized field in scenario file (15:2):
- "gaol"
+ did you mean "goal"?
There was 1 warning generated for the scenario file "MyScenario.yaml"!
**Would you like to play it anyway? Type [y]es or [n]o:**
> no please just fix it for me
**Would you like to play it anyway? Type [y]es or [n]o:**
> n
```
We should check for such warnings in tests of course. :+1:

byorgey:
Well that is a lot fancier than what I was thinking, but sure, why not. :smile:

byorgey:
According to the link, for types where the `FromJSON` instance is derived generically, we can just do something like
```
instance FromJSON MyType where
  parseJSON = genericParseJSON $ defaultOptions
    { rejectUnknownFields = True }
```
For other types where we have a custom `FromJSON` instance, we'll have to do more manual checking.  In the linked StackOverflow answer, this happens by checking the keys of the `HashMap` we get from `aeson`.

kostmo:
This is now implemented: https://github.com/swarm-game/swarm/pull/1475#issuecomment-1694890993

Issue #50, opened 2021-09-24
Saving/loading games


byorgey:
It should be possible to save the current state of your game, and then later to load from a saved game.  There is nothing theoretically difficult about this, but it will take a good deal of engineering.  Off the top of my head, a save would need to include.

* A version number, so that we can do something reasonable with saves created by a previous version of the game
* The current state of the world:
    * The seed used to generate the world
    * Some kind of version number to identify the world function being used, so that we can do something reasonable with old saves
    * Which tiles have been loaded
    * The contents of those tiles (either the complete contents, or perhaps just a catalog of the cells whose entity contents have changed from their initial value)
* All (active and inactive) robots, including their CEK machine state.
* Note, when saving entities (either in the world map, or in an inventory, or a robot entity), if the hash matches the hash of a standard entity, we only have to save the hash.  Otherwise we must serialize the entire entity.
* Probably various other fields of the `GameState` and/or `UIState`.

Note, a related issue is #7 .

xsebek:
Thanks for making this issue! I think a more directly related issue is Generate world by tiles (#29).

I could then save loaded tiles and generate new ones in _whatever is the current world generation function_.
Thus, if I save the game on v1.41.0 and load it on v1.42.0, I could send robots to explore tiles with new features from version 1.42.0. 

If the save was done per square, then changing world generation could introduce problems (:mountain:), but with tiles, I do not care what would have been on the unexplored ones in the previous version.

**TLDR** I would argue for the world state to be the **seed** and **complete tiles**, so versioning world generation is not a headache.

byorgey:
Hmm, thanks, that's definitely worth considering!  That would certainly simplify the need to keep track of versions for world generation.  On the other hand, it might create weird seams at tile boundaries where e.g. half a lake on a previously loaded tile is suddenly right next to half a mountain on a new tile.  But I'm not sure how important that is.

xsebek:
I wonder if we could dramatically simplify this by implementing #29, #7 and give each entity a unique character and actually try storing the game as YAML.
```yaml
- robots:
  ...
- seed: 0
- world:
  - tile:
    - x: 0
    - y: 0
    - map: >
      +------------------------+
      |   *       AAA   ΩTTT   |
      | ~~ *    AAAÅÅ     TT   |
```
Most users likely will not change that many tiles, so the map will not take much space.

I guess with #7, robot definitions should not be an unbound number of lines either, but I do not know how many robots would need to keep a non-default state and how long that would be.

[...]

PR #631, opened 2022-08-07
Save scenario completion status

### LINKED ISSUE #357: Save scenario completion status
From #296 : 

> Save challenge status (Not started/In progress/Finished) in XDG_WHATEVER_DIR (for browsing in the menu)

Some kind of icons could then be shown in the "New game" menu showing the status of each scenario.  For now, there are just two states: completed / not completed.  Once we have #50 we can also have an "in progress" status.

xsebek:
- save the latest and best completion status for a scenario
  - scenarios are either `not started`/`in progress`/`completed` with the latter two also having time (real and in ticks)
- do not save completion status for `--cheat`ers
- show the completion status in the UI
- closes #357 


xsebek:
It does not fully work yet and I have put debugging logs everywhere, but I can finally see the status displayed:
Screenshot from 2022-08-07 19-29-56

src/Swarm/Game/ScenarioStatus.hs
```diff
@@ -0,0 +1,319 @@
...
+
+instance Eq ZonedTime where
+  (==) = (==) `on` zonedTimeToUTC
+
+instance Ord ZonedTime where
+  (<=) = (<=) `on` zonedTimeToUTC
+
+data ScenarioStatus
+  = NotStarted
+  | InProgress
```
byorgey, line 83: Since we can't save games yet (#50), I guess "in progress" at this point really means "you tried it but gave up"?  In that case I don't know why we really need `_scenarioElapsed`.

This whole business with keeping track of elapsed time is tricky, and I don't necessarily have deeply thought reasons for my use of `ZonedTime` and `NominalDiffTime`.  In general, there are a few things at play here.  First, according to the reasoning at https://github.com/swarm-game/swarm/wiki/Speedrunning , I think the way to measure speedruns is to measure wall clock time and disallow adjusting the ticks/second rate.   However, when deciding how many ticks to run each frame etc., we are very careful to use a monotonic clock so that e.g. the game does not freeze while suddenly trying to run an hour's worth of ticks when the time zone changes or whatever.  So simply using `ZonedTime` is actually a bit problematic in case your game takes place across some kind of clock shift.  Of course this is not going to be common, but if we can solve the problem properly, then why not.  A third issue is that people are going to be used to seeing the number of ticks displayed using the clock widget, so it might be nice to also record the number of ticks taken.  Usually this will correspond with the wall clock time, but not if the player explicitly adjusted the tick speed, or if the game slowed down under heavy computational load.

src/Swarm/Game/ScenarioStatus.hs
```diff
@@ -0,0 +1,319 @@
...
+  parseJSON = genericParseJSON scenarioOptions
+
+instance ToJSON ScenarioInfo where
+  toEncoding = genericToEncoding scenarioOptions
+  toJSON = genericToJSON scenarioOptions
+
+scenarioOptions :: Options
+scenarioOptions =
+  defaultOptions
+    { fieldLabelModifier = map toLower . drop (length "_scenario")
```
byorgey, line 121: Cool, didn't know about this!

src/Swarm/Game/ScenarioStatus.hs
```diff
@@ -0,0 +1,319 @@
...
+
+-- | The path of the scenario, relative to @data/scenarios@.
+scenarioPath :: Lens' ScenarioInfo FilePath
+
+-- | The status of the scenario.
+scenarioStatus :: Lens' ScenarioInfo ScenarioStatus
+
+-- | The best status of the scenario.
+scenarioBest :: Lens' ScenarioInfo ScenarioStatus
+updateScenarioInfoOnQuit :: ZonedTime -> Bool -> ScenarioInfo -> ScenarioInfo
```
byorgey, line 134: Can you add some vertical whitespace and a comment here?

src/Swarm/Game/State.hs
```diff
@@ -807,8 +818,16 @@ buildWorld em (WorldDescription {..}) = (robots, first fromEnum . wf)
...
+  (scene, path) <- loadScenario sceneName (g ^. entityMap)
+  gs <- liftIO $ scenarioToGameState scene userSeed toRun g
+  normalPath <- liftIO $ normalizeScenarioPath (gs ^. scenarios) path
+  t <- liftIO getZonedTime
+  liftIO . appendFile "/tmp/debug" $ "TODO: S - Scenario in progress " <> normalPath <> "\n"
+  liftIO . appendFile "/tmp/debug" $ "TODO: S - Originally " <> path <> "\n"
+  return $
+    gs
+      & currentScenarioPath ?~ normalPath
+      & scenarios . scenarioItemByPath normalPath . _SISingle . _2 . scenarioStatus .~ InProgress t 0
```
byorgey, line 830: It seems like this code is repeated several times and should probably be factored out somehow.  Relatedly, I am not 100% sure why we need a separate `initGameStateForScenario` function.  Is it so we avoid having to load all the scenarios?  I honestly don't really remember.  I think it only gets used in `classicGame0` below, which in turn is only used in the test and benchmark suites.  It seems like we should be able to refactor and delete a bunch of code somehow.

src/Swarm/TUI/Model.hs
```diff
@@ -808,8 +818,16 @@ initAppState userSeed scenarioName toRun cheatMode = do
...
+      (scenario, path) <- loadScenario (fromMaybe "classic" scenarioName) (gs ^. entityMap)
+      as <- liftIO $ execStateT (scenarioToAppState scenario userSeed toRun) (AppState gs ui)
+      normalPath <- liftIO $ normalizeScenarioPath (as ^. gameState . scenarios) path
+      liftIO . appendFile "/tmp/debug" $ "TODO: M - Scenario in progress " <> normalPath <> "\n"
+      liftIO . appendFile "/tmp/debug" $ "TODO: M - Originally " <> path <> "\n"
+      t <- liftIO getZonedTime
+      return $
+        as
+          & gameState . currentScenarioPath ?~ normalPath
+          & gameState . scenarios . scenarioItemByPath normalPath . _SISingle . _2 . scenarioStatus .~ InProgress t 0
```
byorgey, line 830: This code definitely shouldn't be duplicated both here and in `startGame`.  Can we just call `startGame` here?

byorgey (commented):
This is coming along nicely.

One general suggestion: we should only show scenario status for those scenarios which have objectives.  For example it seems weird to show the status for the Classic game when there is no way to win.  So it will always show "in progress".

For posterity, recording my IRC comments here re: why the scenario status in the menu currently doesn't update until you exit and re-enter the New Game menu:

```
ah, I noticed that if you quit to the main menu and then enter the "New Game" menu again, the scenario status indicators are updated.
right, so the problem is that the NewGameMenu constructor stores a list of ScenarioItems, which have ScenarioStatus in them
Previously I was thinking of ScenarioItems as static read-only things, loaded at program startup, so it was no big deal to "duplicate" them in the current menu state (under the hood it would just be another pointer to the value of course).
But obviously that doesn't work now.  Maybe the NewGameMenu constructor should store paths instead of actual ScenarioItems, and we can use the fancy new scenarioItemByPath traversal to look up the corresponding ScenarioItems when rendering the menu
that way the ScenarioCollection in the GameState will be the single source of ground truth for scenarios and their status
```

src/Swarm/Game/ScenarioStatus.hs
xsebek: > "you tried it but gave up"?

I would not be so pessimistic. :slightly_smiling_face: Three yellow dots are not that bad and I tweaked the update function so it will show the longest time taken while "in progress". So it just means you had "fun" with classic mode for a number of hours. :wink: 

>  So simply using ZonedTime is actually a bit problematic...

Is it? I convert back to UTC when computing elapsed time, so it should be safe AFAIK. :thinking: 

I wanted to keep the time zone in case we wanted to show the time somewhere in-game which we currently do not.

> ...people are going to be used to seeing the number of ticks displayed using the clock widget...

Absolutely! I already added it to the list of tasks in the description, should be easy to add. :+1:

It also makes more sense to measure in ticks if you are loading the solution from a file, then it is just algorithmic complexity and wall clock time does not matter.

Plus the hex clock is cool! :grinning:

src/Swarm/Game/ScenarioStatus.hs
xsebek: It was mentioned in the SO answer you linked in #587. :wink:

There are more [`Options`](https://hackage.haskell.org/package/aeson-2.1.0.0/docs/Data-Aeson.html#t:Options), and they make the YAML more readable, so I'll tweak the other instances in the future too.

src/Swarm/Game/State.hs
xsebek: Oh yes, I discovered these places incrementally, so I just copy-pasted to get it working.

I think it's just that the initialization functions do not compose well, some take a string, some maybe string and some scenario. It used to not matter because the functions were short, but now it would be annoying to keep in sync.

I will try to refactor this a bit, it should be a small change.

src/Swarm/Game/ScenarioStatus.hs
xsebek: >  So simply using ZonedTime is actually a bit problematic...

Is it? I convert back to UTC when computing elapsed time, so it should be safe AFAIK. :thinking: 

I wanted to keep the time zone in case we wanted to show the time somewhere in-game which we currently do not.

xsebek: > ...people are going to be used to seeing the number of ticks displayed using the clock widget...

Absolutely! I already added it to the list of tasks in the description, should be easy to add. :+1:

It also makes more sense to measure in ticks if you are loading the solution from a file, then it is just algorithmic complexity and wall clock time does not matter.

Plus the hex clock is cool! :grinning:

byorgey: > Is it? I convert back to UTC when computing elapsed time, so it should be safe AFAIK. thinking
> 
> I wanted to keep the time zone in case we wanted to show the time somewhere in-game which we currently do not.

I think I was not clear.  There is nothing wrong with having a time zone, and indeed, we should have it.  What I meant was that if e.g. you start playing a scenario and the computer clock gets reset for some reason then the elapsed time will be incorrect.  I was thinking of e.g. the switch from standard to summer/daylight saving time, though now that I think of it, that will actually be correctly handled since the time zone will change as well.  The elapsed time can still be incorrect if the computer clock gets reset for some other reason (like if you intentionally set it to a different time) but maybe it's not worth worrying about that.

Once we can save games, then we will also need to worry about keeping track of the time elapsed while actually playing the game, as opposed to just the time between start and completion, but we can worry about that when we get to it.

byorgey: > Three yellow dots are not that bad and I tweaked the update function so it will show the longest time taken while "in progress". So it just means you had "fun" with classic mode for a number of hours.

Sure, I have no problem with showing the longest time you have played, that is fun.  But showing three yellow dots still seems strange to me.  If a scenario has no win condition, then it is not really "in progress", it just means "you have played this scenario before", but that does not seem like a useful thing to show in the UI, and three yellow dots does not seem to communicate the right thing.

Concretely, I would suggest:
- Remove the `o` status indicator for scenarios which have not been started.  We don't really need an icon for this state and all the `o`s just seem like visual noise to me.
- Don't show a status indicator for scenarios with no objectives (but we can still show the longest time played on the right).

xsebek: I still do not see the problem with time zones. The way I understand it is that when I ask for the current time, the computer will give me a precise number. I will also ask for the timezone and store it together with the precise time. This way I can show the time in a meaningful format and perform operations like subtracting an hour, where daylight saving time can complicate things.

But our use case is very simple:
1. get time at the start of the game
2. get time at the end of the game
3. compare them to get elapsed time

The last step should not be affected by timezone gimmicks, because we can convert from `ZonedTime` to `UTCTime` and the `time` library will handle leap seconds for us.

At least that's how I optimistically see it. :smile:

byorgey: Yes, everything you said is true, and I was confused about daylight saving time being a problem.  That and leap seconds will be correctly handled. The only issue (which, to reiterate, is a very small issue and not worth dealing with) is that the computer clock can be changed while the game is running.  Either by e.g. the user explicitly setting the clock, or in some more subtle way e.g. `ntp` adjusting the clock.  So if you say "give me the current ZonedTime" and then later say "give me the current ZonedTime", you are not 100% guaranteed that their difference is the actual amount of elapsed time.

xsebek: Well if we did not trust the players, then saving the high score (and game saves in future) in a human-readable YAML might be a bigger problem. :smile:

xsebek:
@byorgey I did not have time to finish this, if you want, you can take it over. 🙂

byorgey:
@xsebek OK, no problem, I think I can probably finish it up.

byorgey:
@xsebek when you wrote "add documentation", what kind of documentation did you have in mind?  Just Haddock comments? Or something else?

xsebek:
@byorgey just haddock :)

byorgey:
I tried just leaving the `ScenarioItem` records in the `NewGameMenu` and updating them to make sure they stayed in sync with the ones in the `ScenarioCollection` --- but that turned out to be a terrible idea, I immediately ran into bugs involving things getting out of sync.  I think we just have to bite the bullet and store canonical paths in the `NewGameMenu` instead of `ScenarioInfo` records.

xsebek:
@byorgey why do we need “path**s**” though?

AFAIK we store a stack of scenarios to know where to return and what was selected. But since the structure is a file _tree_, that should be covered by one file path right?

byorgey:
> But since the structure is a file _tree_, that should be covered by one file path right?

Oh hey, good point!

byorgey:
Hmm, just storing a single file path is conceptually appealing, but it means we would have to reconstruct a brick `List` from scratch every time anything changes.

I ended up going a different way, which I'm reasonably happy with.  I wrote a function that constructs a `NewGameMenu` (with an appropriate stack of menu lists, with each list focused on the appropriate item) from a scenario path.  Then I simply added some code to reconstruct the `NewGameMenu` from scratch every time we update some `ScenarioStatus`, to make sure it has the latest status info.

byorgey:
I think this is finally ready!  @xsebek , want to try it out?

xsebek:
Sure, I will try it on Monday when I get back to my computer. 👀

src/Swarm/Game/State.hs
```diff
@@ -823,8 +834,14 @@ buildWorld em (WorldDescription {..}) = (robots, first fromEnum . wf)
...
-  scene <- loadScenario sceneName (g ^. entityMap)
-  liftIO $ scenarioToGameState scene userSeed toRun g
+  (scene, path) <- loadScenario sceneName (g ^. entityMap)
+  gs <- liftIO $ scenarioToGameState scene userSeed toRun g
+  normalPath <- liftIO $ normalizeScenarioPath (gs ^. scenarios) path
+  t <- liftIO getZonedTime
+  return $
+    gs
+      & currentScenarioPath ?~ normalPath
+      & scenarios . scenarioItemByPath normalPath . _SISingle . _2 . scenarioStatus .~ InProgress t 0 0
```
xsebek, line 844: I am unsure where, but we broke `--run`. For example:
```bash
swarm --run <(echo 'say "Hi!"')
```
works on the main but not this branch. :fearful:

src/Swarm/TUI/View.hs
```diff
@@ -132,17 +136,43 @@ drawNewGameMenuUI (l :| ls) =
...
           ]
       , padLeft (Pad 5) (maybe (txt "") (drawDescription . snd) (BL.listSelectedElement l))
       ]
  where
-  drawScenarioItem (SISingle s) = padRight Max . txt $ s ^. scenarioName
-  drawScenarioItem (SICollection nm _) = padRight Max (txt nm) <+> withAttr robotAttr (txt ">")
+  drawScenarioItem (SISingle s si) = padRight (Pad 1) (drawStatusInfo s si) <+> txt (s ^. scenarioName)
+  drawScenarioItem (SICollection nm _) = padRight (Pad 1) (withAttr boldAttr $ txt " > ") <+> txt nm
+  drawStatusInfo s si = case s ^. scenarioObjectives of
+    [] -> txt "   "
```
xsebek, line 151: Visually, I would like some character for the scenarios that do not have an objective.

How about moving this `case` to the `InProgress` branch and using for example:
- `*`
- `∘`
- [`ю`](https://www.compart.com/en/unicode/U+044E) (Cyrillic yu) or some other modified `o`
- something else?

src/Swarm/Game/ScenarioInfo.hs
```diff
@@ -0,0 +1,328 @@
...
+scenarioPath :: Lens' ScenarioInfo FilePath
+
+-- | The status of the scenario.
+scenarioStatus :: Lens' ScenarioInfo ScenarioStatus
+
+-- | The best status of the scenario.
+scenarioBest :: Lens' ScenarioInfo ScenarioStatus
+
+-- | Update the current @ScenarioInfo@ record when quitting a game.
+updateScenarioInfoOnQuit :: ZonedTime -> Integer -> Bool -> ScenarioInfo -> ScenarioInfo
```
xsebek, line 150: Maybe the docstring should mention that the "best" time is either shortest (once completed) or longest (while playing, which works well for Classic mode).

There is a comment in `updateScenarioInfoOnQuit` saying so, but we should agree on the desired behaviour.

src/Swarm/TUI/View.hs
```diff
@@ -132,17 +136,43 @@ drawNewGameMenuUI (l :| ls) =
...
+        ]
+    Complete _s e t ->
+      withAttr greenAttr . vBox $
+        [ txt $ "completed in " <> formatTimeDiff e
+        , hBox
+            [ txt "("
+            , drawTime t True
+            , txt " ticks)"
+            ]
+        ]
```
xsebek, line 172: I just played with it and realized that the best time in ticks and real-time can be different. :confused: 

Maybe the best thing is to simply have a _best time_ and _best ticks_?
The UI could collapse that to one _best_ if they are the same and otherwise split it into two.

The `updateScenarioInfoOnQuit` would also need to be changed.

src/Swarm/Game/State.hs
xsebek: Found it! The option was not passed on in [`initAppState`.](https://github.com/swarm-game/swarm/pull/631/files#diff-e01da9c26d57a50bfdf719aaf9d0cf53e7f739146a8ec0db5bdea8fc304d3b07R861)

src/Swarm/TUI/View.hs
xsebek: I tried circle characters which IMO look very nice:
|Screenshot from 2022-08-29 18-49-16|image|
|--|-|

What do you think, @byorgey, will these be supported? :grinning:

src/Swarm/TUI/View.hs
xsebek: OK, this will probably be rare, but I think it looks good this way:

|Same|Different|
|-|-|
|image|Screenshot from 2022-08-29 19-39-08|


What do you think, @byorgey? :slightly_smiling_face:

src/Swarm/Game/ScenarioInfo.hs
xsebek: I added a small explanation:

> Note that when comparing "best" times, shorter is not always better!
As long as the scenario is not completed (e.g. some do not have a win condition) we consider having fun _longer_ to be better.

src/Swarm/TUI/View.hs
xsebek: I tried the mono fonts on my system (Gnome Fedora) and they all rendered fine. :man_shrugging:

src/Swarm/TUI/View.hs
byorgey: Well, ultimately I don't know if I want to give people an incentive to complete scenarios in the fewest ticks possible, since that just gives you an incentive to slow the game way down so you can type a lot on each tick, and that seems boring. :yawning_face:   But I think it is fine for now and we can maybe refine it later.

byorgey: To expand on the above comment a bit:
- Measuring *fewest ticks* incentivizes players to slow the game down a lot.  This is boring.
- Measuring *fastest time* incentivizes players to speed the game up a lot.  This favors players with beefy computers.
- So I think a good compromise for speedruns is to measure fastest time, but forbid adjusting the tick speed.
    - Eventually (but not in this PR) I'd like to have it track whether you adjusted the tick speed at all, and have a best time where you didn't adjust the tick speed, and a separate best time where you did adjust the tick speed.  Maybe also a flag `--speedrun` that disables adjusting the tick speed.

src/Swarm/TUI/Model.hs
```diff
@@ -860,18 +860,18 @@ initAppState userSeed scenarioName toRun cheatMode = do
...
-startGame scene si = do
+startGame :: (MonadIO m, MonadState AppState m) => Scenario -> ScenarioInfo -> Maybe FilePath -> m ()
+startGame scene si toRun = do
   t <- liftIO getZonedTime
   ss <- use $ gameState . scenarios
   p <- liftIO $ normalizeScenarioPath ss (si ^. scenarioPath)
   gameState . currentScenarioPath .= Just p
   gameState . scenarios . scenarioItemByPath p . _SISingle . _2 . scenarioStatus .= InProgress t 0 0
-  scenarioToAppState scene Nothing Nothing
+  scenarioToAppState scene Nothing toRun
```
byorgey, line 874: Good catch!

byorgey (approved):
I think this is looking good!  Love the collaborative PR. :smile:

src/Swarm/TUI/View.hs
xsebek: Well for me it was just about getting the best possible tick count for challenges which I can solve with one command/algorithm.

For move its just 3, but in later challenges I will probably care about the number of digits. 😄

Given that tick clock almost matches seconds, it’s just a way to not count pausing e.g. when reading the goal. 🤷

xsebek:
Thanks @byorgey, I enjoyed collaborating on this too! 

We'll just need to get an extra maintainer for review. 😁

src/Swarm/TUI/View.hs
byorgey: Oh, I see, that's a good point.  So you could write some code in a file and then start it with `--run` to get the fewest ticks.  I guess that's not actually so boring after all. :smile:

PR #633, opened 2022-08-08
Improve help modal

### LINKED ISSUE #359: Provide a mechanism for learning the current seed
Currently, if you start a game with a randomly chosen seed, there is no way to find out what seed was used, which means that there is no way to play the same world again later, or tell a friend the seed so they can play the same super-cool world you discovered, etc.  There should be some mechanism for displaying the chosen seed to the player. Perhaps it could simply be displayed somewhere in the UI, or perhaps there should be some kind of key command for showing the seed.  I don't think there should be a robot command for returning the seed, since it seems to be mixing levels in a strange way: the seed used is a meta-aspect at the level of the game, but it is not something robots would ever be able to learn from within the world.

byorgey:
Improve the Help (F1) modal dialog.

- Get rid of the list of commands; they are now available in Commands modal
- Improve formatting of keybindings list, and add more global keybindings to the list
- Add some text with pointers to the wiki and IRC
- Display current seed (closes #359)
- Display current web API port

### DIFF
--- a/src/Swarm/TUI/View.hs
+++ b/src/Swarm/TUI/View.hs
@@ -335,7 +335,7 @@ maybeScroll vpName contents =
 -- | Draw one of the various types of modal dialog.
 drawModal :: AppState -> ModalType -> Widget Name
 drawModal s = \case
-  HelpModal -> helpWidget Nothing Nothing -- XXX save seed & port
+  HelpModal -> helpWidget (s ^. gameState . seed) Nothing -- XXX save port
   RobotsModal -> robotsListWidget s
   RecipesModal -> availableListWidget (s ^. gameState) RecipeList
   CommandsModal -> availableListWidget (s ^. gameState) CommandList
@@ -466,8 +466,8 @@ robotsListWidget s = hCenter table
   debugging = creative && cheat
   g = s ^. gameState
 
-helpWidget :: Maybe Seed -> Maybe Port -> Widget Name
-helpWidget mseed mport =
+helpWidget :: Seed -> Maybe Port -> Widget Name
+helpWidget theSeed mport =
   padTop (Pad 1) $
     hBox . map (padLeftRight 2) $
       [helpKeys, tips <=> padTop (Pad 1) info]
@@ -481,8 +481,8 @@ helpWidget mseed mport =
       ]
   info =
     vBox $
-      [txt $ "Current seed: " <> into @Text (show seed) | Just seed <- [mseed]]
-      ++ [txt $ "Web server running on port: " <> into @Text (show port) | Just port <- [mport]]
+      txt ("Current seed: " <> into @Text (show theSeed)) :
+        [txt $ "Web server running on port: " <> into @Text (show port) | Just port <- [mport]]
   helpKeys =
     vBox
       [ txt "Global Keybindings"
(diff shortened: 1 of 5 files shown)


xsebek:
@byorgey I wonder if it would not be better to add another “Configuration” part to the `AppState`?

Seed is maybe related to current game, but port is global and we could avoid bugs like forgetting port when starting new game. 😅

byorgey:
@xsebek see what you think of the code now.  I don't know if it's worth adding a whole extra Configuration record to the AppState at this point (though it may be a good idea in the long run).  One could argue that the port IS part of the user interface state, since it's one way users can interface with the app...  on the other hand, I take your point that it is never supposed to change so we have to be careful not to overwrite it.

byorgey:
@TristanCacqueray , I ended up refactoring the web startup code quite a bit, since I needed it to actually return information about whether the web server started.  If you have a chance, please take a look to see if I messed anything up.

byorgey:
Also very open to suggestions of more things we could add to the Help dialog!

src/Swarm/TUI/View.hs
```diff
@@ -465,43 +466,56 @@ robotsListWidget s = hCenter table
...
+      , txt " "
+      , txt "  - The Swarm wiki, https://github.com/swarm-game/swarm/wiki"
+      , txt "  - The #swarm IRC channel on Libera.Chat"
+      ]
+  info =
+    vBox
+      [ txt ("Current seed: " <> into @Text (show theSeed))
+      , txt $ case mport of
+          Nothing -> "The web server is not running."
+          Just port -> "Web server running on port " <> into @Text (show port)
```
xsebek, line 487: This looks odd under the simple "Current seed: 0" - why not "Open port: 8080" or "Open port: none"?

xsebek (approved):
Looks good to me. :+1: The help modal definitely needed some love since we added other modals. :heart: 

I have some nitpicks about the style, but those can be addressed in the future if we want to.

xsebek:
Oh, by the way, @byorgey, there is a print of the port to standard output left over somewhere:
```
Web interface listening on :5357
```

src/Swarm/TUI/View.hs
byorgey: Sure, makes sense.  But I decided to put "web server port" instead of "open port" just to be clear what kind of port we're talking about, since new users won't have any idea.

src/Swarm/TUI/View.hs
```diff
@@ -465,43 +466,56 @@ robotsListWidget s = hCenter table
   debugging = creative && cheat
   g = s ^. gameState
 
-helpWidget :: Widget Name
-helpWidget = helpKeys <+> helpCommands
+helpWidget :: Seed -> Maybe Port -> Widget Name
+helpWidget theSeed mport =
+  padTop (Pad 1) $
+    hBox . map (padLeftRight 2) $
+      [helpKeys, tips <=> padTop (Pad 1) info]
```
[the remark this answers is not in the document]
byorgey: Ah, good idea, I like it.

byorgey:
> there is a print of the port to standard output left over somewhere:

Good catch, I got rid of it.

PR #637, opened 2022-08-11
Add `known` field to scenario format

### LINKED ISSUE #387: Add a simpler way to make default entites known

byorgey:
Closes #387 .  Scenario `.yaml` files now accept an extra field, `known`, containing a list of strings which are names of entities that should have the `Known` property added to them.


src/Swarm/Game/Scenario.hs
```diff
@@ -250,21 +251,34 @@ makeLensesWith (lensRules & generateSignatures .~ False) ''Scenario
...
+          revealKnown = appEndo . mconcat . map (Endo . revealEM) $ known
+      localE revealKnown $ do
+        rs <- v ..: "robots"
+        let rsMap = buildRobotMap rs
+        Scenario
+          <$> liftE (v .: "name")
+          <*> liftE (v .:? "description" .!= "")
+          <*> liftE (v .:? "creative" .!= False)
+          <*> liftE (v .:? "seed")
+          <*> pure em
```
xsebek, line 275: Should this be:
```suggestion
          <*> pure $ revealKnown em' -- or 'get' it first
```
Newly created entities will not be known. :grey_question:

byorgey: Oh, good catch!  It should definitely be `em`, not `em'`, since it is supposed to be just the entities that were declared in the scenario, not including all the standard ones etc.  But you're right that we should call `revealKnown` on it.

xsebek:
@byorgey I think we might need to override the standard entities that are known by adding them to the scenario entities.

For example, this does not work as I would expect:
```
cabal run swarm -O0 -- --cheat -c Tutorials/turn
# build {move; create "flower"; place "flower"; move;}
```
Screenshot from 2022-08-12 06-12-34

byorgey:
@xsebek I am not able to reproduce that.  When I try your example, the flower appears as `*`, as I would expect, not as a `?`.  Are you sure you were running the correct executable / had the correct branch checked out / etc?

byorgey:
OK, even if the given example works, I have found others that don't.  For example, suppose we make `branch` known.  If we have a robot `make` a branch and `place` it, then build a second robot and `view` it, the branch shows as a question mark.

Fundamentally the problem is similar to the problem we had with `ScenarioStatus`: values we think of as immutable end up getting duplicated in various data structures, and then we have to make sure we edit them all if we want them to mutate.  For example, in the example I gave above, the problem is that `Entity` values end up getting duplicated inside `Recipe`s, and do not receive the additional `Known` property.  There are two solutions I can think of:

1. Make a single data structure (an `EntityMap`) which has the ground source of truth about all entities, and never store an `Entity` value outside of that, but rather some kind of unique ID number which can be used as a lookup key into the `EntityMap`.  The downside is that this is a big, invasive change, and it's annoying to program against since you are constantly doing map lookups instead of just using values.  Arguably if we wanted to go down this path we should just bite the bullet and switch to some kind of entity-component framework like https://hackage.haskell.org/package/apecs .
2. The simpler solution for now would be to just store the list of `known` entity names as part of the `GameState`.  Then we simply consult that list when calculating how to display the world.

byorgey:
OK, I've implemented option (2) above.  I think this should be much more robust.

xsebek:
@byorgey nice, your solution sounds good to me :)

I think 1) can be simplified to “put entities that need to be changed to scenario entity map”. AFAIK that will override the default entities.

But to be 100% sure we get the entities right, we would probably need to have a single source of truth. Otherwise we could get stuff like recipes on old entities. 🤔

byorgey:
> But to be 100% sure we get the entities right, we would probably need to have a single source of truth. Otherwise we could get stuff like recipes on old entities.

Right, exactly.

@xsebek so do you mind approving this if you think it looks good now?

xsebek (approved):
👍

PR #641, opened 2022-08-16
update to `brick-1.0`

byorgey:


xsebek:
@byorgey you need to regenerate the Haskell CI file, because the cabal project file is recreated there. 😉

byorgey:
@xsebek Ah, thanks!  That would have taken me a while to figure out.

.mergify.yml
```diff
@@ -11,6 +11,7 @@ queue_rules:
           - -files~=^[^/]*\.yaml$
           - -files~=^data/.*\.yaml$
         - and:
+          - check-success=Haskell-CI - Linux - ghc-9.2.4
```
byorgey, line 14: This was missing before, which I assume was just an oversight?  Or was there a good reason not to put `ghc-9.2` in the `queue_rules`?

xsebek: I think it was failing more often when the rule was written. But it doesn’t make sense to check it before and not after so it’s probably a mistake.

Anyway 9.2 has been very stable for a long time, so there is no reason to not check it. 😅

cabal.project
```diff
@@ -3,9 +3,4 @@ source-repository-package
   location: https://github.com/colinhect/hsnoise
   tag: 4ccff11dea7e8d94e6a5fcaf8f43857bd65bd72d
 
-source-repository-package
-  type: git
-  location: https://github.com/jtdaugherty/brick
-  tag: dee90efa8cde1755f435c1f7dd9b6828b1bd0089
-
```
xsebek, line 10: @byorgey could you please add a comment to “regenerate CI file”?

xsebek (approved):

cabal.project
byorgey: Good idea, done!

PR #642, opened 2022-08-17
factor out and improve item checking code from `give` and `install`

### LINKED ISSUE #410: Known entities do not include installed devices

byorgey:
- We now know about installed devices as well (closes #410)
- Add a hint about `create` when in creative mode


xsebek (approved):
Looks great! 👍

PR #646, opened 2022-08-21
add recipe for turning iron into a lodestone

byorgey:
If you rub some iron with a lodestone for a long time, the iron turns into another lodestone.

While playing the game it felt like lodestones were too scarce, given the fact that they are needed to evaluate `not`, which ends up getting used frequently.  This way lodestones are actually renewable.

### DIFF
--- a/data/recipes.yaml
+++ b/data/recipes.yaml
@@ -132,6 +132,14 @@
 #########################################
 
 - in:
+  - [1, rock]
+  out:
+  - [1, lodestone]
+  required:
+  - [1, lodestone]
+  time: 64
+
+- in:
   - [1, boulder]
   out:
   - [3, rock]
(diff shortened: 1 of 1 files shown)


data/recipes.yaml
```diff
@@ -131,6 +131,14 @@
 ##                STONE                ##
 #########################################
 
+- in:
+  - [1, rock]
```
xsebek, line 135: Why not use a renewable iron though?
If you are making new robots that require lodestone, you will need iron for their construction too.

byorgey:
@xsebek Good idea!  How's that?

xsebek (approved):
LGTM, just the PR description needs to be updated as Mergify will use it for commit. 🙂

PR #648, opened 2022-08-22
start the base out with 50 clocks in classic mode

byorgey:
Clocks are somewhat difficult to build, and it's useful to be able to use `wait` early in the game.

### DIFF
--- a/data/scenarios/classic.yaml
+++ b/data/scenarios/classic.yaml
@@ -24,6 +24,7 @@ robots:
       - [70, grabber]
       - [100, solar panel]
       - [50, scanner]
+      - [50, clock]
       - [5, toolkit]
 world:
   seed: null


byorgey:
@xsebek any thoughts?

xsebek (approved):
I do not have an opinion on this, we will see how hard it is once someone finishes the speedrun. 🙂 

If it is not necessary, I would prefer keeping it simple, so that there are not too many devices at the start. 🤷

xsebek:
Btw. can you not busy wait with `turn forward`?

byorgey:
Yes, you can `turn forward`, but that's so ugly, and it's still not as efficient if you want to wait more than one time step.  One would be forced to have two versions of everything, one that uses `turn forward` while you don't have any clocks at the beginning of the game, and one version that properly uses `wait` once you are able to build some clocks.

Also, as for the "story" aspect, the base starts out with a `clock` installed (so you can see the time counting up), so it seems reasonable that it would have a few extras.

xsebek:
Well I must admit I have not played the classic mode as much, so I will defer to you on this point. 😉

Feel free to merge, I will have to try it later. 😁

PR #649, opened 2022-08-22
improved recipe lists

### LINKED ISSUE #418: Order recipes by whether the entity is an input, catalyst, or output
In some circumstances there are a lot of "uninteresting" recipes cluttering up the view, hiding the more "interesting" ones.  A great example is the recipes related to `copper ore`:

copper-recipes

All the recipes about drilling come first, which is tedious because there are many combinations of types of drill, whether you are drilling a mountain or a mine, etc.  The most interesting recipes (that you can make `copper pipe` or `copper wire` from `copper ore`) are buried at the bottom.

I am not quite sure what a good solution would be; I am not even quite sure how to formulate exactly what the problem is.  But it seemed worthwhile to open a discussion.

byorgey:
- Add recipes where the entity is a catalyst to the recipes shown for
  an entity
- Order recipes so we get recipes where the entity is a (1) input (2)
  catalyst (3) output, in that order.

Closes #418.

### DIFF
--- a/src/Swarm/Game/Recipe.hs
+++ b/src/Swarm/Game/Recipe.hs
@@ -25,6 +25,7 @@ module Swarm.Game.Recipe (
   loadRecipes,
   outRecipeMap,
   inRecipeMap,
+  reqRecipeMap,
 
   -- * Looking up recipes
   MissingIngredient (..),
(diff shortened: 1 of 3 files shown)


xsebek (approved):
Nice! I suppose separators would not make sense given that a recipe could have the entity both as an input and output.

byorgey:
@xsebek Yeah, indeed, I'm not sure separators would make sense.

PR #650, opened 2022-08-22
temporary workaround for https://github.com/TomMD/entropy/issues/75

byorgey:


byorgey:
Assuming this works, we can merge this first and it should hopefully fix all the other PR checks if we re-run them.

xsebek (approved):
It seems to work, ~~but what is the failure about actually?~~ Nvm I see now…

@byorgey could you please add a comment that we can update the bound once the entropy macro is fixed?

I am not sure what bounds we will want then or if Cabal will just pick the right ones.

xsebek:
Btw. TomMD/entropy#75 is a shorter syntax supported by Github ;)

byorgey:
@xsebek ah, of course, done.  We don't depend on `entropy` directly, so hopefully if we wait long enough, whatever we depend on that transitively depends on `entropy` will update its bounds so we can just remove ours.

PR #652, opened 2022-08-29
Notify user about newer version

### LINKED ISSUE #66: Add notification when game is not up-to-date

### LINKED ISSUE #627: Display a pop-up when the web service fail to start
When the web service fail to start, it would be useful to add a logentry to the GameState message queue with the error. To implement that, we might need a new type of event coming from the web service (which would also be used for #625 )

xsebek:
- query GitHub Releases for a new version
- add `swarm version` command
- show the version notification in the main menu
- add a new app-wide logging queue
- show the app-wide logs in the main menu as Messages
- catch and show the web API failure in the Messages
- closes #66
- closes #627


xsebek:
Useful links:
- GitHub API: [latest release](https://docs.github.com/en/rest/releases/releases#get-the-latest-release)
- HTTP Client: see the [Making a GET request](https://hackage.haskell.org/package/http-client-0.7.13.1/docs/Network-HTTP-Client.html) example
  - [Blog: why are there no setters in the HTTP Client documentation?](https://www.yesodweb.com/book/settings-types)
  - [TLS package](https://hackage.haskell.org/package/http-client-tls-0.3.6.1/docs/Network-HTTP-Client-TLS.html)
- [GitHubRelease package](https://hackage.haskell.org/package/github-release) handles just uploading, but a query looks almost the same

src/Swarm/App.hs
```diff
@@ -105,3 +119,30 @@ enablePasteMode = do
   when (V.supportsMode output V.BracketedPaste) $
     liftIO $
       V.setMode output V.BracketedPaste True
+
+version :: String
```
TristanCacqueray, line 123: Perhaps this could be `Maybe String`, where a local dirty git status should have a Nothing version. That way we can skip the upstreamVersion query when developing swarm.

xsebek: Good point about skipping the check during development, I will add a check for that.

Why would the status be dirty though?

We have tags both locally and when building in CI, so you will not tell the difference by it being empty.

Instead I think I will have to use the hash that is appended to the tag name and compare it to current commit.

xsebek: @TristanCacqueray well, it did not go as I hoped, but I made another function which gives `Maybe` and another that gives `Either`. Hopefully, that will make using the results easier. :slightly_smiling_face:

TristanCacqueray: I see, that's annoying... How about checking the branch name, if it is `main`, do the new version notification, otherwise assume it is a develop copy?

It might be good to check the values for a `cabal install` build, if I remember correctly, the .git content is not available in that situation.

xsebek:
Now I think the query is actually working. When you run on this branch:
```
> cabal run swarm:swarm -- version
Up to date
Swarm game - pre-alpha version (latest@1ec9d7d495)
Currently on development branch 'latest', skipping check.
```
But in future when we have a new release, you should get:
```
> swarm version
Swarm game - v0.0.1 (main@0123456789)
New upstream release: v0.0.2
```
Now I have to learn how to get this in the main menu...

src/Swarm/App.hs
xsebek: > @TristanCacqueray ...if it's `main`...

That is indeed the approach I have taken. :laughing: 

> It might be good to check the values for a `cabal install` build, if I remember correctly, the .git content is not available in that situation.

The `cabal install` binary is indeed problematic:
```
swarm version
Swarm game - pre-alpha version
"GHEGitRunFailed \".\" [\"rev-parse\",\"--show-toplevel\"] (ExitFailure 128) \"\" \"fatal: not a git repository (or any parent up to mount point /)\\nStopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).\\n\""
```
Maybe we could include the `.git` folder for the build only in cabal? :thinking:

xsebek: Or I could use the readily available `Paths_swarm.version` and save myself the trouble! :grinning:

xsebek:
It is nearing its final state, but I need to handle errors better.

Also, I need to check what will happen if the query result arrives later, the main menu should be updated.

xsebek:
Hey @TristanCacqueray, what do you think about this app-wide log? :slightly_smiling_face: 

### Main menu
If there are any error logs, messages will be highlighted. This could be improved later. :shrug: 

You can also see that the version message is different when on the development branch. :slightly_smiling_face: 

Screenshot from 2022-09-04 20-34-11

### Open messages
It is not much right now, but I added the web service log, as I proposed in #627.

I am not sure how to make the message more meaningful, I do not understand the code there very well. 
Happy to hear any suggestions. :wink: 

image

byorgey:
@xsebek if I choose "Messages" from the main menu and then hit `q`, it quits the game.  I think it should not be possible to quit the game directly from the Messages screen.

byorgey:
Also, the message re: the server does not seem correct.  I see a message saying "failed to start the web API on :5357" but the web API is actually running fine.

src/Swarm/Web.hs
```diff
@@ -70,16 +70,19 @@ defaultPort = 5357
...
   void $ forkIO $ webMain (Just baton) port gsRef
   res <- timeout 500_000 (takeMVar baton)
   case (portM, res) of
     -- User requested explicit port but server didn't start: fail
-    (Just _, Nothing) -> fail $ "Failed to start the web API  on :" <> show port
+    (Just _, Nothing) -> fail $ failMsg port
     -- Otherwise, just report whether the server is running, and if so, on what port
-    _ -> return (port <$ res)
+    (Just _, Just ()) -> return (Right port)
+    _ -> return . Left $ failMsg port
```
byorgey, line 86: This final catch-all case should not be a failure.  If the first value of the tuple is `Nothing` it just means that no explicit `--web` flag was given.  `(Nothing, Just ())` is definitely a success, that means the web server successfully started on the default port.

Actually, now that we have the Messages framework I think we can actually drop the whole thing about ignoring failure when the user didn't explicitly request a port.  In other words this `case (portM, res)` should just be `case res`, we don't care about `portM` at this point.

src/Swarm/Version.hs
```diff
@@ -0,0 +1,119 @@
...
+
+commitInfo :: String
+commitInfo = case gitInfo of
+  Left _ -> ""
+  Right git -> " (" <> giBranch git <> "@" <> take 10 (giHash git) <> ")"
+
+type Hash = String
+
+isSwarmReleaseTag :: String -> Bool
+isSwarmReleaseTag = all (\c -> isDigit c || c == '.')
```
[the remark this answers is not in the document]
xsebek: I can get anything that has the version as a substring to work. The only requirement is that it should be the same forever, so that old versions of the game can display the notification.

I think using the same format as Cabal with no prefix would make it obvious that the tag name is parsed by machines.
Let's use the `0.1.0.0` format for tags and not be tempted to change prefixes to something prettier.

src/Swarm/Version.hs
```diff
@@ -0,0 +1,119 @@
...
+      if giBranch gi /= "main"
+        then return . Left . OnDevelopmentBranch $ giBranch gi
+        else do
+          mUpTag <- upstreamReleaseVersion
+          return $ case mUpTag of
+            Nothing -> Left NoUpstreamRelease
+            Just upTag -> case snd <$> tagVersion of
+              Nothing -> Right upTag
+              Just myTag ->
+                if myTag >= upTag
```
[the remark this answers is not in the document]
xsebek: Heh, it would work for some time and then break on the 10th release. :smile: 

There is `Data.Version` in `base` which is very simple and used in autogenerated `Paths_swarm.version`.
From the docs, it looks like it will cover our needs. :slightly_smiling_face:

src/Swarm/TUI/Controller.hs
```diff
@@ -166,6 +175,13 @@ getTutorials sc = case M.lookup "Tutorials" (scMap sc) of
 advanceMenu :: Menu -> Menu
 advanceMenu = _NewGameMenu . lens NE.head (\(_ :| t) a -> a :| t) %~ BL.listMoveDown
 
+handleMainMessagesEvent :: BrickEvent Name AppEvent -> EventM Name AppState ()
+handleMainMessagesEvent = \case
+  Key V.KEsc -> uiState . uiMenu .= MainMenu (mainMenu Messages)
+  CharKey 'q' -> halt
+  ControlKey 'q' -> halt
```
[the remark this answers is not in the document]
xsebek: Sure, I guess it will feel more natural to make <kbd>Q</kbd> return to the main menu.

After all, it looks the same as messages in-game. :sweat_smile:

src/Swarm/TUI/Model.hs
```diff
@@ -691,6 +690,50 @@ promptUpdateL = lens g s
...
+-- This logging is separate from the logging done during game-play.
+-- If some error happens before a game is even selected, this is the
+-- place to log it.
+eventLog :: Lens' RuntimeState (Seq LogEntry)
+
+-- ----------------------------------------------------------------------------
+--                                   APPSTATE                                --
+-- ----------------------------------------------------------------------------
+
+-- | The 'AppState' just stores together the game state and UI state.
```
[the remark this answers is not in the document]
xsebek: Yep, it's still just a collection of other states, so `AppState` has not gained any status. :smirk: 

I was considering putting the app-wide stuff in `AppState`, but it would look weird to have `Maybe Int` on the same level as `GameState` and I already got used to the `s ^. uiState . uiSomeLens` idiom.

src/Swarm/TUI/View.hs
```diff
@@ -109,18 +111,36 @@ drawUI s
...
+ where
+  dial = dialog (Just "Messages") Nothing maxModalWindowWidth
+  scrollList = withVScrollBars OnRight . vBox
+  drawLogs = map (drawLogEntry True) . toList
+
+drawMainMenuUI :: AppState -> BL.List Name MainMenuEntry -> Widget Name
+drawMainMenuUI s l =
   vBox
     [ maybe emptyWidget drawLogo logo
+    , hCenter . padTopBottom 2 $ newVersioWidget version
```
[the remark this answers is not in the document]
xsebek: :flushed:

src/Swarm/TUI/View.hs
```diff
@@ -109,18 +111,36 @@ drawUI s
...
     ]
+ where
+  logo = s ^. uiState . appData . at "logo"
+  version = s ^. runtimeState . upstreamRelease
+
+newVersioWidget :: Either NewReleaseFailure String -> Widget n
+newVersioWidget = \case
+  Right ver -> txt $ "New version " <> T.pack ver <> " is available!"
+  Left (OnDevelopmentBranch _b) -> txt "Good luck developing!"
+  Left _ -> txt "The game is up-to-date!"
```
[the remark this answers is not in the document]
xsebek: There is always the tension between "it will be nice for the user" and "it will be neat for the developer fixing bugs". :laughing:

I will fix that shortly. :wink:

src/Swarm/Web.hs
xsebek: Ups, my mistake! :sweat_smile:

swarm.cabal
```diff
@@ -127,28 +128,34 @@ library
                       aeson                         >= 2 && < 2.1,
                       array                         >= 0.5.4 && < 0.6,
                       brick                         >= 1.0 && < 1.1,
+                      bytestring                    >= 0.10 && < 0.11,
                       clock                         >= 0.8.2 && < 0.9,
                       containers                    >= 0.6.2 && < 0.7,
                       directory                     >= 1.3 && < 1.4,
                       dotgen                        >= 0.4 && < 0.5,
                       either                        >= 5.0 && < 5.1,
+                      extra                         >= 1.7 && < 1.8,
```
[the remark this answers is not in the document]
xsebek: Some functions like `breakOnEnd` that `Text` has, but `[]` does not.

It would be easy to avoid, but I dislike converting to `Text` and back again just to get a few simple functions.

The `extra` package has a negligible dependency footprint and might already be transitively included. :shrug:

src/Swarm/TUI/View.hs
```diff
@@ -211,11 +231,16 @@ drawNewGameMenuUI (l :| ls) =
...
-drawMainMenuEntry NewGame = txt "New game"
-drawMainMenuEntry Tutorial = txt "Tutorial"
-drawMainMenuEntry About = txt "About"
-drawMainMenuEntry Quit = txt "Quit"
+drawMainMenuEntry :: AppState -> MainMenuEntry -> Widget Name
+drawMainMenuEntry s = \case
+  NewGame -> txt "New game"
+  Tutorial -> txt "Tutorial"
+  About -> txt "About"
+  Messages -> highlightMessages $ txt "Messages"
```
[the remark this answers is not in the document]
xsebek: Yeah, I will have to properly use `Notifications` for that. :unamused: 

I would rather have skipped the menu and ignored it, but I agree it's annoying and it should be easy to fix. :smile:

swarm.cabal
```diff
@@ -127,28 +128,34 @@ library
                       aeson                         >= 2 && < 2.1,
                       array                         >= 0.5.4 && < 0.6,
                       brick                         >= 1.0 && < 1.1,
+                      bytestring                    >= 0.10 && < 0.11,
```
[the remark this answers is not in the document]
xsebek: I will try it, maybe I just copied the bounds wrong from Hackage. :+1:

src/Swarm/Version.hs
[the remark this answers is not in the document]
xsebek: Btw. `versions` is a hilarious library. As long as we stick to PVP in Cabal and in tags, we should not need it.

I might filter Github prereleases from the list just to be safe though, we might want to use those in the future.

src/Swarm/Web.hs
xsebek: Well, I don't think we can entirely get rid of the `portM` in the `case`, in fact, I split it more.

But it should handle errors better now. This is what you get when you run two Swarms:

image

src/Swarm/TUI/View.hs
```diff
@@ -240,8 +240,10 @@ drawMainMenuEntry s = \case
   Messages -> highlightMessages $ txt "Messages"
   Quit -> txt "Quit"
  where
-  allErrors = Seq.filter (\l -> l ^. leSaid == ErrorTrace) (s ^. runtimeState . eventLog)
-  highlightMessages = if null allErrors then id else withAttr notifAttr
+  highlightMessages =
+    if s ^. runtimeState . eventLog . notificationsCount > 0
+      then id
+      else withAttr notifAttr
```
xsebek, line 246: damn it

src/Swarm/App.hs
```diff
@@ -79,9 +78,11 @@ appMain port mseed scenario toRun cheat = do
       gsRef <- newIORef (s ^. gameState)
       eport <- Swarm.Web.startWebThread port gsRef
 
-      let s' = case eport of
-            Left e -> s & runtimeState . eventLog %~ (|> LogEntry 0 ErrorTrace "Web API" (-2) zero (T.pack e))
-            Right p -> s & runtimeState . webPort ?~ p
+      let s' =
+            s & runtimeState
+              %~ case eport of
+                Right p -> webPort ?~ p
```
byorgey, line 84: Should we generate an informational message here saying that the web server was successfully started on port nnn?

byorgey (approved):
Looks great!  I don't really know how to test the case where there is a new release, but it seems reasonable.

xsebek:
@byorgey you can test the release if you use some other repo then swarm and merge to the `main` branch.

For example `haskell/haskell-language-server`:
```
> cabal run swarm:swarm -- version
Up to date
Swarm game - pre-alpha version (main@7ea47f6ad2)
New upstream release: 1.7.0.0
```

[asciicast](https://asciinema.org/a/1RZJLGtChBjW48Tao4dUuOlOR)

src/Swarm/App.hs
xsebek: Done!

image

PR #654, opened 2022-08-31
Add random bitcoin

### LINKED ISSUE #497: Add randomness device
We are missing a device and recipe for the `random` command (and capability).

Proposed recipe:

- `bit (0)` + `bit (1)` + `gold coin` --> `bitcoin`
- `gold` | `furnace` ---> 8x `gold coin`

Gold is almost the hardest thing to get, but I do not want to miss the opportunity to make golden bitcoin the randomness provider, so let's at least give many gold coins from one gold nugget. 

Suggestions are welcome, particularly on the (not so though out) gold part. :slightly_smiling_face:

xsebek:
- add bitcoin (which is random) and gold coin entities
  - `bit (0)` + `bit (1)` + `gold coin` --> `bitcoin`
  - `gold` | `furnace` ---> 8x `gold coin`
- closes #497

### DIFF
--- a/data/recipes.yaml
+++ b/data/recipes.yaml
@@ -487,9 +487,18 @@
   - [1, furnace]
 
 - in:
+  - [1, gold]
+  out:
+  - [8, gold coin]
+  required:
+  - [1, big furnace]
+
+- in:
   - [1, gold coin]
   - [1, bit (1)]
   - [1, bit (0)]
+  required:
+  - [1, metal drill]
   out:
   - [1, bitcoin]
 
--- a/data/recipes.yaml
+++ b/data/recipes.yaml
@@ -479,6 +479,20 @@
   out:
   - [1, 3D printer]
 
+- in:
+  - [1, gold]
+  out:
+  - [8, gold coin]
+  required:
+  - [1, furnace]
+
+- in:
+  - [1, gold coin]
+  - [1, bit (1)]
+  - [1, bit (0)]
+  out:
+  - [1, bitcoin]
+
 ## MAGIC
 
 - in:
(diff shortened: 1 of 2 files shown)


xsebek:
This would be the new hardest-to-craft entity at level 14! :money_mouth_face: 
|Screenshot from 2022-08-31 23-10-21|
|-|

byorgey (approved):
I love it!  @noahyor approves too.

xsebek:
This made me realise that the recipe graph is not exactly correct.

Maybe I should track which entities are obtainable, but also which can be installed. That will be a fun challenge to implement. Note to self - add command line option to delete some arrows, otherwise the graph will get crazy.

data/recipes.yaml
```diff
@@ -486,10 +486,19 @@
...
+  - [8, gold coin]
+  required:
+  - [1, big furnace]
+
 - in:
   - [1, gold coin]
   - [1, bit (1)]
   - [1, bit (0)]
+  required:
+  - [1, metal drill]
```
byorgey, line 501: Why is a metal drill required to make a bitcoin?

xsebek: Because you are mining bitcoin?

byorgey: > Because you are mining bitcoin?

Ohhhhh of course. :smile:

byorgey: > It does not really matter, you already have a drill.

Well, the robot that mined the gold has a drill.  But maybe you have a different robot programmed to mine the bitcoins.  But it's fine, you can give that robot a drill too.

Issue #231, opened 2021-10-15
Handle requirements properly as part of the type system


byorgey:
**Describe the bug**
It seems that programs run by `base` are only checked dynamically for capabilities.  This means that both
1. The base can get quite a ways through a program before crashing due to a missing capability
2. Some capabilities sneak by since they don't correspond directly to a command and are thus never dynamically checked.  For example, recursion is such a capability.

**To Reproduce**
- Run `build "x" {move}; move` to see that the base builds the robot `"x"` first before failing to execute the `move` due to a missing capability.  Ideally, it should instead refuse to run the entire program due to missing capabilities.
- As a different example,
```
def forever = \c. c ; forever c end
build "x" {forever move}
forever (build "x" {move})
```
The `build "x"` fails as it should with an error saying that you don't have the necessarily devices to install on `x` (namely, a lambda and a strange loop).  However, the last line works fine even though it shouldn't, since the `base` doesn't have a lambda or strange loop.

xsebek:
Yup, I was wondering if this was intended, but didn't ask, since it makes getting the first lambda harder :sweat_smile:

xsebek:
It can be hidden behind a lake/mountain and you have to make a program that goes around. :mountain: 

Most cases can be solved with _"move in one direction, move in other direction and then go back"_, but that _"go back"_ can be error prone so I used the `fetchNW`&friends functions. I will just have to write out the first fetch, I guess, no big deal :wink:

byorgey:
So I was trying to work on this today, and ran into what seems like a difficult issue.  I put a basic check in place and it immediately started rejecting things like `build "x" {move}`, because the base does not have treads.  Of course this is nonsense: it is `x` that will be needing treads, not `base`.  But right now capability checking does not treat `build` specially at all: if it sees an application it simply returns the union of the capabilities required by the left- and right-hand sides of the application.

Clearly, capability checking *does* need to treat `build` specially, but I am unsure how to do this in a principled way.  We could specially check things which are *syntactically* arguments to `build`, but that wouldn't be compositional, i.e. it wouldn't properly deal with custom variants of `build`.  For example, we might define 
```
build2 : string -> {cmd ()} -> cmd ()
build2 name p = build name p; build name p
```
and now `base` should be able to execute `build2 "x" {move}`.

[...]

PR #656, opened 2022-08-31
WIP: Infinite count

### LINKED ISSUE #621: Infinite items in inventory
It would be useful to have infinite amounts of entities in inventory.

```
robot chassis  ∞
logger         ∞
compass        ∞
```

This would allow for challenges like BFS search that requires creating an unbounded number of robots. (#608)

It would also be more user-friendly in creative mode (better communication from UI and make commands would not fail)

xsebek:
This approach is buggy in two ways:
- it can **crash** because [`Natural`](https://hackage.haskell.org/package/base-4.14.1.0/docs/GHC-Natural.html#t:Natural) is too strict
- it breaks the fragile hashing laws for inventory

A better approach would be to either do without a `Num` instance and handle every usage carefully or add _negative infinity_ too and use integers.

In that case, we might as well add them to the number values in-game.
We can handle adding infinities together with a `safeAdd` function and they would only become available in creative mode or special scenarios anyway.

- closes #621

### DIFF
--- a/src/Swarm/Game/Entity.hs
+++ b/src/Swarm/Game/Entity.hs
@@ -51,7 +51,7 @@ module Swarm.Game.Entity (
 
   -- * Inventories
   Inventory,
-  Count,
+  Count(..),
 
   -- ** Construction
   empty,
@@ -111,6 +111,8 @@ import Swarm.Util.Yaml
 import Text.Read (readMaybe)
 import Witch
 import Prelude hiding (lookup)
+import Numeric.Natural (Natural)
+import Data.Data (Data)
 
 ------------------------------------------------------------
 -- Properties
@@ -448,9 +450,48 @@ entityInventory = hashedLens _entityInventory (\e x -> e {_entityInventory = x})
 -- Inventory
 ------------------------------------------------------------
 
--- | A convenient synonym to remind us when an 'Int' is supposed to
---   represent /how many/ of something we have.
-type Count = Int
+-- | A type that represent /how many/ of something we have.
+--
+-- In some challenges it is useful to have _an infinite_ amount
+-- of supplies in robots inventory as that is more visible
+-- than hidden default devices.
+data Count = Count Natural | Infinity
+  deriving (Eq, Ord, Show, Read, Generic, Data)
+
+instance ToJSON Count where
+  toJSON (Count a) = toJSON a
+  toJSON Infinity = Null
+
+instance FromJSON Count where
+  parseJSON = \case
+    Number s ->
+      if s - fromInteger (truncate s) == 0
+        then pure $ Count (truncate s)
+        else fail "Count is not a whole number!"
+    Null -> pure Infinity
+    e -> fail $
+      "Expected number or null for count, but got '" <> show e <> "'!"
+
+instance FromJSONE e Count
+
+instance Num Count where
+  Count a + Count b = Count $ a + b
+  _ + _ = Infinity
+  Count a * Count b = Count $ a * b
+  _ * _ = Infinity
+  abs = id
+  signum (Count 0) = 0
+  signum _ = 1
+  fromInteger = Count . fromInteger
+  Count a - Count b = Count $ a - b
+  Infinity - Count _ = Infinity
+  _ - _ = 0
+
+-- | Internal way to shift count to positive integers when hashing.
+hashCount :: Num p => Count -> p
+hashCount c = case c of
+  Infinity -> 1
+  Count a -> 2 + fromIntegral a
 
 -- | An inventory is really just a bag/multiset of entities.  That is,
 --   it contains some entities, along with the number of times each
@@ -531,7 +572,7 @@ insertCount k e (Inventory cs byN h) =
   Inventory
     (IM.insertWith (\(m, _) (n, _) -> (m + n, e)) (e ^. entityHash) (k, e) cs)
     (M.insertWith IS.union (T.toLower $ e ^. entityName) (IS.singleton (e ^. entityHash)) byN)
-    (h + (k + extra) * (e ^. entityHash)) -- homomorphic hashing
+    (h + (hashCount k + extra) * (e ^. entityHash)) -- homomorphic hashing
  where
   -- Include the hash of an entity once just for "knowing about" it;
   -- then include the hash once per actual copy of the entity.  In
@@ -576,8 +617,7 @@ deleteCount k e (Inventory cs byN h) = Inventory cs' byN h'
  where
   m = (fst <$> IM.lookup (e ^. entityHash) cs) ? 0
   cs' = IM.adjust removeCount (e ^. entityHash) cs
-  h' = h - min k m * (e ^. entityHash)
-
+  h' = h - min (hashCount k) (hashCount m) * (e ^. entityHash)
   removeCount :: (Count, a) -> (Count, a)
   removeCount (n, a) = (max 0 (n - k), a)
 
@@ -587,9 +627,10 @@ deleteAll e (Inventory cs byN h) =
   Inventory
     (IM.adjust (first (const 0)) (e ^. entityHash) cs)
     byN
-    (h - n * (e ^. entityHash))
+    (h - hashCount n * (e ^. entityHash))
  where
   n = (fst <$> IM.lookup (e ^. entityHash) cs) ? 0
+
 
 -- | Get the entities in an inventory and their associated counts.
 elems :: Inventory -> [(Count, Entity)]
--- a/src/Swarm/Game/Step.hs
+++ b/src/Swarm/Game/Step.hs
@@ -48,8 +48,8 @@ import Data.Tuple (swap)
 import Linear (V2 (..), zero, (^+^))
 import Swarm.Game.CESK
 import Swarm.Game.Display
-import Swarm.Game.Entity hiding (empty, lookup, singleton, union)
-import Swarm.Game.Entity qualified as E
+import Swarm.Game.Entity hiding (empty, lookup, singleton, union, Count)
+import Swarm.Game.Entity qualified as E 
 import Swarm.Game.Exception
 import Swarm.Game.Recipe
 import Swarm.Game.Robot
@@ -927,7 +927,10 @@ execConst c vs s k = do
     Count -> case vs of
       [VString name] -> do
         inv <- use robotInventory
-        return $ Out (VInt (fromIntegral $ countByName name inv)) s k
+        let n = case countByName name inv of
+              E.Count x -> x
+              E.Infinity -> 42
+        return $ Out (VInt $ fromIntegral n) s k
       _ -> badConst
     Whereami -> do
       V2 x y <- use robotLocation
@@ -1375,22 +1378,22 @@ execConst c vs s k = do
             let salvageInventory = E.union (target ^. robotInventory) (target ^. installedDevices)
             robotMap . at (target ^. robotID) . traverse . robotInventory .= salvageInventory
 
-            let salvageItems = concatMap (\(n, e) -> replicate n (e ^. entityName)) (E.elems salvageInventory)
-                numItems = length salvageItems
 
             -- Copy over the salvaged robot's log, if we have one
             inst <- use installedDevices
             em <- use entityMap
             creative <- use creativeMode
+            system <- use systemRobot
             logger <-
               lookupEntityName "logger" em
                 `isJustOr` Fatal "While executing 'salvage': there's no such thing as a logger!?"
             when (creative || inst `E.contains` logger) $ robotLog <>= target ^. robotLog
 
             -- Immediately copy over any items the robot knows about
-            -- but has 0 of
-            let knownItems = map snd . filter ((== 0) . fst) . elems $ salvageInventory
-            robotInventory %= \i -> foldr (insertCount 0) i knownItems
+            -- but has 0 of and in creative mode just get everything now.
+            let immediate = if system || creative then id else filter ((== 0) . fst)
+            let immediateItems = map snd . immediate . elems $ salvageInventory
+            robotInventory %= \i -> foldr (insertCount 0) i immediateItems
 
             -- Now reprogram the robot being salvaged to 'give' each
             -- item in its inventory to us, one at a time, then
@@ -1399,6 +1402,13 @@ execConst c vs s k = do
             robotMap . at (target ^. robotID) . traverse . systemRobot .= True
 
             ourID <- use @Robot robotID
+            let salvageItems = if system || creative
+                  then []
+                  else concatMap (uncurry replicateCount) (E.elems salvageInventory)
+                numItems = length salvageItems
+                replicateCount n e = e ^. entityName & case n of
+                  E.Count x -> replicate (fromIntegral x)
+                  E.Infinity -> replicate 42
 
             -- The program for the salvaged robot to run
             let giveInventory =
--- a/src/Swarm/TUI/View.hs
+++ b/src/Swarm/TUI/View.hs
@@ -847,7 +847,10 @@ drawItem sel i _ (Separator l) =
   (if sel == Just (i + 1) then visible else id) $ hBorderWithLabel (txt l)
 drawItem _ _ _ (InventoryEntry n e) = drawLabelledEntityName e <+> showCount n
  where
-  showCount = padLeft Max . str . show
+  showCount = padLeft Max . str . prettyCount
+  prettyCount = \case
+    E.Count a -> show a
+    E.Infinity -> "∞"
 drawItem _ _ _ (InstalledEntry e) = drawLabelledEntityName e <+> padLeft Max (str " ")
 
 -- | Draw the name of an entity, labelled with its visual
--- a/src/Swarm/DocGen.hs
+++ b/src/Swarm/DocGen.hs
@@ -33,7 +33,7 @@ import Data.Text (Text, unpack)
 import Data.Text qualified as T
 import Data.Text.IO qualified as T
 import Data.Tuple (swap)
-import Swarm.Game.Entity (Entity, EntityMap (entitiesByName), entityName, loadEntities)
+import Swarm.Game.Entity (Entity, EntityMap (entitiesByName), entityName, loadEntities, Count)
 import Swarm.Game.Entity qualified as E
 import Swarm.Game.Recipe (Recipe, loadRecipes, recipeInputs, recipeOutputs, recipeRequirements)
 import Swarm.Game.Robot (installedDevices, instantiateRobot, robotInventory)
@@ -365,7 +365,7 @@ classicScenario = do
 startingDevices :: Scenario -> Set Entity
 startingDevices = Set.fromList . map snd . E.elems . view installedDevices . instantiateRobot 0 . head . view scenarioRobots
 
-startingInventory :: Scenario -> Map Entity Int
+startingInventory :: Scenario -> Map Entity Count
 startingInventory = Map.fromList . map swap . E.elems . view robotInventory . instantiateRobot 0 . head . view scenarioRobots
 
 -- | Ignore utility entites that are just used for tutorials and challenges.
--- a/src/Swarm/Game/State.hs
+++ b/src/Swarm/Game/State.hs
@@ -116,7 +116,6 @@ import Data.IntMap qualified as IM
 import Data.IntSet (IntSet)
 import Data.IntSet qualified as IS
 import Data.IntSet.Lens (setOf)
-import Data.List (partition)
 import Data.List.NonEmpty (NonEmpty)
 import Data.List.NonEmpty qualified as NE
 import Data.Map (Map)
@@ -764,7 +763,8 @@ scenarioToGameState scenario userSeed toRun g = do
   em = g ^. entityMap <> scenario ^. scenarioEntities
 
   baseID = 0
-  (things, devices) = partition (null . view entityCapabilities) (M.elems (entitiesByName em))
+  entities = M.elems (entitiesByName em)
+  devices = filter (not . null . view entityCapabilities) entities
   -- Keep only robots from the robot list with a concrete location;
   -- the others existed only to serve as a template for robots drawn
   -- in the world map
@@ -781,7 +781,7 @@ scenarioToGameState scenario userSeed toRun g = do
       & ix baseID . robotInventory
         %~ case scenario ^. scenarioCreative of
           False -> id
-          True -> union (fromElems (map (0,) things))
+          True -> union (fromElems (map (Infinity,) entities))
       & ix baseID . installedDevices
         %~ case scenario ^. scenarioCreative of
           False -> id
--- a/src/Swarm/Language/Requirement.hs
+++ b/src/Swarm/Language/Requirement.hs
@@ -40,7 +40,8 @@ import GHC.Generics (Generic)
 import Swarm.Language.Capability (Capability (..), constCaps)
 import Swarm.Language.Context (Ctx)
 import Swarm.Language.Context qualified as Ctx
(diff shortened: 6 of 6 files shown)


src/Swarm/Game/Entity.hs
```diff
@@ -448,9 +450,49 @@ entityInventory = hashedLens _entityInventory (\e x -> e {_entityInventory = x})
...
+--
+-- In some challenges it is useful to have _an infinite_ amount
+-- of supplies in robots inventory as that is more visible
+-- than hidden default devices.
+data Count = Count Natural | Infinity
+  deriving (Eq, Ord, Show, Read, Generic, Data)
+
+instance ToJSON Count where
+  toJSON (Count a) = toJSON a
+  toJSON Infinity = Null
```
byorgey, line 463: Ugh, no, I do not want to have to write  `[tree, null]` in a `.yaml` file when I actually want infinite trees.

src/Swarm/Game/Step.hs
```diff
@@ -927,7 +927,10 @@ execConst c vs s k = do
     Count -> case vs of
       [VString name] -> do
         inv <- use robotInventory
-        return $ Out (VInt (fromIntegral $ countByName name inv)) s k
+        let n = case countByName name inv of
+              E.Count x -> x
+              E.Infinity -> 42
```
byorgey, line 932: :laughing:

byorgey:
I don't understand why `Natural` makes it crash, can you elaborate?

src/Swarm/Game/Entity.hs
xsebek: I guess we can write `infinity` which will be passed as a string?

src/Swarm/Game/Entity.hs
```diff
@@ -448,9 +450,49 @@ entityInventory = hashedLens _entityInventory (\e x -> e {_entityInventory = x})
...
+  fromInteger = Count . fromInteger
+  Count a - Count b = Count $ a - b
+  Infinity - Count _ = Infinity
+  _ - _ = 0
+
+-- | Internal way to shift count to positive integers when hashing.
+hashCount :: Num p => Count -> p
+hashCount c = case c of
+  Infinity -> 1
+  Count a -> 2 + fromIntegral a
```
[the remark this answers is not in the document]
xsebek: Well, how would you hash `Maybe Int`?

src/Swarm/Game/Step.hs
xsebek: If we go the pos/neg infinity way, we could give players the ability to count with actual infinities assuming they have something infinite in inventory (or they divide by zero?).

But without it, I could not really come up with a meaningful answer.

src/Swarm/Game/Entity.hs
```diff
@@ -448,9 +450,49 @@ entityInventory = hashedLens _entityInventory (\e x -> e {_entityInventory = x})
...
+instance Num Count where
+  Count a + Count b = Count $ a + b
+  _ + _ = Infinity
+  Count a * Count b = Count $ a * b
+  _ * _ = Infinity
+  abs = id
+  signum (Count 0) = 0
+  signum _ = 1
+  fromInteger = Count . fromInteger
+  Count a - Count b = Count $ a - b
```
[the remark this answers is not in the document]
xsebek: On `3 - 5` it raises an underflow error.

src/Swarm/Game/Entity.hs
byorgey: Yes, I think that should work.

src/Swarm/Game/Entity.hs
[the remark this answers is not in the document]
byorgey: Eek!  Yeah, I definitely don't want Natural subtraction in there if it can crash the game.  We could make our own `Natural` type which uses saturating subtraction (i.e.  3 - 5 = 0).

src/Swarm/Game/Entity.hs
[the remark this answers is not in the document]
byorgey: Coming up with *a* way to hash it is easy, the problem is that we need some kind of homomorphism.  I will try to write up a detailed explanation somewhere else.

src/Swarm/Game/Step.hs
byorgey: Yeah, and it would be annoying for `count` to always return something of type `() + int` just to deal with the possibility of infinity.  But at least it should be a really big number like a million or a billion or something like that, instead of 42.

byorgey:
Random thought: what if the Swarm `int` type included positive and negative infinity?  That would solve the problem with `count` having to return an arbitrary value when you have infinity of something.  We'd have to consider carefully how arithmetic would work, and we'd probably have to add some new primitives like `isInf`.  But it doesn't seem like an immediately terrible idea.

byorgey:
Let me see if I can write down a coherent explanation of what's going on with the hashing.  The basic issue is with hashing inventories.  We want our hashes to satisfy two main criteria:
1. Two unequal inventories should have a high probability of having different hashes.  
2. It should be possible to update the hash of an inventory in O(1) when an entity is added or removed.

Note that inventories are considered unequal if there is at least one entity that the two inventories have different numbers of.  However, in this case the "number of" an entity can include any nonnegative value but ALSO a special "Nothing" value meaning "I have never heard of this entity".  We are now proposing to add yet another special value, positive infinity.

Criterion (2) above is what makes things interesting.  Without that criterion everything becomes easy, in fact, we could auto-generate the hashing code for an inventory.  But the auto-generated code iterates through the whole inventory and combines hashes for entities in a linearly chained way, so inserting or removing an entity means iterating through the entire inventory to recompute the hash from scratch.  #294 fixed this by switching to a homomorphic hashing scheme for inventories.  Instead of combining hashes by concatenating them and running the hash function again (say), we simply add them.  This is less cryptographically secure but we don't care about that, and it gives us criterion (2): given the existing hash for an inventory, if we insert an entity, just add the entity's hash to the inventory hash.  Also, if we add 100 copies of a given entity, we can just multiply the hash by 100 and then add that, rather than adding it 100 times.

Currently, if we have `k` copies of an entity in the inventory, it contributes (k+1) times the entity's hash to the inventory hash.  The reason for the k+1 is to accommodate the difference between not knowing about an entity at all (in which case its hash is not added to the inventory hash at all) and knowing about it but having 0 (in which case its hash is added once).  This works though since it is homomorphic with respect to adding and removing entities.  If we insert one copy of an entity into an inventory, we just add its hash to the inventory hash.  Say we used to have 3 copies, which means 4 times its hash was contributed to the inventory hash... so now 5 times its hash will be contributed, which corresponds to having 4 copies, just like it should.

However, positive infinity throws a wrench into things.  I think we can make it work but we need to check more cases.  As a concrete proposal, let `h` denote the hash of an entity, and suppose we use
- 0 to represent not knowing about an entity
- 1h to represent having infinity of the entity
- 2h to represent having 0
- 3h to represent having 1
- ... (k+2)h to represent having k

So far, so good.  But one problem is that this is not homomorphic with respect to adding and removing.  Suppose we have infinity copies of an entity, and we insert one more.  Now we should still have infinity, but that means we can't just add a copy of the hash every time we insert.  We have to first check to see how many copies we have before the insert.
- If we didn't know about the entity before, we should add 3h to the inventory hash.
- If we had some finite number of copies, we add h to the inventory hash.
- If we had infinite copies, we don't change the inventory hash.

Taking a union of inventories is also more complex.  Honestly it seems like when unioning two inventories we might as well go ahead and recompute the hash of the result from scratch, since it is not worth the trouble of trying to update incrementally.

xsebek:
> what if the Swarm int type included positive and negative infinity?

I am starting to like that idea. It would make it easier to use code written with the expectation that `Count = Int` and make division safer in the common case `1/0`. There would be one funny case of `1/0 - 1/0`, which is not that likely to bother anyone.

xsebek:
> ... If we had infinite copies, we don't change the inventory hash.

Well, that is annoying. Maybe we can use `updateLookupWithKey` to get the new value and if it is the same infinity we do not add the hash?

byorgey:
> Well, that is annoying. Maybe we can use `updateLookupWithKey` to get the new value and if it is the same infinity we do not add the hash?

Yeah, I think something like that could work.

xsebek:
Rebased and changed numbers to use negative and positive infinity. :slightly_smiling_face:

src/Swarm/Game/Entity.hs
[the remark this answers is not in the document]
xsebek: Solved by using negative infinity. :heavy_check_mark:

src/Swarm/Game/Step.hs
xsebek: This looks like a good candidate for an integration test. :slightly_smiling_face: 

We can test that the base in the creative scenario has an infinite count of items.

xsebek:
Rebased again to see if the VS Code will be fixed :slightly_smiling_face:

src/Swarm/Game/Step.hs
xsebek: :heavy_check_mark:

src/Swarm/Game/Entity.hs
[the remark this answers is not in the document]
xsebek: It is not pleasant, but it works now. :heavy_check_mark:

src/Swarm/Game/Entity.hs
xsebek: :heavy_check_mark:

src/Swarm/Game/Step.hs
```diff
@@ -1399,6 +1400,16 @@ execConst c vs s k = do
...
+            let salvageItems =
+                  if system || creative
+                    then []
+                    else concatMap (uncurry replicateCount) (E.elems salvageInventory)
+                numItems = length salvageItems
+                replicateCount n e =
+                  e ^. entityName & case n of
+                    E.Integer x -> replicate (fromIntegral x)
+                    E.PosInfinity -> replicate 42
+                    E.NegInfinity -> replicate (-42)
```
xsebek, line 1412: Ooops, this is a problem. How should we salvage infinite inventory? :thinking:

byorgey:
Something doesn't look right:

num-printing

src/Swarm/Game/Entity.hs
```diff
@@ -450,9 +452,7 @@ entityInventory = hashedLens _entityInventory (\e x -> e {_entityInventory = x})
 -- Inventory
 ------------------------------------------------------------
 
--- | A convenient synonym to remind us when an 'Int' is supposed to
---   represent /how many/ of something we have.
-type Count = Int
```
byorgey, line 455: We could still keep this comment but just change `Int` to `Number`.

src/Swarm/Game/Entity.hs
```diff
@@ -523,18 +523,38 @@ fromList = foldl' (flip insert) empty
...
+insertInventoryList :: Inventory -> [(Count, Entity)] -> Inventory
+insertInventoryList = foldl' (flip (uncurry insertCount))
+
+-- | Internal way to shift numbers to positive integers when hashing.
+--
+-- THIS DOES NOT FOLLOW ARITHMETIC LAWS, USE EXTREME CAUTION!
+hashCount :: Num p => Number -> p
+hashCount c = case c of
+  PosInfinity -> 1
+  Integer a | a >= 0 -> 1 + fromIntegral a
```
byorgey, line 537: Doesn't this mean `PosInfinity` and `Integer 0` both turn into `1`?  Maybe I'm not understanding what this function is for.

src/Swarm/Game/Entity.hs
```diff
@@ -523,18 +523,38 @@ fromList = foldl' (flip insert) empty
 
 -- | Create an inventory from a list of entities and their counts.
 fromElems :: [(Count, Entity)] -> Inventory
-fromElems = foldl' (flip (uncurry insertCount)) empty
+fromElems = insertInventoryList empty
+
+insertInventoryList :: Inventory -> [(Count, Entity)] -> Inventory
```
byorgey, line 528: Let's add a Haddock comment.

src/Swarm/Game/CESK.hs
```diff
@@ -263,7 +263,7 @@ data CESK
     Up Exn Store Cont
   | -- | The machine is waiting for the game to reach a certain time
     --   to resume its execution.
-    Waiting Integer CESK
+    Waiting Number CESK
```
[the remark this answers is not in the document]
byorgey: Oh, wait, I think I get it --- `Number` is the Haskell type that is used internally to represent Swarm values of type `int`.  So since `wait : int -> cmd unit`, therefore `Waiting` has to take a `Number` because `wait infinity` is now valid Swarm code.  Did I understand it right?

src/Swarm/Game/Step.hs
```diff
@@ -417,8 +418,8 @@ stepCESK cesk = case cesk of
   In (TAntiInt v) _ s k ->
     return $ Up (Fatal (T.append "Antiquoted variable found at runtime: $int:" v)) s k
   -- Require and requireDevice just turn into no-ops.
-  In (TRequireDevice {}) e s k -> return $ In (TConst Noop) e s k
-  In (TRequire {}) e s k -> return $ In (TConst Noop) e s k
+  In TRequireDevice {} e s k -> return $ In (TConst Noop) e s k
+  In TRequire {} e s k -> return $ In (TConst Noop) e s k
```
byorgey, line 422: Did fourmolu insist on getting rid of these parens?  I understand intellectually why they are not needed, but I kinda hate it.  `In TRequire {} e s k` really looks like `In` is taking 5 parameters.

src/Swarm/Game/Step.hs
```diff
@@ -1375,22 +1377,21 @@ execConst c vs s k = do
...
             when (creative || inst `E.contains` logger) $ robotLog <>= target ^. robotLog
 
             -- Immediately copy over any items the robot knows about
-            -- but has 0 of
-            let knownItems = map snd . filter ((== 0) . fst) . elems $ salvageInventory
-            robotInventory %= \i -> foldr (insertCount 0) i knownItems
+            -- but has 0 of and in creative mode just get everything now.
+            let immediate = if system || creative then id else filter ((== 0) . fst)
+            let immediateItems = map snd . immediate . elems $ salvageInventory
+            robotInventory %= \i -> foldr (insertCount 0) i immediateItems
```
byorgey, line 1394: It looks to me like when in system or creative mode, this will identify all items in the salvaged robot's inventory as "immediate" items, but then put them in the salvaging robot's inventory with a count of 0, regardless of how many there were in the salvaged robot's inventory.

src/Swarm/Game/Step.hs
```diff
@@ -1627,13 +1641,13 @@ execConst c vs s k = do
             missingParentInv = neededParentInv `E.difference` parentInventory
             missingMap =
               M.fromList
-                . filter ((> 0) . snd)
+                . filter (\(n, x) -> x > 0 && (x /= PosInfinity || countByName n parentInventory /= PosInfinity))
```
byorgey, line 1644: Can you walk me through the logic here?  I'm not sure I follow.

src/Swarm/Game/Step.hs
```diff
@@ -1950,28 +1964,49 @@ incomparable v1 v2 =
...
+-- | Perform an integer division, but return infinity for division by zero.
+--   Dividing infinity by infinity is also undefined.
+safeDiv :: Has (Throw Exn) sig m => Number -> Number -> m Number
+safeDiv i 0 = return $ signum i * PosInfinity
+safeDiv (Integer a) (Integer b) = return . Integer $ a `div` b
+safeDiv (Integer _a) _ = return 0
+safeDiv i b@(Integer _b) = return $ b * i
+safeDiv _ _ = throwError $ CmdFailed Div "Dividing infinities"
+
+-- | Perform exponentiation, but fail on negative powers and negative number to the poer of infinity.
```
byorgey, line 1996:
```suggestion
-- | Perform exponentiation, but fail on negative powers and negative numbers to the power of infinity.
```

src/Swarm/Game/Step.hs
```diff
@@ -1950,28 +1964,49 @@ incomparable v1 v2 =
...
+safeExp :: Has (Throw Exn) sig m => Number -> Number -> m Number
+safeExp = \case
+  PosInfinity -> \b -> return $ PosInfinity * signum b
+  NegInfinity -> \b -> return $ NegInfinity * signum b
```
byorgey, line 2000: `infinity ^ 0` ought to be 1, not 0.

src/Swarm/Game/Step.hs
```diff
@@ -1627,13 +1641,13 @@ execConst c vs s k = do
...
               M.fromList
-                . filter ((> 0) . snd)
+                . filter (\(n, x) -> x > 0 && (x /= PosInfinity || countByName n parentInventory /= PosInfinity))
                 . map (swap . second (^. entityName))
                 . E.elems
                 $ missingParentInv
 
         -- If we're missing anything, throw an error
-        E.isEmpty missingParentInv
+        null missingMap
```
byorgey, line 1650: Why did this change from checking `missingParentInv` to checking `missingMap`?  Was this a bug before?

src/Swarm/TUI/View.hs
```diff
@@ -881,7 +881,11 @@ drawItem sel i _ (Separator l) =
   (if sel == Just (i + 1) then visible else id) $ hBorderWithLabel (txt l)
 drawItem _ _ _ (InventoryEntry n e) = drawLabelledEntityName e <+> showCount n
  where
-  showCount = padLeft Max . str . show
+  showCount = padLeft Max . str . prettyCount
+  prettyCount = \case
+    E.Integer a -> show a
+    E.PosInfinity -> "∞"
+    E.NegInfinity -> "-∞" -- this shoud never happen
```
byorgey, line 888:
```suggestion
    E.NegInfinity -> "-∞" -- this should never happen
```

byorgey (commented):
The only code I haven't yet gone over carefully is the inventory manipulation code in `Game.Entity`.

src/Swarm/Game/CESK.hs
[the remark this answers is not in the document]
xsebek: Yes, it might be less confusing for us to rename `int` to `number`.

Besides, it will eventually be useful to wait indefinitely and get woken up by some other robot.

src/Swarm/Game/Entity.hs
xsebek: Ups, I should have kept that. :+1:

src/Swarm/Game/Entity.hs
xsebek: Ah, I think I originally had `2 +` and then mistakenly changed that. That will be a pain to adjust again. :slightly_frowning_face:

src/Swarm/Game/Step.hs
xsebek: My HLint does not like it and I got tired of it mentioning it. I will change it to `(TRequire _num _ent)`.

src/Swarm/Game/Entity.hs
byorgey: To be clear, I don't yet fully understand how this is used so it might be correct.

src/Swarm/Game/Step.hs
byorgey: We can also turn off specific hlint hints in `hlint.yaml`.

src/Swarm/Game/Step.hs
xsebek: I don't even remember writing this code anymore. But it indeed looks wrong, I will rewrite it.

src/Swarm/Game/Step.hs
xsebek: To be safe removing an infinite count of items from an infinite count of items is still an infinite count of items.

But here we were being smart by removing the _robot inventory_ from _required items_.

So the infinity stays and is not 0.

It is a hack to keep the changes minimal, but we should eventually clean up the code by changing the data structures.
I will write my thoughts on that in a separate comment.

src/Swarm/Game/Step.hs
xsebek: Well spotted. :+1:

src/Swarm/Game/Step.hs
xsebek: In general, I think it is useful to remove extra parenthesis, but this is an edge case.

Presumably, it is different between some HLint versions because CI did not warn about it.

src/Swarm/Game/Step.hs
xsebek: I am not sure. But if I hack in the infinity check above, I also need to change this.

The inventory version is likely smart and will ignore 0. See my future comment about the overall data structure.

xsebek:
# My comments on the inventory representation
```Haskell
data Inventory = Inventory
  { counts :: IntMap (Count, Entity)
  , -- ...
```

While updating the code to work with infinities, I came to the conclusion that this is a problem.

In the infamous words of every programmer: "It worked."

But there are issues with this representation:
1. it allows an invalid state - count should never be negative
2. it is abused to also work as `known:: Set Entity` (or `IntSet` if we use hash) which causes us to do strange things when counting the hash
3. the stored entity reference is duplicating information that is known globally
4. **(new)** with infinite count the logic of subtraction does not hold

## Doubtful PROs of current representation

Keeping this machinery working is beyond my mental capabilities and I am not sure what are its benefits. 

#### Distributed entity knowledge

From the look of the code, I guess that it tries to solve the future issue of robots having entities that are not known globally.
At this point, I think it might have been premature.

Global knowledge of all entities is easier to reason about. In particular, it would allow us to detect hash collisions that might go unnoticed with distributed knowledge of entities between robots.

#### Performance

I hope that this was never a serious factor in the design. We do not measure the performance, so this could easily be slower and less memory efficient than using different representations designed to be easier to understand.

#### It fit the code somewhere

At some time there was probably a place in the code where the current representation made it easier to write.

That reason sounds good to me. Is it true still? The hacks that were required for `checkRequirements` suggest otherwise.

## KISS representation

I did not want to change the representation here originally, but updating the hashing once more sounds worse.

I propose a new representation that my simple mind could handle:
- global `GameState` will have _all_ entities in its `EntityMap`
- robots will have a special field for known entities of type `IntSet` using entity hashes
- global `GameState` will have known entities also stored as `IntSet` (currently it is `[Text]`)
- inventory will have a separate map for infinities and for normal counts:
  ```Haskell
  data Inventory = Inventory
    { finiteCounts :: IntMap Natural
    , infiniteCounts :: IntSet
    , inventoryHash :: Int
    }
  ```
- item reaching a zero count causes it to be deleted from the map

Hashing will stay holomorphic, but it will separately hash the infinities, the known items and the item counts.

**BONUS:** Containers should not have the ability to know items. That does not make sense and this will make it impossible.

src/Swarm/Game/Step.hs
xsebek: On IRC we agreed that it should take 1 tick. The idea is that an infinite item is like an item generator so we move it like it had count 1.

There is still an issue with how to do that. I think some "game state update" on the stack of the reprogrammed salvaged robot is the cleanest solution. An internal `giveInfinite` function could also make sense.

byorgey:
@xsebek thanks for writing that up!  I love it, and I think that will make a lot of code simpler and solve many of our issues with this PR.

xsebek:
@byorgey, I would like to revive this next, but I might need help. 😄 

As far as I know, the points in my final summary are still valid. I will see if they can be split into smaller issues.

byorgey:
> I would like to revive this next, but I might need help. 😄

Yes, please do! :tada:  According to my current thinking this is a prerequisite for #231 . I will try to read through this again soon and see how I might be able to help.

PR #658, opened 2022-09-01
Add cotton, strings, and nets

### LINKED ISSUE #132: Device to give exceptions capability
Right now the `try` and `raise` commands are available to all robots, but the intention is that there should be a device to give this capability.  I don't think it should be available too early in the game.  Suggestions welcome!

byorgey:
* Add a `cotton` entity that grows in some of the places there used to be flowers
* 4 `cotton` make one `string`, with a `small motor` catalyst.  Currently, `string` serves no purpose except being an ingredient for `net`, but I imagine we can find other fun things to do with `string`s.
* 256 `string` make one `net`.
* A `net` is a device that enables the `try` command.

This is intentionally a bit on the harder side to construct.  You need to set up some good automation to be able to crank out the nets if you want to be able to use `try` on your robots.

Closes #132.

### DIFF
--- a/data/scenarios/Testing/508-capability-subset.yaml
+++ b/data/scenarios/Testing/508-capability-subset.yaml
@@ -13,16 +13,22 @@ objectives:
       } { return false }
 solution: |
   log "Hi, I can build teleporting robots!";
-  build {log "Hi, I can teleport!"; teleport self (5,0); p <- whereami; log $ "I am at " ++ format p}
+  try {
+    build {log "Hi, I can teleport!"; teleport self (5,0); p <- whereami; log $ "I am at " ++ format p}
+  } {
+    fail "Fatal error: BASE: I could not build my robot! What happened?!"
+  }
 robots:
   - name: base
     loc: [0,0]
     dir: [1,0]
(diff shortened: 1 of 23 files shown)


xsebek (approved):
LGTM!

Maybe we could tie it to the string type. Otherwise it might be good to rename it to text.

byorgey:
Hmm, what if `string` is a device that allows `format` and `++`?  Right now a `logger` is required to use those.

Maybe renaming the `string` type to `text` is a good idea anyway.  Then the fact that a `string` is the device for doing stuff with `text` would itself be a kind of joke.

data/scenarios/Testing/508-capability-subset.yaml
```diff
@@ -13,16 +13,22 @@ objectives:
...
   - name: base
     loc: [0,0]
     dir: [1,0]
     devices:
       - logger
+      - net
       - 3D printer
     inventory:
       - [1, mirror]
+      - [1, string]
```
xsebek, line 31: :wink:

xsebek (approved):

PR #659, opened 2022-09-02
WIP: tweak recipe graph

xsebek:


xsebek:
It does not work yet - if someone has an idea how to get IO inside Dot, that would help me with debugging ;)

xsebek:
I would have to start over to remember how it all worked anyway, so closing this for now. :smile:

Issue #574, opened 2022-07-20
Syntax highlighting for code blocks


xsebek:
- part of #545 

As a quick improvement, we should be able to parse markdown code blocks and highlight those in the game UI.

If the first and last line of `Text` is ```, the lines in the middle should be (parsed as swarm code and) highlighted.

Before:
```yaml
  - 'Example:'
  - '  move; turn left; move; turn right'
```

After:
```yaml
  - 'Example:'
  - |
    ```
    move; turn left; move; turn right
    ```
```

As a bonus, we could easily check that the code parses in the CI. #309

byorgey:
As I mentioned in #545, I do not actually think highlighting code blocks will be easy, although I would be very happy to be proven wrong.  But I agree we should do this and at least get #309 .

xsebek:
Well for start I would not semantically highlight the code - just make it bold or italic or something like that - simple attribute of the `txt` block. :slightly_smiling_face: And of course drop the ```.

Highlighting code semantically would require highlighting individual words like in #545, so that is harder even without the extra work of matching parsed elements to code spans and highlighting those.

byorgey:
Ah, that makes sense, yes, that should be possible.

xsebek:
Turns out there is a library for this:
- https://hackage.haskell.org/package/brick-skylighting

xsebek:
Well, it also needs to work with our pretty printing. :thinking: 

The way to do that seems to be to:
1. [`annotate`](https://hackage.haskell.org/package/prettyprinter-1.7.1/docs/Prettyprinter.html#v:annotate) the `Doc`.
2. produce a [`SimpleDocStream`](https://hackage.haskell.org/package/prettyprinter-1.7.1/docs/Prettyprinter.html#t:SimpleDocStream) (we can specify line length)
3. turn that into a list (lines) of lists of text segments each with an assigned [`TokenType`](https://hackage.haskell.org/package/skylighting-core-0.13/docs/Skylighting-Types.html#t:TokenType).

xsebek:
Since I think the implementation is clear now, I will do semantic highlighting too.

Notes:
1. start by splitting the descriptions into plaintext paragraphs and code blocks
2. parse the code blocks
  - if it does not keep the original text so we can print it plainly (e.g. while developing)
3. pretty print the code blocks (consider line length)
4. set up an integration test to check that all code blocks are valid
5. follow the steps in the comment above to highlight the output
6. profit

xsebek:
[...]

PR #663, opened 2022-09-06
Add `ADT calculator` entity, sum and product type capabilities

### LINKED ISSUE #563: Add capability for working with type sums and products
There should be some capability(ies) and device(s) for working with type sums (alternative) and products (tuples):
- sums: **Inl**, **Inr**, **Case**
- products: **Fst**, **Snd**

It would help with organically introducing the language features in-game, provide a reference in entity description and also get notified about it by flashing command help as you obtained new capability.

Ideas:
- `printer + calculator --> typing device` that would allow both sums and products, with the explanation that you are performing arithmetic on types
  - the downside is that `calculator` is hard to make, so maybe the `typing device` could also provide normal arithmetic?

byorgey:
Add `ADT calculator` entity (recipe: `calculator` + `I/O cable` + `typewriter`) which conveys capabilities to work with sum and product types (`inl`, `inr`, `case`; pairs, `fst`, `snd`).

Closes #563 .

### DIFF
--- a/data/entities.yaml
+++ b/data/entities.yaml
@@ -923,19 +923,21 @@
     just a fancy version of the number 2, where the two things happen to be
     labelled `false` and `true`.
   - |
-    The product of two types is a type of pairs, since, for example, if `t` is a type with three elements, then there are 2 * 3 = 6 different pairs containing a `bool` and a `t`, that is, 6 elements
-    of type `bool * t`.
-    For working with products of types, the ADT
-    calculator enables
-    pair syntax `(a, b)` as well as the projection functions
-    `fst : a * b -> a` and `snd : a * b -> b`.
+    The product of two types is a type of pairs, since, for example,
+    if `t` is a type with three elements, then there are 2 * 3 = 6
+    different pairs containing a `bool` and a `t`, that is, 6 elements
+    of type `bool * t`.  For working with products of types, the ADT
+    calculator enables pair syntax `(a, b)` as well as the projection
+    functions `fst : a * b -> a` and `snd : a * b -> b`.
   - |
-    The sum of two types is a type with two options; for example, a value of type `bool + t` is either a `bool` value or a `t` value, and there are 2 + 3 = 5 such values.
-    For working with sums of types, the ADT calculator provides
-    the injection functions `inl : a -> a + b` and `inr : b -> a + b`, as
-    well as the case analysis function `case : (a + b) -> (a -> c) -> (b -> c) -> c`.
-    For example, `case (inl 3) (\x. 2*x) (\y. 3*y) == 6`, and
-    `case (inr 3) (\x. 2*x) (\y. 3*y) == 9`.
+    The sum of two types is a type with two options; for example, a
+    value of type `bool + t` is either a `bool` value or a `t` value,
+    and there are 2 + 3 = 5 such values.  For working with sums of
+    types, the ADT calculator provides the injection functions `inl :
+    a -> a + b` and `inr : b -> a + b`, as well as the case analysis
+    function `case : (a + b) -> (a -> c) -> (b -> c) -> c`.  For
+    example, `case (inl 3) (\x. 2*x) (\y. 3*y) == 6`, and `case (inr
+    3) (\x. 2*x) (\y. 3*y) == 9`.
 
   properties: [portable]
   capabilities: [arith, sum, prod]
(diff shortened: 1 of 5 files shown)


data/entities.yaml
```diff
@@ -908,6 +908,28 @@
...
+    addition to the usual arithmetic on numbers, an ADT calculator can
+    also do arithmetic on types!
+  - |
+    For working with sums of types, the ADT calculator provides
+    the injection functions `inl : a -> a + b` and `inr : b -> a + b`, as
+    well as the case analysis function `case : a + b -> (a -> c) -> (b -> c) -> c`.
+  - |
+    For working with products of types, the ADT calculator enables
+    pair syntax `(a, b)` as well as the projection functions
+    `fst : (a,b) -> a` and `snd : (a,b) -> b`.
```
xsebek, line 928:
```SUGGESTION
    `fst : a * b -> a` and `snd : a * b -> b`.
```
This might need some extra explanation... :thinking:

xsebek (approved):

data/entities.yaml
xsebek: I like the popular explanation that the product/sum also works for the _size_ of the type.

So that `|Bool * Bool|` is 4, `|() + Bool|` is 3, etc.

Maybe a paragraph with such an example would help illuminate the "arithmetic on types" if the player never heard about it.

xsebek:
@byorgey if you feel like the example would be too long, we can leave it for a tutorial.

But I think this should be explained a little for someone who is just playing a classic game and does not know what this is about.

Makes me think this should maybe have its help section since you will see it before crafting the calculator. :thinking:

byorgey:
What do you mean, you will see it before crafting the calculator?

byorgey:
I agree we need more examples/explanation though.

xsebek:
By “it” I meant the sum type which you will see in other commands return type before crafting the calculator.

I might be overthinking it though, maybe we should release Swarm and then fix actually confusing things. 😆

data/entities.yaml
byorgey: I like the funny tone of your suggested examples, but I am afraid for someone who has never seen this before they don't really explain it enough.  I tried adding something that explains it a bit more straightforwardly (with some key terms that someone can google).

I still think this won't be enough for someone who has never seen this kind of thing before, but I don't think we can stuff much more in the description. It's worth thinking about what kinds of optional tutorial scenarios we could make that would guide people through familiar FP concepts like lambdas, pair and sum types, etc. if they have never seen those things before.

byorgey:
@xsebek What do you think?

data/entities.yaml
```diff
@@ -908,6 +908,38 @@
...
+    A calculator with Advanced Display Technology (an attached
+    typewriter that can print out the results).  For some reason, in
+    addition to the usual arithmetic on numbers, an ADT calculator can
+    also do arithmetic on types!  After all, the helpful typewritten manual
+    explains, a type is just a collection of values, and a finite collection
+    of values is just a fancy number.  For example, the type `bool` is
+    just a fancy version of the number 2, where the two things happen to be
+    labelled `false` and `true`.
+  - |
+    The product of two types is a type of pairs, since, for example, if `t` is a type with three elements, then there are 2 * 3 = 6 different pairs containing a `bool` and a `t`, that is, 6 elements
```
xsebek, line 926: Strange indentation. 😁

data/entities.yaml
```diff
@@ -908,6 +908,38 @@
...
+    calculator enables
+    pair syntax `(a, b)` as well as the projection functions
+    `fst : a * b -> a` and `snd : a * b -> b`.
+  - |
+    The sum of two types is a type with two options; for example, a value of type `bool + t` is either a `bool` value or a `t` value, and there are 2 + 3 = 5 such values.
+    For working with sums of types, the ADT calculator provides
+    the injection functions `inl : a -> a + b` and `inr : b -> a + b`, as
+    well as the case analysis function `case : (a + b) -> (a -> c) -> (b -> c) -> c`.
+    For example, `case (inl 3) (\x. 2*x) (\y. 3*y) == 6`, and
+    `case (inr 3) (\x. 2*x) (\y. 3*y) == 9`.
```
xsebek, line 938: Code blocks (#574) would be really useful here. 🙂

xsebek (approved):
LGTM! 👍

data/entities.yaml
byorgey: Good point, fixed!

PR #668, opened 2022-09-08
Fix VSCode installation in CI

xsebek:


byorgey (approved):
No clue what this does but I trust you!

Issue #5, opened 2021-09-20
Add a way to visualize a robot's currently executing program


byorgey:
That is, a way to display the current state of the CEK machine for the focused robot.  This would require first overhauling CEK machine pretty-printing.  I am imagining showing the continuation stack as *nested* frames rather than a sequence of frames.  For example given the program `let x = 3 + 2 in x * 4`, when we are currently focused on evaluating the `3`, we could imagine something like
``` 
[let x = [[3] + 2] in x * 4]
```
or maybe even without all the brackets except the ones around the 3.  This would probably require some creativity in certain cases but I think it could be done in a nice, compact way.

The second part of this would be devising ways both (1) for the user to possibly choose whether they want to see this, and (2) figuring out a good place for it in the UI.

byorgey:
Perhaps it should only be enabled when pausing and single-stepping the game.

byorgey:
Maybe we could have a "debugger" or "tracer" device or something of that sort.  Then, when viewing a given robot and focusing on that installed device, the current machine state would show in the text box at the bottom left of the UI.

byorgey:
I split out the idea of better CESK pretty-printing into #644.

xsebek:
@byorgey as to where it should be placed in the UI, I would suggest the robot modal, but only show it for selected robots.

Imagine you open the list, scroll to the desired robot, hit <kbd>I</kbd> and the one robot line expands to few more lines showing (likely rapidly changing) pretty printed CESK stack.

byorgey:
Having CESK debugging info show up in the robot modal makes sense, and the motivation of wanting to inspect several robots quickly makes sense.  However, it wouldn't be good if the robot modal was the only way to view pretty printed CESK machines, because the modal hides your view of the world.  I can easily imagine wanting to single-step the game while seeing what a robot is doing in the world and also viewing its CESK machine.

xsebek:
Good point, both are good. :+1:

xsebek:
Btw. how about increasing the priority, since it would have been so handy for debugging? :slightly_smiling_face: 

Also, it seems to me like a very interesting thing from players' perspective, as it allows them (me really) to understand the evaluation model better.

byorgey:
[...]

PR #669, opened 2022-09-08
Remove test challenges and add some more interesting ones

### LINKED ISSUE #635: Curate challenge scenarios
We have several testing scenarios that we should get rid of (or move to `Testing/`), and we should add a few more interesting challenges.  A few brainstorming ideas I had:

- towers of hanoi
- sorting
- places to explore
- rpg?
- connect matching things
- carry stuff around in a factory
- sokoban? (pushable entities?)
- 2048 (make 2 1s into 2, 2 2s into 4, etc.  start with a single 1)

byorgey:
Remove a couple challenges which were just there as placeholders (`test` and `drill_test`; the latter is now moved to `Testing`), and add more interesting `2048` and `hanoi` challenges.

Closes #635 .


data/scenarios/Challenges/drill_test.yaml
```diff
@@ -1,113 +0,0 @@
-name: Test drill
```
xsebek, line 1: @byorgey could you move it to tests instead, please? If I have to, I will write down a solution.  :smile:

data/scenarios/Challenges/2048.yaml
```diff
@@ -0,0 +1,187 @@
...
+        as base {has "2048"}
+      } { return false }
+solution: |
+  def makeN : int -> cmd () = \n.
+    if (n == 1)
+      {harvest; return ()}
+      {makeN (n/2); makeN (n/2); make (format n)}
+  end;
+  place "1";
+  makeN 2048
```
xsebek, line 18: Nice solution, took me a while to beat it! :laughing: 

<table>
<tbody>
<tr>
<td>

@byorgey

</td>
<td>

@xsebek

</td>
</tr>
<tr>
<td>

`1:29:d` ticks

</td>
<td>

`1:01:b` ticks

</td>
</tr>
<tr>
<td>

Screenshot from 2022-09-09 01-10-48

</td>
<td>

Screenshot from 2022-09-09 01-20-04

</td>
</tr>
<tr>
<td>

```
def makeN : int -> cmd () = \n.
  if (n == 1)
    {harvest; return ()}
    {makeN (n/2); makeN (n/2); make (format n)}
end;
place "1";
makeN 2048
```

</td>
<td>

```
def for: int -> (int -> cmd a) -> cmd () = \i.\c.
  if (0 > i) {} {c i; for (i - 1) c}
end;

def grow2N: int -> cmd () = \n.
    for (2^n) (\i. harvest);

    for (n-1) (\i.
        for (2^i - 1) (\j.
            make (format $ 2^(n-i))
        )
    );
end;

place "1";
grow2N(0xB)
```

</td>
</tr>
</tbody>
</table>

xsebek: Btw. this is a neat command:
```bash
S=scenarios/Challenges/2048.yaml
cabal run swarm:swarm -O2 -- --scenario $S --run <(yq '.solution' $S)
```

data/scenarios/Challenges/drill_test.yaml
byorgey: Good idea, will do!

data/scenarios/Challenges/2048.yaml
byorgey: Wow, that command came in super handy when testing the solution to the drill test!

data/scenarios/Testing/373-drill.yaml
```diff
@@ -50,7 +61,7 @@ world:
     ┌─────┐                                                                ~~
     │IAAT~                                                                ~L~
     │..AAA│                                                                ~~
-    │....C│                                                                 ~
+    │Ω...C│                                                                 ~
     └─────┘                                                                 ~
 entities:
   - name: copper vein
```
xsebek, line 67: I think we need a regular mountain entity too, otherwise, this can randomly fail.

Some mountain recipes do not give you rocks, e.g. `mountain tunnel`+`lodestone` which is very common now.

data/scenarios/Challenges/hanoi.yaml
```diff
@@ -0,0 +1,91 @@
...
+    condition: |
+      try {
+        teleport self (2,-1);
+        x <- ishere "one";
+        teleport self (2,-2);
+        y <- ishere "blocked two";
+        teleport self (2,-3);
+        z <- ishere "blocked three";
+        return (x && y && z)
+      } { return false }
```
xsebek, line 15: Weird, if I factor it out the checking to a system robot, it works fine. :exploding_head:

data/scenarios/Challenges/2048.yaml
```diff
@@ -0,0 +1,188 @@
...
+    display:
+      attr: gold
+      char: "8"
+    description:
+    - An eight.
+    properties: [portable]
+  - name: "16"
+    display:
+      attr: gold
+      char: "#"
```
xsebek, line 53: How about using powers of two instead? :slightly_smiling_face: So 1 is 0, 2 is 1, 4 is 2,...

data/scenarios/Challenges/hanoi.yaml
xsebek: I just fixed the exception printing and got this:
image

data/scenarios/Challenges/hanoi.yaml
```diff
@@ -63,18 +102,33 @@ entities:
     description:
     - A disk of radius 1.
     properties: [portable]
-  - name: blocked three
+  - name: blocked one
```
xsebek, line 105: The blocked one should not be necessary, but with race conditions, it could come up and it's easier to handle it this way.

data/scenarios/Challenges/hanoi.yaml
xsebek: #675 should fix this. Once that is merged, this PR needs to be rebased on `main` and the tests should pass.

data/scenarios/Challenges/hanoi-invariant.sw
```diff
@@ -0,0 +1,47 @@
...
+      allSorted <- as self {
+        teleport self (undefined);
+        o1 <- ishere "OK";
+        teleport self (undefined);
+        o2 <- ishere "OK";
+        teleport self (undefined);
+        o3 <- ishere "OK";
+        return (o1 && o2 && o3)
+      };
+      if allSorted {grab; place (unlock e)} {}
```
xsebek, line 44: > `grab; place (unlock e)`

@byorgey can this be done atomically? I would like to `swap` the entities without there ever being an empty space.

data/scenarios/Challenges/2048.yaml
byorgey: Oh, good idea.  Using hex digits, of course, so 2048 will look like `b`.

xsebek:
I rebased this and finished the Hanoi tower implementation. At least I hope so. I run into some bugs with the counting robot.

@byorgey do you want to try to solve the challenge? :slightly_smiling_face:

data/scenarios/Challenges/hanoi-invariant.sw
```diff
@@ -0,0 +1,51 @@
...
+end;
+
+repeat (
+me <- scan down;
+case me (\_. return ()) (\e.
+// if
+//  0. I stand on unlocked X
+//  1. place north of me is NOT empty
+// then
+//  - lock X
```
xsebek, line 18: While playing around I found another condition:

If:
- I stand on an unlocked disk
- the count of all placed is not 3

Then:
- I should lock the disk to preserve the invariant

data/scenarios/Challenges/2048.yaml
byorgey: So I tried that and it turned out to just be super confusing when you see them in the inventory.  The `char` attributes are never going to be seen or matter anyway, they just show up in the inventory.  So instead I changed them all to `#`.

data/scenarios/Challenges/hanoi.yaml
byorgey: I could have sworn that I had other objectives specified using `teleport self` before... but maybe not.

data/scenarios/Challenges/hanoi-invariant.sw
byorgey: No, because `grab` and `place` are both tangible commands, and you're allowed at most one of those inside an `atomic` block.  However, I think I would be fine with putting in an exception for system robots.  Not quite sure how that exception would work in practice though, since checking of `atomic` blocks is done at typechecking time, not runtime...

We could also just add a special `swap : str -> cmd str` which does a `grab` and a `place` atomically?

data/scenarios/Challenges/hanoi-invariant.sw
byorgey: How is it possible to get into that state though?

byorgey:
@xsebek I solved the hanoi challenge by hand.  Now I'm trying to write a program to solve it automatically.  After the usual struggles with type errors etc., the first time it ran the system robots weren't fast enough at updating the state so my program crashed when it tried to pick up the 2 which hadn't been changed into a grabbable 2 yet.  So I added a manual `wait`.  Then after grabbing and placing the 1 and the 2 successfully, I got this lovely error:
```
Fatal error: Incompatible comparison of 0 and true
```
Yikes!  Maybe you are right about the stack getting screwed up somehow.  Need to do some debugging here...

data/scenarios/Challenges/hanoi-invariant.sw
xsebek: I think `swap` will be better and also faster since it will only be one command. :+1: 

We can add it later, though the Hanoi tower challenge would look nicer with `swap`.

byorgey:
For reference, here is my proposed automated solution for the hanoi challenge:
```
def until = \p. \c. b <- p; if b {} {c; until p c} end
def rep = \n. \c. if (n == 0) {} {c; rep (n-1) c} end

def fwd = until blocked move end

def onCol = \c.
  fwd;
  res <- c;
  turn back;
  fwd;
  turn back;
  return res
end

def getDisk = onCol grab end
def placeDisk = \d. onCol (place d) end

def moveToCol = \x.
  w <- whereami;
  if (fst w < x) { turn east; rep (x - fst w) move }
  { if (fst w > x) { turn west; rep (fst w - x) move } {} };
  turn south
end

def hanoi : int -> int -> int -> int -> cmd () =
  \n. \a. \b. \c.
  if (n == 0) {}
  {
    hanoi (n-1) a c b;
    moveToCol a;
    wait 32;
    d <- getDisk;
    moveToCol c;
    placeDisk d;
    hanoi (n-1) b a c;
  }
end
```

data/scenarios/Challenges/hanoi-invariant.sw
xsebek: If you have for example:
```
| | | |
| | | |
|1|2|3|
+-+-+-+
```
Then at that moment all of them will be unlocked. Which makes sense (except for the 3).

Once you grab any of them, the others should immediately be locked, so that you can not grab more and swap them.

data/scenarios/Challenges/hanoi-invariant.sw
byorgey: https://github.com/swarm-game/swarm/issues/680

data/scenarios/Testing/373-drill.yaml
byorgey: I changed one of the mountains to a boulder which is guaranteed to yield `rock` when drilled.

data/scenarios/Challenges/hanoi-invariant.sw
byorgey: Ah, I see!  You're right, I can confirm that at the moment it is possible to cheat: put them as in the picture you drew above, then pick them all up, move to the last column, and place them going up the column one after another.

byorgey:
Definitely faster now, but I still get the weird `Fatal error`.  Man, I really wish I had https://github.com/swarm-game/swarm/issues/5 .

xsebek:
Sorry about the system base, @byorgey, I needed that to test that the “creative solution” is at least recognised by the winning condition. 😅

xsebek:
I added your solution @byorgey, but could not get it to work with the last rule.

The game starts oscillating and whatever waiting time I choose, eventually, the robot hits a blocked disk.

I hope the `swap` command will fix that. (#684)

data/scenarios/Challenges/hanoi-invariant.sw
xsebek: > I think swap will be better and also faster...

... and necessary. The last rule makes race conditions much more likely. :weary:

xsebek:
It works! :partying_face: 

@byorgey do you want to add some magic GPS device just for this challenge or modify the solution to work without `whereami`?

After that, the base would not need to be a system robot and we could merge this. :slightly_smiling_face:

byorgey:
Yes, I think we should add a GPS device that enables `whereami`.  We can come up with a recipe later.

byorgey:
@xsebek Do you want to add a `GPS` device or shall I?

xsebek:
Actually, it is really easy to do without `whereami` here, we can add it in another PR. :slightly_smiling_face:

byorgey:
> Actually, it is really easy to do without `whereami` here, we can add it in another PR. :slightly_smiling_face: 

Speak for yourself, I tried doing that before and got myself very confused. :laugh:

data/scenarios/Challenges/2048.yaml
```diff
@@ -0,0 +1,188 @@
+name: "2048"
+description: Make 2048!
+objectives:
+  - goal:
+    - OK, it's not really the same as the classic "2048" game.
+      However, your goal is still to make 2048.
```
xsebek, line 6: This description is technically all you need, but maybe it could be made more exciting. :grinning: 

The "make 2048" could also be expanded upon. In my experience, since you do not see under the robot it is not immediately obvious what you should do with the `1`. IMO, it should be noted that the `1` are harvestable.

xsebek (approved):

byorgey:
Yay!  Thanks for the help on the hanoi challenge!  It seems to work great.   I will edit the PR description then going to merge this.

PR #672, opened 2022-09-09
Fix best time format

xsebek:
- show minutes and seconds modulo 60


xsebek:
Here I noticed something was off: :sweat_smile: 
image

byorgey (approved):

PR #674, opened 2022-09-11
Use system robot for Hanoi challenge checking

xsebek:


xsebek:
With #675 this hack should not be necessary.

PR #675, opened 2022-09-11
Fix various robot bugs

xsebek:
- make sure that when the hypothetical robot crashes the game does not
- allow system robots to pick up anything
- let self teleporting robots touch themselves

### DIFF
--- a/src/Swarm/Game/Step.hs
+++ b/src/Swarm/Game/Step.hs
@@ -133,8 +133,7 @@ gameTick = do
         Left exn -> do
           em <- use entityMap
           time <- use ticks
-          let unused = error "Unexpectedly used field"
-              h = hypotheticalRobot unused unused
+          let h = hypotheticalRobot (Out VUnit emptyStore []) 0
               hid = view robotID h
               hn = view robotName h
               farAway = V2 maxBound maxBound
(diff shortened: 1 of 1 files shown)


src/Swarm/Game/Step.hs
```diff
@@ -1671,13 +1670,17 @@ execConst c vs s k = do
...
-    mother <- robotWithID rid
-    other <- mother `isJustOrFail` ["There is no robot with ID", from (show rid) <> "."]
-    -- Make sure it is in the same location
-    loc <- use robotLocation
-    ((other ^. robotLocation) `manhattan` loc <= 1)
-      `holdsOrFail` ["The robot with ID", from (show rid), "is not close enough."]
-    return other
+    cid <- use robotID
+    if cid == rid
+      then get @Robot
```
byorgey, line 1675: I don't quite understand why this didn't work before - isn't a robot within Manhattan distance 1 of itself?  But this code is certainly sensible.

byorgey (approved):
LGTM!

src/Swarm/Game/Step.hs
xsebek: Hah, not if it is not inside the robot map. :sweat_smile:

byorgey: Ahhh, of course!

PR #677, opened 2022-09-11
Add easy mazes

xsebek:
- add one maze shaped like a cave and one shaped like a spiral
- part of #635


data/scenarios/Challenges/Mazes/easy_spiral_maze.yaml
```diff
@@ -0,0 +1,104 @@
...
+      if b {
+        return true; 
+      } {
+        move;
+        return false;
+      }
+    );
+    grab
+  }
+robots:
```
byorgey, line 30: My solution, similar to the previous:
```
def until = \p. \c. b <- p; if b {} {c; until p c} end
def fwd = until blocked move end
build {turn west; until (ishere "goal") (fwd; turn right); grab}
```

data/scenarios/Challenges/Mazes/easy_spiral_maze.yaml
```diff
@@ -0,0 +1,104 @@
...
+      - [50, solar panel]
+      - [50, treads]
+      - [50, compass]
+      - [50, scanner]
+      - [50, lambda]
+      - [50, branch predictor]
+      - [50, strange loop]
+      - [50, logger]
+      - [50, grabber]
+      - [0, boulder]
```
byorgey, line 51: Another 0 boulders.  Was this a copy-pasto?

byorgey (approved):
These are great!

data/scenarios/Challenges/Mazes/easy_spiral_maze.yaml
xsebek: Of course, I don't know how to type Omega and also don't want to type every needed device. :smile:

It's probably leftover from the other scenarios that use boulders as walls. I will fix that. :wink:

data/scenarios/Challenges/Mazes/easy_cave_maze.yaml
```diff
@@ -0,0 +1,102 @@
...
+      - [50, solar panel]
+      - [50, treads]
+      - [50, compass]
+      - [50, scanner]
+      - [50, lambda]
+      - [50, branch predictor]
+      - [50, strange loop]
+      - [50, logger]
+      - [50, grabber]
+      - [0, boulder]
```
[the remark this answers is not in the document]
xsebek: :sweat_smile:

data/scenarios/Challenges/Mazes/easy_cave_maze.yaml
```diff
@@ -0,0 +1,102 @@
...
+      b <- if b { turn back;  blocked } { return false };
+      if b {
+        return true; 
+      } {
+        move;
+        return false;
+      }
+    );
+    grab
+  }
```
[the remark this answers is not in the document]
xsebek: Nice 3-liner in REPL, I respect that. :+1:

I think it's even faster and shorter than mine when pretty printed, but I have not measured it yet. I will probably use yours, it looks nicer. :smile:

data/scenarios/Challenges/Mazes/easy_spiral_maze.yaml
xsebek: Weird, in REPL it works, but when I `--run` it, the robot stops in the wrong direction at the start.

byorgey: Did you have `turn west` or did you copy-paste the `turn right` from the previous solution?  ...it shouldn't matter though...

byorgey: Weird. I added a logger and a command to `view` the built robot, and then executed it using `run`, and got an error about needing to install a `branch predictor`.  Then in the same REPL session, I simply hit "up" and executed the exact same `run` command again... and the second time, it worked. :fearful:

byorgey: To be more specific, the robot is getting built, and it executes the `turn west`, and then when it goes to execute the `until` it fails on the runtime capability check for `if`, because it indeed does not have a `branch predictor` installed.  (It has a `compass`, `grabber`, `logger`, `scanner`, `solar panel`, and `treads`, but no branch predictor).

xsebek: Oh, so that is what it was. 😮 

Thanks for debugging it @byorgey. 👍

TristanCacqueray (approved):

data/scenarios/Challenges/Mazes/easy_spiral_maze.yaml
xsebek: I used your solution @byorgey with `require` as a temporary workaround. :+1:

PR #678, opened 2022-09-11
Fallback to XDG data directory

xsebek:
- when `datadir` is not available, try using the XDG data directory

This way the game can be installed as an executable and data files unpacked to  `~/.local/share/swarm/data`.
Notice that the XDG data folder is  `~/.local/share/swarm`; inside it is the unpacked `data`.

The alternative approach is to use the environment variable `swarm_datadir` and set that to the unpacked data folder.
That works (even after this change) but is not very beginner friendly.

Ideally, we would like to set this in Cabal when building executable, for example to `/usr/share/swarm/<version>`.
Reading through haskell/cabal#5997, it looks like that is not supported.

### DIFF
--- a/src/Swarm/Util.hs
+++ b/src/Swarm/Util.hs
@@ -63,7 +63,10 @@ module Swarm.Util (
 
   -- * Utilities for NP-hard approximation
   smallHittingSet,
-getDataDirSafe, getDataFileNameSafe, dataFileNotFound) where
+  getDataDirSafe,
+  getDataFileNameSafe,
+  dataFileNotFound,
+) where
 
 import Control.Algebra (Has)
 import Control.Effect.State (State, modify, state)
@@ -101,8 +104,10 @@ import System.Clock (TimeSpec)
 import System.Directory (
   XdgDirectory (XdgData),
   createDirectoryIfMissing,
+  doesDirectoryExist,
+  doesFileExist,
   getXdgDirectory,
-  listDirectory, doesDirectoryExist, doesFileExist
+  listDirectory,
  )
 import System.FilePath
 import System.IO
@@ -218,7 +223,7 @@ getDataDirSafe = do
     then return $ Just d
     else do
       xd <- (</> "data") <$> getSwarmDataPath False
-      xde <- doesDirectoryExist xd 
+      xde <- doesDirectoryExist xd
       return $ if xde then Just xd else Nothing
 
 getDataFileNameSafe :: FilePath -> IO (Maybe FilePath)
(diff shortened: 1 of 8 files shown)


xsebek:
@byorgey I still have to test this, but it looks like the easiest solution to me.

We can add a small bash snippet to the release description. Something like:
```bash
# disclaimer: I have not tried this yet and do not remember the correct flags
wget github.com/swarm-release/swarm
wget github.com/swarm-release/data.zip
mkdir -p ~/.local/share/swarm
unzip data.zip ~/.local/share/swarm/data
```

xsebek:
The "right way" to do this sort of stuff seems to be to use a wrapper script that sets the env var and runs the executable:
```bash
export swarm_datadir=$HOME/.local/share/swarm/data
swarm $@
```
The problem is that the swarm binary would have to be put _somewhere_ not in PATH and then called by this script...

That may be the path to go if we ever package Swarm so that can be installed by `apt` or other Linux distribution packager, but right now I want something simple that works. :confused:

src/Swarm/Util.hs
```diff
@@ -210,6 +215,30 @@ readFileMayT = catchIO . T.readFile
...
+  dir <- getDataDirSafe
+  case dir of
+    Nothing -> return Nothing
+    Just d -> do
+      let fp = d </> name
+      fe <- doesFileExist fp
+      return $ if fe then Just fp else Nothing
+
+dataFileNotFound :: Text
+dataFileNotFound = T.pack ""
```
xsebek, line 240: Note to self: Add some meaningful error message teling the user that the file/directory was not found and that they might want to download the data archive and extract it to `<xdg data>/data`.

byorgey (approved):
This seems good to me, though we should definitely make sure to test it as part of the final release checklist.  I assume you're going to fill in a more reasonable error message for `dataFileNotFound` before merging.

xsebek:
I tried this in a clean ubuntu bionic container and it seems to work*! :partying_face: 

---

*) There are a few issues:
- asking GitHub about releases fails with an exception that is written on the screen - yikes
  image
  we can fix that in another PR, it can probably be captured somehow and properly shown as an error message.
- the data is correctly loaded from XDG directory, but an exception is written to the terminal:
  ```
  root@d5b29255a7db:/# swarm
  /root/.cabal/store/ghc-8.10.7/swarm-0.0.0.1-5c2b0652808f070b61c234cf4e3209c1a263e73e5e3869130aa61d2b62d466a6/share: getDirectoryContents:openDirStream: does not exist (No such file or directory)root@d5b29255a7db:/#
  ```
  That is not so bad but should be fixed in this PR if possible.
- with the default `TERM=xterm` only basic colours work:
  image
  We should suggest setting `TERM=xterm-256color` in the installation section of the README.

Issue #480, opened 2022-06-24
Add optional `upgrades` field to entities


xsebek:
I would like to have a note in entities saying which entity (if any) they are a better version of.

For example "metal drill" upgrades "drill".

### Gains

This should
- resolve preference of entity when making a recipe, drilling or installing
- simplify documentation (namely recipe graph from #476)
- cut down on boilerplate in recipes (see below)

### Proposed solution

Add an optional **upgrades** field to entities that would be a structure with the following fields:
- **name** - string - name of the lesser entity (e.g. drill)
- **time multiplier** - optional float - how long will the recipes of the original as the catalyst take with this upgraded version
- **output multiplier** - optional float - how much will output from making or drilling with this entity as the catalyst increase compared to original

These fields would be parsed into a `Maybe Upgrades` field of the new upgraded entity*.
We could add capabilities of the original entity to the upgraded one or at least check it is a subset in CI.

After loading all recipes to the game, we would find those that have the original entity as a catalyst and add a copy with the upgraded entity, applying the multipliers if they were specified.

---

<sup>*) To keep passing simple we could not require resolving the entity name immediately and keep it as text.</sup>

byorgey:
I think it will be interesting to see how much we end up wanting this feature.  Currently, there are few enough "upgradeable" entities that it seems like implementing this might be more trouble than it's worth.  But if we came up with more uses it could be worth it.  On the other hand I definitely don't want Swarm to become the kind of game where there's a tree of upgrades for every entity.

On a more technical note, we would have to make sure that this interacts nicely with `require`.  For example, if you `build { drill forward }` and you have both a `drill` and a `metal drill`, we could install the `metal drill` by preference.  But if you `build { require "drill"; drill forward }` then we should make sure to install just a normal `drill` as requested and not also a `metal drill`.

xsebek:
@byorgey good points 🙂

I hope this will help with required entities - as a defaulting rule we could prefer installing one or the other.

Also if you write a definition that requires a drill and another one that requires a metal drill, then when you compose those, you get both.
If we knew which entity upgrades which, we could delete the “downgrades” from the set (`s \\ fmap down s`).
That way you would get only one drill installed.

I am not huge fan of upgrading everything, but here it would make the game logic easier to understand and more consistent. Rather then making ad-hoc rules that are hardcoded or recipes that we have to write by hand, we could make sure that upgrades follow a contract.

[...]

PR #685, opened 2022-09-14
Add swap command

### LINKED ISSUE #680: Add `swap` command to do `grab` + `place` atomically
We could also just add a special `swap : str -> cmd str` which does a `grab` and a `place` atomically?

_Originally posted by @byorgey in https://github.com/swarm-game/swarm/pull/669#discussion_r967874176_

### LINKED ISSUE #684: Add swap command
When writing robot programs it is easy to get into race conditions.

Consider this simplified example from #669:
- robot A at (1,1) watches location (2,3) and if it has an entity "three" it will grab and place
- robot B at (2,3) watches location (1,1) and if it is empty it will grab and place "two" otherwise "three"

The worst part is that in extended programs race conditions rarely happen because robots will spend some ticks not looking.

Adding an atomic `swap` command that works like `a <- grab; place e; return a` would solve these issues.

xsebek:
- closes #684
- closes #680

### DIFF
--- a/data/entities.yaml
+++ b/data/entities.yaml
@@ -656,8 +656,25 @@
   - "The 'install' command takes two arguments: the robot
     on which to install a device (which can be at most 1 cell away),
     and the name of the device to install."
-
   capabilities: [grab, give, place, install]
+  properties: [portable]
+
+- name: fast grabber
+  display:
+    attr: device
+    char: '≪'
+  description:
+  - A fast grabber is an improved version of the basic grabber - not only
+    can it 'grab', 'place', 'give', and 'install', it can also 'swap'.
+  - The 'swap' command allows the robot to execute grab and place at the
+    same time so that the location where the robot is standing does not
+    become empty.
+  - You can use this to prevent failures where multiple robots
+    are trying to grab, place or scan a given location.
+  - In addition you retain the capability to use the 'atomic' command,
+    with which you can implement other commands that are safe when
+    run in parallel.
+  capabilities: [grab, swap, give, place, install, atomic]
   properties: [portable]
 
 - name: harvester
@@ -1000,7 +1017,7 @@
     rock to grab.  However, it is actually possible for the 'grab' to
     fail, if some other robot B snatches the rock right after robot A sensed
     it and before robot A got around to grab it on the next game tick."
-  - This will make robot A very sad and it will crash.
+  - "This will make robot A very sad and it will crash."
   - "To prevent this situation, robot A can wrap the commands in 'atomic', like so:"
   - |
     atomic (b <- ishere "rock"; if b {grab} {})
@@ -1020,4 +1037,3 @@
   - will attempt to move, but if that fails, turn left instead.
   properties: [portable]
   capabilities: [try]
-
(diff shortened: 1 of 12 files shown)


src/Swarm/Language/Capability.hs
```diff
@@ -193,6 +193,7 @@ constCaps = \case
   Exp -> Just CArith
   Whoami -> Just CWhoami
   Self -> Just CWhoami
+  Swap -> Just CAtomic
```
xsebek, line 196: @byorgey this seems wrong, but I didn't want to add it to the already overloaded grabber. :confused: What do you think?

Whichever I choose, the device description needs to be updated too.

byorgey: Maybe make a new device, `grabber` + `rubber band` yields a... um... `rubber grabber`?  And a `rubber grabber` is an upgraded (#480) `grabber` that additionally provides `swap` capability?

xsebek: I made a fast grabber :+1: It looks like two grabbers (`≪`) and requires two grabbers and a rubber band to build. :grinning:

data/entities.yaml
```diff
@@ -656,10 +656,27 @@
   - "The 'install' command takes two arguments: the robot
     on which to install a device (which can be at most 1 cell away),
     and the name of the device to install."
-
   capabilities: [grab, give, place, install]
   properties: [portable]
 
+- name: fast grabber
```
byorgey, line 662: I like it!

src/Swarm/Game/Step.hs
```diff
@@ -774,8 +774,28 @@ execConst c vs s k = do
         flagRedraw
         return $ Out VUnit s k
       _ -> badConst
-    Grab -> doGrab False
-    Harvest -> doGrab True
+    Grab -> doGrab Grab'
+    Harvest -> doGrab Harvest'
+    Swap -> case vs of
+      [VText name] -> do
+        inv <- use robotInventory
```
byorgey, line 781: Some of this code seems like it's duplicated from elsewhere; are there any additional opportunities for factoring out common utility functions?

byorgey (approved):
Looks great!

data/scenarios/Testing/684-swap.yaml
```diff
@@ -0,0 +1,59 @@
...
+    dir: [1,0]
+    system: true
+    devices:
+      - logger
+    program: |
+      def repeat = \c. c; repeat c end;
+      repeat (
+        d <- scan down;
+        case d (\_.
+          say "Fatal error: swap does not work atomically!"
```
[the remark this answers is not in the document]
xsebek: You win, but the integration test fails with this message. :slightly_smiling_face:

src/Swarm/Game/Step.hs
xsebek: Yes, it is what `place` does, with some changes and some omissions.

I thought about refactoring this further, but it is not as similar to `place` as `grab` is to `harvest`, so I do not think it would make the code clearer.

xsebek: OK, I found a small thing to improve. :grin:

PR #688, opened 2022-09-16
preserve the focused scenario when rebuilding the NewGameMenu

### LINKED ISSUE #682: Tutorial levels repeat
**Describe the bug**
Tutorial challenges seem to play twice.

**To Reproduce**
Start the app and select Tutorial from the main menu.  Complete the backstory challenge (by typing `say "Ready!"`), select "Next Challenge!", then complete the move challenge (`move; move`) and select "Next Challenge!" again.

**Expected behavior**
It should move on to the next challenge (`turn`) but instead it repeats the `move` challenge.  If you go on and beat the `turn` challenge it plays twice too.

First reported by `zzz` in the `#swarm` IRC channel.

byorgey:
Fixes #682.

I still don't understand why it was doing the thing before where it repeated each challenge twice, instead of just repeating the same challenge over and over again.  So that has me a little worried.  But I do understand why there was a problem --- because when we win a scenario, the menu automatically advances to the next one, but the `currentScenarioPath` is still the one we just won, as it should be so we can save the updated `ScenarioInfo` associated to the correct scenario.  But then the menu is rebuilt in order to get all the updated `ScenarioInfo` --- so we must be sure to carefully preserve whatever scenario the menu is focused on, which may or may not be the same as the one we just exited, depending on whether we won or just quit.

### DIFF
--- a/src/Swarm/TUI/Controller.hs
+++ b/src/Swarm/TUI/Controller.hs
@@ -175,7 +175,7 @@ getTutorials sc = case M.lookup "Tutorials" (scMap sc) of
 
 -- | If we are in a New Game menu, advance the menu to the next item in order.
 advanceMenu :: Menu -> Menu
-advanceMenu = _NewGameMenu . lens NE.head (\(_ :| t) a -> a :| t) %~ BL.listMoveDown
+advanceMenu = _NewGameMenu . neHead %~ BL.listMoveDown
 
 handleMainMessagesEvent :: BrickEvent Name AppEvent -> EventM Name AppState ()
 handleMainMessagesEvent = \case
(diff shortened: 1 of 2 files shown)


src/Swarm/TUI/Controller.hs
```diff
@@ -399,9 +399,14 @@ saveScenarioInfoOnQuit = do
           Nothing -> return ()
           Just si -> liftIO $ saveScenarioInfo p si
 
-        -- rebuild the NewGameMenu so it gets the updated ScenarioInfo
+        -- See what scenario is currently focused in the menu.  Depending on how the
+        -- previous scenario ended (via quit vs. via win), it might be the same as
+        -- currentScenarioPath or it might be different.
+        curPath <- preuse $ uiState . uiMenu . _NewGameMenu . neHead . BL.listSelectedElementL . _SISingle . _2 . scenarioPath
```
xsebek, line 405: Could we use [`first1Of`](https://hackage.haskell.org/package/lens-5.2/docs/Control-Lens-Combinators.html#v:first1Of) or [`_head`](https://hackage.haskell.org/package/lens-5.2/docs/Control-Lens-Combinators.html#v:_head) or `ix 0` since you only need a traversal?

This seems like such a common thing that I would expect the `lens` package to provide a safe function to access the first element of a non-empty list. :thinking:

xsebek (approved):
I can confirm this fixes the issue for me! :+1:

src/Swarm/TUI/Controller.hs
byorgey: I definitely tried using `_head` before, but there is no `Cons` instance for `NonEmpty`.  I hadn't thought of doing `ix 0`, I will give that a try.

Issue #664, opened 2022-09-07
Optional tutorial scenarios to explain FP concepts


byorgey:
It's worth thinking about what kinds of optional tutorial scenarios we could make that would guide people through familiar FP concepts like lambdas, pair and sum types, etc. if they have never seen those things before.

_Originally posted by @byorgey in https://github.com/swarm-game/swarm/pull/663#discussion_r964883223_

byorgey:
- lambdas: it's tricky
- sum types: I'm imagining a scenario where there is a row of invisible items and you have to do something with certain ones, to force you to use `scan` and then deal with the output.  I'm not sure how to do it in such a way that you can't just use `ishere`.
- product types: maybe something requiring you to use `whereami`?

byorgey:
If we had #94 , I could imagine a scenario where there is a robot which is going to do something, but it needs some help, in the form of a term with a specific type.  e.g. you have to send it a term of type `Cmd Unit -> Cmd Unit` which does something, and once you send it such a term, it will use it to do something.  The robot will only be able to accomplish its task (thus completing the scenario) if the term you sent it does what it is supposed to.

byorgey:
I am imagining an overarching story where you are working some kind of desk job writing code for robots, maybe you are some kind of junior programmer-in-training, but the point is that you are not writing *all* the robot code, you are just tasked with writing *some* of it, with a well-specified interface in the form of a type + specification that you have to implement.  The first challenge could involve e.g. writing a function to increment a number, and the challenges could get more and more involved from there.

byorgey:
With #2504 we could also add a section in the help system that explains FP concepts in a bit more detail.  Perhaps not as good as an interactive tutorial but still better than nothing.

PR #697, opened 2022-09-20
rename unit type to `unit`

### LINKED ISSUE #696: Rename unit type
Haskell's `() : ()` is confusing.  I propose:
- The unit value should be named `()` because it can be explained as analogous to tuple values like `(1,2)`.
- The unit *type* should get a new name.  I suggest `unit`.

Additional required tasks if we make this change:
- [x] Update any code in entity descriptions
- [x] Regenerate cheatsheets on the wiki
- [x] Update types tutorial with a short explanation of this type

byorgey:
Closes #696.


data/scenarios/Tutorials/bind.yaml
```diff
@@ -6,9 +6,9 @@ objectives:
       - |
         Every command returns a value.  However, some simple commands, like
         `move`, do not have any meaningful
-        value to return.  Since the unit type `()` has only one value, also
+        value to return.  Since the unit type `unit` has only one value,
```
xsebek, line 9: Should we mention the motivation for the syntax here?
```suggestion
        - You can think of `()` as an empty tuple, so that a pair `(false,2)` has two meaningful components, normal value `(true)` has one component and the unit `()` has none.
```

xsebek (approved):

data/scenarios/Tutorials/bind.yaml
byorgey: Hmm, at this point we haven't talked about tuples in Swarm at all, although most people are probably familiar with the idea of tuples in general...

byorgey: I'm thinking at this point explaining the motivation behind the syntax would just be extra noise for a lot of people.   They are probably happy just being told that it exists.  I think later if we make a tutorial going more in depth into algebraic data types (#664) that would be the place to explain the notation for `()`.

However, your suggestion made me take a close look at this tutorial challenge and I think I have improved it in several other ways.

xsebek: Nice, I like it :+1:

PR #701, opened 2022-09-26
Unify move check and fix bugs

### LINKED ISSUE #699: Salvaged robots don't disappear
**Describe the bug**
Salvaged robots are properly salvaged and don't show up in the robots list any more, but they can still be seen in the world view.

**To Reproduce**
- Start a new classic game. 
- `build {}`
- `salvage`

**Expected behavior**
The robot should disappear, so that we can only see the base again. 

**Screenshots**
salvaged-robot

xsebek:
- unify move checking for move and teleport
- make `destroyIfNotBase` able to destroy system robots
- add test for destructive teleport
- closes #699


src/Swarm/Game/Step.hs
```diff
@@ -1873,7 +1875,9 @@ onTarget rid act = do
         Nothing -> return ()
         Just tgt -> do
           tgt' <- execState @Robot tgt act
-          robotMap . ix rid .= tgt'
+          if tgt' ^. selfDestruct
+            then deleteRobot rid
+            else robotMap . ix rid .= tgt'
```
xsebek, line 1880: @byorgey this took me a while to figure out, but now it's fixed and destroyed robots will be correctly destroyed immediately, instead of hanging around on the map for a while. :slightly_smiling_face:

data/scenarios/Testing/699-teleport-blocked.yaml
```diff
@@ -0,0 +1,45 @@
...
+          robotNamed "one"
+        } {
+          robotNamed "two"
+        };
+        return false;
+      } {
+        return true; // both robots must be destroyed
+      }
+solution: |
+  o <- robotNamed "one"; reprogram o {t <- robotNamed "two"; teleport t (3,0); teleport self (3,0)}
```
xsebek, line 19: @byorgey what do you think about this test? 🙂

Does it cover the original issue enough? I guess I could test `move` and surviving base too. 🤔

byorgey: I like it!  If we want to test additional stuff maybe we should put that in a separate test.

byorgey (approved):
Looks good to me!

data/scenarios/Testing/699-teleport-blocked.yaml
xsebek: Done :+1:

PR #704, opened 2022-09-26
Use type constraint for step functions

xsebek:
Uses custom `HasRobotStepState` to make the monad constraints for step functions easier to read.

### DIFF
--- a/src/Swarm/Game/Step.hs
+++ b/src/Swarm/Game/Step.hs
@@ -1,4 +1,5 @@
 {-# LANGUAGE BlockArguments #-}
+{-# LANGUAGE ConstraintKinds #-}
 {-# LANGUAGE DataKinds #-}
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE QuasiQuotes #-}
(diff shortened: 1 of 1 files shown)


src/Swarm/Game/Step.hs
```diff
@@ -1972,7 +1961,7 @@ safeExp a b
 ------------------------------------------------------------
 
 -- | Update the global list of discovered entities, and check for new recipes.
-updateDiscoveredEntities :: (Has (State GameState) sig m, Has (State Robot) sig m) => Entity -> m ()
+updateDiscoveredEntities :: (HasRobotStepState sig m) => Entity -> m ()
```
byorgey, line 1964: We do lose something here (and in similar places): we used to know statically that `updateDiscoveredEntities` does not throw any exceptions, but now in theory it could.  I think it's not really a big deal and it's worth the improved readability, but wanted to make a note.

byorgey (approved):

src/Swarm/Game/Step.hs
xsebek: Good point :+1: I did not include the `IO` for a similar reason, but the `Error` will always be safely handled and shown to the player so including it feels not so bad to me.

If we ever come up with something better, then it will be easy to replace that. :slightly_smiling_face:

PR #709, opened 2022-10-01
Make robots direction optional

xsebek:
When writing scenarios, specifying the direction for system robots is annoying.
With this change robots without direction will have a default direction `down`.


data/scenarios/README.md
```diff
@@ -260,7 +260,7 @@ table.
...
-| `dir`         |          | `int × int`           | The starting orientation of the robot, expressed as a vector.  Every time the robot executes a `move` command, this vector will be added to its position.  Typically, this is a unit vector in one of the four cardinal directions, although there is no particular reason that it has to be.                                                                                                                                  |
+| `dir`         |          | `int × int`           | An optional starting orientation of the robot, expressed as a vector.  Every time the robot executes a `move` command, this vector will be added to its position.  Typically, this is a unit vector in one of the four cardinal directions, although there is no particular reason that it has to be. When ommited, the robots direction will be the zero vector.                                                                                                 |
```
byorgey, line 263: We should also add `[0,0]` to the second column of the table, which is used to indicate default values for optional fields.
```suggestion
| `dir`         | `[0,0]`  | `int × int`           | An optional starting orientation of the robot, expressed as a vector.  Every time the robot executes a `move` command, this vector will be added to its position.  Typically, this is a unit vector in one of the four cardinal directions, although there is no particular reason that it has to be. When omitted, the robot's direction will be the zero vector.                                                                                                 |
```

data/schema/robot.json
```diff
@@ -41,7 +41,7 @@
                     "type": "number"
                 }
             ],
-            "description": "The starting orientation of the robot, expressed as a vector. Every time the robot executes a move command, this vector will be added to its position. Typically, this is a unit vector in one of the four cardinal directions, although there is no particular reason that it has to be."
+            "description": "An optional starting orientation of the robot, expressed as a vector.  Every time the robot executes a `move` command, this vector will be added to its position.  Typically, this is a unit vector in one of the four cardinal directions, although there is no particular reason that it has to be. When ommited, the robots direction will be the zero vector."
```
byorgey, line 44:
```suggestion
            "description": "An optional starting orientation of the robot, expressed as a vector.  Every time the robot executes a `move` command, this vector will be added to its position.  Typically, this is a unit vector in one of the four cardinal directions, although there is no particular reason that it has to be. When omitted, the robot's direction will be the zero vector."
```

byorgey (approved):

PR #708, opened 2022-09-27
Make multistep move tutorial

xsebek:
The goal is to make the move tutorial more interesting, by using multiple objectives in sequence:

- _move twice_ (the current **Moving, part 1**) teaches the `move` command
- _move six times_ (**new**) will teach about using <kbd>↑</kbd> to get previous commands in REPL and chaining with `;`
- _turn and move_ (part of the current **Moving, part 2**) teaches the `turn` command and `left`/`right`
- _repeat_ (part of the current **Moving, part 2**) will let the player enjoy reusing previous commands

The map of the next part will be revealed (:magic_wand: :mage:) upon fulfilling the objective.
```
+--+
|> |
+--+
```
move seven times:
```
+--+------+
| >       |
+--+------+
```
turn and move:
```
       +--+
       |  |
       +  |
       |  |
       |  |
+--+---+  +
|        >|
+--+------+
```
repeat:
```
+---------+
|         |
|  +------+
|  |
|  |
|  +---+--+
|        ^|
+------+  |
       |  |
       |  |
+--+---+  +
|         |
+--+------+
```

---
<sup>The number of moves and the shape itself is subject to change. :wink: </sup>

PR #711, opened 2022-10-01
Allow multiple robots in palette

### LINKED ISSUE #710: Allow multiple robots in one cell
Currently, we only allow specifying one robot in a cell palette. Any additional robots have to be added using location.

In some scenarios, it would be easier to specify multiple robots in the same cell using the palette and not bothering with the location.

For example:
```yaml
  palette:
    '.': [blank]
    '┌': [blank, upper left corner]
    '┐': [blank, upper right corner, upright]
    'L': [blank, "2", upleft, upright2]
```

Formally, the "cell" JSON list would have a length _greater than one_, with the first mandatory element being the terrain, the second optional parameter being the entity and the rest will be the names of robots.

xsebek:
- support specifying multiple robots in a cell
- add a test with multiple robots revealing new rooms
  - add T shaped walls
- closes #710


xsebek:
Here is how the test scenario looks:
[asciicast](https://asciinema.org/a/mzLIbjq233XoipoGj6hK6GIsw)

It is basically a simplified move tutorial (#708). Amazingly the revealing works fast enough that no waiting is necessary! :open_mouth:

data/scenarios/README.md
```diff
@@ -228,18 +228,19 @@ at various locations.
...
   terrain values are `stone`, `dirt`, `grass`, `ice`, or `blank`.
 - The second item (if present) specifies the name of an entity which
   should be present in the cell.  This may be a built-in entity, or a
   custom entity specified in the `entities` section.  `null` may be
   used to explicitly specify no entity in the cell.
-- The third item (if present) specifies the name of a robot which
-  should be present in the cell.  This must be the name of a robot
-  specified in the `robots` section.  A copy of the robot will be
+- The third item and later (if present) specifies the names of the robots
+  which should be present in the cell.  This must be the names of the robots
```
byorgey, line 242:
```suggestion
  which should be present in the cell.  These must be names of robots
```

This reads a bit better to me, previously it could be read to imply you needed to list *all* the robots.

src/Swarm/Game/Scenario.hs
```diff
@@ -164,7 +165,7 @@ data Cell = Cell
 instance FromJSONE (EntityMap, RobotMap) Cell where
   parseJSONE = withArrayE "tuple" $ \v -> do
     let tup = V.toList v
-    when (null tup || length tup > 3) $ fail "palette entry must have length 1, 2, or 3"
+    when (null tup) $ fail "palette entry must have length 1, 2, or more"
```
byorgey, line 168: Why not just say "nonzero length"?

byorgey (approved):
Looks good to me!

src/Swarm/Game/Scenario.hs
xsebek:
```suggestion
    when (null tup) $ fail "palette entry must nonzero length (terrain, optional entity and then robots if any)"
```

PR #714, opened 2022-10-01
Catch release query errors

xsebek:
- catch `HttpException` thrown when the query to GitHub API fails
- refactor the version error handling


xsebek:
This should fix the HTTP error getting written across the screen as seen here:
 - https://github.com/swarm-game/swarm/pull/678#issuecomment-1264414358

For the record, that particular error was caused by bad certificates on the ubuntu bionic image. It can be fixed ([SO answer](https://askubuntu.com/a/1145374)) by:
```bash
apt update
apt install ca-certificates
```
Good thing that happened, otherwise I would not think about fixing that query error handling.

byorgey (approved):
LGTM

PR #716, opened 2022-10-01
Add release action

### LINKED ISSUE #673: Automate release
- it would be nice if everything was built automatically and pushed to Hackage and GitHub release
- presumably, other projects have GitHub Actions that we could use for inspiration :slightly_smiling_face:
  - a blogpost with action snippet: https://vrom911.github.io/blog/github-actions-releases

xsebek:
- create the release on pushed PVP-style tag
- publish the binary, zipped data and license to the release
- publish to Hackage
- closes #673


xsebek:
@byorgey I wonder if a separate pipeline is a good way to do this.

Maybe I could remove the cache since we will not use it as often. :thinking: 

The alternative way would be to use the new release action and conditionally make a release in the normal pipeline.
But IMO there is already a lot going on there.

Either way, I will probably try it on some temporary repo so that I don't merge it blindly. :worried:

byorgey:
@xsebek I have no idea what any of that means.  I trust you. :smile:

.github/workflows/haskell-release.yml
```diff
@@ -0,0 +1,115 @@
...
+        run: cabal v2-sdist
+      
+      # - name: Generate documentation for Hackage
+      #   # The hackage-server attempts to build documentation for library packages, but this can fail.
+      #   # If it does we can do it ourselves
+      #   run: cabal v2-haddock --builddir=docs --haddock-for-hackage --enable-doc
+
+      - uses: haskell-actions/hackage-publish@v1
+        with:
+          hackageToken: "${{ secrets.HACKAGE_AUTH_TOKEN }}"
```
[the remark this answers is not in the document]
byorgey: @xsebek Done!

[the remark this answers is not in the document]
byorgey: For future reference, yes, you have to log into Hackage, go to User Accounts, then the Account Management page, and generate a new token.

.github/workflows/haskell-release.yml
```diff
@@ -0,0 +1,115 @@
...
+        # cabal: ["3.2"]
+        include:
+          - compiler: ghc-8.10.7
+            compilerKind: ghc
+            compilerVersion: 8.10.7
+            cabal: "3.6"
+            # It might be good idea to use older runner for building binary:
+            # on latest runners like fedora, it would link symbols in newer
+            # system libraries, so users could not run it on older OSes
+            os: ubuntu-20.04
```
byorgey, line 60: Can we somehow use this to build on other OSes besides just Ubuntu?

byorgey (approved):

.github/workflows/haskell-release.yml
xsebek: Like with another item in the `include` section? Hopefully yes, I would like to add macOS with GHC 9.2 in the future.

byorgey: Right, I meant like macOS and Windows, I realize the Ubuntu build will probably work on other GNU/Linux variants.

byorgey: Well, I guess Windows doesn't matter since you would need to run Swarm under something like WSL anyway.

xsebek: Yes, but macOS might be worth a little effort.

At the moment [GitHub does not have ARM macOS runners](https://github.com/actions/runner-images/issues/2187), so maybe we can wait a little while longer.

xsebek:
For future reference, I tested this on [my fork of `ghci-pretty`](https://github.com/xsebek/ghci-pretty). The workflow was basically:
1. edit & commit
3. tag with a different last digit (`0.1.0.X`)
4. push and push tags
5. see what does not work and repeat

byorgey (approved):
I don't see anything obviously wrong,  for whatever that's worth!

PR #718, opened 2022-10-02
set up config for issue template chooser, with link to COMMUNITY

byorgey:


byorgey:
The idea is to add a link to the COMMUNITY page from the issue template chooser, in case someone actually wants to just chat on IRC instead of opening an issue.  

Unfortunately there doesn't seem to be any way to test this or see what it will look like without just merging it to `main`.

xsebek (approved):
Looks just like the tutorial :+1: 
- [Configuring the template chooser](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository#configuring-the-template-chooser)

PR #719, opened 2022-10-04
Add `author` (optional) and `version` (required) fields to scenarios

### LINKED ISSUE #611: Add `version` and `author` fields to scenario format
A couple things that occurred to me while writing up documentation for scenario description files (#354):

- We should add a (free-form) `author` field, since hopefully in the near future we will have scenarios written by many different people.
- It might also be a good idea to add a `version` field containing an `int`, so that in the future if we change the format for scenario descriptions, we can easily make things backwards compatible, not by going to contortions to make the new format backwards compatible with the old, but simply by leaving in the code to parse the old format, and choosing which code to run based on the version number.

byorgey:
Closes #611 .


src/Swarm/TUI/View.hs
```diff
@@ -217,17 +217,24 @@ drawNewGameMenuUI (l :| ls) =
...
-      , noSame $ -- hide best game time if it is same as best real time
-          padTop (Pad 1) $
-            padRight (Pad 1) (txt "best game time:") <+> describeStatus (si ^. scenarioBestTicks)
-      , Just $
-          padTop (Pad 1) $
-            padRight (Pad 1) lastText <+> describeStatus (si ^. scenarioStatus)
+    vBox
+      [ txtWrap (nonBlank (s ^. scenarioDescription))
+      , vBox . catMaybes $
+          [ padTop (Pad 1)
```
xsebek, line 223: Why use the inner `vBox`?

If we wanted to get something out of it, we could at least refactor out the padding:
```Haskell
    let padTop1Right1 = padTop (Pad 1) . padRight (Pad 1)
    vBox
      [ txtWrap (nonBlank (s ^. scenarioDescription))
      , vBox . catMaybes . fmap padTop1Right1 $
```

src/Swarm/TUI/View.hs
```diff
@@ -217,17 +217,24 @@ drawNewGameMenuUI (l :| ls) =
...
+              . padRight (Pad 1)
+              . (txt "Author: " <+>)
+              . txt
+              <$> (s ^. scenarioAuthor)
```
xsebek, line 227: The author field could be coloured differently, as it's only one line from the description and kind of melds with it.

I would suggest the [`dim`](https://hackage.haskell.org/package/vty-5.37/docs/Graphics-Vty-Attributes.html#v:dim) attribute. :slightly_smiling_face:

xsebek (approved):
Works for me, thanks for taking the time to fill in all the names @byorgey :+1:

data/schema/scenario.json
```diff
@@ -94,4 +102,4 @@
         "world",
         "robots"
```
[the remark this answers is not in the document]
byorgey: Ah, I didn't see the `required` section at the bottom, thanks!

data/scenarios/Tutorials/conditionals.yaml
```diff
@@ -1,3 +1,4 @@
+version: 1
 name: Conditionals
```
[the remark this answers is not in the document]
byorgey: Yeah, I was thinking that for something as "official" as the tutorial, it's better not to put individual author names on the scenarios, to make it clear that no one "owns" different parts of it, so everyone feels free to contribute improvements, and to emphasize that writing and maintaining the tutorial is a collective effort.  (I also can't actually remember who wrote which parts of the tutorial, which just goes to show that it has indeed been a collective effort so far.)

src/Swarm/TUI/View.hs
byorgey: Hmm, I don't know, I'll put it back into a big `vBox`.

Factoring out `padTop (Pad 1) . padRight (Pad 1)` doesn't actually work, since the `padRight` calls are not being applied to the entire widget.

PR #720, opened 2022-10-05
Add split and text length functions

xsebek:
- add `split: int -> text -> (text * text)`
- add `chars: text -> int`
- test using property tests


src/Swarm/Language/Syntax.hs
```diff
@@ -633,6 +637,13 @@ constInfo c = case c of
   Geq -> binaryOp ">=" 4 N "Check that the left value is greater or equal to the right one."
   Format -> function 1 "Turn an arbitrary value into a string."
   Concat -> binaryOp "++" 6 R "Concatenate the given strings."
+  Chars -> function 1 "Counts the number of caracters in the text."
```
byorgey, line 640:
```suggestion
  Chars -> function 1 "Counts the number of characters in the text."
```

src/Swarm/Language/Syntax.hs
```diff
@@ -633,6 +637,13 @@ constInfo c = case c of
...
+  Split ->
+    function 2 . doc "Split the text into two at given position." $
+      [ "To be more specific, the following holds:"
+      , "`\\s1.\\s2. (s1,s2) == split (chars s1) (s1 ++ s2)`"
```
byorgey, line 644:
```suggestion
      [ "To be more specific, the following holds for all `text` values `s1` and `s2`:"
      , "`(s1,s2) == split (chars s1) (s1 ++ s2)`"
```

src/Swarm/Util.hs
```diff
@@ -373,11 +373,14 @@ number _ = plural
...
-squote t = T.concat ["'", t, "'"]
+squote t = T.concat ["'", escape t, "'"]
 
 -- | Surround some text in double quotes.
 quote :: Text -> Text
-quote t = T.concat ["\"", t, "\""]
+quote t = T.concat ["\"", escape t, "\""]
+
+escape :: Text -> Text
+escape = T.replace "\"" "\\\"" . T.replace "\\" "\\\\"
```
byorgey, line 383: I'm not quite sure I understand the motivation for adding `escape` here.  `quote` and `squote` are only used for formatting messages, not for pretty-printing, right?  So why does it matter that we escape quote marks and backslashes?  In fact, I think I'd rather *not* have escape sequences showing up in messages shown to the player.

test/unit/TestEval.hs
```diff
@@ -210,6 +212,19 @@ testEval g =
         , testCase
             "try / div by 0"
             ("try {return (1/0)} {return 3}" `evaluatesTo` VInt 3)
+        , testProperty
+            "number of characters"
+            ( \s ->
+                ("chars " <> quote (T.pack s)) `evaluatesToP` VInt (fromIntegral $ length s)
```
byorgey, line 218: Oh, is this why you added escaping to `quote`, because you used it to write a test case?  I think we should use a different function here (you should be able to just say `T.pack (show s)`?), remove the escaping from `quote` and `squote`, and better document what `quote` and `squote` are supposed to be for.

byorgey (approved):
Looks good, other than the stuff about `quote` and `escape` --- see my other comments.

test/unit/TestEval.hs
xsebek: Yes, that was what I did at first, but got confused by a different mistake. :sweat_smile: 

Now it should be fixed.

test/unit/TestEval.hs
```diff
@@ -245,7 +263,7 @@ testEval g =
   evaluatesToP :: Text -> Value -> Property
   evaluatesToP tm val = ioProperty $ do
     result <- evaluate tm
-    return $ Right val == (fst <$> result)
+    return $ Right val === (fst <$> result)
```
byorgey, line 266: Ooh, fancy!

byorgey (approved):

PR #724, opened 2022-10-06
Refactor move tutorial using split and whoami

xsebek:


xsebek:
I like how this makes the system robots more uniform and I can no longer mess up inventory/place.

There are a couple of improvements that could still be made:
1. default `robot.display.invisible` to `robot.system` (so system robots are invisible by default)
2. let system robots use `place` and `swap` without having to `create` (easier to write and one tick faster)
3. remove flowers once the objective is reached (just for fun)

The naming system for grab/place/swap looks IMO quite readable, so maybe it will be useful in other multistep challenges. :slightly_smiling_face:

byorgey:
> 2\. let system robots use `place` and `swap` without having to `create` (easier to write and one tick faster)

`create` shouldn't take a tick, does it?

PR #725, opened 2022-10-06
Add release badges

xsebek:


xsebek:
Lets merge this after we make a release though :sweat_smile:

PR #730, opened 2022-10-06
Run data files

xsebek:
- add fallback to the swarm data directory when the run file is not found


byorgey (approved):
Looks good to me.  I tried it out (at least by doing something like `run "scenarios/Tutorials/crash-secret"`) and it seems to work.

Issue #304, opened 2022-02-15
Ability to access past REPL results


byorgey:
From #303 ; depends on implementing #98 first.

The idea might be to have a special primitive array available which is notionally appended to every time a new REPL value is returned. We could change the REPL output to show the index, e.g.
```
> 3 + 5
res[0] = 8
> build {move}
res[1] = <r0>
> def r = res[1] end
```
or something like that.

byorgey:
A couple related questions come to mind: (1) how early do we want players to be able to access past REPL results? (2) What device(s) will be necessary to use arrays?

byorgey:
I just realized that even without an array to access all past REPL results, it would already be extremely useful to have a special variable `it`, like GHCi has, which is always bound to the most recent result.  We would need to make sure that something like
```
build {move; move}
def mover = it end
1 + 2
view mover
```
works properly --- the (potential) problem is that the definition of `mover` isn't evaluated immediately, so by the time it is actually used the value of `it` will be different.  However, I *think* this should actually work properly --- evaluating `1 + 2` will shadow the definition of `it`, but the definition of `mover` should still refer to the `it` that was in scope at the time it was defined.

valyagolev:
I want to note that if there's only plans to implement arrays with fixed type, it will be impossible to use them to implement this feature. In that case there's no need to wait for the arrays to land, and we can simply introduce indexed variables like `it0`, `it1`, etc

byorgey:
@valyagolev oh, of course, that's a very good point!

PR #733, opened 2022-10-07
Persist the type of the last expression in the REPL UI

valyagolev:
- adds `PolyUnit` pattern (just seemed useful?)
- persists the last type/value into the `REPLDone (Maybe (Polytype, Value))` constructor
    - `value` is not needed right now, but I'm thinking about using it for #304
    - I'm also considering if pretty-printing the last value in that cozy corner could also be a good idea
- notion of the "active" type (last or currently-executing) and the corresponding `Getter`: `replActiveType`
- show this active type in the REPL UI (fixes #692)

Issue #707, opened 2022-09-27
Split large Haskell modules


xsebek:
Some of the Swarm Haskell modules have grown exceedingly large, which makes them hard to maintain.

Notably:
```bash
find src -name '*.hs' -exec wc -l {} \; | sort -h | tail -n 6
#  857 src/Swarm/Game/State.hs
#  897 src/Swarm/Language/Syntax.hs
#  942 src/Swarm/TUI/Controller.hs
#  999 src/Swarm/TUI/Model.hs
# 1130 src/Swarm/TUI/View.hs
# 2008 src/Swarm/Game/Step.hs
```

It also makes compilation slightly longer, since any change causes the recompilation of the entire file and its dependencies.

## Proposed solution

The `Step` module is the largest one and the trickiest to split. But I think it could be reasonably split into three parts:

- `Exec`: `execConst` function which currently accounts for ~1000 loc
- `Step`: running CESK machines (`stepCESK`, `evalPT`,...)
- various utility functions - some of them are general enough to go in some `Util` module, others could be moved to `Robot` or `GameState`

I think we could make room for the utility functions on robots and game state if we moved the records themselves into an internal module.

For example, if we moved the `GameState` record and its lenses:
```Haskell
module Swarm.Game.State.Internal (GameState, creativeMode, ...)
```
```Haskell
module Swarm.Game.State (module Swarm.Game.State.Internal, initGameState, scenarioToGameState, ...)
import Swarm.Game.State.Internal
```
This could of course be tweaked for documentation so that the `Step` module renders nicely.

byorgey:
Update:
```
diophantus :: ~/projects/swarm » find src -name '*.hs' -exec wc -l {} \; | sort -h | tail -n 6
852 src/Swarm/Game/Step.hs
1126 src/Swarm/Language/Typecheck.hs
1208 src/Swarm/Language/Syntax.hs
1467 src/Swarm/TUI/View.hs
1503 src/Swarm/TUI/Controller.hs
1693 src/Swarm/Game/Step/Const.hs
```
We managed to split up `Step`, though `Step.Const` is still the longest module.  Maybe the next place to focus some energy would be the `TUI` modules as well as syntax + typechecking.

byorgey:
Update, as of August 2025 (**edited to add**: relatively unchanged as of August 2026):

```
diophantus :: ~/projects/swarm » find src -name '*.hs' -exec wc -l {} \; | sort -h | tail -n 6
[...]

PR #734, opened 2022-10-07
expose the last evaluated result as `it` in REPL

valyagolev:
This depends on #733.

This partially addresses #304.

There's one essential commit with the functionality: 8666e55d104821a5b026eb4b5da031c563e6462d . And then there's a small refactoring that I'm proposing, but since I'm unfamiliar with the codebase, I have no idea if it's useful or not.


xsebek (commented):
The code and even the refactoring seem reasonable to me.

I just have no idea how the `it` variable works in the code. I guess you grab the result from the finished REPL computation and set it in the base robot context?

byorgey:
@valyagolev , since this has merge conflicts, and now that you have push access, it's probably easiest to just close this PR and create a new one from a branch in the `swarm` repo itself rather than from a branch in your fork.  (I'll assume you know what that means but feel free to ask for clarification/help if not.)

xsebek:
@byorgey in this case the conflict is small enough and should be easy to resolve.

IIRC the only reason for creating branches in the swarm repo was the Restyled Action worked better that way or was there another reason?

valyagolev:
It just needs a rebase :) The branch is the continuation of the one
squash-merged. I will do the rebase when I wake up.

Thank you for the welcome!

byorgey:
Yes, I'm sure fixing the merge conflicts will be easy, I just meant if some "administrative" work (such as rebasing) has to happen anyway, it could be an opportunity to reopen a PR from a swarm branch.  Not required, just a suggestion.  

> IIRC the only reason for creating branches in the swarm repo was the Restyled Action worked better that way or was there another reason?

That is one reason, but the other reason is that it makes collaboration easier.  For example #669 was a very collaborative effort, and that is harder when PRs are opened from people's private forks where no one else can push.  (Not that we are going to immediately start pushing commits to others' PRs!  But as trust develops it becomes a possibility.)

valyagolev:
I kept this branch & PR for this one to avoid duplicating the conversation. But next time I'll gladly push a new branch to the main repo directly. Thank you! 

The conflict and the comments were resolved.

src/Swarm/Game/Robot.hs
```diff
@@ -431,6 +433,8 @@ tickSteps :: Lens' Robot Int
 
 -- | Is the robot currently running an atomic block?
 runningAtomic :: Lens' Robot Bool
+emptyRobotContext :: RobotContext
```
byorgey, line 436: Can we put a blank line in between `runningAtomic` and `emptyRobotContext`?  A Haddock comment might be nice too, though I guess `emptyRobotContext` is pretty self-explanatory.

src/Swarm/Language/Pipeline.hs
```diff
@@ -36,6 +37,11 @@ import Swarm.Language.Typecheck
 import Swarm.Language.Types
 import Witch
 
+-- | A value, or something like a value, along with its
+--   type and requirements
+data Processed val = Processed val Polytype Requirements
```
byorgey, line 42: What types will end up being substituted for `val`?  I think I saw `Value` and `Maybe Value`, will there be others?

I wonder if we can come up with a better name than `Processed`.  To me, the name doesn't really help me guess what the definition is.  I guess you named it in parallel with `ProcessedTerm`, but a `ProcessedTerm` actually results from processing a `Term`, whereas that's not the case for a `Processed Value`. (Though perhaps we should rename `ProcessedTerm` as well!)  Also, seeing the type `Processed` made me expect that `ProcessedTerm` is now a type synonym for `Processed Term` or something like that, but it's not.

`Decorated`, perhaps?  `Typeful`?  `Typed`?  I'm open to suggestions.

src/Swarm/Game/Robot.hs
```diff
@@ -509,6 +513,28 @@ instance FromJSONE EntityMap TRobot where
     mkMachine Nothing = Out VUnit emptyStore []
     mkMachine (Just pt) = initMachine pt mempty emptyStore
 
+type instance Index RobotContext = Ctx.Var
```
byorgey, line 516: I like the idea of adding `Ixed` and `At` instances for `RobotContext`, but I would tend to put them right next to the definition of `RobotContext`.  Was there a particular reason for having them down here?

src/Swarm/Game/Robot.hs
```diff
@@ -69,6 +69,7 @@ module Swarm.Game.Robot (
   runningAtomic,
 
   -- ** Creation & instantiation
+  emptyRobotContext,
```
byorgey, line 72: I think this should be listed in the "Robot context" section earlier in the export list.

src/Swarm/Game/Robot.hs
byorgey: Also, is there a particular reason for putting `emptyRobotContext` here?  I would tend to put it near the definition of `RobotContext`.

src/Swarm/TUI/Controller.hs
```diff
@@ -694,21 +706,17 @@ handleREPLEvent = \case
...
-          fromMaybe emptyStore $
-            s ^? gameState . robotMap . at 0 . _Just . robotContext . defStore
-        startBaseProgram t@(ProcessedTerm _ (Module ty _) _ _) =
-          (gameState . replStatus .~ REPLWorking ty Nothing)
-            . (gameState . robotMap . ix 0 . machine .~ initMachine t topValCtx topStore)
+        topCtx = topContext s
+
+        startBaseProgram t@(ProcessedTerm _ (Module ty _) reqs _) =
+          (gameState . replStatus .~ REPLWorking (Processed Nothing ty reqs))
+            . (gameState . robotMap . ix 0 . machine .~ initMachine t (topCtx ^. defVals) (topCtx ^. defStore))
```
byorgey, line 713: Love it, I knew this code could be simplified somehow.

byorgey (commented):
This is looking nice!  I just have a few comments about cosmetic issues.

We should probably also talk about the `it` variable in the tutorial, but we can make a separate issue for that.

src/Swarm/Language/Pipeline.hs
valyagolev: > What types will end up being substituted for val? I think I saw Value and Maybe Value, will there be others?

I'd also not be too surprised if I saw there: `()` or `Term`; `[Var]` (e.g. as a list of suggestions for a typed hole)... It has a functorial feel to it. Obviously I'm not familiar enough with the codebase to have any idea if it's good or not. I also thought maybe there could be a way to have `ProcessedTerm` be part of this already, but I'm not sure.

> I wonder if we can come up with a better name than Processed. 

Completely agree – it's more of a placeholder name, I was hoping for a feedback.

Actually I'm wondering why is it that there's such a hard separation between `Type` and `Requirements`. I guess they "mean" different things, but are they actually ever separated from each other anyway? Doesn't their calculation follow similar paths? Wouldn't they be essentially a decoration over a type in other PLs?

So with that in mind, I do like `Typed`. 

Maybe there could be a name for a generalised type like this, e.g... `AssemblableType`? 

then we could have:

    data AssemblableType = AssemblableType Requirements Polytype
    data Assemblable v = Assemblable AssemblableType v

or even:

    data Assemblable v = AssemblableType Requirements Polytype v
    type AssemblableType = Assemblable ()

And yeah, whatever we decide about the meaning of this should determine which file it goes to, as well.

src/Swarm/Game/Robot.hs
valyagolev: I'm just a bit lost in big files. Also I think the best placement for everything will get clearer if we clarify the specific types proposed here... 

I'm also not very sure if I was right not to include the contents of the `Store`...

There's another way to approach this particular change, namely by changing `RobotContext` into a product of `Ctx (Typed Value)` (or whatever we agree on) and the `Store`. I do think the three `Ctx`s should go together somehow. Three different maps potentially mean three different sets of `Var`s: do we want them?

(Such a change will remove the need for those instances)

src/Swarm/Language/Pipeline.hs
xsebek: I like `Typed`, that sounds most natural to me. I might consider using a record though:
```Haskell
data Typed v = Typed
  { value :: v,
  , typed :: Polytype
  , requires :: Requirements
  }
```
We use lenses a lot so we might as well use them for this too.


Btw. if you want to get feedback on something in PR, it is a good idea to insert some `TODO` comments so that we don't forget about it. Alternatively, you can make comments on the code here on GitHub or in your editor if you have some plugin for that. :wink:

src/Swarm/Game/Robot.hs
xsebek: If you have any suggestions on how to split the Robot source file to be easier to orient yourself, feel free to chime in:
- #707 :wink: 

I did not mention this file in the list because it is only over 500 loc, but it is still in our top 10. :grinning:

src/Swarm/Language/Pipeline.hs
byorgey: > Actually I'm wondering why is it that there's such a hard separation between Type and Requirements. I guess they "mean" different things, but are they actually ever separated from each other anyway? Doesn't their calculation follow similar paths? Wouldn't they be essentially a decoration over a type in other PLs?

Yes, very good observation.  Right now requirements are calculated separately, after typechecking, but actually there are several bugs related to the fact that it doesn't work quite right (#540, #394) and I want to redesign things so that requirements are part of types, as they should be (#231).

But let's not get too crazy with refactoring in this PR (what you have so far is fine).  I'd rather merge something that works to give us `it`, and then contemplate further refactoring as a separate PR.

src/Swarm/Game/Robot.hs
byorgey: > There's another way to approach this particular change, namely by changing RobotContext into a product of Ctx (Typed Value) (or whatever we agree on) and the Store.

Yes, we contemplated this previously, in #190 .  We abandoned it because it got too complex.  But maybe if we try again we can find a simpler way to do it.

byorgey:
@valyagolev This is looking good so far.  Let us know if you feel like you need more guidance to know how to proceed from here.

valyagolev:
what's the verdict on `Processed`? I guess I could just rename it to `Typed` and extract to its own module for now

i will apply other suggested changes soon and merge. thanks!

xsebek:
I think we agreed on `Typed`, so please do so 🙂

valyagolev:
added 9ca4e4666f5adf68125e13063d0e8050020d23b7 to fix #304 
Screenshot 2022-10-11 at 11 05 26

src/Swarm/Language/Typed.hs
```diff
@@ -0,0 +1,20 @@
...
+module Swarm.Language.Typed (Typed (..), value, polytype, requires) where
+
+import Control.Lens (makeLenses)
+import Data.Aeson (ToJSON)
+import Data.Aeson.Types (FromJSON)
+import GHC.Generics (Generic)
+import Swarm.Language.Requirement (Requirements)
+import Swarm.Language.Types (Polytype)
+
+-- | A value, or a hole, or something else that has its type & requirementss fixed
```
byorgey, line 12:
```suggestion
-- | A value, or a hole, or something else that has its type & requirements fixed
```

src/Swarm/Game/State.hs
```diff
@@ -361,6 +365,10 @@ robotsInArea o d gs = map (rm IM.!) rids
   rl = gs ^. robotsByLocation
   rids = concatMap IS.elems $ getElemsInArea o d rl
 
+-- | The base robot, if it exists.
+baseRobot :: Traversal' GameState Robot
+baseRobot = robotMap . ix 0
```
byorgey, line 370: Eventually I want to stop assuming that the base is always ID 0, and this refactoring is a great first step!

byorgey (commented):
This doesn't seem to work quite right when executing commands.  For example, if I type `build {move}` at the prompt, it says `it1  :=  <r1> : robot`, which makes sense.  But then when I type `it1` at the prompt, it says it has type `cmd robot` instead of `robot`.  And if I try to actually use `it1`, it causes a fatal crash.  We do have some special case behavior built in for things of `cmd` type entered at the REPL, and I suspect we're running into that somehow, though I haven't yet figured out what's going on exactly.

We should also reset the `replNextValueIndex` to 0 every time we start a new game (see the `startGameWithSeed` function in `Swarm.TUI.Model`).

valyagolev:
> This doesn't seem to work quite right when executing commands. For example, if I type build {move} at the prompt, it says it1 := <r1> : robot, which makes sense. But then when I type it1 at the prompt, it says it has type cmd robot instead of robot. And if I try to actually use it1, it causes a fatal crash. We do have some special case behavior built in for things of cmd type entered at the REPL, and I suspect we're running into that somehow, though I haven't yet figured out what's going on exactly.

Oh, this I didn't test. Yeah I will test a few weirder scenarios, maybe we'll need a test for those...

CONTRIBUTING.md
```diff
@@ -216,7 +216,11 @@ suggestions.
   to configure your favorite editor to have `fourmolu` automatically
   applied to your code; but if you don't know how to set that up, don't
   worry!  The [`restyled.io` bot](https://restyled.io/) automatically
-  reformats pull requests as necessary.
+  reformats pull requests as necessary. Here's the command to apply
+  the formatting that *exactly* matches what the bot would do:
+
+      cabal install fourmolu-0.4.0.0
+      find src/ app/ test/ -name "*.hs" | xargs fourmolu --mode=inplace --cabal-default-extensions
```
xsebek, line 223: I was not aware you also added it here, @valyagolev. :laughing: Meanwhile, I opened #744.

Technically this is not true - because of our restyled configuration file from times of Fourmolu 0.3 it runs with:
```bash
find src/ app/ test/ -name "*.hs" | xargs fourmolu --mode=inplace '--ghc-opt=-XBangPatterns' \
         '--ghc-opt=-XFlexibleContexts' \
         '--ghc-opt=-XFlexibleInstances'\
         '--ghc-opt=-XGADTSyntax'\
         '--ghc-opt=-XMultiParamTypeClasses'\
         '--ghc-opt=-XNumericUnderscores'\
         '--ghc-opt=-XRankNTypes'\
         '--ghc-opt=-XTupleSections'\
         '--ghc-opt=-XTypeApplications'\
         '--ghc-opt=-XTypeOperators'\
         '--ghc-opt=-XImportQualifiedPost'\
         '--ghc-opt=-XLambdaCase'
```
Which should be the same, but I should probably update it anyway.

While it is nice that you noted it down, it's kind of silly to put everything in this one PR, please feel welcome to open other small PRs. :wink:

valyagolev: oh that explains why i had to manually undo some of its changes then :)

I just copied what I was suggested. I will remove it, thanks for your PR

valyagolev:
> [9ca4e46](https://github.com/swarm-game/swarm/commit/9ca4e4666f5adf68125e13063d0e8050020d23b7) to fix #304 Screenshot 2022-10-11 at 11 05 26

are we okay with this syntax?

xsebek:
I think `it0: int = 1` would look more like the def/let syntax we use currently, but I don't have a strong opinion.

CONTRIBUTING.md
xsebek: Yes, you will have to revert the change to this file, because I merged first, sorry about that. :sweat_smile: 

You can do that automatically but in this case, it's quite easy either way.

valyagolev:
>  We do have some special case behavior built in for things of cmd type entered at the REPL, and I suspect we're running into that somehow, though I haven't yet figured out what's going on exactly.

Ok, yes, now I see the problem and it requires making a decision.

Basically if I run a line that returns `cmd a`, REPL will run the line to get `cmd a`, then run the command to return `a` (e.g. a particular `robot`). So the question is: do we want to store the `cmd a` or `a` in `it*`?

If we store `cmd a`, we will recreate (e.g.) a robot every time we run `it`. Because one of the main use-cases is to avoid losing references to the robots, probably it's better to store `a`. `cmd a` is available through the repl history anyway, I suppose.

ghci stores `a`:
```
λ> print "hello" >> return 3
"hello"
3

λ> it
3
```

valyagolev:
Screenshot 2022-10-15 at 12 18 30

This looks better

byorgey:
> So the question is: do we want to store the `cmd a` or `a` in `it*`?

Yes, absolutely we want to store the returned `a` value, just like `ghci` does.

By the way, relatedly, the syntax
```
it4 : robot <- <r2>
```
looks strange, because `<r2>` itself does not have a `cmd` type, so it's not something you could bind with `<-`.  It should just be displayed with an `=` sign like everything else.  The fact that we executed a command to get that value, instead of just evaluating an expression, should be irrelevant to how we display the resulting value.

valyagolev:
I still think it's good to mark it differently somehow, even if the particular syntax is questionable

byorgey:
> I still think it's good to mark it differently somehow, even if the particular syntax is questionable

Why? What difference does it make?

xsebek:
@valyagolev I understand that it could be useful to make it different in order to more easily debug this new feature.

A technically correct syntax would be:
```
it0 = 0
it1 <- return 1
```

Honestly the debug features I previously introduced were never useful after merging. But it may just be because that they required recompiling to enable - it would be better if we could e.g. turn them on in the Main Menu.

I am fine with either way, but consider that making the result look different could be more confusing (GHCi does not do it either) and will require more code. 🤷

valyagolev:
No, it's not really about debugging this feature, rather about avoiding
loss of information in REPL. If I run `some_func`, I probably also want to
see what exactly it returned, value-wise and type-wise ([there's][1] a very
useful `:set +t` in ghci). Otherwise there's no visual distinction between
two fundamentally different results. I'm sure the output syntax can be
tweaked.

That said, if the general opinion is still to  hide the distinction between
`cmd a` and `a` in repl, I'll remove the extra code that shows it.

[1]:
https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/ghci.html#ghci-options

xsebek:
@valyagolev but there is no distinction between the values _that you get_:
```
> return 1
it0: int = 1
> 1
it1: int = 1
> it0 == it1
it3: bool = true
```

Sure the code at _the REPL line_ has different type, but you can see that in the corner and from the perspective of later lines the results are the same. 🤔 

If you have some other difference in mind, please give an example where it would have an impact. I might be missing something. 😅

byorgey:
To offer a bit more perspective: there are essentially two different things you can do at the REPL.  You can use it as a "comand center" where you enter instructions (i.e. things with a `cmd` type) to be executed.  Or, you can use it as a "calculator" where you enter expressions to be evaluated.  For convenience, both work by entering an expression at the prompt: we just process them differently depending on the type. Mostly, I think players will not even think about the distinction, though it could be useful to have some way to help them understand the difference.  However, both modes result in a value, and once you get a value as output there is no longer any difference in how those values may be used.  So I think displaying the result values with different syntax is not the right way to go about helping players understand the distinction between executing commands and evaluating expressions.  I believe it is more likely to make them think there is some difference in how they can or should use the values later, which is not the case, so it would be more confusing than helpful.

byorgey:
Also, I'm very open to thoughts/suggestions on whether the distinction between execution and evaluation could be confusing to players, and how we might make it less so.

valyagolev:
I pushed a commit to remove the discussed syntax thing

byorgey (approved):
This looks great, thanks @valyagolev for your hard work on this!

Now we just need to implement #60 so we can actually see the previous `it{n}` bindings!

byorgey:
@valyagolev in the future remember you can merge by adding the `merge me` label, then the Mergify bot will do everything (including making sure that the branch still passes all the tests after merging the current state of the `main` branch).

valyagolev:
oh right, sorry, forgot about that

PR #735, opened 2022-10-08
Add void type

### LINKED ISSUE #665: Void type
We currently have sum types, product types, and unit type, but no void type.  We should include one to complete the algebraic type story.  We already have `fail` and `undefined` which can be used to implement `absurd : void -> a`, so literally all we have to do is add a constructor for the `void` type, and add it to the parser and pretty-printer.

Originally suggested by @xsebek in IRC:
```
(00:01) <   xsebek_> I wonder if Swarm should have a Void type
(00:04) <   xsebek_> We always have fail function to inhabit it, so it would not need any extra support
```

ussgarci:
Closes #665


byorgey:
Looking good so far.  You'll also need to add `void` to the parser for types.  We'll also want to add a test or two to the test suite.

We might want to also mention `void` somewhere so players can discover it, for example, in the description of an `ADT calculator`.

ussgarci:
Thank you for the feedback and direction.  I will review the parser code this evening.

byorgey:
@ussgarci I missed you in IRC, so just wanted to say: reading about megaparsec and parser combinators is great, don't let me stop you!  But for this particular issue all you should have to do is figure out where the `unit` type is parsed, and then just copy-paste it for `void`.  They should be handled in exactly the same way.

ussgarci:
@byorgey Thank you for the tip.

ussgarci:
Hi @byorgey , I've pushed an initial attempt at writing a test for the `void` type.  Any feedback is appreciated.

test/unit/TestVoid.hs
```diff
@@ -0,0 +1,19 @@
...
+
+-- | Swarm unit tests
+module TestVoid where
+
+import Test.Tasty
+import Test.Tasty.HUnit
+import Swarm.Language.Types
+import Swarm.Language.Typecheck (isSimpleUType)
+
+testVoid :: TestTree
```
byorgey, line 11: This test is fine --- good job noticing that `isSimpleUType` had better return `True` for `void`, I wouldn't have thought of that.  Fortunately, the way `isSimpleUType` is implemented, we get this for free.

test/unit/Main.hs
```diff
@@ -26,6 +26,7 @@ import TestLanguagePipeline (testLanguagePipeline)
 import TestModel (testModel)
 import TestNotification (testNotification)
 import TestPretty (testPrettyConst)
+import TestVoid (testVoid)
```
byorgey, line 29: It's probably overkill to have a separate test module just for testing `void`.  I would put the `void` tests in `TestLanguagePipeline`, perhaps?

byorgey (commented):
Looking good so far.  

- We could add some tests just to make sure that `void` is a valid type, for example, a test like `valid "def f : void -> a = \x. undefined end"` or something like that.
- We could also put a test in `TestPretty` to make sure that the `void` type pretty-prints properly.

byorgey (approved):
This looks great.  We could update the `ADT calculator` description (or brainstorm other ways for the player to learn about `void`) in this issue, or it might be simpler to just merge this and split out in-game documentation of `void` into a separate issue.

ussgarci:
Hi @byorgey merging this and splitting out the documentation of `void` sounds good to me.

ussgarci:
@byorgey I can create the issue.  Unless I'm mistaken, it appears I can't add a label to this PR.

byorgey:
Oh, right, I was confused and thought you were already added as a contributor.

byorgey:
Oops, forgot to edit the PR description before merging, oh well :smile:

PR #741, opened 2022-10-10
Add blank world

### LINKED ISSUE #740: Add a scenario for creative mode with a blank world
We currently have a scenario for playing in creative mode with a procedurally generated world.  We should also add a scenario for creative mode with a *blank* world.  This would be useful for creating other scenarios (#558), and also might just be nice not having to worry about stuff being in the way if you want to play with creating some big farm or whatever.

_Originally posted by @byorgey in https://github.com/swarm-game/swarm/issues/558#issuecomment-1272619369_

xsebek:
- closes #740

### DIFF
--- a/data/scenarios/blank.yaml
+++ b/data/scenarios/blank.yaml
@@ -12,4 +12,3 @@ robots:
       attr: robot
 world:
   default: [grass]
-  offset: true
--- a/data/scenarios/blank.yaml
+++ b/data/scenarios/blank.yaml
@@ -1,0 +1,15 @@
+version: 1
+name: Blank world
+description: No constraints and no obstacles!  Do whatever you want, in this empty void!
+creative: true
+robots:
+  - name: base
+    loc: [0,0]
+    dir: [0,1]
+    heavy: true
+    display:
+      char: Ω
+      attr: robot
+world:
+  default: [grass]
+  offset: true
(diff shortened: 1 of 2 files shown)


data/scenarios/blank.yaml
```diff
@@ -0,0 +1,15 @@
...
+  - name: base
+    loc: [0,0]
+    dir: [0,1]
+    heavy: true
+    display:
+      char: Ω
+      attr: robot
+world:
+  default: [grass]
+  offset: true
```
byorgey, line 15: Should be able to remove `offset: true`, it does nothing if the world is blank (`offset: true` means we try to find a starting location for the base with various starting resources nearby).

byorgey (approved):

data/scenarios/blank.yaml
byorgey: Also, maybe the default should be `blank` instead of `grass`?  After all, the world is supposed to be... blank. :smile:

byorgey: Of course, we need to add a way to change the terrain... #742 .

xsebek: Sure, done! I purposely did not make it entirely blank... at least for now. There are a couple of reasons:
1. There are more games with default flat green world, like [Superflat in Minecraft](https://minecraft.fandom.com/wiki/Superflat). The first programming game I played was a gardener robot puzzle, so that might have influenced me.
2. It's harder to tell apart from the side panels.
3. A truly blank void is depressing.

PR #746, opened 2022-10-13
upgrade to `hsnoise-0.0.3`

### LINKED ISSUE #745: Hackage version has square worlds
I noticed that the worlds in the Hackage release version have visible squares:
Screenshot from 2022-10-13 00-12-13

The binary release does not have this defect.

This is caused by `hsnoise` having an old release on Hackage, without the fix for negative coordinates:

- colinhect/hsnoise#1

Using `source-repository-package` should not be a long-term solution. Cabal Issue with related discussion:
- haskell/cabal#8039

byorgey:
Closes #745.


.github/workflows/haskell-ci.yml
```diff
@@ -172,7 +172,7 @@ jobs:
           cabal-docspec --version
       - name: install hlint
         run: |
-          if [ $((HCNUMVER < 90000)) -ne 0 ] ; then HLINTVER=$(cd /tmp && (${CABAL} v2-install -v $ARG_COMPILER --dry-run hlint  --constraint='hlint >=3.3 && <3.4' |  perl -ne 'if (/\bhlint-(\d+(\.\d+)*)\b/) { print "$1"; last; }')); echo "HLint version $HLINTVER" ; fi
+          if [ $((HCNUMVER < 90000)) -ne 0 ] ; then HLINTVER=$(cd /tmp && (${CABAL} v2-install -v $ARG_COMPILER --dry-run hlint  --constraint='hlint >=3.5 && <3.6' |  perl -ne 'if (/\bhlint-(\d+(\.\d+)*)\b/) { print "$1"; last; }')); echo "HLint version $HLINTVER" ; fi
```
xsebek, line 175: Looks like this breaks GHC8 build - perhaps we need to change the condition so that this HLint version runs on GHC9?

I only wanted to run HLint on one GHC (otherwise versions get complicated), so I wrote this in `cabal.haskell-ci`:
```yaml
hlint-job: 8.10.7
```

Practically it's not important which GHC it runs on, just that it builds and runs on at least one, so feel free to change (or revert) to anything that makes sense. :slightly_smiling_face:

cabal.project
```diff
@@ -1,9 +1,4 @@
 -- If you edit this file, be sure to regenerate the CI configuration
 -- by running 'haskell-ci regenerate'.
 
-source-repository-package
-  type: git
-  location: https://github.com/colinhect/hsnoise
-  tag: 4ccff11dea7e8d94e6a5fcaf8f43857bd65bd72d
-
 packages: .
```
xsebek, line 4: Can we just delete the file, now that there is nothing left?

xsebek (approved):
Thanks, @byorgey! I was about to just start working on this, but you were faster. :smile:

.github/workflows/haskell-ci.yml
byorgey: OK, I'll try running it on 9.2.4.

PR #747, opened 2022-10-13
CHANGELOG for 0.1.1 release

byorgey:
I propose making a 0.1.1.0 release with the Hackage hsnoise fix and the few extra features we've accumulated so far.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,20 @@
 # Revision history for swarm
 
+## **0.1.1.0** - 2022-10-13
+
+A couple new features and an important bugfix for the Hackage release.
+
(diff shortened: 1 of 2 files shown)


xsebek (approved):
Might as well change the version in cabal. Speaking of which, I checked with the [policeman](https://github.com/kowainik/policeman) if we follow the PVP and it said we do:

image

byorgey:
> Might as well change the version in cabal.

Oh, right, we *must* change the version in cabal, or else the Hackage release will not work properly!

Re: PVP, I am not *too* concerned about it --- I think it is unlikely that people will be building things on top of swarm-the-library --- but we might as well follow it.

PR #749, opened 2022-10-15
Move TH GitInfo back to Main

xsebek:
It is probably safer to have Template Haskell GitInfo in Main and not depend on it in Swarm modules.

Depending on how the build system and TH interact we were either recompiling too often or not often enough.
- if the git info was evaluated once and not again after making a commit, then it was not up to date
- if the git info was reevaluated on every commit then we would needlessly recompile dependent modules

I believe it was only the former, but this helps even in that case (any code change recompiles Main).

### DIFF
--- a/src/Swarm/Version.hs
+++ b/src/Swarm/Version.hs
@@ -1,5 +1,4 @@
 {-# LANGUAGE OverloadedStrings #-}
-{-# LANGUAGE TemplateHaskell #-}
 
 -- |
 -- Module      :  Swarm.Version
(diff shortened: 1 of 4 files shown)


src/Swarm/Version.hs
```diff
@@ -58,32 +49,12 @@ import Text.ParserCombinators.ReadP (readP_to_S)
...
-type CommitHash = String
-
 -- | Check that the tag follows the PVP versioning policy.
 --
 -- Note that this filters out VS Code plugin releases.
 isSwarmReleaseTag :: String -> Bool
 isSwarmReleaseTag = all (\c -> isDigit c || c == '.')
 
-tagVersion :: Maybe (CommitHash, String)
-tagVersion = case gitInfo of
```
byorgey, line 78: What happened to `tagVersion`?  Was it dead code?

byorgey (approved):
Looks good to me.

src/Swarm/Version.hs
xsebek: Yes, we could not use it for release. Instead, we have to make sure that the version and tag are the same.

PR #750, opened 2022-10-18
improved handling of REPL mouse clicks

### LINKED ISSUE #470: Pass on mouse clicks to the REPL form
When the user clicks on text typed at the REPL prompt, it would be nice if it caused the cursor to jump to the place where they clicked.  I don't think this should be hard; brick forms apparently already support mouse events, so all we need to do is notice and pass along the mouse events appropriately.

byorgey:
- Never clear the `uiWorldCursor`.
    - In practice it seemed annoying and glitchy to understand when it would or wouldn't clear.  After clicking somewhere on the map to see the coordinates I was always afraid I was going to do the wrong thing and cause the cursor display to clear before I could make use of it.
- Pass on mouse clicks to the REPL input form 
    - This means you can click in the middle of the input to move the cursor 
    - Closes #470.

### DIFF
--- a/src/Swarm/TUI/Controller.hs
+++ b/src/Swarm/TUI/Controller.hs
@@ -281,10 +281,8 @@ handleMainEvent ev = do
         WorldPanel -> do
           mouseCoordsM <- Brick.zoom gameState (mouseLocToWorldCoords mouseLoc)
           uiState . uiWorldCursor .= mouseCoordsM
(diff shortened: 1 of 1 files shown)


xsebek (approved):
That’s a surprisingly small change 😮

byorgey:
> That’s a surprisingly small change

I know, isn't it lovely? :smile:

PR #762, opened 2022-10-19
Fix backstory loop

### LINKED ISSUE #756: "Backstory" tutorial freezes and crashes upon incorrect input
**Describe the bug**
In the first tutorial, the game pegs the CPU, consumes all memory, and is killed after typing: `say "whatever"` [enter] (instead of `say "Ready!"` [enter])

**To Reproduce**

   1. Start the "Backstory" tutorial, dismiss the Goal dialog.
   2. Type `say "z"` and hit Enter
   3. Observe CPU pegged, memory usage increase until the process is killed

**Expected behavior**
Memory/CPU should stay flat

**Screenshots**

Screenshot from 2022-10-18 21-50-03

Screenshot from 2022-10-18 21-40-26

**Additional context**
This is from the `main` branch (commit `b895668`), compiled and run via `stack run`.

xsebek:
- closes #756

### DIFF
--- a/data/scenarios/Tutorials/backstory.yaml
+++ b/data/scenarios/Tutorials/backstory.yaml
@@ -35,7 +35,7 @@ objectives:
         as l {has "READY"}
       } { return false }
 solution: |
-  say "I said something wrong."; say "Nobody types new messages this fast."; say "Ready!"
+  say "I said something wrong."; say "Ready!"
 entities:
   - name: READY
     display:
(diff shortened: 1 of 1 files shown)


byorgey:
This seems like an overly complex way to solve #756. Why not just use `log` instead of `say`, like you originally suggested?  We can discuss how to improve the `listen` command (and make a bunch of regression tests like this) in a separate issue.

xsebek:
@byorgey I wanted to keep the functionality of letting the player know that his message was wrong.

However, since it is not displayed in a logger it is not very visible anyway, so I reverted it and used `log`.

Now that I think about it, we could give the base a logger and `build` a robot that would log the message and `upload` it to the base. :thinking: But that may be a bit too complicated. :smile:

byorgey (approved):

byorgey:
> Now that I think about it, we could give the base a logger and `build` a robot that would log the message and `upload` it to the base. :thinking: But that may be a bit too complicated. :smile:

s/may/would definitely/   s/a bit/much/  :smile:

PR #767, opened 2022-10-21
Add a button to the Quit dialog to restart a scenario.

### LINKED ISSUE #549: Ability to quickly restart scenarios with win conditions
With some scenarios, it may be possible to literally get stuck so that it is impossible to win.  With others, it may be possible but very annoying.  In either case, the player should be able to restart somehow.  Maybe it should just be an additional option in the quit (^Q) menu, or maybe there should be a different key shortcut for restarting.

See related discussion at https://github.com/swarm-game/swarm/pull/533/files#r912874095 .

kostmo:
* Introduces a type alias `ScenarioInfoPair`.
* Introduces a new member `_scenarioRef` to the `UIState` record.
* Add a "Start over" button to the Quit dialog, which carries a reference to the current scenario.

Closes #549

### DIFF
--- a/src/Swarm/TUI/Model.hs
+++ b/src/Swarm/TUI/Model.hs
@@ -97,6 +97,7 @@ module Swarm.TUI.Model (
   uiInventoryShouldUpdate,
   uiTPF,
   uiFPS,
+  scenarioRef,
   appData,
 
   -- ** Initialization
@@ -129,6 +130,7 @@ module Swarm.TUI.Model (
   AppOpts (..),
   initAppState,
   startGame,
+  restartGame,
   scenarioToAppState,
   Seed,
 
@@ -167,10 +169,11 @@ import Linear (zero)
 import Network.Wai.Handler.Warp (Port)
 import Swarm.Game.Entity as E
 import Swarm.Game.Robot
-import Swarm.Game.Scenario (Scenario, loadScenario)
+import Swarm.Game.Scenario (loadScenario)
 import Swarm.Game.ScenarioInfo (
   ScenarioCollection,
   ScenarioInfo (..),
+  ScenarioInfoPair,
   ScenarioItem (..),
   ScenarioStatus (..),
   normalizeScenarioPath,
@@ -441,7 +444,7 @@ data ModalType
   | GoalModal [Text]
   deriving (Eq, Show)
 
-data ButtonSelection = CancelButton | KeepPlayingButton | QuitButton | NextButton (Scenario, ScenarioInfo)
+data ButtonSelection = CancelButton | KeepPlayingButton | StartOverButton ScenarioInfoPair | QuitButton | NextButton ScenarioInfoPair
 
 data Modal = Modal
   { _modalType :: ModalType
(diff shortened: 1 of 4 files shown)


src/Swarm/TUI/Model.hs
```diff
@@ -987,10 +999,11 @@ scenarioToAppState scene userSeed toRun = do
     l .= x'
 
 -- | Modify the UI state appropriately when starting a new scenario.
-scenarioToUIState :: Scenario -> UIState -> IO UIState
-scenarioToUIState _scene u =
-  return $
-    u
+scenarioToUIState :: (Scenario, ScenarioInfo) -> UIState -> IO UIState
+scenarioToUIState siPair u =
+  return $ newU {_scenarioRef = Just siPair}
```
byorgey, line 1004: Instead you can add `& scenarioRef ?~ siPair` below.

src/Swarm/TUI/Model.hs
```diff
@@ -947,6 +950,15 @@ initAppState AppOpts {..} = do
 startGame :: (MonadIO m, MonadState AppState m) => Scenario -> ScenarioInfo -> Maybe FilePath -> m ()
 startGame = startGameWithSeed Nothing
 
+
+-- | Re-initialize the game from the stored reference to the current scenario
+restartGame :: (MonadIO m, MonadState AppState m) => m ()
+restartGame = do
+  g <- use uiState
+  let Just (scene, si) = _scenarioRef g
```
byorgey, line 958: Might be simpler to just say `s <- use $ uiState . scenarioRef` and then pattern-match on `s`.  We should also handle the case when the result is `Nothing` --- I know in theory that shouldn't happen but we still don't want the game to crash if it does happen for some reason.

xsebek:
Hi @kostmo, thanks for working on this! :+1: Just so you know for the PR to be merged it must follow our formatting style.

Luckily it is really easy - the restyled action prepared a patch that you can apply. See the [Details](https://restyled.io/gh/swarm-game/repos/swarm/jobs/2415114) button next to it.

Alternatively, you can format the files yourself - see CONTRIBUTING about how to install and run Fourmolu 0.4.

byorgey:
@kostmo FYI, if you click on `Details` next to the restyled check you will see a command you can use to automatically apply the restyling fixes.

src/Swarm/TUI/Model.hs
kostmo: I've now handled the `Nothing` case by pushing the check all the way upstream to the population of buttons in the dialog.  The "Start over" button shall carry a reference to the scenario.   If the `scenarioRef` is `Nothing`, then the "Start over" button is not added.

byorgey (approved):
This looks good to me, thanks @kostmo !  Once you're 100% happy with it you can kick off the merge process by adding the `merge me` label to this PR.

byorgey:
@kostmo remember you should just add the `merge me` label rather than clicking the merge button manually.  Then the Mergify bot will make sure e.g. that the merged version of the branch will pass all the tests, make sure multiple merges at the same time won't conflict, etc. and merge automatically.  Seems like everything worked fine in this case though.

kostmo:
Of course, brain fart.  Will do.

byorgey:
No worries!  We used to have it set so you *couldn't* merge manually, but there were enough annoying situations where the merge bot got stuck or couldn't be used for some reason that we relaxed it.

xsebek:
@byorgey is there a merge commit template that GitHub allows us to set?

I wonder if we could add a note about using the label and Mergify there. So if you attempt to merge via GitHub button you get warned.

byorgey:
Good question, I don't know.

kostmo:
> We used to have it set so you _couldn't_ merge manually, but there were enough annoying situations where the merge bot got stuck or couldn't be used for some reason that we relaxed it.

Is Mergify Bot getting stuck still an issue?  Filed #843 to capture discussion.

PR #768, opened 2022-10-21
Default invisibity to system robot value

### LINKED ISSUE #726: Make system robots invisible by default
Default `robot.display.invisible` to `robot.system` (so system robots are invisible by default)

_Originally posted by @xsebek in https://github.com/swarm-game/swarm/issues/724#issuecomment-1270041526_

xsebek:
- closes #726


src/Swarm/Game/Display.hs
```diff
@@ -98,14 +99,19 @@ displayPriority :: Lens' Display Priority
...
-    Display
-      <$> v .:? "char" .!= ' '
-      <*> v .:? "orientationMap" .!= M.empty
-      <*> v .:? "curOrientation"
-      <*> (fmap (worldPrefix <>) <$> v .:? "attr") .!= entityAttr
-      <*> v .:? "priority" .!= 1
-      <*> v .:? "invisible" .!= False
+  parseJSON v = runE (parseJSONE v) (defaultEntityDisplay ' ')
+
+instance FromJSONE Display Display where
```
byorgey, line 104: I don't understand the purpose of making this into a `FromJSONE Display Display` instance.  It seems unnecessarily complicated.  Why not just keep the simple `FromJSON Display` instance and write something like 
```
do
  let defD = defaultEntityDisplay ' '
  Display
    <$> v .:? "char" .!= (defD ^. defaultChar)
```
and so on?

byorgey: Oh, wait, never mind, I get it now, it's so that you can pass in a different default display from the `FromJSONE EntityMap TRobot` instance, depending on the value of `system`.

byorgey (approved):
Looks good!

byorgey (changes requested):
Wait, I rescind my approval.  The base now looks like `^`.

src/Swarm/Game/Display.hs
xsebek: The real reason is that I want something better than `FromJSONE Bool` and the only reasonable wrapper seemed to be `Display` itself. The idea is really to provide defaults and in code it seemed to work OK.

xsebek:
Thanks for spotting that! :fearful: I will look into it.

xsebek:
@byorgey OK, I think what was going on is that the base robot was using _entity defaults!_

Originally once you set any part of the display the rest came from the defaults of entities.

Now it should use the robot defaults if you set just a part of it. :slightly_smiling_face: 

---

However, that means all robots that set just part of the display could have expected entity defaults, so now I have to check them out too.

byorgey (approved):
Looks good to me!

PR #769, opened 2022-10-21
Generate other wikis

### LINKED ISSUE #344: Create command line tools to automate documentation

xsebek:
- generate wiki for:
  - entities
  - recipes (table and added recipe graph image)
  - capabilities (only table)
- properly parse command line argument for `generate cheatsheet`
- closes #344


byorgey:
Links on the new recipes and capabilities pages seem to be broken, they are missing `/wiki/`.

byorgey (approved):
I don't plan to read every line of code carefully. :smile:  But the generated wiki pages look good (other than the broken links, which I suspect is not a problem with the code but rather with the flags passed on the command line).

xsebek:
@byorgey I am passing `--entities-page='../Entities-cheat-sheet'` because it is the relative address on our Wiki.

But if we or someone else wants to use it outside the Wiki then absolute links will be necessary.

---

I will check why the links are broken, I must have made some silly mistake. 😅

xsebek:
OK, so what confused me is that relative links work in the Wiki editor preview but then not on the page once saved.

I fixed it by using absolute links. 🙂

byorgey:
Would it not work to just say e.g. `--commands-page='Commands-Cheat-Sheet' ?

Issue #519, opened 2022-06-30
Ability to filter / sort the inventory


byorgey:
It would be nice to have some improvements to the way the player can manipulate the inventory.
- As a start, we should sort alphabetically (case-insensitive) instead of just by ASCII.
- We can already toggle showing entities of which we have 0: #301.
- There is a discussion of search / narrowing at #126
- It might be nice to be able to sort by other criteria.  Off the top of my head:
    - Sort by how many copies you have
    - Sort/categorize by type, i.e. growable, etc.
    - Sort by color (because why not =)

Based on [a comment](https://github.com/swarm-game/swarm/pull/503#discussion_r911380332_) discussing #503 .

xsebek:
Numerical or some sort of “human” sort would also be handy for the 2048 challenge. 😁

byorgey:
Like https://en.wikipedia.org/wiki/Natural_sort_order ?  Yes, that would be nice!

xsebek:
Nice I did not know about that. There seems to be a package available that is _old_ but has very few dependencies and they are not very constrained:
- https://hackage.haskell.org/package/natural-sort

byorgey:
@kostmo do you think this can be closed now, or is there still more we'd like to implement?  For example, I'm not sure if the existing sort uses "natural sort order" as discussed above.

kostmo:
Yes, it [does use natural sort](https://github.com/swarm-game/swarm/blob/a6abebda522782f9d81621bafc854de976ae64e8/src/Swarm/TUI/Inventory/Sorting.hs#L16).  Closing.

PR #775, opened 2022-10-23
make inventory sort case-insensitive

byorgey:
Part of #519.


byorgey:
In general I like to leave small, easy improvements like this for newcomers to tackle, but this one was bothering me so much I just had to do it myself. :grin:

xsebek (approved):
We have 34 <kbd>Low hanging fruit</kbd>, no harm in picking a few. :smile: :green_apple:

PR #781, opened 2022-10-23
Better advertise pop-out for entity descriptions

### LINKED ISSUE #764: Better advertise ability to pop up entity description
> > We should add a way to show the description in a popup modal so that such long descriptions are easier to read. 

> That already exists, just hit `Enter` on an inventory item.  But clearly we need to advertise it better!

_Originally posted by @byorgey in https://github.com/swarm-game/swarm/issues/753#issuecomment-1283999856_

- [x] Change `[Ret] focus` hint to something like `[Ret] pop out`
- [x] Add hint to the tutorial
- [x] Highlight `[Ret] pop out` when the info panel scrolls
- [x] Highlight all panel-specific keyboard shortcuts

byorgey:
Closes #764.

- Hint now says `[Ret] pop out` instead of `[Ret] focus`.
- `pop out` is now highlighted whenever the info panel can scroll.
- Add hints to the tutorial about popping out or scrolling entity descriptions.
- All panel-specific keys are now highlighted the same color as the focused panel.
- Added a panel-specific key hint for `f` on the World panel toggling FPS info.

### DIFF
--- a/src/Swarm/TUI/View.hs
+++ b/src/Swarm/TUI/View.hs
@@ -791,7 +791,7 @@ drawKeyMenu s =
   keyCmdsFor (Just REPLPanel) =
     [ ("↓↑", "history")
     ]
-      ++ [("Ret", "execute") | not isReplWorking]
+      ++ [("Enter", "execute") | not isReplWorking]
       ++ [("^c", "cancel") | isReplWorking]
   keyCmdsFor (Just WorldPanel) =
     [ ("←↓↑→ / hjkl", "scroll") | creative
@@ -799,7 +799,7 @@ drawKeyMenu s =
       ++ [("c", "recenter") | not viewingBase]
       ++ [("f", "FPS")]
   keyCmdsFor (Just RobotPanel) =
-    [ ("Ret", "pop out")
+    [ ("Enter", "pop out")
     , ("m", "make")
     , ("0", (if showZero then "hide" else "show") <> " 0")
     ]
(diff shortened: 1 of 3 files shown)


src/Swarm/TUI/View.hs
```diff
@@ -793,20 +797,29 @@ drawKeyMenu s =
     [ ("←↓↑→ / hjkl", "scroll") | creative
     ]
       ++ [("c", "recenter") | not viewingBase]
+      ++ [("f", "FPS")]
   keyCmdsFor (Just RobotPanel) =
-    [ ("Ret", "focus")
+    [ ("Ret", "pop out")
```
xsebek, line 802: We have enough space for clearer message:
```suggestion
    [ ("Enter", "pop out entity description")
```

xsebek (approved):
:+1:

src/Swarm/TUI/View.hs
byorgey: Depends on how big the terminal is. :thinking: I wonder if we could make things "responsive" so e.g. it uses abbreviations, shorter messages,  etc when there's not much space.

byorgey: I think I'll stick with the shorter `pop out` for now.  These are supposed to be small hints to remind you what the keys do, not full descriptions, and I'd rather not have one be that long.  However, changing `Ret` to `Enter` is a good idea.

PR #784, opened 2022-10-23
add lambdas to `requireinv` scenario and change solution to use them

byorgey:
This is a pretty simple change --- we've already covered lambdas previously in the tutorial so it's annoying not to be able to use them to define things like `x4 = \c. c;c;c;c end`.  But just wanted to make sure there's nothing I'm missing here.


kostmo (approved):
Makes sense!

PR #787, opened 2022-10-24
Preserve seed upon restart

kostmo:
Part of #549

Before this change, if the Classic scenario was started:

```
stack run -- --scenario data/scenarios/classic.yaml
```
and then "Start over" is selected from the Quit dialog, a new seed was generated.

Now, the previous seed is re-used.

### DIFF
--- a/src/Swarm/TUI/Controller.hs
+++ b/src/Swarm/TUI/Controller.hs
@@ -371,7 +371,7 @@ handleModalEvent = \case
     case dialogSelection <$> mdialog of
       Just (Just QuitButton) -> quitGame
       Just (Just KeepPlayingButton) -> toggleModal KeepPlayingModal
-      Just (Just (StartOverButton siPair)) -> restartGame siPair
+      Just (Just (StartOverButton currentSeed siPair)) -> restartGame currentSeed siPair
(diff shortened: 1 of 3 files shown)


byorgey (commented):
The code looks fine, but what's the motivation for keeping the seed the same across restarts?  Of course this only makes a difference in scenarios with a procedurally generated world, which at this point includes Classic and Creative modes, as well as some of the speedrun scenarios.  Personally, one of the reasons I might want to restart is that I don't like the seed and want to restart with a new one!  So I would be annoyed if I can't use the restart button for that purpose but have to quit out to the menu and choose the scenario again.  But that's just the way that I personally play --- so I'm interested to hear your motivation for adding this.

byorgey:
Hmm, however, I can definitely see the argument that if you start with an explicit seed, like `stack run -- --scenario classic -s 25`, then we ought to preserve it across restarts.  In that case we would have to remember whether the seed was randomly generated or explicitly specified by the user.

xsebek:
@byorgey If you want to try a new seed, then it makes sense to quit the scenario entirely and run the command again.

Similarly, we don't remove other command line options like `--cheat`, so the proposed change makes sense to me.

xsebek (approved):

byorgey:
OK, I'm convinced!

PR #792, opened 2022-10-27
Introduce autoplay flag

kostmo:
This PR introduces an `--autoplay` flag that causes the solution stored in the `solutions` field of a scenario file (which was selected on the command line via the `--scenario` option) to be automatically run when the game starts, identically to the `--run` option.

The purpose is to:
1. more easily showcase scenarios to the community
2. facilitate a quick a sanity check that the solution text in the yaml file is correct

# Test procedure

    stack run -- --scenario data/scenarios/Tutorials/conditionals.yaml --autoplay

This has the same effect as manually copying the solution stored in `conditionals.yaml` to `myfile.sw`, and then running:

    stack run -- --scenario data/scenarios/Tutorials/conditionals.yaml --run myfile.sw

<a href="https://asciinema.org/a/mGB5VDKhsLLPdrItZWi3qsHUV" target="_blank"></a>

### DIFF
--- a/src/Swarm/TUI/Model.hs
+++ b/src/Swarm/TUI/Model.hs
@@ -146,7 +146,7 @@ import Brick.Focus
 import Brick.Forms
 import Brick.Widgets.Dialog (Dialog)
 import Brick.Widgets.List qualified as BL
-import Control.Applicative (Applicative (liftA2))
+import Control.Applicative (Applicative (liftA2), (<|>))
 import Control.Lens hiding (from, (<.>))
 import Control.Monad.Except
 import Control.Monad.State
@@ -181,6 +181,7 @@ import Swarm.Game.ScenarioInfo (
   scenarioCollectionToList,
   scenarioItemByPath,
   scenarioPath,
+  scenarioSolution,
   scenarioStatus,
   _SISingle,
  )
@@ -927,7 +928,9 @@ data AppOpts = AppOpts
   , -- | Scenario the user wants to play.
     userScenario :: Maybe FilePath
   , -- | Code to be run on base.
-    toRun :: Maybe FilePath
+    scriptToRun :: Maybe FilePath
+  , -- | Automatically run the solution defined in the scenario file
+    autoPlay :: Bool
   , -- | Should cheat mode be enabled?
     cheatMode :: Bool
   , -- | Explicit port on which to run the web API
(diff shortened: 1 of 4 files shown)


src/Swarm/TUI/Model.hs
```diff
@@ -939,20 +942,28 @@ data AppOpts = AppOpts
...
   case skipMenu of
     False -> return $ AppState gs ui rs
     True -> do
       (scenario, path) <- loadScenario (fromMaybe "classic" userScenario) (gs ^. entityMap)
+
+      let maybeAutoplay = do
+            guard autoPlay
+            soln <- scenario ^. scenarioSolution
+            return $ SuggestedSolution soln
+      let realToRun = maybeAutoplay <|> (ScriptPath <$> scriptToRun)
```
byorgey, line 959:
```suggestion
      let realToRun = (guard autoPlay >> SuggestedSolution <$> (scenario ^. scenarioSolution)) <|> (ScriptPath <$> scriptToRun)
```

Something like this (possibly split across a couple lines) seems like it would be simpler/clearer to me, but it's just a matter of taste, I don't feel strongly about it.

byorgey (approved):
I like it!  I only had one minor style suggestion but you're welcome to take it or leave it.

byorgey:
By the way, you can just write `--scenario Tutorials/conditionals`, you don't have to put `data/scenarios/` on the front and the trailing `.yaml` is also optional.  Though I suppose they are technically slightly different: `--scenario data/scenarios/Tutorials/conditionals.yaml` will load it from the file as it exists in the repo, whereas `--scenario Tutorials/conditionals` will load it from the data directory that has been installed with the built package.  So if you edit the `.yaml` file, using the full `data/scenarios/...` path will immediately pick up the change, whereas just using `Tutorials/conditionals` will not pick up the change until the package has been rebuilt.

byorgey:
BTW, @kostmo , thanks for all the great PRs and issues recently.   Sorry I haven't been as responsive as I'd like, I recently had to take on some extra responsibilities at work when a colleague left unexpectedly for medical reasons.  So it's been a little crazy. :scream:  But looking at PRs is actually fun (unlike the extra work responsibilities) so I'm slowly looking at things as I can.

xsebek:
Wow, I totally missed this PR and only learned about it from the release. 😮 

Thanks @kostmo!

Does this imply a cheating flag, so that the autoplay does not count as best time?

kostmo:
Not currently, but we should definitely make it so. #816

PR #793, opened 2022-10-27
User-controllable sort criteria

kostmo:
When the inventory pane is selected (Alt+e), one can use the `;` and `:` keys to cycle sort criteria and sort direction, respectively.
Screenshot from 2022-10-26 21-52-11

This functionality is implemented in a new module to avoid exacerbating #707.

Note that `entityProperties` may be an unreliable criteria to sort, because it is a list that itself may contain duplicates or items in unpredictable order.  Perhaps this field should be made a `Set` (#794).

### DIFF
--- a/src/Swarm/TUI/Inventory/Sorting.hs
+++ b/src/Swarm/TUI/Inventory/Sorting.hs
@@ -1,0 +1,77 @@
+{-# LANGUAGE OverloadedStrings #-}
+
+module Swarm.TUI.Inventory.Sorting (
+  InventorySortOptions (..),
+  InventorySortDirection (..),
+  InventorySortOrder (..),
+  cycleSortOrder,
+  cycleSortDirection,
+  defaultSortOptions,
+  sortInventory,
+  renderSortMethod,
+) where
+
+import Algorithms.NaturalSort (sortKey)
+import Control.Lens (view)
+import Data.List (sortBy)
+import Data.Ord (Down (Down), comparing)
+import Data.Text qualified as T
+import Swarm.Game.Entity as E
+import Swarm.Util (cycleEnum)
+
+data InventorySortDirection
+  = Ascending
+  | Descending
+  deriving (Enum, Bounded, Eq)
+
+data InventorySortOrder
+  = ByNaturalAlphabetic
+  | ByQuantity
+  | ByType
+  deriving (Enum, Bounded, Eq)
+
+data InventorySortOptions = InventorySortOptions InventorySortDirection InventorySortOrder
+
+defaultSortOptions :: InventorySortOptions
+defaultSortOptions = InventorySortOptions Ascending ByNaturalAlphabetic
+
+renderSortMethod :: InventorySortOptions -> T.Text
+renderSortMethod (InventorySortOptions direction order) =
+  T.unwords [prefix, label]
+ where
+  prefix = case direction of
+    Ascending -> "↑"
+    Descending -> "↓"
+  label = case order of
+    ByNaturalAlphabetic -> "name"
+    ByQuantity -> "count"
+    ByType -> "type"
+
+cycleSortOrder :: InventorySortOptions -> InventorySortOptions
+cycleSortOrder (InventorySortOptions direction order) =
+  InventorySortOptions direction (cycleEnum order)
+
+cycleSortDirection :: InventorySortOptions -> InventorySortOptions
+cycleSortDirection (InventorySortOptions direction order) =
+  InventorySortOptions (cycleEnum direction) order
+
+-- | All non-alphabetic sort criteria perform alphabetic tie-breaking.
+-- "Reverse ordering" only applies to the *primary* sort criteria; the secondary
+-- alphabetic sort is always in ascending order.
+getSortCompartor :: Ord a => InventorySortOptions -> (a, Entity) -> (a, Entity) -> Ordering
+getSortCompartor (InventorySortOptions direction order) = case order of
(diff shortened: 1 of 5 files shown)


swarm.cabal
```diff
@@ -152,6 +153,7 @@ library
                       minimorph                     >= 0.3 && < 0.4,
                       mtl                           >= 2.2.2 && < 2.3,
                       murmur3                       >= 1.0.4 && < 1.1,
+                      natural-sort,
```
kostmo, line 156: How to determine appropriate bounds for this package?

byorgey: Hmm, well, it was uploaded in 2012 and has never been updated since (and probably never will).  Since it is so old, there's no reason to allow anything but the latest version.  We might as well just put `>= 0.1.2 && < 0.2`.

src/Swarm/TUI/Inventory/Sorting.hs
```diff
@@ -0,0 +1,79 @@
+{-# LANGUAGE OverloadedStrings #-}
+
+-- |
+-- Module      :  Swarm.TUI.Inventory.Sorting
+-- Copyright   :  Brent Yorgey
```
byorgey, line 5: It's actually a little silly to say that I hold the copyright to this file that I didn't write.  By contributing to this repository you're agreeing to license your contributions under a BSD open-source license, of course, but you're certainly not actually giving me the copyright.  You should either put your own name, or maybe we should discuss something more systematic like putting "Swarm contributors", or not listing a copyright field at all.

src/Swarm/TUI/Inventory/Sorting.hs
```diff
@@ -0,0 +1,79 @@
...
+) where
+
+import Control.Lens (view)
+import Data.List (sortBy)
+import Data.Ord (comparing)
+import Data.Text qualified as T
+
+import Algorithms.NaturalSort (sortKey)
+
+import Swarm.Game.Entity as E
```
byorgey, line 26: I think what we've agreed on most recently is to list all the imports in one big block with no blank lines in between, because these days imports can just be handled automatically by your favorite editor/IDE, so there's no particular reason any more to separate them out into logical groups to make them easier for humans to read.  I don't know if we've actually converted all the source files in the repo to use that format or whether there are still some that have separate blocks of imports.

src/Swarm/TUI/Inventory/Sorting.hs
```diff
@@ -0,0 +1,79 @@
...
+ where
+  prefix = case direction of
+    Ascending -> "↑"
+    Descending -> "↓"
+  label = case order of
+    ByNaturalAlphabetic -> "name"
+    ByQuantity -> "count"
+    ByType -> "type"
+
+succCyclic :: (Enum a, Bounded a, Eq a) => a -> a
```
byorgey, line 53: We already have this function in `Swarm.Util.cycleEnum`.  (`succCyclic` is probably a better name, we can definitely rename it if we want, but in any case we shouldn't duplicate it.)

src/Swarm/TUI/Inventory/Sorting.hs
```diff
@@ -0,0 +1,79 @@
...
+  InventorySortOptions direction (succCyclic order)
+
+cycleSortDirection :: InventorySortOptions -> InventorySortOptions
+cycleSortDirection (InventorySortOptions direction order) =
+  InventorySortOptions (succCyclic direction) order
+
+getSortCompartor :: Ord a => InventorySortOrder -> (a, Entity) -> (a, Entity) -> Ordering
+getSortCompartor order = case order of
+  ByNaturalAlphabetic -> comparing $ sortKey . T.toLower . view entityName . snd
+  ByQuantity -> comparing fst
```
byorgey, line 70: How does it sort among entities with the same quantity?  It seems kind of random.  For example if I start Creative mode and sort by quantity I get everything in a pretty random-seeming order.  I guess it is by hash or whatever order the entities would already be in without applying any sort?  Maybe we should add a secondary natural sort, like  `comparing fst <> (comparing $ sortKey . T.toLower . view entityName . snd)`?

src/Swarm/TUI/Inventory/Sorting.hs
```diff
@@ -0,0 +1,79 @@
...
+  ByType -> comparing $ view entityProperties . snd
```
byorgey, line 71: The same comment goes for sorting by type: among entities with the same type the order seems very random, I think it would be nice to secondarily sort by name.

src/Swarm/TUI/Model.hs
```diff
@@ -836,6 +840,7 @@ initUIState showMainMenu cheatMode = liftIO $ do
       , _uiReplHistory = newREPLHistory history
       , _uiReplLast = ""
       , _uiInventory = Nothing
+      , _uiInventorySort = InventorySortOptions Ascending ByNaturalAlphabetic
```
byorgey, line 843: I think we should also reset the `uiInventorySort` in `scenarioToUIState`, so the sort order always starts out as ascending alphabetic every time you start a game.  Currently, whatever the sort order was in your previous game persists into the next.

byorgey (commented):
This is looking very nice!

src/Swarm/TUI/Inventory/Sorting.hs
kostmo: I'd prefer omitting it entirely.  Maximize information density in the source code by eliminating redundancies.  The root-level copyright doc certainly covers all of the source files.

kostmo:
All comments resolved.

byorgey (approved):
:rocket:

PR #794, opened 2022-10-27
Use Set instead of a list for entity properties, for sortability

kostmo:
It is possible for list to contain duplicates or items in unpredictable order.


byorgey (approved):
Makes sense, not sure why I didn't do it this way from the beginning. :smile:

Issue #795, opened 2022-10-27
Generalize representation of dependencies between goals


kostmo:
This builds upon #378.

**Is your feature request related to a problem? Please describe.**

I'd like to be able define "optional" goals (e.g. hidden achievements, or "secondary" or "tertiary" goals) for a scenario, aside from the primary goals.

**Describe the solution you'd like**
We could use a `Tree` rather than a list to represent the goals.  Conveniently, YAML can already represent trees.

Goals that must be completed *in sequence* would be represented as a chain of increasing depth, while goals that are independent of each other shall be siblings in the YAML file.  A slight drawback is that for each of the existing scenario files, multi-goal sequences will need to be indented into a "terrace" of children.

Perhaps we would define the following enum to label each goal:

    data GoalType = Primary | Secondary | Tertiary | Hidden

`Primary` shall be the "default" goal type when omitted.
The scenario would be considered "complete" once all of the `Primary` goals are met.  The "Keep playing" option will then be useful to pursue the other goals.

byorgey:
Hmm, I'm not convinced --- that is, I'm not convinced it's worth it to use *trees* in particular.  What about a scenario where you have to complete goal A to unlock goals B and C, and then you can only unlock goal D once you have completed *both* B and C?  Or what about something where you need to complete any two of A, B, or C in order to unlock D?  Or where secondary goal B is only available to you if you have already completed a different tutorial scenario?  etc.?  Switching to trees would be a lot of work and would complicate the code, but there would be lots of potential scenarios we still could not express.  Really the only reason to use trees in particular would be that YAML can already represent them.  So it would be a lot of added complexity with only a small gain in expressiveness.

If we want to generalize the goal system (and I'm very open to that), I think we should jump straight past trees to something much more general.  I'm not sure exactly what, but here are some ideas off the top of my head:
* If we give goals unique identifiers, then each goal could have a list of other goals which are prerequisites
    - This by itself would let us express not just trees but arbitrary directed acyclic graphs
* Or each goal could have some *code* to define when it should become active/available, just like goals already have arbitrary code to determine when they have been met
* The idea of having "primary" goals that have to be met to complete a scenario makes sense, but I don't think we would need to make up a `GoalType` enum; each scenario could simply have a list of goals which must be achieved.  Or, even simpler, each scenario would be required to list *one* goal as its primary goal.  That goal could then encode whatever logic makes sense, such as completing a list of other goals, or completing two out of three of certain goals, etc.

[...]

PR #796, opened 2022-10-27
achievements board

kostmo:
Towards #797.

This PR implements defines the achievements schema, defines a few example achievements, and implements saving and loading.

Main menu entry:
Screenshot from 2022-10-26 23-33-47

Achievements screen:
image

### DIFF
--- a/src/Swarm/Game/Step.hs
+++ b/src/Swarm/Game/Step.hs
@@ -48,6 +48,7 @@ import Data.Set (Set)
 import Data.Set qualified as S
 import Data.Text (Text)
 import Data.Text qualified as T
+import Data.Time (getZonedTime)
 import Data.Tuple (swap)
 import Linear (V2 (..), zero)
 import Swarm.Game.CESK
@@ -68,6 +69,8 @@ import Swarm.Language.Pipeline.QQ (tmQ)
 import Swarm.Language.Requirement qualified as R
 import Swarm.Language.Syntax
 import Swarm.Language.Typed (Typed (..))
+import Swarm.TUI.Model.Achievement.Attainment
+import Swarm.TUI.Model.Achievement.Definitions
 import Swarm.Util
 import Swarm.Util.Location
 import System.Clock (TimeSpec)
@@ -336,7 +339,11 @@ hasCapabilityFor cap term = do
 
 -- | Create an exception about a command failing.
 cmdExn :: Const -> [Text] -> Exn
-cmdExn c parts = CmdFailed c (T.unwords parts)
+cmdExn c parts = CmdFailed c (T.unwords parts) Nothing
+
+-- | Create an exception about a command failing, with an achievement
+cmdExnWithAchievement :: Const -> [Text] -> GameplayAchievement -> Exn
+cmdExnWithAchievement c parts a = CmdFailed c (T.unwords parts) $ Just a
 
 -- | Raise an exception about a command failing with a formatted error message.
 raise :: (Has (Throw Exn) sig m) => Const -> [Text] -> m a
@@ -596,15 +603,27 @@ stepCESK cesk = case cesk of
   -- First, if we were running a try block but evaluation completed normally,
   -- just ignore the try block and continue.
   Out v s (FTry {} : k) -> return $ Out v s k
-  -- If an exception rises all the way to the top level without being
-  -- handled, turn it into an error message.
-
-  -- HOWEVER, we have to make sure to check that the robot has the
-  -- 'log' capability which is required to collect and view logs.
-  --
-  -- Notice how we call resetBlackholes on the store, so that any
-  -- cells which were in the middle of being evaluated will be reset.
   Up exn s [] -> do
+    -- Here, an exception has risen all the way to the top level without being
+    -- handled.
+    case exn of
+      CmdFailed _ _ (Just a) -> do
+        currentTime <- sendIO getZonedTime
+        gameAchievements
+          %= M.insertWith
+            (<>)
+            a
+            (Attainment (GameplayAchievement a) Nothing currentTime)
+      _ -> return ()
+
+    -- If an exception rises all the way to the top level without being
+    -- handled, turn it into an error message.
+
+    -- HOWEVER, we have to make sure to check that the robot has the
+    -- 'log' capability which is required to collect and view logs.
+    --
+    -- Notice how we call resetBlackholes on the store, so that any
+    -- cells which were in the middle of being evaluated will be reset.
     let s' = resetBlackholes s
     h <- hasCapability CLog
     em <- use entityMap
@@ -736,7 +755,7 @@ execConst c vs s k = do
         return $ Waiting (time + d) (Out VUnit s k)
       _ -> badConst
     Selfdestruct -> do
-      destroyIfNotBase
+      destroyIfNotBase $ Just AttemptSelfDestructBase
       flagRedraw
       return $ Out VUnit s k
     Move -> do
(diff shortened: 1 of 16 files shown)


xsebek:
How about adding the first achievements for completing some set of scenarios? 🙂 

We do not need to store new data for it, except for which scenarios are needed for the achievements.

@kostmo you could hardcode it in the app for now or add a `scenario-achievements.yaml` file if you want.

I think that would make this actually useful and also populate the Achievements view with some entries for a relatively little cost.

But I leave that to you, we had empty panels before, so we can merge it even as a mock-up. 👍

src/Swarm/TUI/Model/Achievement/Description.hs
```diff
@@ -0,0 +1,40 @@
+{-# LANGUAGE OverloadedStrings #-}
+
+module Swarm.TUI.Model.Achievement.Description where
+
+import Swarm.TUI.Model.Achievement.Definitions
+
+describe :: CategorizedAchievement -> AchievementInfo
```
byorgey, line 7: Would there be any benefit to having all this information stored in an external `.yaml` file?  I know we still need to have a data type enumerating the possible achievements, since the code for deciding when we have obtained them is arbitrary and can't live in an external data file.  But the titles, flavor text, explanation, etc. could all live in a data file.

src/Swarm/TUI/Model/UI.hs
```diff
@@ -163,6 +169,10 @@ uiModal :: Lens' UIState (Maybe Modal)
 --   has been displayed to the user initially.
 uiGoal :: Lens' UIState (Maybe [Text])
 
+-- | Status of the scenario goal: whether there is one, and whether it
+--   has been displayed to the user initially.
```
byorgey, line 173: This comment seems like a copy-pasto.

src/Swarm/TUI/Model/UI.hs
```diff
@@ -270,6 +281,7 @@ initUIState showMainMenu cheatMode = liftIO $ do
       , _uiError = Nothing
       , _uiModal = Nothing
       , _uiGoal = Nothing
+      , _uiAchievements = M.fromList $ map (\x -> (x ^. achievement, x)) achievements
```
byorgey, line 284:
```suggestion
      , _uiAchievements = M.fromList $ map (view achievement &&& id) achievements
```

I find this clearer, but YMMV.  Feel free to use or ignore this suggestion as you wish.

src/Swarm/TUI/View/Achievement.hs
```diff
@@ -0,0 +1,84 @@
...
+getCompletionIcon = \case
+  False -> txt " ○  "
+  True -> withAttr greenAttr $ txt " ●  "
+
+drawAchievementsMenuUI :: AppState -> BL.List Name CategorizedAchievement -> Widget Name
+drawAchievementsMenuUI s l =
+  vBox
+    [ hCenter $ padTopBottom 1 $ str "🏆  Achievements 🏆 "
+    , hBox
+        [ hLimitPercent 50 $
```
byorgey, line 32: This wastes a lot of space and doesn't look very good on bigger terminals:
achievements-menu

Perhaps there should be some proportional padding to the left and right so the menu and achievement details don't extend all the way to the edges of the window?

src/Swarm/TUI/View/Achievement.hs
```diff
@@ -0,0 +1,84 @@
...
+  details = describe x
+
+singleAchievementDetails ::
+  Map CategorizedAchievement Attainment ->
+  CategorizedAchievement ->
+  Widget Name
+singleAchievementDetails attainedMap x =
+  padRight (Pad 1) $ borderWithLabel titleWidget $ padAllEvenly 1 innerContent
+ where
+  wasAttained = M.member x attainedMap
```
byorgey, line 59: Since we store the time at which the achievement was attained, it might be nice to show that information somewhere.

src/Swarm/TUI/Controller.hs
```diff
@@ -150,10 +153,13 @@ handleMainMenuEvent menu = \case
...
                 _ -> error "No first tutorial found!"
           startGame firstTutorial Nothing
+        Achievements -> uiState . uiMenu .= AchievementsMenu (BL.list AchievementList (V.fromList listAchievements) 1)
         Messages -> do
           runtimeState . eventLog . notificationsCount .= 0
           uiState . uiMenu .= MessagesMenu
-        About -> uiState . uiMenu .= AboutMenu
+        About -> do
+          uiState . uiMenu .= AboutMenu
+          attainAchievement PersistImmediate $ GlobalAchievement LookedAtAboutScreen
```
byorgey, line 162: This is the only achievement out of the four that is actually achievable at the moment, right?  Or did I miss something?

src/Swarm/TUI/Model/StateUpdate.hs
```diff
@@ -98,6 +105,24 @@ scenarioToAppState siPair@(scene, _) userSeed toRun = do
...
+  (uiState . uiAchievements)
+    %= M.insertWith
+      (<>)
+      a
+      (Attainment a Nothing currentTime)
+  case persist of
+    PersistImmediate -> do
+      newAchievements <- use $ uiState . uiAchievements
+      liftIO $ saveAchievementsInfo $ M.elems newAchievements
+    PersistDeferred -> return ()
```
byorgey, line 124: When will deferred achievements actually be saved?  Upon winning a scenario?  I don't think I saw any calls to `saveAchievementsInfo` anywhere else.  What kind of achievements do you envision being deferred vs immediate?

byorgey (commented):
This is really nice so far!

src/Swarm/TUI/Model/Achievement/Description.hs
kostmo: > we still need to have a data type enumerating the possible achievements

One of the members of this enumeration's sum type could have a string field, which makes it open-ended for achievements defined in config files.  Which raises the question, should all of the achievements be stored in config?  At the moment I think I prefer a hybrid: define as many achievements as feasible in code to obtain exhaustiveness-checking from the compiler, and let the rest be in `.yaml`.

byorgey: Makes sense, although if we allow achievements to be defined completely in a `.yaml` file, wouldn't we need some sort of little DSL to describe what the achievement is for?  Or are you imagining that achievements could be defined via some Swarm code just like goal conditions?

kostmo: Yes, they would be specially-annotated goals.  #795 is probably a prerequisite.

src/Swarm/TUI/Model/StateUpdate.hs
kostmo: I added this speculatively, but I suppose I'll remove it for now.

src/Swarm/TUI/View/Achievement.hs
kostmo: Here's the new look:

| Narrow | Wide |
| --- | --- |
| Screenshot from 2022-12-17 14-53-55 | Screenshot from 2022-12-17 14-49-16 |

src/Swarm/TUI/Controller.hs
kostmo: At this moment there are two achievements implemented out of seven defined:

* "About Time!"
* "Call of the Void"

The latter demonstrates propagation of the achievement from within the GameState, out to the UI and persistence layers.

src/Swarm/Game/Step.hs
```diff
@@ -736,7 +755,7 @@ execConst c vs s k = do
         return $ Waiting (time + d) (Out VUnit s k)
       _ -> badConst
     Selfdestruct -> do
-      destroyIfNotBase
+      destroyIfNotBase $ Just AttemptSelfDestructBase
```
byorgey, line 758: Note this is currently only possible in creative mode, since it is not possible to craft a `detonator`.  That's fine though.

byorgey (approved):
This is looking great!  I'll leave it up to you whether you want to merge it as-is and incrementally work on adding more achievements, or wait until more has been implemented.

PR #798, opened 2022-10-28
autocomplete entity names in the repl

kostmo:


src/Swarm/TUI/Controller.hs
```diff
@@ -788,18 +788,35 @@ handleREPLEvent = \case
...
+
+  completeWith w m = T.append t (T.drop (T.length w) m)
   lastWord = T.takeWhileEnd isIdentChar t
   names = s ^.. gameState . baseRobot . robotContext . defTypes . to assocs . traverse . _1
   possibleWords = reservedWords ++ names
-  matches = filter (lastWord `T.isPrefixOf`) possibleWords
+  kwMatches = filter (lastWord `T.isPrefixOf`) possibleWords
+
+  -- checks the "parity" of the number of quotes. If odd, then there is an open quote.
+  hasOpenQuotes = (== 1) . (`mod` 2) . T.count "\""
```
byorgey, line 814: I don't know if it's worth fixing this, but technically this will break if there are any escaped quote characters in a string literal.

xsebek: This approach sounds good enough for now. 👍 

Realistically, I think complex programs using escaped quotes will be written in the editor, not REPL.
If we add suggestions to the LSP, then we can think about a more robust check.

src/Swarm/TUI/Controller.hs
```diff
@@ -788,18 +788,34 @@ handleREPLEvent = \case
...
+  theEntityMap = s ^. gameState . entityMap
+  entityNames = M.keys $ entitiesByName theEntityMap
+  lastQuotedString = T.takeWhileEnd (/= '"') t
+  strMatches = filter (lastQuotedString `T.isPrefixOf`) entityNames
```
xsebek, line 818: I feel like this could be nicer if we filtered using some Map function and then got the list of keys.

But maybe that is just useless optimization. 🤷

src/Swarm/TUI/Controller.hs
```diff
@@ -788,18 +788,34 @@ handleREPLEvent = \case
 tabComplete :: AppState -> REPLPrompt -> REPLPrompt
 tabComplete _ p@(SearchPrompt {}) = p
 tabComplete s (CmdPrompt t mms)
-  | (m : ms) <- mms = CmdPrompt (replaceLast m t) (ms ++ [m])
+  | (m : ms) <- mms = CmdPrompt (replacementFunc m t) (ms ++ [m])
+  | hasOpenQuotes t = case strMatches of
+    [] -> CmdPrompt t []
+    [m] -> CmdPrompt (completeWith lastQuotedString m) []
+    (m : ms) -> CmdPrompt (completeWith lastQuotedString m) (ms ++ [m])
```
xsebek, line 795: I forgot what the idea was with this completing code.

I would expect it to cycle on the suggestions, which makes me wonder why the "one element" and "more element" branches can not be the same.

kostmo: I've added explanatory comments.

kostmo:
Code is now cleaned up.  Will merge if I get an "Approve".

byorgey (approved):
This looks great, and thanks for adding the detailed comments!

src/Swarm/TUI/Controller.hs
```diff
@@ -783,23 +783,64 @@ handleREPLEvent = \case
...
+  (replacementCandidates, replacementBoundaryPredicate) = case completionType of
+    EntityName -> (entityNames, (/= '"'))
+    FunctionName -> (possibleWords, isIdentChar)
+
   names = s ^.. gameState . baseRobot . robotContext . defTypes . to assocs . traverse . _1
   possibleWords = reservedWords ++ names
-  matches = filter (lastWord `T.isPrefixOf`) possibleWords
+
+  theEntityMap = s ^. gameState . entityMap
+  -- Enhancement idea: Filter entities to those applicable to the preceding function name
```
xsebek, line 842: This comment looked to me like you implemented it, but I think you meant it as `TODO`?

Let's just remove it and not bother with it, there are only a few entities anyway. 😉

PR #801, opened 2022-10-29
Make REPLPanel not blink when it is not selected

### LINKED ISSUE #776: REPL cursor should stop blinking when focus is elsewhere
It also has been bothering me that the cursor continues to blink in the repl pane when focus is in another pane.

_Originally posted by @fryguybob in https://github.com/swarm-game/swarm/pull/151#discussion_r720825641_

xsebek:
- closes #776


xsebek:
@byorgey and @kostmo, feel free to have a look. 🙂 I would in particular be interested in better naming and further refactoring.

I might have introduced some bugs while porting the code from `Form` to `Editor`, but they should be easy to fix. 🐛 
Likely I have just left searching on instead of switching to command mode or something.

xsebek:
Well never mind, it looks like I broke both searching and completion. 😆

src/Swarm/TUI/Model.hs
```diff
@@ -1009,6 +977,19 @@ nextScenario = \case
           else BL.listSelectedElement nextMenuList >>= preview _SISingle . snd
   _ -> Nothing
 
+-- | Context for the REPL commands to execute in. Contains the base
+--   robot context plus the `it` variable that refer to the previously
+--   computed values. (Note that `it{n}` variables are set in the
+--   base robot context; we only set `it` here because it's so transient)
+topContext :: AppState -> RobotContext
+topContext s = ctxPossiblyWithIt
```
xsebek, line 985: I just moved this from controller to model, where it belongs. 😉

xsebek:
This also addresses some of the refactoring suggestions from 23817a5 by @lsmor.

In particular, it moves the REPL parts to one data type.

byorgey:
Everything seems to work, except that #750 introduced the ability to click in the middle of the REPL input to move the cursor there (#470) and that no longer seems to do anything.

byorgey:
Completion also seems to be broken.  Once you complete something, the completion queue persists even when you erase what you were typing and type something else.  For example, try typing `"t` at the REPL and then hitting Tab; it should expand to `"tank treads` as expected.  But now erase everything at the prompt, type `m`, and hit Tab. Instead of expanding to something like `move` it expands to `teeter-totter`.

xsebek:
Thanks, @byorgey! I was not even aware that clicking worked in REPL, I guess that is the only reason we need `REPLInput` as a special `Name` as otherwise it just acts as part of `REPLPanel`.

I wish we had some mock tests for this, that would make finding that bug faster. 😩

xsebek:
@byorgey should be fixed now 😉

src/Swarm/TUI/Controller.hs
```diff
@@ -282,7 +282,7 @@ handleMainEvent ev = do
         WorldPanel -> do
           mouseCoordsM <- Brick.zoom gameState (mouseLocToWorldCoords mouseLoc)
           uiState . uiWorldCursor .= mouseCoordsM
-        REPLInput -> handleREPLEvent ev
+        REPLPanel -> handleREPLEvent ev
```
byorgey, line 285: So this still isn't a perfect solution: now, clicking anywhere on the entire REPL panel can cause the cursor to jump around inside the input.  To see this, type some long input, click on a different panel, and then click somewhere inside the REPL panel on the left side, but not on the line of input.  The cursor will jump to somewhere in the middle of the input --- and the position where it jumps doesn't even correspond exactly with the x-coordinate of where you click.

I was thinking of keeping `REPLInput` and simply wrapping the REPL editor widget in `clickable REPLInput` --- does that not work?

src/Swarm/TUI/Controller.hs
```diff
@@ -837,55 +828,65 @@ tabComplete s (CmdPrompt t mms)
...
-  getCurrEntry = fromMaybe replLast . getCurrentItemText . view repl
-  oldEntry = getCurrEntry s
-  newEntry = getCurrEntry ns
+  moveREPL :: REPLState -> REPLState
+  moveREPL repl =
+    newREPL
+      & (if replIndexIsAtInput (repl ^. replHistory) then saveLastEntry else id)
+      & (if oldEntry /= newEntry then showNewEntry else id)
+   where
+    -- & validateREPLForm
```
byorgey, line 878: Can we get rid of this stray commented code now?

src/Swarm/TUI/Model.hs
```diff
@@ -387,48 +382,74 @@ lastEntry t h =
...
+-- | The way we interpret text typed by the player in the REPL prompt.
+replPromptType :: Lens' REPLState REPLPrompt
+
+-- | The prompt where the user can type input at the REPL.
+replPromptEditor :: Lens' REPLState (Editor Text Name)
+
+-- | Convinience lens to get text from editor and replace it with new
+--   one that has the provided text.
+--
+-- TODO: Editor uses a list of Text meaning something? It does not seem to be lines or cursor position.
```
byorgey, line 432: What's the status of this TODO?

It looks like `getEditContents` returns the result of `getText` from the `text-zipper` package, which in turn seems to have something to do with returning the part before the focus, the focus, and the part after the focus.  But it looks like just concatenating it is probably fine.

byorgey (approved):
I think this is looking great.  I still think clicking in the REPL is not perfect, and I *hope* it is easy to fix, but I haven't tried yet.  If you're sick of this and just want to merge I can play with it in a separate PR. :smile:

src/Swarm/TUI/Model.hs
xsebek: I think they may be lines but with the `'\n'` character. Concatenating is the only thing that works.

I will remove the comment and we can investigate it more when implementing multiline REPL.

src/Swarm/TUI/View.hs
```diff
@@ -1149,21 +1149,39 @@ drawLogEntry addName e = withAttr (colorLogs e) . txtWrapWith indent2 $ if addNa
...
+      | otherwise -> txt $ "[found: \"" <> lastentry <> "\"] "
+
+renderREPLPrompt :: FocusRing Name -> REPLState -> Widget Name
+renderREPLPrompt focus repl = ps1 <+> replE
+ where
+  prompt = repl ^. replPromptType
+  replEditor = repl ^. replPromptEditor
+  color = if repl ^. replValid then id else withAttr redAttr
+  ps1 = replPromptAsWidget (T.concat $ getEditContents replEditor) prompt
+  replE = withFocusRing focus (renderEditor (color . vBox . map txt)) replEditor
```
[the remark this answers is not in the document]
xsebek: It seems to work with my fix, so this may be unnecessary. 🙂

xsebek:
@byorgey I am indeed quite tired of this but it all seems to work now. 🤞 

If you find the clicking does not work as expected, then feel free to improve it in another PR. 🙂

PR #802, opened 2022-10-29
Hide robots for two seconds

### LINKED ISSUE #761: A way to "peek" under the robots
At least from the tutorials, it's not clear whether it's possible to see what's "underneath" a robot without moving the robot.  It is desirable to see underneath the robot especially for plants like flowers or trees that change their state over time.

Perhaps we should be able to hold down some keyboard combo to "peek underneath" (hide the robots) while the keys are depressed.  Or just have the key combo *toggle* robot visibility.  Or when the key combo is pressed, make the robots invisible for N seconds.  While the robot is hidden, can re-color the background of the tile that the robot is on.

* Provide the REPL command to see what is underneath the robot
* Continuously "blink" the robot when it is atop something

**Additional context**
This is from the `main` branch (commit `b895668`), compiled and run via `stack run`.

xsebek:
- closes #761


src/Swarm/TUI/View.hs
```diff
@@ -785,6 +787,7 @@ drawKeyMenu s =
       , Just (NoHighlight, "^p", if isPaused then "unpause" else "pause")
       , Just (NoHighlight, "^o", "step")
       , Just (NoHighlight, "^zx", "speed")
+      , Just (if s ^. uiState . uiShowRobots then NoHighlight else Alert, "M-h", "hide robots")
```
xsebek, line 790: I am not sure if it is important enough to show in the UI (it is also in help) but it was useful while debugging. 😁

src/Swarm/TUI/Controller.hs
```diff
@@ -261,7 +261,8 @@ handleMainEvent ev = do
     MetaKey 'h' -> do
       t <- liftIO $ getTime Monotonic
       uiState . uiHideRobotsUntil .= t + TimeSpec 2 0 -- hide for two seconds
-      -- pausing and stepping
+      invalidateCacheEntry WorldCache
```
byorgey, line 264: @xsebek I wondered why it took up to a second or so for robots to disappear after I hit M-h.  Then I figured it out --- we try hard NOT to redraw the world on every event because it is slow.  So when the world actually needs to be redrawn we have to signal it explicitly.  I fixed it by calling `invalidateCacheEntry WorldCache` here.

I also realized we should do the same fix for Ctrl-v.  Don't know whether you want to just include that fix with this PR or put it in a separate one.

src/Swarm/TUI/View.hs
byorgey: I think it's fine.  We don't show some "obvious" keys in the UI (like keys for doing stuff at the REPL) but this one is definitely not obvious.

src/Swarm/TUI/View.hs
```diff
@@ -844,21 +848,25 @@ drawWorld g =
...
+displayLoc :: Bool -> GameState -> W.Coords -> Display
+displayLoc showRobots g coords =
+  sconcat $ terrain NE.:| entity <> robots
  where
+  terrain = terrainMap M.! toEnum (W.lookupTerrain coords (g ^. world))
+  entity = maybeToList (displayForEntity <$> W.lookupEntity coords (g ^. world))
+  robots =
+    if showRobots
+      then map (view robotDisplay) (robotsAtLocation (W.coordsToLoc coords) g)
+      else []
```
byorgey, line 868: Looks much cleaner than my original code, thanks. :smile:

byorgey (approved):
Looks good to me after the small fix I pushed.  Take a look and see what you think.

src/Swarm/TUI/Controller.hs
xsebek: Interesting, I originally had code to not change anything for one second.

Might as well include it again if some player decides to hold down the <kbd>H</kbd> key. 😅

src/Swarm/TUI/View.hs
xsebek: STAN shows me warnings about unsafe functions in HLS and in this case, it got me thinking about refactoring to use the `NonEmpty` list constructor instead of `fromList`.

I usually just ignore them (they are not intrusive) but here it worked nicely. 😁

PR #805, opened 2022-10-29
Disambiguate choice of base robot

### LINKED ISSUE #790: Order of robots on map determines base
I've constructed three minimum reproducible test cases to demonstrate, with a `2x1`-sized map.
In the `palette`, I've mapped the character `b` to what is supposed to be the "base" robot, and the character `a` to a second robot.
In the screenshots, the contents of the "F2" menu is displayed.

Differences:
* The only difference in the scenario file between the **first** and **second** column is the `system: true` property added to the robot named "fake" (`a` in the palette).
* The only difference in the scenario file between the **second** and **third** column is the positioning of the robots in the map.

| Two normal robots | One system robot, to the right of base robot | One system robot, to the left of base robot |
| ---| --- | --- |
| Screenshot from 2022-10-24 23-00-48 | Screenshot from 2022-10-24 23-11-31 | Screenshot from 2022-10-24 23-13-35 |
| [code](https://gist.github.com/kostmo/96ab073e9f626da2ec67d54ff3adcf40) | [code](https://gist.github.com/kostmo/ba0c8ecc6efc30c61aae36c8d4d66036)  | [code](https://gist.github.com/kostmo/4e229b76ff8a283fad56a1e2467a694c) |
| Map layout: `ba` | Map layout: `ba` | Map layout: `ab` |

In the third column, I would have expected the system robot to not be listed in the "F2" menu.

kostmo:
When none of the robot definitions are assigned explicit coordinates (i.e. all of the robots are specified via the map and palette), this allows scenario authors to determine the "base" robot via the order of definitions as they exist in the scenario file.

A comment detailing the base selection logic has also been added.

closes #790

### DIFF
--- a/data/scenarios/README.md
+++ b/data/scenarios/README.md
@@ -271,6 +271,19 @@ table.
 | `system`      | `False`  | `boolean`             | Whether the robot is a "system" robot.  System robots can do anything, without regard for devices and capabilities. System robots are invisible by default.                                                                                                                                                                                                                                                                                                            |
 | `heavy`       | `False`  | `boolean`             | Whether the robot is heavy.  Heavy robots require `tank treads` to `move` (rather than just `treads` for other robots).                                                                                                                                                                                                                                                                                                        |
 
+#### Base robot
+
+There must be at most one **base** robot in the world. Since concrete robots can be created
+either via the `loc` attribute or via the map and palette, use the following guide to
+ensure the base robot is the one you intended:
+
+1. Always list the intended **base** as the first robot definition in your scenario.
+2. The first robot with a `loc` attribute will become the base, even if other robots are defined earlier.
+3. Without any located robots, if multiple robots are instantiated on the map from
+   the first robot definition, the first robot in
+   [row-major order](https://en.wikipedia.org/wiki/Row-_and_column-major_order)
+   shall be the base.
+
 ### Objectives
 
 The top-level `objectives` field contains a list of objectives that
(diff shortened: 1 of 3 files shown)


data/scenarios/README.md
```diff
@@ -271,6 +271,26 @@ table.
...
+       Scenario file.
+    2. Note that if a robot is both given a `loc` AND is specified in the world map, then
+       two instances of the robot shall be created.  The instance with the `loc` shall be
+       preferred as the base.
+2. Fall back to robots generated from templates via the map and palette.
+    1. If multiple robots are specified in the map, prefer the one that is defined first
+       within the Scenario file.
+    2. If multiple robots are instantiated from the *same* template, then prefer
+       the one closest to the upper-left of the screen,
+       with higher rows given precedence over columns.
```
xsebek, line 292: Please make this shorter.

Only a few people will read this long description about which robot will be chosen as the base.
It will also be annoying to keep up-to-date.

I would suggest not explaining the behaviour in such detail but instead giving an example that will make the robot base for sure. From what I understand, there are two strategies:
1. when the robot is first with a location in the list, it is base for sure
2. when no robots have a location, base is the leftmost topmost on the map

When there are more robots in the topmost leftmost cell, base is implementation-defined.

xsebek (approved):
Thanks, @kostmo! 👍 

I think we will eventually have to rewrite the logic to support robot IDs, but this works for now.

data/scenarios/README.md
kostmo: Ok, I've shortened the user-facing text, and relegated the verbose text to a comment in the code.

PR #806, opened 2022-10-31
Fix lodestone test

xsebek:
Turns out the test was not actually blocked by #540, I just did not investigate it properly.


byorgey (approved):
:tada:

PR #808, opened 2022-10-31
set focus to REPL panel when clicking on REPL input

byorgey:
Ensure that focus changes to the REPL panel also when clicking on the REPL input itself.

Cleaning up a small leftover issue from #801.


xsebek (approved):
👍

PR #810, opened 2022-11-01
Fix listening logic and ignore messages by self

### LINKED ISSUE #763: Robots should not be able to hear what they `say` with `listen`
In #756 we realized that if a robot executes `say` and then immediately executes `listen`, it will hear the thing it just said, which seems silly.  I mean, sure, technically you can hear what you yourself say, but it seems sensible that `listen` would "filter it out" and only listen for things coming from other robots.

> We should _definitely_ change `listen`, it is absurd that you would hear the same thing over and over again!
Let's do that in a separate issue and discuss how exactly it should work. :confused:

_Originally posted by @xsebek in https://github.com/swarm-game/swarm/issues/756#issuecomment-1283847374_

xsebek:
- add predicate checking if the message was produced by this robot
- compose the predicates correctly (system robots ignored time)
- closes #763

### DIFF
--- a/data/scenarios/Tutorials/backstory.yaml
+++ b/data/scenarios/Tutorials/backstory.yaml
@@ -38,7 +38,7 @@ objectives:
         as l {has "READY"}
       } { return false }
 solution: |
-  say "I said something wrong."; say "Ready!"
+  say "Ready!"
 entities:
   - name: READY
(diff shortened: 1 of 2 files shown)


byorgey (approved):
LGTM!

PR #812, opened 2022-11-01
Update editors with void and tweak VSCode coloring

xsebek:


xsebek:
Once this is merged, tag a VSCode plugin release. 🙂

byorgey:
Actually, it looks like both editors are missing `unit` as well...

PR #817, opened 2022-11-02
copy parent robot context to child when executing `build`

### LINKED ISSUE #394: Inner build does not correctly require Drill

byorgey:
Fixes #394.

### DIFF
--- a/data/scenarios/Testing/394-build-drill.yaml
+++ b/data/scenarios/Testing/394-build-drill.yaml
@@ -18,7 +18,14 @@ solution: |
   r <- build {
     wait 2;
     log "Hi, I am builder";
-    forever (build {log "Hi, I am pusher"; turn forward; forever push}; log "- robot built")
+    forever (
+      build {
+        log "Hi, I am pusher";
+        require "drill";          // #540
+        forever push
+      };
+      log "- robot built"
+    )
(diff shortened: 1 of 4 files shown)


xsebek (approved):
Wow, I am surprised that the robots even worked. 🤯

byorgey:
I was too at first, but I posted an explanation over at #394.

Issue #683, opened 2022-09-13
I/O console


byorgey:
My original motivation was contemplating the idea of sending robots into places you can't see (off the screen, into structures #112 or sub-worlds #144 ...) and programming them to wander around, build a map of the area (probably using some kind of mutable array #98 ), then come back and give you the map.  How would you actually view the resulting map?

I propose having a "virtual I/O console" (perhaps in a different "tab" in the same window as the world view, so you can switch back and forth between them, or perhaps in the same window as the REPL) and adding some commands to output to the console.  One might think of adding built-in commands like `putStr`, but actually, since I'd like to keep the number of new primitives to a minimum, and because I want to be able to do cool things like draw color maps, we might contemplate going even lower-level like a command to set the color and a command to put a character at a specific (row,column) position.  Then you could build your own functions like `putStr` on top of that.

I am not sure how reading keyboard input would work, or whether we should, but doing output to the virtual console seems like a good place to start.

xsebek:
To tie it back to the map idea, maybe you could "place entities" on some (finite) canvas?

You would not actually put the entity there, just print the entity character with its attributes.

So the resulting command could be `print :: entity -> (int*int) -> entity -> cmd ()` where the `entity` type is from #455 and the third argument would be the particular "map canvas" in inventory.

byorgey:
The "canvas" idea is cool, but it seems like that would require doing #455 first; at the current rate I don't know how soon that will actually happen.  For now it seems like it could be a relatively straightforward, "low hanging fruit"-type task to add a couple commands to print to the virtual console and then make a way to visualize it in the UI.

Issue #558, opened 2022-07-14
Scenario editor


byorgey:
It would be super cool to have an in-game graphical scenario editor, where you could move a cursor around, draw terrain and entities, place robots, create custom entities and code, etc...

Of course the scenarios would still be stored in `.yaml` files, so creating them by hand will always be an option.

This would obviously be a massive undertaking, and I don't think it needs to happen any time soon, but I just thought of it today so writing it down!

xsebek:
Selecting a cell with right click and then drawing by holding left mouse button would be pretty sweet. 🎨

xsebek:
For inspiration, [asciiflow](https://asciiflow.com/) has a simple editor with some line drawing logic that could be useful for mazes.

byorgey:
Wait, I just had an epiphany --- we don't need to make a scenario editor, because you should be able to just *program your own*!  A few ingredients would go into this:

- #102 would let you move a robot around with arrow keys, program your own hotkeys for placing certain items, drawing certain structures, etc., and of course you could always 
- We should add a built-in scenario that starts with a *blank* world
- There needs to be a way to save the current world in the format expected by a scenario description (related to #50).

byorgey:
Well, OK, being able to draw stuff with the mouse would still be pretty cool, too.  But maybe we can think of an in-game way to do that---like some device that lets you receive mouse events and respond to them somehow, similar to #102 .  If we also had the ability to teleport entities in creative mode, you could then write a little program that responds to mouse events by teleporting certain entities to the proper coordinates.

byorgey:
The overarching principle here is that this is a *programming* game --- instead of building in more and more cool features into the game itself, we should just build in programming language features that let players build the cool game features they want. (It even already says this in [DESIGN.md](https://github.com/swarm-game/swarm/blob/main/DESIGN.md),  believe it or not. :smile: )

xsebek:
That is quite mind blowing and I love it! 🤯 Making mouse events an upgrade sounds good to me, I guess you just get coordinates and a boolean. 🤔 

Making saves compatible with scenarios would be great and this interactive workflow could really save time.

byorgey:
> Making mouse events an upgrade sounds good to me, I guess you just get coordinates and a boolean.

I guess it depends how sophisticated you want to be.  Many real GUI frameworks differentiate between more than two kinds of mouse events, e.g. mouse down, mouse up, mouse move, etc.

xsebek:
[...]

Issue #102, opened 2021-09-28
Input handler command


byorgey:
I don't know yet exactly how it should work, but I think it's a really fun idea.  There should be a command something like
```
handleInput : (key -> maybe (cmd ())) -> cmd ()
```
where `key` is some sort of type to represent keypress events, the idea being that you specify how to handle each possible key press (with `nothing` indicating that the function should quit).  If you run `handleInput` at the REPL, it is now listening directly to any keypresses you make while the REPL is focused.  There are lots of fun uses you could imagine for this.

* You could create a set of "hotkeys" where you can press a single key to do various common actions.
* If you are eventually able to install some `treads` on your base, then you could drive your base around manually using arrow keys!
* Or you could remotely pilot another robot by broadcasting instructions to it!  Or you could remotely pilot a whole swarm of robots!

I don't know what happens if a robot other than the base tries to execute this command.  I guess it just sits there waiting for input forever, since you are not sitting there to push any keys.

byorgey:
I guess this would require:

1. Adding a built-in primitive `key` type with lots of built-in constants to represent possible keypresses.  I guess this should be auto-generated somehow from the list of X11 keysyms or something like that.
2. Writing code to interpret `handleInput` by putting the REPL into a special state that just listens for keypresses and sends them directly to the provided callback.  This would be the easy part, I think.
3. We would also have to think carefully about what facilities we would provide for writing a function of type `key -> ...`.  Right now we have sum types and the `case` function.  But writing a handler for lots of different keypresses by nesting `case`s would be horrendous.  I guess there would need to be some kind of pattern-matching or at the very least a kind of built-in `switch` expression.

byorgey:
Maybe we don't need special syntax for `switch`, pattern-matching, etc. after all --- users could just make their own function like  `lookup : list (a * b) -> (a -> maybe b)` and then use that to write a list of key/cmd pairs.

xsebek:
I like that, we could just put a flag to game state that we would check in input handler. 

I think we should still handle a few global inputs, like Ctrl-Q, Ctrl-C or time control otherwise we could freeze the base. Luckily that’s easy to do, we’ll just put the flag check on the right line in input handler.

byorgey:
So let me try to summarize the current design in one place:

* Create a new type `key` to represent keypresses
[...]

PR #819, opened 2022-11-03
Add a "driving mode" to the REPL

kostmo:
Ctrl+d in the REPL will toggle into "driving mode".  In this mode, the arrow keys simply submit predefined commands to the REPL:
| Key | REPL input |
| --- | --- |
| ↑ | `move;` |
| ↓ | `turn back;` |
| ← | `turn left;` |
| → | `turn right;` |

This yields a pretty natural driving experience:

<a href="https://asciinema.org/a/pZQ7wexMllA5S3v9VrsrdjtsN" target="_blank"></a>

### DIFF
--- a/src/Swarm/TUI/Controller.hs
+++ b/src/Swarm/TUI/Controller.hs
@@ -93,10 +93,13 @@ import Witch (into)
 pattern Key :: V.Key -> BrickEvent n e
 pattern Key k = VtyEvent (V.EvKey k [])
 
-pattern CharKey, ControlKey, MetaKey :: Char -> BrickEvent n e
+pattern CharKey, ControlChar, MetaChar :: Char -> BrickEvent n e
 pattern CharKey c = VtyEvent (V.EvKey (V.KChar c) [])
-pattern ControlKey c = VtyEvent (V.EvKey (V.KChar c) [V.MCtrl])
-pattern MetaKey c = VtyEvent (V.EvKey (V.KChar c) [V.MMeta])
+pattern ControlChar c = VtyEvent (V.EvKey (V.KChar c) [V.MCtrl])
+pattern MetaChar c = VtyEvent (V.EvKey (V.KChar c) [V.MMeta])
+
+pattern ShiftKey :: V.Key -> BrickEvent n e
+pattern ShiftKey k = VtyEvent (V.EvKey k [V.MShift])
 
 pattern EscapeKey :: BrickEvent n e
 pattern EscapeKey = VtyEvent (V.EvKey V.KEsc [])
@@ -168,7 +171,7 @@ handleMainMenuEvent menu = \case
         About -> uiState . uiMenu .= AboutMenu
         Quit -> halt
   CharKey 'q' -> halt
-  ControlKey 'q' -> halt
+  ControlChar 'q' -> halt
   VtyEvent ev -> do
     menu' <- nestEventM' menu (handleListEvent ev)
     uiState . uiMenu .= MainMenu menu'
@@ -187,7 +190,7 @@ handleMainMessagesEvent :: BrickEvent Name AppEvent -> EventM Name AppState ()
 handleMainMessagesEvent = \case
   Key V.KEsc -> returnToMainMenu
   CharKey 'q' -> returnToMainMenu
-  ControlKey 'q' -> returnToMainMenu
+  ControlChar 'q' -> returnToMainMenu
   _ -> return ()
  where
   returnToMainMenu = uiState . uiMenu .= MainMenu (mainMenu Messages)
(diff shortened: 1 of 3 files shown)


kostmo:
Things to improve:
* The keyboard shortcut list above the REPL pane is quite crowded.  Should think of re-organizing it.
* In driving mode, another row should be displayed below the keyboard shortcut list, explaining the arrow key meanings.

src/Swarm/TUI/Controller.hs
```diff
@@ -719,7 +719,38 @@ resetREPL t r ui =
 
 -- | Handle a user input event for the REPL.
 handleREPLEvent :: BrickEvent Name AppEvent -> EventM Name AppState ()
-handleREPLEvent = \case
+handleREPLEvent x = do
+  s <- get
+  let controlMode = s ^. uiState . uiREPL . replControlMode
+  case x of
+    ControlKey 'd' -> do
```
xsebek, line 726: We already use Control D for quit (when the REPL is empty) to make it easier for players coming from shell. How about Meta D?

xsebek (changes requested):
Thanks @kostmo, it looks great and I will try it out soon! 👍

AFAIK we planned to add event handlers to the language that would be more general. Then you could use keys for grabbing etc.

But I like how your approach shows the commands, this system looks very user friendly and we could still do both. 🙂

src/Swarm/TUI/Controller.hs
kostmo: I've renamed the mode from "driving" to "piloting".  Changed the mode switch key to `Meta-p`.  Added other commands like grab, harvest.  Now is only available via Creative mode:

1. Run in `cheat mode`
    stack run -- --scenario data/scenarios/Tutorials/move.yaml --cheat
2. Then type Ctrl+v to enter creative mode
3. Then type Meta+p to enter piloting mode

byorgey:
I love the idea of driving robots around, and I tried this out and it works well.  It is a little annoying trying to use it in a Creative game though, because the base always looks like `Ω` no matter which way it is facing so you have to keep track of which way it is facing in your head.  I see the appeal of having Left/Right arrow keys correspond to `turn left/right` and Up correspond to `move`, but for driving around something that always looks like `Ω` it might work better for e.g. Left to correspond to `turn west; move` and so on.  I'm not necessarily advocating that particular solution, just pointing out that it probably needs to be addressed somehow for this to be usable beyond solving early tutorial challenges more easily.

---

In general, I do want to keep this game focused on *programming*; the relevant design principle is that instead of adding complex built-in features, we should strive instead to provide language constructs that allow players to program the cool features themselves.  For example, see #102 for a previously discussed feature which would allow players to program their own driving mode.  See also #558 for another discussion of this design principle.  However, I am not against including this PR, especially if it is limited to creative mode.  In fact I guess there could be something fun about seeing a feature like driving mode that's only available in cheat/creative mode, then much later realizing that you can actually program it yourself in classic mode!

---

There are some things I like about this that make me think of improvements to #102, or more general features we could provide in the spirit of the design principle I mentioned above.  This is just me brainstorming and should probably eventually be moved into new issues etc.
- I like having a key shortcut to toggle pilot mode off/on.  Maybe we could have a dedicated key shortcut to toggle the most recently installed input handler on/off.  Then you couldn't get stuck running an input handler where you forgot to program in a way to exit, and you could easily toggle it off temporarily to run some other commands etc., then toggle it back on without having to type in a whole command to run it again.
- Maybe you should be able to programmatically choose what is shown in the key shortcut menu, so that when e.g. running an input handler you could show help to remind yourself (or anyone else who wants to use your input handler) of the available keys.
- Heck, maybe the _whole UI should be programmable_!!?  Like there would be some standard built-in commands corresponding to the usual content of the UI, but you could (1) program a different UI layout using those standard built-in commands, or even (2) program your own completely new UI widgets showing whatever info you want.

xsebek (approved):
I tried it and it's great! 👍 I imagine it will at least make testing challenges easier. 🙂 

I generally agree with @byorgey's comments here, but I think we can fix the mentioned issues separately. Meanwhile, this adds a lot of value to the `--cheat` mode. 😁 

Feel free to merge @kostmo unless you want to improve anything still.

xsebek:
Btw. I do not think `turn west; move` is a good idea. Once you can move and turn the base with tank treads, the direction issue will come up, so making a hack for it here is unnecessary.

We used to have the base orientation map with normal characters for absolute directions. The trick was that the base was facing down by default which sometimes made other robots look like the base. We can now customize this through scenarios or allow players to set it themselves. 🤷 

But let's _not_ do that here, we should aspire to have small commits. 😄

src/Swarm/TUI/Controller.hs
```diff
@@ -719,7 +719,46 @@ resetREPL t r ui =
 
 -- | Handle a user input event for the REPL.
 handleREPLEvent :: BrickEvent Name AppEvent -> EventM Name AppState ()
-handleREPLEvent = \case
+handleREPLEvent x = do
+  s <- get
+  let controlMode = s ^. uiState . uiREPL . replControlMode
+  case x of
+    MetaKey 'p' ->
```
xsebek, line 726: Slight tweak: let's add an empty REPL check so that you do not accidentally overwrite the command you are writing.

We could also add it to the `drawKeyMenu` so that the players know it is not available while writing.

Just a suggestion though, if you think it is fine, feel free to leave it as is @kostmo. 😉

kostmo:
Related: #683

Issue #625, opened 2022-08-06
Implement run command for the web service 


TristanCacqueray:
**Is your feature request related to a problem? Please describe.**
When playing with a text editor, it is not convenient to use the REPL to load new definitions.

**Describe the solution you'd like**
I would like to load swarm code using the web service directly from my text editor.
The editor integration could provide a `send-buffer` or something equivalent to automate that process.

xsebek:
That sounds cool! Should this only accept definitions or also scripts?

I can imagine wanting to restart the scenario when writing a script to solve a maze, but only loading new definitions when playing the long game. :thinking:

byorgey:
In swarm-lang there is no difference between definitions and scripts; a sequence of definitions *is* a script.  I don't think there's any reasonable way to restrict something to only accept definitions.

xsebek:
@byorgey I was thinking that we could easily restrict that if you for example wanted to reload as you type, but not run any commands.

But that might not be a problem or could be offered as a config option later.

xsebek:
@TristanCacqueray how would this work on the application state though?

The current Web API only has access to GameState, would this be a `Post '[Text] ()` type of endpoint where the text is parsed as `Term` and run by the base robot?

The alternative I see is adding some generic channel to the Web service, through which it would pass a `Term` and receive the `Value` once that term has finished executing. In the current brick application, we could route this through an event, so that it can be added to the `UIState` REPL history. But in principle, we could cut the brick part and run Swarm on a server, with the run term being executed in the base robot directly.

Honestly, I am really intimidated by the Servant types, so if you could add at least a minimal scaffolding for one POST method with error handling, that would make it much easier. 😉

TristanCacqueray:
@xsebek I don't know how exactly how the text term would be submitted to the game state, but from the API perspective, the handler would look like this:

```haskell
type SwarmAPI =
  ... 
  :<|> "run" :> ReqBody '[JSON] Text :> Post '[JSON] Text

runHandler userSuppliedCommand = do
  term <- parse userSuppliedCommand
  eval term
  pure "completed"
```

- `ReqBody '[JSON] Text` indicates that the request will have a body of type `application/json` containing a text.
- `Post '[JSON] Text` indicate the path returns a text encoded in json.

The user would use the handler with `curl -XPOST -d'"the program"' -H 'Content-Type: application/json' localhost/run`

PR #821, opened 2022-11-04
Expose REPL history via web interface

kostmo:
This makes it possible to dump the commands typed into the REPL to a text file, which makes it easier to start writing an external `.sw` script based on experiments written in the REPL.

E.g.

    curl http://localhost:5357/repl/history/full | jq .[] -r

This is related to (and may be a stepping stone towards) #625.

### DIFF
--- a/src/Swarm/Web.hs
+++ b/src/Swarm/Web.hs
@@ -25,41 +25,51 @@ import Control.Exception (Exception (displayException), IOException, catch, thro
 import Control.Lens ((^.))
 import Control.Monad (void)
 import Control.Monad.IO.Class (liftIO)
+import Data.Foldable (toList)
 import Data.IORef (IORef, readIORef)
 import Data.IntMap qualified as IM
 import Data.Maybe (fromMaybe)
+import Data.Text qualified as T
 import Network.Wai qualified
 import Network.Wai.Handler.Warp qualified as Warp
 import Servant
 import Swarm.Game.Robot
 import Swarm.Game.State
+import Swarm.TUI.Model
 import System.Timeout (timeout)
 
 type SwarmApi =
   "robots" :> Get '[JSON] [Robot]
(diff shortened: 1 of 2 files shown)


src/Swarm/Web.hs
```diff
@@ -68,19 +79,21 @@ webMain baton port gsRef = catch (Warp.runSettings settings app) handleErr
 defaultPort :: Warp.Port
 defaultPort = 5357
 
+data StateRefs = StateRefs (IORef GameState) (IORef UIState)
```
byorgey, line 82: What's the benefit of having two separate IORefs?  Why not just have a single `IORef  AppState`?

kostmo: Good point, will do.

src/Swarm/Web.hs
```diff
@@ -25,41 +25,51 @@ import Control.Exception (Exception (displayException), IOException, catch, thro
...
 import Servant
 import Swarm.Game.Robot
 import Swarm.Game.State
+import Swarm.TUI.Model
 import System.Timeout (timeout)
 
 type SwarmApi =
   "robots" :> Get '[JSON] [Robot]
     :<|> "robot" :> Capture "id" Int :> Get '[JSON] (Maybe Robot)
+    :<|> "repl" :> Get '[JSON] [T.Text]
```
xsebek, line 44: Maybe we should add a length (or start+end) parameter?

My history is quite long, so I imagine it might be useful to shorten it if the Swarm binary was not running on my machine.

Or alternatively, how about only showing the history from the current game (as we do in REPL) unless you explicitly request it?

xsebek (approved):
Good idea @kostmo! 👍 I have some suggestions for improvement, but I would not really object to merging this as is.

src/Swarm/Web.hs
kostmo: I've now made the path longer for the current output (`repl/history/full`) to future-proof for different varations/parameters.

PR #822, opened 2022-11-04
use two rows for keyboard commands list

kostmo:
There are too many keyboard commands to list on one line, so they get cut off even with a very wide terminal.  This PR splits panel-specific commands into a separate row from global commands.

| Version | Screenshot |
| --- | --- |
| Before: | Screenshot from 2022-11-04 01-30-49 |
| After: | Screenshot from 2022-11-04 01-30-08 |


byorgey (approved):
Looks good to me!

PR #824, opened 2022-11-04
Ctrl+C should clear the REPL text

kostmo:
Currently there's no (obvious?) shortcut to clear the currently typed text in the REPL.

Mapping CTRL+c to this action would be familiar to terminal users.


xsebek (approved):
Ah, my bad. This is a regression that I caused while refactoring REPL.

We should really test this stuff. 😅

PR #825, opened 2022-11-04
Autogenerate web api docs

kostmo:
The docs are viewable from URL path that isn't already defined, e.g.:

    http://localhost:5357/

As an minor update to #821, the CLI invocation to dump REPL commands is now:

    curl http://localhost:5357/repl/history/full | jq .[].in -r

Implemented by following [this tutorial](https://docs.servant.dev/en/stable/tutorial/Docs.html).

Screenshot from 2023-04-02 12-40-38


src/Swarm/Util/Bench.hs
```diff
@@ -0,0 +1,34 @@
...
+import Swarm.Language.Pipeline (ProcessedTerm)
+import Swarm.Language.Pipeline.QQ (tmQ)
+import System.Clock (fromNanoSecs)
+
+-- | The program of a robot that does nothing.
+idleProgram :: ProcessedTerm
+idleProgram = [tmQ| {} |]
+
+fakeRobot :: Robot
+fakeRobot =
```
xsebek, line 20: Why do you need `Util.Benchmark` with an example robot to use in Web API?

Please feel welcome to add stuff to the modules where it belongs. 🙂 They are a bit long, but making new `Util` modules will not help us orient ourselves better in the code.

src/Swarm/Web.hs
```diff
@@ -25,24 +25,68 @@ import Control.Exception (Exception (displayException), IOException, catch, thro
...
+instance ToSample T.Text where
+  toSamples _ =
+    [ ("When foo", "foo")
+    , ("When bar", "bar")
+    ]
+
+swarmApi :: Proxy SwarmApi
+swarmApi = Proxy
+
+type DocsAPI = SwarmApi :<|> Raw
```
xsebek, line 74: That does not make sense to me. The "raw" API is not part of the "docs" API.

Is there a reason why an extra type is necessary?

xsebek (commented):
I am interested to see how this will turn out. 🙂 

Please just try to make this as simple as possible - I am not familiar with servant applications and adding extra levels of indirection will make it harder for me to contribute to and maintain the Web API.

src/Swarm/Util/Bench.hs
kostmo: I got rid of the samples for `Robot`, since they're not easily human-readable anyway.  So this change is gone.

src/Swarm/Web.hs
kostmo: Fixed the naming.

swarm.cabal
```diff
@@ -158,6 +160,7 @@ library
                       prettyprinter                 >= 1.7.0 && < 1.8,
                       random                        >= 1.2.0 && < 1.3,
                       servant                       >= 0.19 && < 0.20,
+                      servant-docs                  >= 0.12 && < 0.13,
```
kostmo, line 163: Not sure why the GHC 9.4 CI is failing, since `servant-docs` [added support for GHC 9.4 in October](https://github.com/haskell-servant/servant/pull/1592).

kostmo: @TristanCacqueray I see you were involved with bumping the upper bound for lens [here](https://github.com/haskell-servant/servant/pull/1614).  However, the lens bound for [`servant-docs` on Hackage is still `< 5.2`](https://hackage.haskell.org/package/servant-docs).  Do you know how that can be fixed?

TristanCacqueray: I think the maintainers need to do a new release or revision, e.g. https://hackage.haskell.org/package/servant-docs-0.12/revisions/ does not include the recent bump.

byorgey: https://github.com/haskell-servant/servant/issues/1636

byorgey: Looks like the maintainers just made a new `servant-docs` release.

kostmo: Yes, the build now succeeds.

kostmo:
@byorgey, shall we get this merged?  It at least accomplishes the objective of listing all of the endpoints.

stack.yaml
```diff
@@ -2,6 +2,7 @@ extra-deps:
 - fused-effects-lens-1.2.0.1@sha256:675fddf183215b6f3c1f2a0823359a648756435fd1966284e61830ec28ad61fa,1466
 - hsnoise-0.0.3@sha256:260b39175b8a3e3b1719ad3987b7d72a3fd7a0fa99be8639b91cf4dc3f1c8796,1476
 - simple-enumeration-0.2.1@sha256:8625b269c1650d3dd0e3887351c153049f4369853e0d525219e07480ea004b9f,1178
+- servant-docs-0.12@sha256:73fd8901a9dbe4d2c563b7da0b94c75f7755ee9059eec80ff621f415a8d3ef6a,3383
```
byorgey, line 5: Note that the specific revision of `servant-docs` identified here by hash has an upper bound of `base < 4.17` so it won't build with GHC 9.4 (which comes with our selected resolver).  However, the most recent revision of `servant-docs` on Hackage does work.  I was able to build simply by removing the `@sha256` part; I am not sure how to get `stack` to tell me the right hash to replace it with.

kostmo: A couple of things:
1. I'm not sure what's up with the `base` upper bound, but it builds locally for me with `stack` already:
    ```
    ~/github/swarm$ stack ghc -- --version
    The Glorious Glasgow Haskell Compilation System, version 9.4.4
    ~/github/swarm$ stack build --fast
    ~/github/swarm$ echo $?
    0
    ```
2. Our Cabal-based CI builds are all passing, but if there is an additional merge precondition of successful `stack` builds, we should add `stack` to CI.

byorgey: Weird.  This is what I get:
```
diophantus :: ~/projects/swarm » stack --version
Version 2.9.3, Git revision 6cf638947a863f49857f9cfbf72a38a48b183e7e x86_64 hpack-0.35.1
diophantus :: ~/projects/swarm » stack clean --full
diophantus :: ~/projects/swarm » stack ghc -- --version
The Glorious Glasgow Haskell Compilation System, version 9.4.4
diophantus :: ~/projects/swarm » stack build --fast

Error: [S-4804]
       Stack failed to construct a build plan.
       
       While constructing the build plan, Stack encountered the following errors:
       
       In the dependencies for servant-docs-0.12:
           base-4.17.0.0 from Stack configuration does not match >=4.9 && <4.17 (latest matching version is 4.16.4.0)
           lens-5.2 from Stack configuration does not match >=4.17 && <5.2 (latest matching version is 5.1.1)
       needed due to swarm-0.3.0.1 -> servant-docs-0.12
       
       Some different approaches to resolving this:
       
         * Set 'allow-newer: true' in /home/brent/.stack/config.yaml to ignore all version constraints and build anyway.
       
         * Build requires unattainable version of base. Since base is a part of GHC, you most likely need to use a different GHC version with the matching
           base.    
```

> 2\. if there is an additional merge precondition of successful `stack` builds, we should add `stack` to CI.

Yes, you're right.  I think the intention has always been to support building both with `cabal` and with `stack`, so our CI should test that.  I'll look into it.

byorgey (approved):

PR #827, opened 2022-11-05
Copy requirements map to robot context when loading a new `ProcessedTerm`

### LINKED ISSUE #540: Definition capabilities are not available later in the same term

byorgey:
Closes #540.

When we typecheck and capability check a new term, resulting in a `ProcessedTerm`, we already know the `Requirement`s for all the `def`s it contains.  This PR simply takes all those definition requirements and adds them to `robotContext . defReqs` just before installing the new `ProcessedTerm` in the CESK machine, so that if we ever encounter the name of any of the definitions inside an argument to `build` or `reprogram`---which we capability check at runtime---we will be able to properly match up names with their requirements.

This is a hack for several reasons:
- We really shouldn't be capability-checking arguments to `build` and `reprogram` at runtime in the first place---ideally we should be able to check everything statically.  But fixing that will require implementing #231 .
- We have to do this in three places in the code: when loading a new term into the base when the player hits Enter at the REPL; when executing `run`; and when running the solution from a scenario in the integration tests.  Ideally, there would be only one place, but I don't have a good idea at the moment how to refactor things properly.
- Technically, the names being defined shouldn't be in scope until after their corresponding `def` runs, so it's incorrect to dump them all into the `defReqs` prior to executing any of the `def`s.
    - However, doing it properly, with each name coming into scope with its requirements right after the `def` runs, is very tricky/annoying to implement for more reasons than I care to write about here (believe me, I tried, and it made my brain hurt).  For starters, we don't really have access to the robot state when stepping the CESK machine.
    - The only scenario where this would be a problem is if (1) there is already a name `x` defined, and then we (2) `run` a `.sw` file containing, first, a `build` command whose argument references `x`, and second, a redefinition of `x`.  In that case the `build` would incorrectly decide that the `x` in the argument to `build` has the requirements of the *second* `x` (the one that will be redefined later) even though the first `x` is the one that should still be in scope.  However, this seems like a very unlikely scenario (who would write a `.sw` file that depends on some specific name already being defined before it is run, but then redefines that same name later?) and I'd *much* rather have that obscure problem than the current very relevant and annoying one.

However, it's a simple hack that solves the issue and will be easy to get rid of later if and when we do something better.  I'm a big believer in doing things the right way, but in this instance I definitely think we should just do the simple thing that mostly works and then continue to make it better in the future.

### DIFF
--- a/data/scenarios/Challenges/Mazes/easy_cave_maze.yaml
+++ b/data/scenarios/Challenges/Mazes/easy_cave_maze.yaml
@@ -15,7 +15,6 @@ solution: |
   def until = \p. \c. b <- p; if b {} {c; until p c} end;
   def fwd = until blocked move end;
   build {
-    require "branch predictor"; // #540
     turn east;
     until (ishere "goal") (fwd; turn right; fwd; turn left);
     grab
--- a/data/scenarios/Challenges/Mazes/easy_spiral_maze.yaml
+++ b/data/scenarios/Challenges/Mazes/easy_spiral_maze.yaml
@@ -16,7 +16,6 @@ solution: |
   def until = \p. \c. b <- p; if b {} {c; until p c} end;
   def fwd = until blocked move end;
   build {
-    require "branch predictor"; // #540
     turn west;
     until (ishere "goal") (fwd; turn right);
     grab
--- a/data/scenarios/Challenges/Mazes/invisible_maze.yaml
+++ b/data/scenarios/Challenges/Mazes/invisible_maze.yaml
@@ -25,9 +25,6 @@ solution: |
   def rightBlocked = tR; b <- blocked; tL; return b end;
   def mazeStep = ifC rightBlocked {ifC blocked {tL} {move}} {tR; move} end;
   build {
-    require "treads"; require "branch predictor"; require "lambda";
-    require "strange loop"; require "scanner"; require "grabber";   // #540
-
     forever (mazeStep; ifC (ishere "goal") {grab; return ()} {})
   }
 robots:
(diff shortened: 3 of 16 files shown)


byorgey:
Hmm, the `lodestone` test is failing, not sure why yet.

xsebek (approved):
Kudos @byorgey! 👍

I would add that it will also be easier to optimize in the Web API future, so simple solutions now make sense. 😉

PR #830, opened 2022-11-05
Drills by capability

kostmo:
This PR consists of two commits:
1. Refactor list of capabilities to sets (continuation of #794)
2. Instead of hard-coding drill names, select equipped entities by their `drill` capability

This allows scenario authors to define their own drills.

### DIFF
--- a/src/Swarm/Game/Entity.hs
+++ b/src/Swarm/Game/Entity.hs
@@ -69,6 +69,7 @@ module Swarm.Game.Entity (
   isSubsetOf,
   isEmpty,
   inventoryCapabilities,
+  extantElemsWithCapability,
 
   -- ** Modification
   insert,
@@ -97,7 +98,7 @@ import Data.Map (Map)
 import Data.Map qualified as M
 import Data.Maybe (fromMaybe, isJust, listToMaybe)
 import Data.Set (Set)
-import Data.Set qualified as Set (fromList, toList, unions)
+import Data.Set qualified as Set (fromList, member, toList, unions)
 import Data.Text (Text)
 import Data.Text qualified as T
 import Data.Yaml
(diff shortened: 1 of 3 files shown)


src/Swarm/Game/Entity.hs
```diff
@@ -313,7 +313,7 @@ buildEntityMap :: [Entity] -> EntityMap
 buildEntityMap es =
   EntityMap
     { entitiesByName = M.fromList . map (view entityName &&& id) $ es
-    , entitiesByCap = M.fromListWith (<>) . concatMap (\e -> map (,[e]) (e ^. entityCapabilities)) $ es
+    , entitiesByCap = M.fromListWith (<>) . concatMap (\e -> map (,[e]) (Set.toList $ e ^. entityCapabilities)) $ es
```
byorgey, line 316: I wonder if `entitiesByCap` should change to have type `Map Capability (Set Entity)`?  I don't mean we should change it in this PR, it just made me wonder.

src/Swarm/Game/Entity.hs
```diff
@@ -567,7 +567,14 @@ isEmpty = all ((== 0) . fst) . elems
 
 -- | Compute the set of capabilities provided by the devices in an inventory.
 inventoryCapabilities :: Inventory -> Set Capability
-inventoryCapabilities = setOf (to elems . traverse . _2 . entityCapabilities . traverse)
+inventoryCapabilities = Set.unions . map (^. entityCapabilities) . nonzeroEntities
+
+nonzeroEntities :: Inventory -> [Entity]
+nonzeroEntities = map snd . filter ((> 0) . fst) . elems
+
+elemsWithCapability :: Capability -> Inventory -> [Entity]
```
byorgey, line 575: It would be good to add Haddock comments to `nonzeroEntities` and `elemsWithCapability`.  In particular the fact that `elemsWithCapability` only returns entities with a nonzero count is not obvious from just the name.

byorgey (approved):
Looks good to me.  The heuristic re: preferring drills with longer names makes sense for now.  Perhaps we will eventually have a more principled way to choose if we implement #480.

byorgey:
Now I'm curious to see what scenario with custom drill entities you have up your sleeve. :smile:

src/Swarm/Game/Entity.hs
kostmo: Added.

src/Swarm/Game/Entity.hs
```diff
@@ -567,7 +567,17 @@ isEmpty = all ((== 0) . fst) . elems
...
-inventoryCapabilities = setOf (to elems . traverse . _2 . entityCapabilities . traverse)
+inventoryCapabilities = Set.unions . map (^. entityCapabilities) . nonzeroEntities
+
+-- | List elements that have at least one copy in the inventory.
+nonzeroEntities :: Inventory -> [Entity]
+nonzeroEntities = map snd . filter ((> 0) . fst) . elems
+
+-- | List elements that possess a given Capability and
+-- exist with nonzero count in the inventory.
+extantElemsWithCapability :: Capability -> Inventory -> [Entity]
```
xsebek, line 578: Extend?

kostmo: [extant](https://www.merriam-webster.com/dictionary/extant): meaning "existing"

PR #833, opened 2022-11-05
fix local paths for new scenario scripts

kostmo:


byorgey:
@kostmo I've been trying out the new scenario and it seems really fun (aside from the fact that it doesn't work without fixing the paths of course).  But I've got some additional feedback that I think will improve it.  Rather than pushing a bunch of individual fixes, I think it probably makes more sense to revert the commit that added the new scenario, and then open a traditional PR where we can discuss and improve the scenario as a whole.  Let me know if you want to do that or if you'd like me to.

kostmo:
Agreed: #834 for reversion

kostmo:
New PR: #835

Issue #797, opened 2022-10-27
Achievements system


kostmo:
Creating this issue as a nucleation point for ideas around an achievements system.

* There should be a toplevel "trophy case" accessible from the main menu.
* Achievements should be definable within scenario files as some combination of goals.  #795 would probably be a prerequisite to make this work.
    * Goals would then probably also need to support some "identifier" to make them referenceable by the achievements definitions elsewhere in the file.
* Since scenario files may come and go, would probably represent the global achievements list as **hashes** of the distributed achievement definitions.
* In addition to scenario-level achievements, it would also be nice to earn achievements through either **non-scenario** or **cross-scenario** activities.  For example:
    * Completing your first tutorial
    * Completing *all* of the tutorials
    * Surpassing *N* hours of total game time
    * Exploring the complete menu of tutorials
    * Driving a robot into the water
    * (Attempting to) self-destruct your base

I recognize that peppering the game code with achievement-persistence functions has a maintainability cost, i.e. if it breaks encapsulation or purity of some part of the code.  So if it can't be done cleanly, we may want to avoid such achievements.

If the "Classic" game is considered the "canonical" way to play Swarm, perhaps some of these latter examples would just be defined in the `classic.yaml` file, and only achievable in "Classic" mode.

xsebek:
I love this idea! 👍 Let's make Swarm feel like a proper game! 🎮 💻 

We can definitely add some system-wide achievements like the ones you mentioned. If they are checked once at the end of the tick or even at the end of the game, they will likely not even be noticeable.

Some of them, like self-destructing, can be checked in one place in the step function. In this case, when we catch exceptions. In other cases, it will be "the first time you execute teleport" or something like that.

Checking custom achievements can use the same infrastructure as the objectives. Maybe they could even _be_ parallel _optional_ objectives marked as achievements? 🤔

---

> Since scenario files may come and go, would probably represent the global achievements list as hashes of the distributed achievement definitions.

Noooooooo! 🤣 We already have unique names of scenarios and _our players and developers are not attackers_. To me, this looks similar to how packages are distributed. They simply have unique names and a centralized repository from where you can download them.

This does not prevent distributed achievements, you can have anything you want on your machine and when sharing with others you just pick a unique name. Or just add a number to it, or underscore or change it a little.

I think everyone in this time and age knows the importance of picking a good name on a platform first. 😆 It is its own kind of game and using hashes takes away all the fun.

[...]

Issue #506, opened 2022-06-29
`robotsWhich` and other ways to get references to existing robots


byorgey:
This is a problem that has been talked about in various places. This is intended to be a sort of omnibus issue bringing together all the previous discussions and also offering a concrete proposal.

The problem
-----------------

There are lots of robots in the world, and sometimes you need to refer to them (in order to tell them to halt, to communicate with them, to `give` them something, to program another robot to `give` them something, to `view` them...).  The status quo is "if you ever want to refer to a robot in the future, or even think you might, you must bind it to a variable at the REPL; otherwise you're out of luck".  This is, of course, ridiculously limiting and annoying, for many reasons:
- You might forget to bind a robot to a name (I have done this many times)
- You might not realize that you will want to refer to a particular robot in the future
- You might want to call a function which builds multiple robots, or build a robot that builds other robots.  In both these cases, it is not even possible to bind the resulting robots to variables in the REPL.
- Even if you create robots one by one, and remember to bind them to variables, it is annoying to have to remember the right variable names for all the robots.  It would be much nicer to be able to just create them and then select the one you want to refer to later without having to remember the exact variable name you happened to assign it in the past.

We need a more general solution for referring to robots after the fact.

Related discussions/partial solutions
--------------------------------------------------

- #304 
    - This would help in the case that you execute `build`, and you wanted/intended to bind the result to a variable, but forgot.  However, it doesn't really solve the problem (though it's a nice idea on its own merits and we should definitely implement it).
- #60 
    - REPL results are now bound to variables like `itN`, but without being able to scroll the REPL panel you can only see the past few results.
- #306
- #462
    - This would help you refer to robots that you create at the REPL, whether or not you bind them to a variable.  However, it doesn't really solve the problem either.  For example, what if you call a function that `build`s multiple robots?  What if you build a robot that builds other robots?  In these cases you would never see the robot value printed at the REPL, so you would not know how to refer to them.  I actually now think we should *not* implement a robot registry as described in #462 , but something more general instead (see below).
- #343
    - My proposal in this issue would (I think) also solve this.  In the discussion I was worried about functions that would let you manufacture robot references out of thin air, but I think there is a way we can have these functions without being able to manufacture robot references you shouldn't; see below.
    - #405
    - #339 

What robots do you "know about"?
----------------------------------------------

[...]

Issue #306, opened 2022-02-16
Commands to meet other robots


byorgey:
This is an idea that came out of the discussion on #212 .  The basic version would be to have a command like `meet : cmd (() + robot)` which gives you a reference to another robot in the same cell as you (or perhaps up to a certain distance away), if there is one.  With lists (see #97 and #154), there could also be some kind of `meetAll : cmd (list robot)`.

byorgey:
Instead of a function that returns a list, we could have a function that lets you fold over the robots nearby.  This is just as good as having a list but without requiring any special data structures.  Something like `meetAll : (robot -> b -> b) -> b -> cmd b`.

xsebek:
@byorgey I like the second `meetAll` that takes the function `(robot -> b -> cmd b)`.

I take it will not be possible to use it with atomic? :smile:

byorgey:
I had almost convinced myself that the version with `robot -> b -> b` is better, but after playing with things for a while I'm convinced that the `robot -> b -> cmd b` version is superior.
- It is easy to implement the `robot -> b -> b` version in terms of the `robot -> b -> cmd b` version:
    ```
    def meetAllPure : (robot -> b -> b) -> b -> cmd b = \f. meetAll (\r. \b. return (f r b)) end
    ```
    Going the other direction is also possible but not as nice.
- The version with `cmd` will often be convenient when you want to run some command for each nearby robot, e.g. `give` each one a `bitcoin`:
    ```
    def makeItRain : cmd unit = meetAll (\r. \b. give r "bitcoin") () end
    ```
- The version with `cmd` is implicitly running the resulting commands in some particular order under the hood, but since it's going to iterate through the nearby robots in some arbitrary order anyway, you can't really complain about that.
- If you want complete control over the order in which the resulting commands get run or anything more fancy, you can always use `b = cmd unit`. (You just might have to throw in a call to `join` or two.)

byorgey:
Any ideas for what device could provide the `meet` capability?  Currently I added it to `scanner` just to put it somewhere but `scanner` is already rather overloaded...

xsebek:
I would suggest something made out of mid game material like iron or silver.

- shaky iron arm (for handshakes)
- antenna (for communication)
- radar (to scan around for robots)

byorgey:
Good idea.  We've talked about adding `antenna` before for other things, so I added it.  Current recipe is 2 `silver` + 2 `copper wire` but I'm open to other ideas.

Issue #115, opened 2021-09-29
Container commands


byorgey:
Every entity is in theory capable of holding an inventory of other entities, but at the moment we only take advantage of this with robots themselves.  However, the idea has always been to have other entities besides just robots that can contain other entities.  As a first example the plan is to make a `box` capable of holding (possibly a limited number of) other entities.  This will be useful for several reasons:
* Each cell in the world holds at most one entity; but if you want to place multiple items somewhere for another robot to pick up, you can put the items in a box and then place the box.
* The `give` command takes a single tick to execute, so it takes a while to give a bunch of entities one at a time.  Using a box, you can instead put the items you want to give into a box (putting an item in a box will not take a tick), and then just `give` the box.
* In conjunction with #114 , nested boxes can be used to simulate recursive data types such as lists and trees.  I kind of doubt we literally want to represent data types this way, but it's fun to contemplate. =)

So we need some commands for manipulating container entities; here's what I propose.
* Add a `container` property, which governs entities that can contain other entities.
* `insert : string -> string -> cmd ()` will insert the first entity into the second, throwing an exception if you don't have either one, or if the second entity is not a container.
* `remove : string -> string -> cmd ()` is the converse.  Note that the entity to be removed gets taken out of the container and placed in your inventory.
* `unpack : string -> cmd ()` unpacks a container, rendering it empty and placing its entire contents in your inventory.  This way you don't have to know in advance what you expect to get in the box.
* It might be nice to have a way to iterate over all the things in a box without taking them out; I am not sure what that would look like.
* None of these commands take a tick to execute.  The general principle is that commands which have some external effect on the world take a tick; those that merely sense the world or affect some change internal to a robot do not.
* I don't know what capabilities these commands should require, or what device should provide them.  We might consider just letting you do these things with a grabber, i.e. as soon as

byorgey:
If we end up switching to a specific `entity` type instead of just using `string`, as described in #114 , the types of a bunch of the above functions should of course change to match.

byorgey:
Given an `entity` type, and the fact that entities must be immutable, as discussed in #455 , here's my current thinking on the types the above commands should have:
* `insert : entity -> entity -> cmd entity` will insert the first entity into the second, throwing an exception if you don't have either one, or if the second entity is not a container.  It *returns a new reference to the container entity*, since entity values are immutable.
* `remove : entity -> entity -> cmd entity`
[...]

PR #835, opened 2022-11-05
New ranching scenario

kostmo:
Demo with:

    stack run -- --scenario Challenges/Ranching/gated-paddock --autoplay

The primary element of this scenario is "testing for a closed loop" in the objective condition.  As written, this test is performed every "tick", and introduces noticeable rendering lag at higher "ticks / s" speeds.

What may be better is for the player to "ask" for the condition to be checked, perhaps by standing somewhere specific, picking up a particular object, or even with `say "Ready!"`.

### DIFF
--- a/data/scenarios/Challenges/Ranching/_gated-paddock/meandering-sheep.sw
+++ b/data/scenarios/Challenges/Ranching/_gated-paddock/meandering-sheep.sw
@@ -1,0 +1,144 @@
+// A "sheep" that wanders around randomly.
+
+/** A "gate" is walkable, so we need to supplement the "blocked" check with this function.
+Since fences are "unwalkable", they do not need to be mentioned in this function.
+*/
+def isFenced =
+    s <- scan forward;
+    return (
+        case s
+            (\_. false)
+            (\x. x == "gate")
+    );
+    end;
+
+def isBlockedOrFenced =
+    b <- blocked;
+    f <- isFenced;
+    return (b || f);
+    end;
+
+def elif = \p.\t.\e. {if p t e} end;
+
+
+def turnToClover = \direction.
+
+  x <- scan direction;
+  case x (\_. return false;) (\y.
+    if (y == "clover") {
+      turn direction;
+      return true;
+    } {
+      return false;
+    };
+  );
+  end;
+
+
+/**
+If there is adjacent clover,
+turn that direction.
+*/
+def turnCloverDirection =
+
+    foundN <- turnToClover north;
+    if (foundN) {return true} {
+      foundE <- turnToClover east;  
+      if (foundE) {return true} {
+        foundS <- turnToClover south;
+        if (foundS) {return true} {
+          turnToClover west;
+        }
+      }
+    }
+    end;
+
+def decideDirection =
+
+  let randdir : cmd dir =
+    d <- random 4;
+    return $ if (d == 0) {
+        north
+      } $ elif (d == 1) {
+        east
+      } $ elif (d == 2) {
+        south
+      } {
+        west
+      }
+    in
+
+  cloverCount <- count "clover";
+  if (cloverCount > 4) {
+    d <- randdir;
+    turn d;
+  } {
+    nearClover <- turnCloverDirection;
+    if (nearClover) {} {
+      d <- randdir;
+      turn d;
+    }
+  }
+  end;
+
+let forever : cmd unit -> cmd unit = \c. c ; forever c in
+let repeat : int -> cmd unit -> cmd unit =
+  \n. \c. if (n == 0) {} {c ; repeat (n-1) c} in
+
+
+forever (
+  n <- random 30;
+  wait (30 + n);
+
+  decideDirection;
+
+  dist <- random 3;
+  repeat dist (
+
+    b <- isBlockedOrFenced;
+    if b {} {
+      move;
+    };
+
+    // Sheep can drown.
+    hasWater <- ishere "water";
+    if hasWater {
+      say "The jolly jumbuck sprang into the billabong. \"You'll never catch me alive!\", said he.";
+      selfdestruct;
+    } {};
+
+    // Eat clover.
+    x <- scan down;
+    case x return (\y.
+        if (y == "clover") {
+            harvest;
+            cloverCount <- count "clover";
+            if (cloverCount < 2) {
+              say "yum!"
+            } {};
+        } {};
+    );
+  );
+  r <- random 30;
+  if (r == 0) { say "baaa" } {};
+
+  hasClover <- has "clover";
+  if (hasClover) {
+    r <- random 8;
+    if (r == 0) { 
+      let item = "wool" in
+      hasWool <- has item;
+      if (hasWool) {
+        // Make sure nothing's in the way before we place
+        // our wool:
+        x <- scan down;
+        case x return (\_.
+          grab;
+          return ();
+        );
+
+        place item;
+      } {};
+    } {};
+  } {};
+)
--- a/data/scenarios/Challenges/Ranching/gated-paddock.yaml
+++ b/data/scenarios/Challenges/Ranching/gated-paddock.yaml
@@ -1,0 +1,536 @@
+version: 1
+name: Pastoral Island
+author: Karl Ostmo
+description: |
+  Start a wool industry from the local fauna.
+creative: false
+objectives:
+  - goal:
+      - |
+        You've homesteaded on a small island in the ocean.
+        It's time to gather resources to trade.
+      - |
+        You encounter some feral sheep (@). They slowly wander the island and eat grass.
+        Your mind wanders to textiles...
+      - |
+        First, paddock at least one sheep so they don't drown.
+        Make sure there are no gaps in the fence!
+      - |
+        Note that you can use the "drill" command (by way of the "post puller" tool)
+        to demolish a fence that has been "placed".
+    condition: |-
+      // Algorithm:
+      // ----------
+      // Maintain current direction until a wall is encountered.
+      // Then enter "wall-following mode".
+      // This mode presumes the wall is not a loop.
+      // Wall-following mode exploits recursion to keep track of how many left turns were made
+      // and then unwinds them again by ensuring each is paired with a right turn.
+      // Once the recursion is fully unwound, the robot proceeds along its original direction
+      // (though it may now be laterally displaced).
+      //
+      // (If it was a loop, then an "oriented breadcrumb" would need to be left.
+      // The breadcrumb is oriented in case a single-width passage is backtracked
+      // along the opposite wall.)
+
+      /** A "gate" is walkable, so we need to supplement the "blocked" check with this function.
+      Since fences are "unwalkable", they do not need to be mentioned in this function.
+      */
+      def isFenced =
+          s <- scan forward;
+          return (
+              case s
+                  (\_. false)
+                  (\x. x == "gate")
+          );
+          end;
+
+      def isBlockedOrFenced =
+          b <- blocked;
+          f <- isFenced;
+          return (b || f);
+          end;
+
+      // Returns true if we've already placed two
+      // breadcrumbs on a given tile, false otherwise.
+      def leaveBreadcrumbs =
+
+          let bc1 = "fresh breadcrumb" in
+          let bc2 = "treaded breadcrumb" in
+
+          wasTraversedOnce <- ishere bc1;
+          if wasTraversedOnce {
+              _crumb <- grab;
+              make bc2;
+              place bc2;
+              return false;
+          } {
+              wasTraversedTwice <- ishere bc2;
+              if wasTraversedTwice {
+                  return true;
+              } {
+                  // Make sure nothing's in the way before we place
+                  // our breadcrumb:
+                  x <- scan down;
+                  case x return (\y.
+                      // If we're on a water tile, get rid of
+                      // it with our special "drilling" recipe
+                      if (y == "water") {
+                          drill down;
+                          // Nothing will remain on the ground.
+                          // after making the "steam" via
+                          // the drilling recipe.
+                      } {
+                          grab;
+                          return ();
+                      };
+                  );
+
+                  make bc1;
+                  place bc1;
+                  return false;
+              };
+          };
+          end;
+
+      def goForwardToPatrol = \wasBlocked.
+          b <- isBlockedOrFenced;
+          if b {
+              turn left;
+              goForwardToPatrol true;
+              turn right;
+              goForwardToPatrol false;
+          } {
+              if wasBlocked {
+                  isLoop <- leaveBreadcrumbs;
+                  if isLoop {
+                      fail "loop";
+                  } {};
+              } {};
+              move;
+          };
+          end;
+
+      /**
+      There should only be one place in the
+      code where an exception is thrown: that is,
+      if a treaded breadcrumb is encountered.
+      */
+      def checkIsEnclosedInner =
+          try {
+              goForwardToPatrol false;
+              // Water is the outer boundary
+              hasWater <- ishere "water";
+              if hasWater {
+                  return false;
+              } {
+                  checkIsEnclosedInner;
+              };
+          } {
+              return true;
+          };
+          end;
+
+      def checkIsEnclosed =
+
+          // The "evaporator" drill is used
+          // to clear water tiles.
+          let specialDrill = "evaporator" in
+          create specialDrill;
+          install self specialDrill;
+
+          // NOTE: System robots can walk on water
+          // so we only need this if we want to
+          // demo the algorithm with a player robot.
+      //    create "boat";
+      //    install self "boat";
+
+          checkIsEnclosedInner;
+          end;
+
+      def boolToInt = \b. if (b) {return 1} {return 0}; end;
+
+      def countAdjacentBlockages =
+
+          turn left;
+          b1 <- isBlockedOrFenced;
+          c1 <- boolToInt b1;
+
+          turn left;
+          b2 <- isBlockedOrFenced;
+          c2 <- boolToInt b2;
+
+          turn left;
+          b3 <- isBlockedOrFenced;
+          c3 <- boolToInt b3;
+
+          turn left;
+          b4 <- isBlockedOrFenced;
+          c4 <- boolToInt b4;
+
+          return $ c1 + c2 + c3 + c4;
+          end;
+
+      // Step forward, observing left and right.
+      def observeLeftAndRight =
+          move;
+          turn left;
+          amBlockedLeft <- isBlockedOrFenced;
+          val1 <- boolToInt amBlockedLeft;
+
+          turn back;
+          amBlockedRight <- isBlockedOrFenced;
+          val2 <- boolToInt amBlockedRight;
+
+          turn right;
+          move;
+          return $ val1 + val2;
+          end;
+
+
+      /** If the four cardinal directions have at most
+      one blockage, then there will exist an orientation
+      where both that direction and its opposite direction
+      are clear.
+      So we can step that direction, check to the left and
+      right of us, then step in the opposite direction
+      and do the same. This allows us to check the 4
+      blocks that touch the corners of the center block.
+      */
+      def countDiagonalBlockages =
+          // First, orient to the clear front-to-back path
+          amBlocked <- isBlockedOrFenced;
+          if amBlocked {turn left;} {};
+
+          // Second, step to both sides
+          fwdCount <- observeLeftAndRight;
+          backCount <- observeLeftAndRight;
+          return $ fwdCount + backCount;
+          end;
+
+      def isStandingOnBridge =
+          onFence <- ishere "fence";
+          onGate <- ishere "gate";
+          if (onFence || onGate) {
+              adjCount <- countAdjacentBlockages;
+              if (adjCount > 1) {
+                  return true;
+              } {
+                  diagCount <- countDiagonalBlockages;
+                  return $ (adjCount + diagCount) > 1;
+              };
+          } {return false};
+          end;
+
+      def getValForSheepIndex = \predicateCmd. \i.
+          try {
+              // This will throw an exception if
+              // the sheep has already drowned.
+              r <- robotnumbered i;
+              didSucceed <- as r {predicateCmd};
+
+              boolToInt didSucceed;
+          } {
+              return 0;
+          }
+          end;
+
+      /**
+      There are 3 sheep.
+      They have indices 1, 2, 3.
+      (The base has index 0).
+
+      THIS DOES NOT WORK!
+      */
+      def countSheepWithRecursive = \predicateCmd. \i.
+
+          if (i > 0) {
+              val <- getValForSheepIndex predicateCmd i;
+              recursiveCount <- countSheepWithRecursive predicateCmd $ i - 1;
+              return $ val + recursiveCount;
+          } {
+              return 0;
+          }
+          end;
+
+
+      def countSheepWith = \predicateCmd.
+
+          val1 <- getValForSheepIndex predicateCmd 1;
+          val2 <- getValForSheepIndex predicateCmd 2;
+          val3 <- getValForSheepIndex predicateCmd 3;
+          return $ val1 + val2 + val3;
+
+          end;
+
+
+      justFilledGap <- as base {
+          isStandingOnBridge;
+      };
+
+      if (justFilledGap) {
+          enclosedCount <- countSheepWith checkIsEnclosed;
+          return $ enclosedCount >= 1;
+      } {
+          return false;
+      }
+  - goal:
+      - |
+        Safe! Your sheep are now hungry.
+        Offer them something tasty and you may be rewarded.
+      - |
+        The sheep will move toward something edible on
+        an adjacent tile and will eat it if they walk over it.
+      - |
+        You may want to add a gate to the fence
+        to give yourself easier access.
+    condition: |-
+      def getTruthForSheepIndex = \predicateCmd. \i.
+          try {
+              // This will throw an exception if
+              // the sheep has already drowned.
+              r <- robotnumbered i;
+              as r {predicateCmd};
+          } {
+              return false;
+          };
+          end;
+
+      def anySheep = \predicateCmd. \i.
+
+          if (i > 0) {
+              didSucceed <- getTruthForSheepIndex predicateCmd i;
+              if didSucceed {
+                  return true;
+              } {
+                  anySheep predicateCmd $ i - 1;
+              };
+          } {
+              return false;
+          };
+          end;
+
+      anySheep (has "clover") 3;
+  - goal:
+      - |
+        Yum! Contented, well-fed sheep may drop wool.
+      - |
+        Winter is coming! Collect three wool bundles to make a sweater.
+      - |
+        Each sheep drops a finite amount over
+        their lifetime.
+    condition: |-
+      as base {
+        has "sweater";
+      };
+robots:
+  - name: base
+    dir: [0, 1]
+    devices:
+      - treads
+      - scanner
+      - dictionary
+      - branch predictor
+      - strange loop
+      - clock
+      - ADT calculator
+      - comparator
+      - workbench
+      - grabber
+      - lambda
+      - logger
+      - hearing aid
+      - counter
+      - mirror
+      - post puller
+    inventory:
+      - [0, fence]
+      - [0, gate]
+      - [10, hinge]
+  - name: sheep
+    description:
+      - meandering livestock
+    display:
+      invisible: false
+      char: '@'
+    system: true
+    dir: [0, 1]
+    inventory:
+      - [4, wool]
+    program: |
+      run "scenarios/Challenges/Ranching/_gated-paddock/meandering-sheep.sw";
+entities:
+  - name: fence
+    display:
+      char: '#'
+    description:
+      - Keeps sheep in. And some other things out.
+    properties: [known, portable, unwalkable]
+  - name: post puller
+    display:
+      char: 'P'
+      attr: rock
+    capabilities: [drill]
+    description:
+      - Good for dismantling fences.
+    properties: [known, portable]
+  - name: scrap wood
+    display:
+      char: '\'
+    description:
+      - Scrap wood. Can be reconditioned into boards.
+    properties: [known, portable]
+  - name: sweater
+    display:
+      attr: gold
+      char: 'S'
+    description:
+      - A warm wool sweater. Just in time for winter!
+    properties: [known, portable]
+  - name: clover
+    display:
+      attr: flower
+      char: '%'
+    description:
+      - A tasty stack for fluffy ruminants.
+    properties: [portable, growable]
+    growth: [80, 100]
+  - name: gate
+    display:
+      char: '/'
+      attr: rock
+    description:
+      - A gate permits the player to pass through, but sheep cannot.
+    properties: [known]
+  - name: hinge
+    display:
+      char: 'U'
+      attr: rock
+    description:
+      - Facilitates swinging action.
+    properties: [known, portable]
+  - name: cabin
+    display:
+      char: Π
+      attr: rock
+    description:
+      - Home sweet home.
+    properties: [known, unwalkable]
+  - name: pier
+    display:
+      char: 且
+      attr: rock
+    description:
+      - Docking area for ships
+    properties: [known]
+  - name: fresh breadcrumb
+    display:
+      char: '.'
+    description:
+      - A marker that can be put down and found again. Used only by judge robot.
+    properties: [portable]
+  - name: treaded breadcrumb
+    display:
+      char: 'x'
+    description:
+      - A marker that can be put down and found again (for a second time). Used only by judge robot.
+    properties: [portable]
+  - name: evaporator
+    display:
+      char: 'E'
+    description:
+      - A tool that allows clearing a water tile. Used only by judge robot.
+    properties: [portable]
+    capabilities: [drill]
+  - name: wool
+    display:
+      char: 'ω'
+      attr: gold
+    description:
+      - A bundle of raw animal fiber.
+    properties: [portable]
+  - name: steam
+    display:
+      char: 'Z'
+    description:
+      - What's left after evaporating water. Used only by judge robot.
+    properties: [portable]
+recipes:
+  - in:
+      - [2, board]
+    out:
+      - [1, fence]
+  - in:
+      - [3, wool]
+    out:
+      - [1, sweater]
+  - in:
+      - [1, scrap wood]
+    out:
+      - [1, board]
+  - in:
+      - [0, fresh breadcrumb]
+    out:
+      - [1, fresh breadcrumb]
+  - in:
+      - [1, fresh breadcrumb]
+    out:
+      - [1, treaded breadcrumb]
+  - in:
+      - [1, fence]
+    out:
+      - [1, scrap wood]
+    required:
+      - [1, post puller]
+  - in:
+      - [1, water]
+    out:
+      - [1, steam]
+    required:
+      - [1, evaporator]
+  - in:
+      - [1, hinge]
+      - [1, fence]
+    out:
+      - [1, gate]
+known: [mountain, tree, water]
+seed: 0
+solution: |
+  run "scenarios/Challenges/Ranching/_gated-paddock/fence-construction.sw"
+world:
+  default: [dirt, water]
+  palette:
+    'B': [grass, null, base]
+    '.': [grass]
+    't': [dirt, tree]
+    'x': [stone, mountain]
+    'c': [stone, cabin]
+    's': [grass, null, sheep]
+    '%': [grass, clover, null]
+    'H': [stone, pier, null]
+    '~': [dirt, water]
+  upperleft: [-34, 11]
+  map: |-
+    ~~~~.......~~~~~~~~~~~~~.......................~~~~~~
+    ~~.............~~~~.......%.%%%%%%..%%%%%.%.......~~~
+    ~........................%%%%%%%%%%%%%..%%%%.......~~
+    ~...................................................~
+    .............................x.xx........tttttttt...~
+    .........................................tttttttt...~
+    ..........................xxxxx..........tttttttt...~
+    .........................................tttttttt...~
+    ...........s.............................tttttttt....
+    ....x....................................tttttttt....
+    ...xx................................................
+    ..xx.....................................B......c....
+    ...x.................................................
+    ..............................s......................
+    .....................................................
+    ....................................................~
+    ~..............s...................................~~
+    ~~.......xxx...........x.........................~~~~
+    ~~~............xx............x............~~~~H~~~~~~
+    ~~~.....................................~~~~~~H~~~~~~
+    ~~.....................................~~~~~~~H~~~~~~
+    ~~..............................~~~~~~~~~~~~~~H~~~~~~
+    ~~~~...................~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
--- a/test/integration/Main.hs
+++ b/test/integration/Main.hs
@@ -165,6 +165,10 @@ testScenarioSolution _ci _em =
             , testSolution Default "Challenges/Mazes/invisible_maze"
             , testSolution Default "Challenges/Mazes/loopy_maze"
             ]
+        , testGroup
+            "Ranching"
+            [ testSolution (Sec 30) "Challenges/Ranching/gated-paddock"
+            ]
         ]
     , testGroup
         "Regression tests"
--- a/data/scenarios/Challenges/00-ORDER.txt
+++ b/data/scenarios/Challenges/00-ORDER.txt
@@ -3,3 +3,4 @@ teleport.yaml
 2048.yaml
 hanoi.yaml
 Mazes
+Ranching
--- a/data/scenarios/Challenges/Ranching/00-ORDER.txt
+++ b/data/scenarios/Challenges/Ranching/00-ORDER.txt
@@ -1,0 +1,1 @@
+gated-paddock.yaml
--- a/data/scenarios/Challenges/Ranching/_gated-paddock/enclosure-checking.sw
+++ b/data/scenarios/Challenges/Ranching/_gated-paddock/enclosure-checking.sw
@@ -1,0 +1,255 @@
+// Algorithm:
+// ----------
+// Maintain current direction until a wall is encountered.
+// Then enter "wall-following mode".
+// This mode presumes the wall is not a loop.
+// Wall-following mode exploits recursion to keep track of how many left turns were made
+// and then unwinds them again by ensuring each is paired with a right turn.
+// Once the recursion is fully unwound, the robot proceeds along its original direction
+// (though it may now be laterally displaced).
+//
+// (If it was a loop, then an "oriented breadcrumb" would need to be left.
+// The breadcrumb is oriented in case a single-width passage is backtracked
+// along the opposite wall.)
+
+/** A "gate" is walkable, so we need to supplement the "blocked" check with this function.
+Since fences are "unwalkable", they do not need to be mentioned in this function.
+*/
+def isFenced =
+    s <- scan forward;
+    return (
+        case s
+            (\_. false)
+            (\x. x == "gate")
+    );
+    end;
+
+def isBlockedOrFenced =
+    b <- blocked;
+    f <- isFenced;
+    return (b || f);
+    end;
+
+// Returns true if we've already placed two
+// breadcrumbs on a given tile, false otherwise.
+def leaveBreadcrumbs =
+    
+    let bc1 = "fresh breadcrumb" in
+    let bc2 = "treaded breadcrumb" in
+
+    wasTraversedOnce <- ishere bc1;
+    if wasTraversedOnce {
+        _crumb <- grab;
+        make bc2;
+        place bc2;
+        return false;
+    } {
+        wasTraversedTwice <- ishere bc2;
+        if wasTraversedTwice {
+            return true;
+        } {
+            // Make sure nothing's in the way before we place
+            // our breadcrumb:
+            x <- scan down;
+            case x return (\y.
+                // If we're on a water tile, get rid of
+                // it with our special "drilling" recipe
+                if (y == "water") {
+                    drill down;
+                    // Nothing will remain on the ground.
+                    // after making the "steam" via
+                    // the drilling recipe.
+                } {
+                    grab;
+                    return ();
+                };
+            );
+
+            make bc1;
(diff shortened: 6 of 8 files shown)


byorgey:
I understand why you want to have a `gated-paddock` subdirectory to hold all the `.sw` files for the scenario.  But currently this means the test suite spits out a ton of warnings about not finding an `00-ORDER.txt` file in that directory.  We should either add an empty `00-ORDER.txt` file in `gated-paddock` just to make it shut up, or (probably the better long-term solution) make the scenario loading code ignore any directories that don't contain any `.yaml` files.

byorgey:
One of the first things I did was make a `logger` and install it on myself so I could see errors.  Is that intended, or should we install a `logger` on the base robot to start?

kostmo:
> We should either add an empty 00-ORDER.txt file in gated-paddock just to make it shut up, or (probably the better long-term solution) make the scenario loading code ignore any directories that don't contain any .yaml files.

I think we'll have to go with the latter; I tried creating an empty `data/scenarios/Challenges/Ranching/gated-paddock/00-ORDER.txt` file, and the warnings are still there:

    Warning: while processing Ranching/00-ORDER.txt: files not listed in 00-ORDER.txt will be ignored
    - gated-paddock

kostmo:
> One of the first things I did was make a `logger` and install it on myself so I could see errors. Is that intended, or should we install a `logger` on the base robot to start?

Good idea, will add it.

byorgey:
Completed in 4m37s! :grin:  Two of the stupid sheep had drowned themselves by the time I finished making fences but one was still around.

Based on the goal description I was kind of surprised that there wasn't another step after building the fence (like harvesting wool or something).  Are you intending to extend the scenario in the future?

kostmo:
> Are you intending to extend the scenario in the future?

Yes, good catch.  Figured I'd get some early feedback first.

data/scenarios/Challenges/Ranching/gated-paddock.yaml
```diff
@@ -0,0 +1,163 @@
...
+  - name: gate
+    display:
+      char: '/'
+      attr: rock
+    description:
+    - A gate.
+    properties: [known]
+  - name: cabin
+    display:
+      char: 🏠
```
byorgey, line 87: This looks really cute, but unfortunately in gnome-terminal it is displayed as a double wide character so the alignment of everything to the right of it is messed up.  We should probably use a different character.

data/scenarios/Challenges/Ranching/gated-paddock/meandering-sheep.sw
```diff
@@ -0,0 +1,36 @@
...
+      hasWater <- ishere "water";
+      if hasWater {
+        say "whoops!";
+        selfdestruct;
+      } {};
+
+    }
+  );
+  r <- random 5;
+  if (r == 0) { say "baaa" } {}
```
byorgey, line 35: I don't think there is any way to hear the baas.  Maybe install a `hearing aid` on the base robot?

byorgey (commented):
I love it!

data/scenarios/Challenges/Ranching/gated-paddock.yaml
xsebek: Capital Pi might work.

There are also a few houses for nerdfonts:
https://www.nerdfonts.com/cheat-sheet

xsebek:
@kostmo if you wait a little, I can look into optimizing the goal condition. 😉 

But I would like to play it first without getting spoiled by having seen the solution. 😅

byorgey:
Does it require you to have *all* the sheep inside the fence, or at least one?  I played again and made a small fence around one sheep but nothing happened... so I went to make a screenshot to report a bug, and when I was done making the screenshot it suddenly congratulated me on winning.  @noahyor suggested it might be because all the other sheep died.  Seems a little strange that you can win when a sheep dies rather than winning when you have actually completed something.

data/scenarios/Challenges/Ranching/gated-paddock.yaml
kostmo: Pi works. Π

data/scenarios/Challenges/Ranching/gated-paddock/meandering-sheep.sw
```diff
@@ -0,0 +1,36 @@
...
+
+let forever : cmd unit -> cmd unit = \c. c ; forever c in
+let repeat : int -> cmd unit -> cmd unit =
+  \n. \c. if (n == 0) {} {c ; repeat (n-1) c} in
+let randdir : cmd dir =
+  d <- random 4;
+  return (
+    if (d == 0) {north}
+    {if (d == 1) {east}
+    {if (d == 2) {south} {west}}})
```
xsebek, line 11: TIP:
```
def elif = \p.\t\e. {if p t e} end;
```
```
if (d == 0) {
  north
} $elif (d == 1) {
  east
} $elif (d == 2) {
  south
} {
  west
}
```

data/scenarios/Challenges/Ranching/gated-paddock/meandering-sheep.sw
```diff
@@ -0,0 +1,36 @@
...
+  dist <- random 3;
+  repeat dist (
+    b <- blocked;
+    if b {} {
+      move;
+
+      // Sheep can drown.
+      hasWater <- ishere "water";
+      if hasWater {
+        say "whoops!";
```
xsebek, line 28: Oh, so that is how drowning sheep sound. 😆  🌊 🐑 🌊 

As an alternative (e.g. if sheep were normal robots and just drowned) a referee robot could check if the sheep still exist and say something if they were not found.

data/scenarios/Challenges/Ranching/gated-paddock/enclosure-checking.sw
```diff
@@ -0,0 +1,87 @@
+// Algorithm:
```
xsebek, line 1: How about requiring that _the fence must not branch and must connect the north and south solid fence_?

That would simplify the checking a lot, just follow the path using a scanner. Then at least one sheep has to be surrounded by the fence:
```
 +------+
++######|
|## ^  #v
|#  @  /
|#  v  #^
|#######|
+---|---+
    ^
start here and go left and right
```

So if you go north, find a fence and go left, you must reach the south solid fence and if you go right the north solid fence.
Similarly from the fence that must be south of the sheep.

This check will short-circuit a lot faster if the fence is not connected or a sheep can just go south or north to the ocean.

data/scenarios/Challenges/Ranching/gated-paddock/meandering-sheep.sw
```diff
@@ -0,0 +1,36 @@
...
+forever (
+  n <- random 30;
+  wait (30 + n);
+  d <- randdir;
+  turn d;
+  dist <- random 3;
+  repeat dist (
+    b <- blocked;
+    if b {} {
+      move;
```
xsebek, line 23: The sheep could crash in rare cases, consider atomic:
```suggestion
  atomic (b <- blocked; if b {} {move});
```

xsebek (changes requested):
It took me a while to finish, but I enjoyed it a lot. 👨‍🌾 🐑

I have a lot of small suggestions for improvement, if you disagree with any of them please explain why so that I can understand the challenge better. 😉

data/scenarios/Challenges/Ranching/gated-paddock/meandering-sheep.sw
xsebek: I think @kostmo fixed that. 🙂

data/scenarios/Challenges/Ranching/gated-paddock.yaml
```diff
@@ -0,0 +1,165 @@
...
+      - |
+        You encounter some feral sheep. They look like a promising
+        source of wool.
+      - |
+        First, paddock the sheep so they don't drown.
+        Construct a fence using the nearby gate as a starting point.
+        Make sure there are no gaps!
+      - |
+        Note that you can use the "drill" command (by way of the "post puller" tool)
+        to demolish a fence that has been "placed".
```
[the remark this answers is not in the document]
kostmo: One cannot walk on top of the fence, and `grab` requires the robot to be on the same tile as the thing being grabbed.

data/scenarios/Challenges/Ranching/gated-paddock.yaml
```diff
@@ -0,0 +1,165 @@
...
+      - |
+        You've homesteaded on a small Pacific island.
+        It's time to gather resources to trade.
+      - |
+        You encounter some feral sheep. They look like a promising
+        source of wool.
+      - |
+        First, paddock the sheep so they don't drown.
+        Construct a fence using the nearby gate as a starting point.
+        Make sure there are no gaps!
```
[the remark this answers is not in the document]
kostmo: I may revise the language to:
> You may using the nearby gate as a starting point.

The `gate` tile is special as it permits the player to walk through, but the sheep cannot.  I'm thinking of exploiting that fact for a "feeding" or "fleecing" challenge.

I should describe this behavior to the `gate` description, and probably to the challenge description.  And perhaps I should make a recipe for a gate...

data/scenarios/Challenges/Ranching/gated-paddock.yaml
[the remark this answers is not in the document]
xsebek: @kostmo that is not true, I remember I specifically allowed system robots to do that:
https://github.com/swarm-game/swarm/blob/29594abfa23e0ad10e36267a4192443f1a0aea0c/src/Swarm/Game/Step.hs#L1676-L1677

data/scenarios/Challenges/Ranching/gated-paddock.yaml
```diff
@@ -0,0 +1,165 @@
...
+      - |
+        Note that you can use the "drill" command (by way of the "post puller" tool)
+        to demolish a fence that has been "placed".
+    condition: |
+      r <- robotnamed "sheep";
+      as r {run "scenarios/Challenges/Ranching/gated-paddock/enclosure-checking.sw"};
+robots:
+  - name: base
+    dir: [0, 1]
+    devices:
```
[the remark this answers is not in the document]
kostmo: OK, I'll add `fence` and `gate` to the inventory.  I'm actually not sure I need the "solid fence" anymore.

data/scenarios/Challenges/Ranching/gated-paddock.yaml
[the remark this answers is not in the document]
kostmo: But it is the `base` robot that may need to dismantle the fence, not a system robot.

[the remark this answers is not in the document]
xsebek: Oh, sorry I agree that drill is necessary (now I see why). I just wanted to object to the statement because I thought you were responding to my other comments, where walking on the fence is relevant. 😅

data/scenarios/Challenges/Ranching/gated-paddock.yaml
```diff
@@ -0,0 +1,165 @@
...
+      - treads
+      - scanner
+      - dictionary
+      - branch predictor
+      - strange loop
+      - clock
+      - ADT calculator
+      - comparator
+      - workbench
+      - lodestone
```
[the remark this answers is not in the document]
kostmo: Ah, vestigial from when I was using "bits" as breadcrumbs... will remove

data/scenarios/Challenges/Ranching/gated-paddock.yaml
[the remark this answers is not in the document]
xsebek: I like it, see my other comment about how it could simplify the objective check. 😉

data/scenarios/Challenges/Ranching/gated-paddock.yaml
```diff
@@ -0,0 +1,165 @@
...
+      - grabber
```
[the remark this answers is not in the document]
kostmo: I suppose it's up to the player whether they want to construct a harvester.  They can "choose their own adventure"---weighing the effort of constructing a harvester vs. short term progress.

data/scenarios/Challenges/Ranching/gated-paddock.yaml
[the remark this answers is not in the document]
xsebek: I do not see why the sheep should not be able to walk through a gate.

You could add a check so that the sheep _choose_ not to wander through the gate or fence. That would also allow you to get rid of the drill hack as the player could walk over the fence and just grab it.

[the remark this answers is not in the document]
byorgey: I like having a distinction between fence that no one can walk through, and gate that the player can walk through but sheep cannot.  Although I do not see anything in the current code that would prevent sheep from walking through a gate?

data/scenarios/Challenges/Ranching/gated-paddock.yaml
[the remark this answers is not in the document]
byorgey: Perhaps you could add a program to the base robot which just `def`ines a new command like `dismantle` which is just a synonym for drill?

data/scenarios/Challenges/Ranching/gated-paddock/enclosure-checking.sw
byorgey: More simply (perhaps), just (1) start at the north gate and follow fence until reaching the south gate (making sure it does not branch and there are no gaps), then (2) start at a sheep and go in a straight line until reaching the ocean, counting fence+gates encountered (being careful to count contiguous sections as only one).  If a fence was encountered an odd number of times, the sheep is inside.

data/scenarios/Challenges/Ranching/gated-paddock.yaml
```diff
@@ -0,0 +1,161 @@
+version: 1
+name: Pastoral Island
```
byorgey, line 2: I'm sure you have already spent many, many hours on this, so you should definitely add an `author: ` field with your name! :smile:

data/scenarios/Challenges/Ranching/gated-paddock.yaml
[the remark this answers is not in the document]
byorgey: A recipe for a gate would be nice.  Maybe the player will want to put more gates in.

data/scenarios/Challenges/Ranching/gated-paddock.yaml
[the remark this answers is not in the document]
byorgey: Personally I kind of liked having to figure out how to make a fence, I think it is just a matter of taste.  Either way seems fine.

data/scenarios/Challenges/Ranching/gated-paddock.yaml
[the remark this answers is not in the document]
byorgey: I definitely made a harvester since it was not at all clear to me whether I would have enough trees otherwise, especially since I did not know the recipe for a fence.  I can see now that you included plenty of trees for making the fence, but even if I had known that, I was not sure what else I might need the trees for in the future.  Actually, I think having to make a little tree farm seems very much in keeping with the "homesteading" idea.  I wonder if you might consider starting with just a small stand of trees (like 6 or 7) and *requiring* the player to use a `harvester`.

data/scenarios/Challenges/Ranching/gated-paddock/meandering-sheep.sw
byorgey: @xsebek now it seems like when sheep say "baaa", it flashes instantly in the Messages dialog, like for one tick, and then disappears.  I am not sure exactly how the Messages dialog is supposed to work re: `say` but having things disappear after one tick does not seem very useful.

xsebek:
@kostmo I wrote #836 which could be used to speed up the objective checking if we used Haskell for pathfinding. 🙂

It could be useful in other situations too, but it would be a bit harder to implement.

Also, it would allow us to taunt sheep with food like in Minecraft if that is something you are interested in. 😄

data/scenarios/Challenges/Ranching/gated-paddock/enclosure-checking.sw
xsebek: @byorgey but if you add any other line of fence unconnected to the gate, your check is not valid, right? 🤔 

I did not want to check all island locations for unconnected fences, so going south and north seemed simpler.

data/scenarios/Challenges/Ranching/gated-paddock.yaml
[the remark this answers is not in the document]
kostmo: Ah, I realized that I had not coded the sheep to treat the gate special yet, though my loop-checking algorithm did.  Fixed now.  And added a gate recipe.

data/scenarios/Challenges/Ranching/gated-paddock/meandering-sheep.sw
kostmo: I tried adding `atomic`, but now the blocked-checking logic is larger, and the compiler refuses to allow atomic around it.

xsebek: @kostmo AFAIK, we do not allow definitions. If you inline it then it should be fine, because only the move is observable.

kostmo:
> Does it require you to have _all_ the sheep inside the fence, or at least one?

@byorgey The buggy behavior is an artifact of using `robotnamed`, which is kind of a hack:

      r <- robotnamed "sheep";

Perhaps it would have worked if `robotnamed` did a dynamic lookup based on the currently existing robots, but as is, the name is fixed to an (arbitrary) sheep's robot ID upon scenario initialization.

I explored using `r <- robotnumbered i` and iterating up to an a-priori-known index (hard-coded to how many robots I've defined in the scenario), and checking if `whoami` is `"sheep"`.  However, then I was no longer able to satisfy the typechecker with the `run` command.

data/scenarios/Challenges/Ranching/gated-paddock/meandering-sheep.sw
xsebek: @byorgey you should still see it if you have a hearing device. The idea is that you should not hear things from before you were able to hear anything at all. So newly heard messages go to the logger from which they are shown in the Messages modal.

But I see it does not work now. Either @kostmo accidentally broke it, or I made a mistake in the `listen`/`say` implementation. 😄

xsebek:
@kostmo maybe you could name the individual sheep? Either funny names like `"Shaun"` or just use the dull `"sheep #1"`.

With the latter, you can use `"sheep #" ++ format i`.

kostmo:
> maybe you could name the individual sheep?

Yes, that's one workaround.  I guess I am being stubborn because I feel I should not have to create a new `robot` entry in the scenario for each sheep---I prefer to have just one sheep template robot.

data/scenarios/Challenges/Ranching/gated-paddock.yaml
```diff
@@ -118,6 +126,11 @@ recipes:
     - [1, scrap wood]
     required:
     - [1, post puller]
+  - in:
+    - [1, hinge]
+    - [1, fence]
+    out:
+    - [1, gate]
```
byorgey, line 133: :+1:

data/scenarios/Challenges/Ranching/gated-paddock.yaml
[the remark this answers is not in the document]
kostmo: > Perhaps you could add a program to the base robot which just `def`ines a new command like `dismantle` which is just a synonym for drill?

That could be nice.  Filed #838; currently if add the following to `base` in the scenario file:
```
    program: |
      def dismantle = drill; end;
```
When I try to `dismantle` in the REPL, I get this error:
```
Fatal error: Reference to unknown memory cell 0
```

data/scenarios/Challenges/Ranching/gated-paddock.yaml
```diff
@@ -0,0 +1,165 @@
...
+      - logger
+      - hearing aid
+      - post puller
+  - name: sheep
+    description:
+      - meandering livestock
+    display:
+      invisible: false
+      char: '@'
+    system: true
```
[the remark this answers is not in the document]
kostmo: > If the sheep were not system robots, I would be able to see them in the robot panel.
I think I would prefer that they don't appear in the robot panel; they are "NPCs".

data/scenarios/Challenges/Ranching/gated-paddock.yaml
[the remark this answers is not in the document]
byorgey: oh no :scream:

byorgey:
@kostmo I just merged #837 which means if you rename your `gated-paddock` subdirectory to `_gated-paddock` it will no longer generate warnings.  Any subdirectory starting with `_` is ignored for the purposes of loading scenarios.

data/scenarios/Challenges/Ranching/gated-paddock.yaml
```diff
@@ -0,0 +1,175 @@
...
+  - in:
+    - [1, hinge]
+    - [1, fence]
+    out:
+    - [1, gate]
+known: [mountain, tree, water]
+seed: 0
+solution: |
+  run "scenarios/Challenges/Ranching/gated-paddock/fence-construction.sw"
+world:
```
xsebek, line 139: @kostmo maybe add seed, so that the sheep always do the same thing.

I hope it is unlikely, but it could save us some headaches in the future if the sheep randomly came up with some clever escape plan. 🐑

byorgey: It would also make the solution more stable.  Currently, there is always the unlikely possibility the sheep will all decide to run straight into the ocean before you build the fence. :laugh:

xsebek:
> I guess I am being stubborn because I feel I should not have to create a new robot entry in the scenario for each sheep---I prefer to have just one sheep template robot.

@kostmo indeed we run into that before and there are issues discussing this:
- #306
- #506

If you want, you can add a `TODO` with those issue numbers so that once those are implemented we remember to also improve this scenario. 🙂

xsebek:
> Perhaps it would have worked if robotnamed did a dynamic lookup based on the currently existing robots, but as is, the name is fixed to an (arbitrary) sheep's robot ID upon scenario initialization.

@kostmo are you sure? I was able to view the last surviving `robotNamed "sheep"` just fine.

It very much should do a dynamic lookup based on the current robot map in the game state. If not then that is a bug in my implementation of `robotNamed`. 😄

data/scenarios/Challenges/Ranching/gated-paddock/enclosure-checking.sw
byorgey: Oh, good point.  You could somehow mark every fence section you traversed while walking along it, and then only count marked fence sections.

byorgey:
I guess the point is that each time step, there is one specific "magic" sheep which has to be enclosed, but there is no way to know which one it is.  If you guess wrong and enclose one of the other sheep, you will not win until such time as the magic sheep dies; then the next time the check runs a different sheep will be selected.  If it is the one inside your fence, you win!  Otherwise, you have to wait for that one to die too.

data/scenarios/Challenges/Ranching/gated-paddock.yaml
kostmo: > add seed

Actually I have `seed: 0` already on line 136.  But I wonder if there's global influence on the random number generation that causes the sheep to behave non-reproducibly?  Perhaps robot-local random number generators would help?

kostmo:
> are you sure?
[...]
It very much should do a dynamic lookup

You're right - I made the wrong inference about the behavior of `robotWithName` (should have read the code first).  @byorgey explains it properly.  When there still exist multiple sheep, the one retrieved with `robotnamed "sheep` is arbitrary.

xsebek:
Ah, now I understand what the problem is. I paused the game and stepped through it so that the sheep would not have time to escape. I saved all of them, as it did not occur to me that one would be enough. (Feels like that should be an achievement. #797)

Screenshot from 2022-11-06 14-25-18

Admittedly in my hurry to save all sheep, I also forgot about the gate. 🤣 

---

I think you should be able to safely assume that sheep have robot IDs 1,2 and 3 as they are added after base.

If you don't like that (I certainly don't) you can implement `meetAll` from #306. Honestly, it's probably not that hard, but I may have taken pride in coming up with convoluted hacks to avoid the proper solution. Choose which path you will take @kostmo. 😈 😇

data/scenarios/Challenges/Ranching/gated-paddock.yaml
xsebek: @kostmo Oh, my mistake then. 👍 The only other source of nondeterminism I can think of would be the order of elements in an unordered container. That may or may not have an effect here.

You can check that by listing the sheep location at the end and then testing that they are always the same
using `testSolution'`.

byorgey: > But I wonder if there's global influence on the random number generation that causes the sheep to behave non-reproducibly?

If there is, it's a bug.

data/scenarios/Challenges/Ranching/gated-paddock.yaml
```diff
@@ -0,0 +1,175 @@
...
+    - [1, post puller]
+  - in:
+    - [1, hinge]
+    - [1, fence]
+    out:
+    - [1, gate]
+known: [mountain, tree, water]
+seed: 0
+solution: |
+  run "scenarios/Challenges/Ranching/gated-paddock/fence-construction.sw"
```
xsebek, line 138: @kostmo you need to add the scenario to integration tests:
```Haskell
        , testGroup
            "Ranching"
            [ testSolution Default "Challenges/Ranching/gated_paddock"
            ]
```

data/scenarios/Challenges/Ranching/gated-paddock.yaml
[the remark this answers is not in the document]
xsebek: Sure, but system robots have the powers of gods... such sheep could _literally walk through walls_. 🤣 

Besides they belong to the farmer who owns the island, so they are more pets than NPCs. 🙂

data/scenarios/Challenges/Ranching/gated-paddock/meandering-sheep.sw
xsebek: It's a bug. 🐛 Fix incoming.

xsebek: Fixed in #842. ✅

test/integration/Main.hs
```diff
@@ -165,6 +165,10 @@ testScenarioSolution _ci _em =
             , testSolution Default "Challenges/Mazes/invisible_maze"
             , testSolution Default "Challenges/Mazes/loopy_maze"
             ]
+        , testGroup
+            "Ranching"
+            [ testSolution Default "Challenges/Ranching/gated_paddock"
```
xsebek, line 170:
```suggestion
            [ testSolution Default "Challenges/Ranching/gated-paddock"
```

kostmo:
> I also forgot about the gate

I've decided to get rid of the pre-built gate and fence stubs.

data/scenarios/Challenges/Ranching/gated-paddock/enclosure-checking.sw
kostmo: One thing I'd intended but haven't stated yet is for the player to be able to incorporate natural barriers (e.g. mountains) into their fencing strategy.  For instance, the player might extend the fence a bit further than they would otherwise to reach a mountain range that, in the end, saves more fencing after all.

I hadn't demoed this yet because the robot program would be more complicated than just drawing a rectangular fence without obstacles.

Point being, whatever alternative algorithm is used to check for enclosure should account for non-fence barriers (of potentially multi-tile thickness.

xsebek: @kostmo I would suggest using more complicated checks in more complicated challenges.

Notice that chess horse does not implement full chess rules and not all mazes implement the general shortest path algorithm in an undirected graph.

Using solutions tailored to the specifics of the challenge leads to better performance and robots can `say` to the player what is wrong. Ideally, the Swarm code should also be maintainable, so that when things break other people can fix it.

---

What I am advocating for here is making the challenge easy for players and developers alike.

Small things like deciding that the fence has to connect to the starting gate improve the performance by a lot.
That in turn makes it unnecessary to add extra requirements like `say "Please check now"`.

xsebek:
> I've decided to get rid of the pre-built gate and fence stubs.

Why? They can be used to improve performance and allow us to at least compare with other ways to check the objective.

Also `scan`ing them allowed me to learn how to make a fence in the first place, so I feel a bit sentimental about that gate. 😢

kostmo:
Building upon @xsebek's optimization ideas, I've sped up the goal checking significantly.

Using the `stack test` "Test scenario solutions" time as a benchmark:
* Inlining the script into the `.yaml` file brought the time down from `6.22` seconds to `3.83` seconds.
* Adding new short-circuiting pre-check brought the time down even further to `0.68` seconds.

## How does the new pre-check work?

1. We know that the solution will be achieved at the exact moment that a "gap" is filled by a fence tile.  So instead of running the full loop check on every tick, we only do it when a gap is filled.
2. The only time that the player can be "on top" of a fence tile is when that fence has just been placed.  Since fences are "unwalkable", once the player moves off of the fence tile, they cannot get back on.
3. A "gap" is filled when the currently placed fence tile bridges **at least two** adjacent tiles in the [Moore neighborhood](https://en.wikipedia.org/wiki/Moore_neighborhood) (i.e. the surrounding 8 tiles).  We use 8-connectivity instead of 4-connectivity ([Von Neumann neighborhood](https://en.wikipedia.org/wiki/Von_Neumann_neighborhood)) because diagonally-adjacent barriers are enough to block a sheep's movement.
4. So if the current tile underneath the `base` robot is a fence and at least two of the surrounding tiles are "unwalkable" (including any "gate" tiles), we have just filled a gap and we proceed to the full loop check.

This new pre-check does not require a pre-placed gate on the map.

kostmo:
Ok, my solution checker now accounts for all three sheep.
I also added two more goals and tweaked the map.

I'm pretty happy with the scenario now.  Give it another try?

data/scenarios/Challenges/Ranching/gated-paddock.yaml
```diff
@@ -0,0 +1,484 @@
...
+          }
+          end;
+
+
+      justFilledGap <- as base {
+          isStandingOnBridge;
+      };
+
+      if (justFilledGap) {
+          allSheep checkIsEnclosed 3;
```
xsebek, line 252: Well, that is a lot harder now, previously it was enough to save some of the sheep. :confused: 

Now if just one of the sheep drowns, I can no longer win.

Imagine the following scenario:
1. Sheep number 1 randomly runs straight into the ocean
2. The player does not notice
3. Player successfully paddocks all remaining sheep
4. Player waits for the checks to finish
5. The game slows down as it checks the condition every tick
6. Player wonders what is wrong

---

Please do at least one:
1. note what is actually required in the goal description
2. simplify the goal condition to require just `anySheep`
3. move the goal condition to a helper "judge" robot
    - judge robot is loaded once so it can use `run`
    - judge robot could run the condition in `as` on each tick
    - the goal condition would check that judge has "win" or something
    - the judge can interact with the environment, in particular, it can `say` what the player is doing wrong
      - e.g. `say "One sheep has drowned, please start over!"`

data/scenarios/Challenges/Ranching/gated-paddock.yaml
```diff
@@ -0,0 +1,484 @@
...
+  Start a wool industry from the local fauna.
+creative: false
+objectives:
+  - goal:
+      - |
+        You've homesteaded on a small island in the ocean.
+        It's time to gather resources to trade.
+      - |
+        You encounter some feral sheep (@). They slowly wander the island and eat grass.
+        Your mind wanders to textiles...
```
xsebek, line 14: I like this description. :+1: :smile: 9/10 capitalists would play again.

data/scenarios/Challenges/Ranching/gated-paddock.yaml
```diff
@@ -0,0 +1,484 @@
...
+
+          // Second, step to both sides
+          fwdCount <- observeLeftAndRight;
+          backCount <- observeLeftAndRight;
+          return $ fwdCount + backCount;
+          end;
+
+      def isStandingOnBridge =
+          onFence <- ishere "fence";
+          onGate <- ishere "gate";
```
xsebek, line 213: Honestly, there is not much of a motivation to place gates now. :man_shrugging: 

I had to race to save all sheep from drowning (which the description does not mention) I totally forgot that gates could be used to enclose sheep, but let me walk through.

Some things that would make the gate more useful:
- have one already placed - if it is free I will be happy to use it and save precious time
- make its recipe produce more gates from fences - this would allow me to have just gates all around the sheep
- require that there be at least one gate in the fence surrounding sheep

data/scenarios/Challenges/Ranching/gated-paddock.yaml
```diff
@@ -0,0 +1,484 @@
...
+                  return true;
+              } {
+                  // Make sure nothing's in the way before we place
+                  // our breadcrumb:
+                  x <- scan down;
+                  case x return (\y.
+                      // If we're on a water tile, get rid of
+                      // it with our special "drilling" recipe
+                      if (y == "water") {
+                          drill down;
```
xsebek, line 79: I don't get it, the sheep should not be able to reach water so why do you need to vaporise it? :confused:

xsebek (changes requested):
The game feels much more responsive now, great job @kostmo! :+1: 

Please just clarify the goal condition in the description or make it easier again.

I have a few other notes, but those are not blockers and I think we could merge this afterwards. :slightly_smiling_face:

data/scenarios/Challenges/Ranching/_gated-paddock/meandering-sheep.sw
```diff
@@ -0,0 +1,141 @@
...
+
+    b <- isBlockedOrFenced;
+    if b {} {
+      move;
+    };
+
+    // Sheep can drown.
+    hasWater <- ishere "water";
+    if hasWater {
+      say "whoops!";
```
xsebek, line 106: @kostmo are these the last words of the sheep? :cry:  They could say something funny or helpful instead.

data/scenarios/Challenges/Ranching/_gated-paddock/meandering-sheep.sw
```diff
@@ -0,0 +1,141 @@
...
+    x <- scan down;
+    case x return (\y.
+        if (y == "clover") {
+            harvest;
+            say "yum!"
+        } {};
+    );
+  );
+  r <- random 5;
+  if (r == 0) { say "baaa" } {};
```
xsebek, line 120: I noticed the sheep baa quite a lot and it is hard to for example read the recipes because it switches to the logger.

data/scenarios/Challenges/Ranching/gated-paddock/meandering-sheep.sw
xsebek: @kostmo did you try inlining the functions? I don't care as much about the sheep crashing into a fence as I would like to know if the `atomic` command is usable. :wink:

data/scenarios/Challenges/Ranching/gated-paddock.yaml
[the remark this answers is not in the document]
xsebek: IMO this would not be possible in time to save all sheep.

What do you think, @kostmo?

data/scenarios/Challenges/Ranching/gated-paddock.yaml
xsebek: Also, the judge robot could remember that it already checked the fence connection and not check again on the next tick until the player moves. This would be another speedup. :thinking:

data/scenarios/Challenges/Ranching/gated-paddock.yaml
```diff
@@ -0,0 +1,484 @@
...
+  - goal:
+      - |
+        Safe! Your sheep are now hungry.
+        Offer them something tasty and you may be rewarded.
+      - |
+        The sheep will move toward something edible on
+        an adjacent tile and will eat it if they walk over it.
+      - |
+        You may want to add a gate to the fence
+        to give yourself easier access.
```
xsebek, line 265: >  I also added two more goals and tweaked the map.

Neat! I have not saved all of them, so I will have to start over and actually finish the challenge. :open_mouth:

data/scenarios/Challenges/Ranching/gated-paddock.yaml
```diff
@@ -0,0 +1,484 @@
...
+    'c': [stone, cabin]
+    's': [grass, null, sheep]
+    '%': [grass, clover, null]
+    'H': [stone, pier, null]
+    '~': [dirt, water]
+  upperleft: [-34, 11]
+  map: |-
+    ~~~~.......~~~~~~~~~~~~~.......................~~~~~~
+    ~~.............~~~~.......%.%%%%%%..%%%%%.%.......~~~
+    ~........................%%%%%%%%%%%%%..%%%%.......~~
```
xsebek, line 464: Would it not be better to have the clovers more spread out?

Also they are not known, is that intended?

data/scenarios/Challenges/Ranching/_gated-paddock/meandering-sheep.sw
byorgey: Yeah, I was going to complain about that too.

data/scenarios/Challenges/Ranching/gated-paddock.yaml
byorgey: Ohh, I ran into exactly this scenario.  One sheep drowned and I did not realize that would prevent me from winning.  What's worse, when I completed the fence but was still sitting on top of the last fence I placed, it pegged my CPU at 100% and the game at 30 ticks per frame, 0.1 fps. =(  It was so sluggish I had to wait like 5-10 seconds just to switch to a different panel in the UI.  As soon as I moved off the fence (with great difficulty, waiting 5-10 seconds in between each UI action) the game suddenly sped up again.

byorgey:
I have now restarted four times and have not yet been able to get all the sheep in a fence before they drown.  It seems like the only way to win is to write a program that will automatically do all the actions of harvesting trees, making fence, and placing the fence so that it can happen fast enough to catch the sheep.  But this feels very irritating.

byorgey:
I also don't understand the point of the post puller and the fact that it leaves behind scrap wood that you cannot do anything with and *cannot pick up*.  It means that once you have placed a fence somewhere you will never be able to use that spot for anything else.  For example if you create an unbroken fence and then decide you want to place a gate in the fence, it is kind of difficult to do because you cannot place a gate on top of a pile of scrap wood.

Why not just let players pick up scrap wood?

byorgey:
Finally managed to beat it.  The other two objectives are fun, though after the fence building the last objective seemed too easy.  How about requiring a certain number of wool (say, 10?) instead of just one?

data/scenarios/Challenges/Ranching/_gated-paddock/meandering-sheep.sw
byorgey: Now that we have the Messages dialog which gets highlighted when there is a new message, maybe it's time to stop jumping to the logger every time there's a new message?  That's going to be irritating in more scenarios than just this one.

xsebek:
Btw. sorry for being so perfectionist @kostmo, please understand that I am simply trying to make this as good and fun as possible. :sweat_smile:  I have limited time due to work and school, so my comments might seem a little brisk.

If you ever feel too tired to deal with all of my nitpicks, just comment which one you want to do and others I can do myself. :wink:

data/scenarios/Challenges/Ranching/gated-paddock.yaml
kostmo: In the enclosure checker, I need to place a breadcrumb on the ground to mark a route that I've traversed once.  It is not possible to place an item on an already-occupied tile.  "Grab"-ing water does not remove it from the tile, so instead I "drill" it. :laughing:

data/scenarios/Challenges/Ranching/_gated-paddock/meandering-sheep.sw
kostmo: I've now made the baaa-ing less frequent, but I agree with @byorgey that we should have a way to disable jumping to the logger upon new messages.

data/scenarios/Challenges/Ranching/gated-paddock.yaml
kostmo: > is that intended?
Yes, it's intended that they are not known initially; I make an oblique reference in the goal text to find something "tasty".

I bunch them up as a convenience for the player to harvest.

data/scenarios/Challenges/Ranching/gated-paddock.yaml
kostmo: I have now relaxed the goal to just save one sheep.

kostmo: > the judge robot could remember

Is there a way to have a persistent "judge" robot that can execute an entire program in one tick?
Also, my judge-algorithm modifies/destroys the environment, so it would have to execute in a "sandboxed" state that is rolled back/discarded upon completion of its program.

What may be useful (and would be facilitated by #795) is if there existed a `god`-capability command that can check if a specific (named) goal has been already achieved.

data/scenarios/Challenges/Ranching/_gated-paddock/meandering-sheep.sw
kostmo: Added an allusion to a [song](https://en.wikipedia.org/wiki/Waltzing_Matilda).

data/scenarios/Challenges/Ranching/_gated-paddock/meandering-sheep.sw
byorgey: See https://github.com/swarm-game/swarm/pull/856.

kostmo:
> It seems like the only way to win is to write a program

Ah that's a pitfall of "writing to the integration test" as a scenario designer.  I tend to forget that the player will have to re-invent the program that I spent so long perfecting :)

I hope it's more reasonable now with the new single-sheep requirement.

> don't understand the point of the post puller

Fixed.  Using the `post puller` now lets one recover a `board`.

> How about requiring a certain number of wool

I've modified the goal to knit `sweater` from 3 wool.  I actually at first tried to code the goal to win directly on obtaining 3 wool, but the following goal evaluation code does not work.  A bug, perhaps?

      as base {
        woolCount <- count "wool";
        return (woolCount >= 3);
      };

Finally, for code reviews which may become large, may we adopt a practice of making every comment a "threaded" comment to make the replies easier to follow?  The way to do this in GitHub seems to be to attach the comment to a line of code (that line may have to be arbitrary).  The toplevel (non-code-attached) comments do not facilitate long reply chains well.

byorgey:
> I hope it's more reasonable now with the new single-sheep requirement.

Yes, indeed.  The most recent time I played one sheep pretty much immediately drowned itself but I was able to fence the other two.

I think this also solves my issue with the CPU being pegged at 100%, since that would only happen when the base is sitting on top of a fence, we just closed a gap, but you haven't won.  I guess that could still happen in theory if there are NO sheep inside the fence?  But that seems less likely.

> Fixed. Using the `post puller` now lets one recover a `board`.

Yes, that's much nicer, thanks!

> I've modified the goal to knit `sweater` from 3 wool.

I like it.

> I actually at first tried to code the goal to win directly on obtaining 3 wool, but the following goal evaluation code does not work. 

Hmm, strange.  Not sure what's going on there.

> Finally, for code reviews which may become large, may we adopt a practice of making every comment a "threaded" comment to make the replies easier to follow? The way to do this in GitHub seems to be to attach the comment to a line of code (that line may have to be arbitrary). The toplevel (non-code-attached) comments do not facilitate long reply chains well.

Ah, good idea.  I made top-level comments since they did not really pertain to a specific line of code, but you're right that the threading is much nicer for comments attached to lines of code.

byorgey (approved):
I think it works great now and all my concerns have been addressed!  Beat the most recent version in 6m 52s. :smile:

kostmo:
@xsebek can you update the "requested changes" status of your review?  As is, merging is blocked.

data/scenarios/Challenges/Ranching/gated-paddock.yaml
kostmo: For now I suppose the gate is not essential.  Perhaps in the future it would be a hidden/optional achievement (#797).  Or maybe another goal could be added that necessitates going in and out of the paddock frequently.  Like continuously delivering bundles of wool to the `pier`.

byorgey: It definitely feels to me like we could keep adding more objectives, like shipping sweaters or something. (Some little boats could come up to the pier and take any waiting boxes (#115) of sweaters... :smile: )

kostmo:
> > I actually at first tried to code the goal to win directly on obtaining 3 wool, but the following goal evaluation code does not work.
> 
> Hmm, strange. Not sure what's going on there.

Created #858

data/scenarios/Challenges/Ranching/gated-paddock.yaml
xsebek: > Is there a way to have a persistent "judge" robot that can execute an entire program in one tick?

We normally use system robots with the `as` command, which is as powerful as you can get.

data/scenarios/Challenges/Ranching/_gated-paddock/meandering-sheep.sw
```diff
@@ -0,0 +1,144 @@
...
+            } {};
+        } {};
+    );
+  );
+  r <- random 30;
+  if (r == 0) { say "baaa" } {};
+
+  hasClover <- has "clover";
+  if (hasClover) {
+    r <- random 10;
```
xsebek, line 127: It takes quite a while for the two sheep to produce enough wool, lowering the value to 8 seemed to be quite a bit faster.

xsebek (approved):
Thanks for the great challenge @kostmo!

image

I look forward to other ranching challenges. :cowboy_hat_face:

data/scenarios/Challenges/Ranching/_gated-paddock/meandering-sheep.sw
kostmo: Changed to `8`.

byorgey:
:tada: :rocket: :heart: 

Pretty sure 118 comments is a new record. :smile:

PR #837, opened 2022-11-06
Ignore scenario subdirs starting with underscore

byorgey:
Such subdirectories can be used to e.g. organize `.sw` files, and will be ignored when recursively loading scenarios.

This resolves an issue we noticed while reviewing #835, which uses a subdirectory to hold multiple `.sw` files for a scenario and was generating a lot of spurious warnings.

### DIFF
--- a/src/Swarm/Game/ScenarioInfo.hs
+++ b/src/Swarm/Game/ScenarioInfo.hs
@@ -57,7 +57,7 @@ import Data.Aeson (
   genericToEncoding,
   genericToJSON,
  )
-import Data.Char (isSpace, toLower)
+import Data.Char (isLower, isSpace, toLower)
 import Data.Function (on)
 import Data.List (intercalate, stripPrefix, (\\))
 import Data.Map (Map)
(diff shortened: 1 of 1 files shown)


byorgey:
This is probably one of the simplest ways to resolve the issue, but it's kind of a blunt approach.  If others are unhappy with this we could also consider doing something more refined like allowing lowercase things as long as they are listed in `00-ORDER.txt`, or ignoring all subdirectories that recursively do not contain any `.yaml` files.

src/Swarm/Game/ScenarioInfo.hs
```diff
@@ -287,7 +287,11 @@ loadScenarioDir em dir = do
   let morder' = filter (`elem` fs) <$> morder
   SC morder' . M.fromList <$> mapM (\item -> (item,) <$> loadScenarioItem em (dir </> item)) fs
  where
-  keepYamlOrDirectory = filter (\f -> takeExtensions f `elem` ["", ".yaml"])
+  -- Keep only files which are .yaml files or directories that start
+  -- with something other than a lowercase letter.
```
xsebek, line 291: Lowercase is a bit surprising, how about using some standard way to hide directories?

E.g. `.hidden`, `_private`, `internal_...` or something like that?

xsebek (approved):

src/Swarm/Game/ScenarioInfo.hs
byorgey: Ah, good point, I think a leading underscore would work well.  I don't think we want `.hidden` since it would be annoying having your OS hide the directories if you are using them to organize `.sw` files.

PR #839, opened 2022-11-06
Account for files that lack a dot-extension

kostmo:
It's possible to have files that lack a dotted extension.

This is more OCD than anything, since we should have control over which files are committed to this directory.

### DIFF
--- a/src/Swarm/Game/ScenarioInfo.hs
+++ b/src/Swarm/Game/ScenarioInfo.hs
@@ -49,7 +49,7 @@ import Control.Algebra (Has)
 import Control.Carrier.Lift (Lift, sendIO)
 import Control.Carrier.Throw.Either (Throw, runThrow, throwError)
 import Control.Lens hiding (from, (<.>))
-import Control.Monad (unless, when)
+import Control.Monad (filterM, unless, when)
 import Data.Aeson (
   Options (..),
   defaultOptions,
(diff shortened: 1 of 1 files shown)


src/Swarm/Game/ScenarioInfo.hs
```diff
@@ -289,9 +289,15 @@ loadScenarioDir em dir = do
...
-      (\f -> takeExtensions f == ".yaml" || (takeExtensions f == "" && head f /= '_'))
+  keepYamlOrPublicDirectory = filterM . isCatalogEntry
+
+  -- Whether the directory entry should be included in the scenario catalog.
+  isCatalogEntry d f = do
+    isDir <- doesDirectoryExist $ d </> f
+    return $
+      if isDir
+        then not $ "_" `isPrefixOf` f
+        else takeExtensions f == ".yaml"
```
xsebek, line 300: The description does not seem to match filtering YAML-only files. 😕

xsebek (approved):

src/Swarm/Game/ScenarioInfo.hs
kostmo: Do you mean the comment?

    -- Whether the directory entry should be included in the scenario catalog.

A directory "entry" is anything output by the `listDirectory` function.  So "directory" in that phrasing is referring to the argument to the `listDirectory` function.  But I recognize the ambiguity, and have re-worded the comment to:

    -- Whether the directory or file should be included in the scenario catalog.

PR #842, opened 2022-11-06
Fix: add first heard message to log

xsebek:
When the logger is empty, newly heard messages would have been ignored.


xsebek:
@kostmo @byorgey With this fix, all sheep shall **baaaaaa!**
Screenshot from 2022-11-07 00-19-45

kostmo (approved):

Issue #687, opened 2022-09-14
Location watch command


xsebek:
A neat optimisation would be to allow robots to let themselves be woken up when a selected location changes.

For example:
- in the game of life, the robots can wait until the neighbouring locations change
- in the Hanoi tower, robots only check the location where they are and a few known control locations

I propose adding a command `watch : int * int -> cmd ()`. If the specified location changes and the robot is
in the waiting robot queue, then it will be woken up like it reached the timeout. Otherwise, nothing happens.

---

Originally posted in the `#swarm` IRC channel:
```
<xsebek> I wonder if we could speedup swarm robots with something like "filewatch"
<xsebek> I am thinking about letting robots register to wait for changes to entity at some location
<xsebek> Not sure what the type would be 🤔
<xsebek> A simple solution would be to just wake up the robots on change
<xsebek> But to let robots register for multiple locations, just registering would not put them to waiting queue
<xsebek> So something like `repeat (watch (0,0); watch (3,1); wait 1024; doMyJob)`
<xsebek> Sorry I just feel bad for all the hanoi robots looking around, while I do nothing 😅
```

byorgey:
I think this is a great idea, but I'm not sold on the proposed semantics for `watch` (or at least I think it needs to be clarified).  My initial reaction was to simplify things and just have `watch : int * int -> cmd ()` immediately sleep and wait until the specified cell changes.  But I guess you want to be able to watch multiple cells, and wake up if any one of them changes?

Once a robot executes `watch`, will it be watching that cell forever, i.e. until it is destroyed/salvaged?  Or do the watches get cleared when a robot becomes idle?  Or when it gets woken up?  Or is there a way to explicitly un-watch a cell?

xsebek:
I would propose that watching a set of locations stops once the first location changes or the robot is woken up by timeout.

If the robot was waiting at that point it would be woken up and if it is doing something, nothing happens to it.

To keep it efficient we would need some data structure that has:
- an insertion of `RID` and `Location` and multiple insertions add more locations
- a query of `Location` that returns `Set RID`
- deletion of `RID` that removes the `RID` from all locations

---

But if there is no such thing, I can live with a `watch` that immediately sleeps - just add more robots and signalling `Location` to simulate OR. :grinning:

byorgey:
[...]

Issue #495, opened 2022-06-26
Change `run` to `import` and typecheck it properly


byorgey:
Elsewhere (*e.g.* #7, #208, #328) I have talked about how I don't like the `run` command and want to get rid of it.  However, I'd like to explore an alternative: maybe I don't like it because it's not typechecked properly!

In particular, right now `run : string -> cmd ()`, and typechecking it is utterly trivial: just make sure it is given a string as an argument.  However, this leads to a few problems:
- In #208 , we see that typechecking (and in particular name resolution) does not take into account any definitions that are made in a file which is run.  So if `x.sw` defines `go : cmd ()`, then `run("x.sw"); go` would work perfectly fine at runtime, but currently does not typecheck, with an error about `go` being unbound.
- In #328 , the type of the value in the file is not taken into account, giving us bizarre things being printed out like `true : ()`.
- In fact, the way `run` works now can break type safety entirely: #2068.

So here's an alternative we might contemplate:
- Rename `run` to `import` and make it special built-in syntax (requiring an argument which is a string constant) rather than a command constant.
- To typecheck `import "foo"`, read `foo.sw` from disk and typecheck it (unless it has already been imported before, in which can we can just look it up in a map).  The type of `import "foo"` is simply the type of the Swarm expression in `foo.sw`, and also binds the names of any definitions it contains.  In other words it should behave exactly as if the contents of `foo.sw` had been typed in place of the `import` (modulo the fact that files can be imported more than once but only need to be checked once).
    - Note this might require us to add `IO` to the type inference monad.  Or actually, maybe not --- perhaps we could add a phase in between parsing and type checking where we gather up all the `import` statements and load and parse their referents into a `Map`, recursively processing any imports encountered; then we pass that map to the typechecker too.  In any case it seems like it requires us to add `IO` to the `processTerm` function in one way or another.

This would still not be a true module system (in particular there's no way to do qualified imports, or hide names, etc.) but it would be way better than what we have, and it would at least solve #208 and #328 .

xsebek:
Thanks for writing this @byorgey, this sounds great and could be a huge improvement.

For example, my earlier error with not included sources (#411) would be caught at type checking!

byorgey:
OK, in that case I've marked it `S-Critical`!

TristanCacqueray:
Perhaps `import` could add the definitions to the environment using qualified names, for example, given a file `my.sw`:

```
// my.sw
def m2 = move; move end
[...]

PR #844, opened 2022-11-07
Faster sheep check

xsebek:
This seeks to improve the performance of contained sheep checks.

The results were surprising, discuss below. 😕

### DIFF
--- a/data/scenarios/Challenges/00-ORDER.txt
+++ b/data/scenarios/Challenges/00-ORDER.txt
@@ -3,3 +3,4 @@ teleport.yaml
 2048.yaml
 hanoi.yaml
 Mazes
+Ranching
--- a/data/scenarios/Challenges/Ranching/00-ORDER.txt
+++ b/data/scenarios/Challenges/Ranching/00-ORDER.txt
@@ -1,0 +1,1 @@
+gated-paddock.yaml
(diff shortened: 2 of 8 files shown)


kostmo:
FYI I just fixed a bug in my own solution and force-pushed my branch, so you may want to rebase.  I'll stop force-pushing for now while we look at this PR.

xsebek:
@kostmo I am tired of this and it's broken (I was switching between alternatives), but I will write down my notes:

- checking all sheep makes everything **3x slower** and it is unavoidable (while they all live free)

- following a path and checking for left, right and forward is **slower than expected**
     - opportunistic going forward if possible would likely speed this up

- parsing the condition check once **makes a huge difference, 2x to 3x speedup**
    - this should be solved by #495
    - in the meantime inlining the file to the condition works

- checking if there are fences added from both sides results in **no lag until the fence is finished**
    - this would be best implemented using #687
    - here I broke the solution, as I was tired, but I think it worked for me briefly 
    - this is an automated version of "say when ready" assuming that the goal is to connect to the original gate

---

@kostmo I think we should measure the solutions and choose the most correct and fastest one. 

Yours was pretty fast, but it only checked one sheep so it did not suffer from the 3x slowdown.

A properly optimized correct solution could run under 5s, but when it's around 10s it at least feels fast enough.

kostmo:
Awesome, thank you for the thorough investigation!  I will incorporate these optimizations.

xsebek:
Closing, as @kostmo implementation is now pretty optimised. :slightly_smiling_face:

PR #847, opened 2022-11-09
Add a `foldText` constant as a generic eliminator for the `text` type

byorgey:
`foldText` lets you do a left fold over the ASCII values of the characters in a `text`.  For example,
```
foldText (\c. \n. n+1) 0 "hello" == 5
```
To make this really useful we probably also need a way to convert back from ASCII values to `text`.

I wanted this because... actually, I forget why now.  I had a specific reason.  I'll add it when I think of it again.  In any case a fold in theory lets you do anything at all with a `text` value.

This is a draft PR.  Still need to add some tests, maybe also add an `int -> text` conversion as well.  Would also love feedback on anything at all.  Should we keep the `chars` function, now that you can just implement it with `foldText`?

### DIFF
--- a/src/Swarm/Game/CESK.hs
+++ b/src/Swarm/Game/CESK.hs
@@ -130,6 +130,10 @@ data Frame
     -- application) in environment @e@.  We will also push an 'FApp'
     -- frame on the stack.
     FArg Term Env
+  | -- | @FVArg v@ is like @FArg@ except that the argument is already
+    -- a value; when we are done evaluating the left-hand side of
+    -- the application we just directly apply it.
+    FVArg Value
   | -- | @FApp v@ says that we were evaluating the right-hand side of
     -- an application; once we are done, we should pass the resulting
     -- value as an argument to @v@.
@@ -341,6 +345,7 @@ prettyFrame :: Frame -> String
 prettyFrame (FSnd t _) = "(_, " ++ prettyString t ++ ")"
 prettyFrame (FFst v) = "(" ++ from (prettyValue v) ++ ", _)"
 prettyFrame (FArg t _) = "_ " ++ prettyString t
+prettyFrame (FVArg v) = "_ " ++ from (prettyValue v)
 prettyFrame (FApp v) = prettyString (valueToTerm v) ++ " _"
 prettyFrame (FLet x t _) = "let " ++ from x ++ " = _ in " ++ prettyString t
 prettyFrame (FTry t) = "try _ (" ++ prettyString (valueToTerm t) ++ ")"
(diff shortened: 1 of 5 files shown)


byorgey:
No, we can't.  For example, try writing something like `isLower` to check whether every letter in a `text` value is lowercase.  `split` lets you go through a `text` piece by piece but it doesn't really let you do anything with the characters themselves.  Here's a challenge list of functions that players ought to be able to write; I'd like to think about what minimal language features we need to include to be able to write these.  I'm not even fully happy with `foldText`, and as I mentioned it's even incomplete without some way to turn characters (represented as `int` values?) back into `text`.

- `isLower : text -> bool`
- `toLower : text -> text` to convert all letters to lowercase
- `capitalize : text -> text` to capitalize the first letter of a string
- `mapText : (int -> int) -> text -> text`
- `filterText : (int -> char) -> text -> text`
- `count : int -> text -> int` to count the number of occurrences of a particular character
- `append : text -> text -> text`
- `length : text -> int`
- `charAt : int -> text -> int` to get the character at a particular index
- `isPrefix/Suffix/InfixOf : text -> text -> bool`

Fundamentally, we need (1) some way to construct `text` values and (2) some way to eliminate them.  `foldText` is the canonical eliminator.  For constructing we have a few options.  We could provide a `cons : int -> text -> text` but the problem is that's very inefficient on Haskell `Text` values.  Other options might be to provide something like `buildText : int -> (int -> int) -> text` which takes a length and a function from each valid index to a character.  Or something like `unfoldText`.

xsebek:
@byorgey so the fold is a way to get `charAt`? I still do not see why a fold should be the API to convert characters to integers.

This seems backwards to me and unlike anything I know from Haskell/Python/C/...

I think that if we had `charAt` (and `length` to be efficient) all the other functions in the list and `foldText` could be implemented from what we have. My vision is to have an `import swarm` instead of adding 10 more reserved words. 🙂

byorgey:
> My vision is to have an import swarm instead of adding 10 more reserved words. 

Oh, I think you misunderstood me --- that is my vision too!  I am *not* suggesting we add all those things in my list as reserved words, just trying to figure out what are the *bare minimum language features* we need to be able to implement all of them in a reasonable way.

> I think that if we had `charAt` (and `length` to be efficient) all the other functions in the list and `foldText` could be implemented from what we have.

Thanks, I think this was the insight I needed!  `foldText` comes from thinking about text as a *sequence*, i.e. list, of chars.  It is the natural eliminator for something like `type String = [Char]` in Haskell.  However, Swarm doesn't have lists, so a list of chars is not the most natural way to think about `text`.  Even from an implementation perspective, we use `Text`, not `[Char]`.  The other way of thinking about text is as an *array* of chars, i.e. a function from index to char.  The natural eliminator for this perspective would be `charAt`+`length`: `length` tells you which indices are valid, and `charAt : int -> text -> int` can be thought of as turning a `text` value into an `int -> int` lookup function.  And the construction method that corresponds to this would be something like `buildText : int -> (int -> int) -> text`.

Notice that you can actually implement `split` in terms of `charAt` + `buildText`.  Something like
```
split = \k. \t. (buildText k (\i. charAt i t), buildText (chars t - k) (\i. charAt (i - k) t))
```
(but perhaps with a bit more detail to handle the case when e.g. `k` is bigger than the length of the text, etc.) Whether we should *actually* do this is open to debate, it might be nice to provide `split` as a built-in for convenience?

In any case, I will close this and open a new PR soon.

Issue #845, opened 2022-11-08
Normalizing .yaml files


kostmo:
I was looking into using `yq`[^1] to stuff content from a `.sw` script into the `program` field of a `.yaml` file.
This in fact works, with a bit of finagling[^2]:

    PROGRAM=$(cat data/scenarios/Challenges/Ranching/_gated-paddock/enclosure-checking.sw | sed -e 's/[[:blank:]]\+$//') yq -i '.objectives[0].condition = strenv(PROGRAM) | .objectives[].condition style="literal"' data/scenarios/Challenges/Ranching/gated-paddock.yaml

Then I was thinking that we may want to facilitate some kind of automatic modification, or perhaps auto-reformatting (ensuring normalization), of scenario files.  It is possible to do so in bulk:

    find data/scenarios -type f -name '*.yaml' -print0 | xargs -0 --max-args 1 yq --inplace

However, there are a couple of issues:
1. Indentation of lists gets deepened by one level (this is idempotent).  Both the Go and Python implementations of `yq` do this, unfortunately.  If we made a normalization pass across the whole repo to re-standardize on this indentation, it would be a non-issue going forward.  Diffs look like this:
    ```
       - in:
    -    - [1, hinge]
    -    - [1, fence]
    +      - [1, hinge]
    +      - [1, fence]
         out:
    -    - [1, gate]
    +      - [1, gate]
    ```
3. Maps that have a right edge that is not "flat" (i.e. the map makes use of *spaces* (` `) in the palette) get converted to double-quoted YAML style, messing up the formatting.

[^1]: The [Go implementation](https://mikefarah.gitbook.io/yq/), not the [Python implementation](https://pypi.org/project/yq/).  It is available via `snap` on Ubuntu:
        ```
        snap install yq
        ```
[^2]: There are two aspects in this command line invocation that ensure proper formatting of the program as an unquoted literal:
    1. There must be **no** trailing whitespace on any line of the input text (hence, using `sed`)
    2. The replacement text will adopt the pre-existing "style" of the input `.yaml` file for a given field.  So if for some reason (*this bit me, actually*) the field in the original `.yaml` file is double-quoted, then the program will get stuffed into double quotes, all on a single line, with escaped newlines (`\n`), **unless** `style="literal"'` is explicitly supplied to override the original style.

xsebek:
That is actually better than I expected - I thought `yq` would expand our “tuples as lists”.

[...]

PR #848, opened 2022-11-10
prepare maps for yaml normalization

kostmo:
Towards #845

`yq` requires that string literals do not have trailing spaces.
This PR replaces space characters with commas so that the right edge of the map rectangle is straight.


byorgey (approved):
Seems reasonable.

PR #849, opened 2022-11-10
apply normalization to all scenario files

kostmo:
Towards #845

Using the Go implementation of `yq`:

    find data/scenarios/Challenges -name '*.yaml' | xargs --max-args 1 yq --inplace

The changes produced by this normalization include:
* lists may have deeper indentation by one level
* word wrap is eliminated in `description` and `goal` string literals
* spaces after commas in lists

### DIFF
--- a/data/scenarios/Challenges/2048.yaml
+++ b/data/scenarios/Challenges/2048.yaml
@@ -4,13 +4,8 @@ author: Brent Yorgey
 description: Make 2048!
 objectives:
   - goal:
-    - OK, OK, it's not really the same as the classic "2048" game.
-      However, your goal is still to make 2048!  You start with a `1`
-      which regrows immediately when
-      it is harvested, so if you plant it, you can get as many as you want.  Your
-      job is to combine `1`s in order to make a `2048`
-      entity.
-    - "Hint: the `format` command can turn numbers into strings!"
+      - OK, OK, it's not really the same as the classic "2048" game. However, your goal is still to make 2048!  You start with a `1` which regrows immediately when it is harvested, so if you plant it, you can get as many as you want.  Your job is to combine `1`s in order to make a `2048` entity.
+      - "Hint: the `format` command can turn numbers into strings!"
     condition: |
       try {
         as base {has "2048"}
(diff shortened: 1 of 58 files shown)


data/scenarios/Challenges/2048.yaml
```diff
@@ -4,13 +4,8 @@ author: Brent Yorgey
...
 objectives:
   - goal:
-    - OK, OK, it's not really the same as the classic "2048" game.
-      However, your goal is still to make 2048!  You start with a `1`
-      which regrows immediately when
-      it is harvested, so if you plant it, you can get as many as you want.  Your
-      job is to combine `1`s in order to make a `2048`
-      entity.
-    - "Hint: the `format` command can turn numbers into strings!"
+      - OK, OK, it's not really the same as the classic "2048" game. However, your goal is still to make 2048!  You start with a `1` which regrows immediately when it is harvested, so if you plant it, you can get as many as you want.  Your job is to combine `1`s in order to make a `2048` entity.
```
xsebek, line 7: I do not have that wide monitor.

xsebek (changes requested):
We could make the files less likely to be changed by automated tool edits.

But _should we do it_ before we use it? If this PR does not include a Restyled.io or similar, then we will have to manually keep up the style. 😕

data/scenarios/Challenges/2048.yaml
kostmo: I noticed that the [Python implementation of `yq`](https://kislyuk.github.io/yq/) has a `--width` option to set line wrap for string literals.  The concern would be if it also ends up wrapping swarm-lang code.  I may have to research a bit more.

kostmo:
> But _should we do it_ before we use it?

I'm OK with postponing this normalization indefinitely until we get CI figured out and/or a pressing need for normalization arises.

data/scenarios/Challenges/2048.yaml
xsebek: Swarm lang ignores extra whitespace so that should not break.

Wrapping maps would be a problem, but those should already use the raw format.

kostmo: I tried the Python implementation of `yq` with mixed results:

1. The existing line wraps in `description` and `goal` keys are preserved
2. Quotes are removed from most keys when allowed, e.g.
    ```
    -    'λ': [grass, lambda]
    +    λ: [grass, lambda]
    ```
    but not
    ```
     ' ': [ice, water]
    ```
3. All comments are stripped from the `.yaml` file

Comments right now are fairly pervasive, e.g. in [this file](https://github.com/swarm-game/swarm/blob/9be30c5450f87f96834eb157dac89e065a64a215/data/scenarios/Testing/479-atomic-race.yaml#L1-L16).

An argument could be made that any information that we care about in the `.yaml` file should be programmatically extractable, i.e. assigned as a keyed value within the tree structure of the document.  If you buy that, we should create a new toplevel key called "developer-notes" or something, and move the toplevel `#` comments there.

byorgey:
I'm fine with the idea of standardizing on some normalized form in general, but adamantly opposed to merging this until and unless we have CI in place to enforce it.  Because otherwise it's just one more tool to install and one more thing to remember, and a barrier for new contributors.

@kostmo , did you take a look at restyled.io's `prettier-yaml` restyler? I wonder how well it does with our files.  If it works, it would definitely be the easiest option.  Links:
- https://docs.restyled.io/available-restylers/
- https://prettier.io/blog/2018/07/29/1.14.0.html

PR #851, opened 2022-11-11
Add full color mode option

### LINKED ISSUE #815: Sometimes colors don't work
**Describe the bug**
Rarely, upon launching the game, none of the color-styling shows up.
This causes grass in the map to be gray instead of green, and dialog buttons to have no highlighting.

**To Reproduce**
This has happened perhaps in 1/100 launches of the game (at least twice for me, so far).
Last time it happened, I had passed both the `--scenario` and `--run` options.

    stack run -- --scenario my-scenario.yaml --run my-script.sw

The contents of `my-script.sw` were as follows:
```
def plod = move; end

def rotate = turn left; end;

plod; rotate; plod;
```
Note the absence of the semicolon on the first line, if that may be relevant.  The program still ran, though.
The failure is nondeterministic; I ran it one time, and got no colors.  I quit and ran it again with no changes, and the colors were back.

**Expected behavior**
Normal coloring.

**Screenshots**
| Normal | Failure mode |
| --- | --- |
| Screenshot from 2022-11-01 10-38-26 | Screenshot from 2022-11-01 10-37-55 |

**Additional context**
Version: commit `1d928810`.

xsebek:
- closes #815

### DIFF
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -1,5 +1,6 @@
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE TemplateHaskell #-}
+{-# LANGUAGE LambdaCase #-}
 
 module Main where
 
@@ -14,11 +15,12 @@ import Swarm.App (appMain)
 import Swarm.DocGen (EditorType (..), GenerateDocs (..), PageAddress (..), SheetType (..), generateDocs)
 import Swarm.Language.LSP (lspMain)
 import Swarm.Language.Pipeline (processTerm)
-import Swarm.TUI.Model (AppOpts (..))
+import Swarm.TUI.Model (AppOpts (..), ColorMode(..))
 import Swarm.Version
 import Swarm.Web (defaultPort)
 import System.Exit (exitFailure, exitSuccess)
 import System.IO (hPrint, stderr)
+import Text.Read (readMaybe)
 
 gitInfo :: Maybe GitInfo
 gitInfo = either (const Nothing) Just ($$tGitInfoCwdTry)
@@ -115,8 +117,17 @@ cliParser =
   autoplay = switch (long "autoplay" <> short 'a' <> help "Automatically run the solution defined in the scenario, if there is one. Mutually exclusive with --run.")
   cheat :: Parser Bool
   cheat = switch (long "cheat" <> short 'x' <> help "Enable cheat mode. This allows toggling Creative Mode with Ctrl+v and unlocks \"Testing\" scenarios in the menu.")
-  color :: Parser Bool
-  color = switch (long "full-color" <> short 'c' <> help "Use full color mode.")
+  color :: Parser (Maybe ColorMode)
+  color = optional $ option colorModeParser (long "full-color" <> short 'c' <> metavar "MODE" <> help "Use none/8/16/full color mode.")
+  colorModeParser = Data.Foldable.asum
+    [ ColorMode8 <$ text "8"
+    , ColorMode16 <$ text "16"
+    , ColorMode240 <$> maybeReader (\case ('2':'4':'0':'_':w) -> readMaybe w; _ -> Nothing)
+    , FullColor <$ text "full"
+    , NoColor <$ text "none"
+    ]
+  text t = maybeReader (\x -> if x == t then Just x else Nothing)
+
 
 cliInfo :: ParserInfo CLI
 cliInfo =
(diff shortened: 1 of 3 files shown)


app/Main.hs
```diff
@@ -98,13 +108,15 @@ cliParser =
...
-  scenario = optional $ strOption (long "scenario" <> short 'c' <> metavar "FILE" <> help "Name of a scenario to load")
+  scenario = optional $ strOption (long "scenario" <> short 's' <> metavar "FILE" <> help "Name of a scenario to load")
   run :: Parser (Maybe String)
   run = optional $ strOption (long "run" <> short 'r' <> metavar "FILE" <> help "Run the commands in a file at startup")
   autoplay :: Parser Bool
   autoplay = switch (long "autoplay" <> short 'a' <> help "Automatically run the solution defined in the scenario, if there is one. Mutually exclusive with --run.")
   cheat :: Parser Bool
   cheat = switch (long "cheat" <> short 'x' <> help "Enable cheat mode. This allows toggling Creative Mode with Ctrl+v and unlocks \"Testing\" scenarios in the menu.")
+  color :: Parser Bool
+  color = switch (long "full-color" <> short 'c' <> help "Use full color mode.")
```
xsebek, line 119: I was considering giving more options or loading some settings file, but right now it does not seems as important.

If it becomes a nuisance or we can at least reproduce it, then we can come up with a more complex solution.

byorgey:
What does "full color mode" mean, exactly? And why would we not want it to be the default?

xsebek:
@byorgey I think it means showing "a full 24-bit truecolor".
- https://wiki.archlinux.org/title/Color_output_in_console
- https://hackage.haskell.org/package/vty-5.37/docs/Graphics-Vty-Attributes-Color.html#t:ColorMode

Not all terminals support this. Presumably, Brick/Vty try to autodetect this, but sometimes it does not work for an unknown reason.

I could allow setting any of the Vty colour values so that even the terminals limited to a lower colour range can experiment with this. Or maybe someone could have (not set) a TERM value which is interpreted by Vty as only having 8bit colours.

byorgey:
OK, sounds good.  But in what sense does this solve #815 ?

xsebek:
It allows you to override the default on the application side, isolating the problem.

Consider this scenario:

A hypothetical player named Peter will try to play Swarm, but it will show no colours.
Peter runs `swarm --help` and learns about `--full-color` option.
Once he runs Swarm again with the option, he will either:
1. see colours, confirming that the problem is with how Swarm detected his terminal capabilities (not setting TERM is on him though)
2. still see no colours, which means that he is using a Windows console or a revived IBM 2741
3. see colours 99% of the time and can not figure out why it does not work in rare cases, so he shrugs and goes on with his life, never thinking about it again (or he can write an Issue upstream)

byorgey:
Sure, that all sounds great, and I am in favor of adding this flag.  But in case (3) we still have exactly the same issue represented by #815.  So I still don't see why adding this flag resolves #815.  Unless the idea is "well, we did what we could and it's not worth worrying about a rare, transient, unreproducible issue".

app/Main.hs
```diff
@@ -115,8 +117,17 @@ cliParser =
   autoplay = switch (long "autoplay" <> short 'a' <> help "Automatically run the solution defined in the scenario, if there is one. Mutually exclusive with --run.")
   cheat :: Parser Bool
   cheat = switch (long "cheat" <> short 'x' <> help "Enable cheat mode. This allows toggling Creative Mode with Ctrl+v and unlocks \"Testing\" scenarios in the menu.")
-  color :: Parser Bool
-  color = switch (long "full-color" <> short 'c' <> help "Use full color mode.")
+  color :: Parser (Maybe ColorMode)
+  color = optional $ option colorModeParser (long "full-color" <> short 'c' <> metavar "MODE" <> help "Use none/8/16/full color mode.")
```
byorgey, line 121:
```suggestion
  color = optional $ option colorModeParser (long "color" <> short 'c' <> metavar "MODE" <> help "Use none/8/16/full color mode.")
```

src/Swarm/TUI/Model.hs
```diff
@@ -934,6 +938,8 @@ data AppOpts = AppOpts
     autoPlay :: Bool
   , -- | Should cheat mode be enabled?
     cheatMode :: Bool
+  , -- | Use full colours.
```
byorgey, line 941:
```suggestion
  , -- | What colour mode should be used?
```

xsebek:
@byorgey well, I am actually doing more than I could - I just added more options to select 8/16/full colour to see which one we really need. It turns out that with 8/16bit colours you get something like this:

image

I did this to better understand #815 and also to improve the player experience when dealing with any terminal colour issue.

---

I feel this should be enough to close our issue, given the response upstream. :slightly_smiling_face:

byorgey (approved):

Issue #516, opened 2022-06-30
Fix scenario loading to game state


xsebek:
See the discussion in #515:

> Changes to the entity map are permanent. For example, to reproduce:
>
> 1. Start Swarm in cheat mode
> 2. Load the new testing scenario containing Tardis ("Install one device")
> 3. Quit to the menu
> 4. Start a creative mode game
> 5. Execute create "Tardis": it should say "I've never heard of a Tardis", but instead, it works. That's because the Tardis was added to the entityMap when loading the test scenario, and it was never removed.
>
> We will have the same problem with recipes.
>
> The problem is that scenarioToGameState modifies the existing GameState instead of creating a brand new one from scratch. I am actually not sure right now whether there are good reasons for this. But I think the "right" solution, assuming it is not too difficult for some reason I can't think of right now, is to create a new GameState from scratch every time we load a new scenario. But if we don't want to actually reread the entities and recipes from disk every time, we will have to save copies of the "pristine" versions somewhere, which is what my comment do we need to keep an old entity map around??? was supposed to mean.

### Solution

Store the original game state (or at least entity map and recipes) and update it when loading the scenario.

Issue #689, opened 2022-09-16
Custom entities persist when quitting scenario


byorgey:
This is a placeholder for now to remind me to fix custom entities persisting across games by storing generic entitymap in new runtimestate, rebuilding GameState from scratch every time.

---

This is a duplicate of #516, but leaving it open since the discussion below about the new `RuntimeState` is relevant.

xsebek:
Is this #516?

Anyway, recently I added an extra `RuntimeState` to `AppState`, do we want to put the persisting parts of `GameState` there?

byorgey:
Oh, yes, this is #516 , I forgot I had already filed it.  Anyway, yes, seeing the new `RuntimeState` is exactly what made me think of this --- it's the perfect way to fix this issue.  Any state that should persist across games (such as the `EntityMap` loaded from `entities.yaml`) should go in the `RuntimeState`, and the `GameState` should be regenerated from scratch every time we load a new scenario.

kostmo:
Just encountered this bug, so adding my data point.  After playing the `2048` challenge, the inventory for a `Creative` game is full of powers of two:

![image](https://user-images.githubusercontent.com/261693/201751537-d939e041-9ebc-4ec7-a85c-abe590323540.png)

xsebek:
@kostmo interesting, they might be visible in inventory because they are _known_, I guess? :thinking:

This can serve as a manual test case for #516. 👍

PR #855, opened 2022-11-12
Reset `lastFrameTime` when starting a new scenario

### LINKED ISSUE #853: Newly initialized game tries to "catch up" after sitting in the menu for a while
**Describe the bug**
After spending a while letting the game sit in the menu, when starting a new scenario the game runs really fast for a little bit, as if it thinks it should have been running while the menu was active and it is now trying to "catch up".

**To Reproduce**
I first noticed this when testing out #835 (it was very noticeable since the sheep ran around really fast and quickly drowned themselves) but it can be seen in any scenario where the base has a `clock` installed.
- Start the app
- Let it sit in the menu for a while (e.g. a few minutes)
- Now start the Classic game and look at the clock in the upper right.  It will run fast for a little while and then at some point start running normally.

**Expected behavior**
The game should run at a normal speed regardless of how long you have spent in the menu.

byorgey:
Fixes #853.

### DIFF
--- a/src/Swarm/TUI/Model.hs
+++ b/src/Swarm/TUI/Model.hs
@@ -1039,7 +1039,8 @@ scenarioToAppState siPair@(scene, _) userSeed toRun = do
 
 -- | Modify the UI state appropriately when starting a new scenario.
 scenarioToUIState :: ScenarioInfoPair -> UIState -> IO UIState
-scenarioToUIState siPair u =
+scenarioToUIState siPair u = do
+  curTime <- getTime Monotonic
   return $
     u
       & uiPlaying .~ True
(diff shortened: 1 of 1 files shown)


src/Swarm/TUI/Model.hs
```diff
@@ -1053,3 +1054,4 @@ scenarioToUIState siPair u =
       & uiREPL .~ initREPLState (u ^. uiREPL . replHistory)
       & uiREPL . replHistory %~ restartREPLHistory
       & scenarioRef ?~ siPair
+      & lastFrameTime .~ curTime
```
xsebek, line 1057: This makes me think we should have something like a `UIGameState` that is reset on a new game. :thinking: 

It should not be possible to carry over things between games both in `GameState` and in `UIState`.

xsebek (approved):
:+1:

src/Swarm/TUI/Model.hs
byorgey: Or perhaps it means `lastFrameTime` really ought to live in `GameState` (once we fix #516 and #689).

Issue #858, opened 2022-11-12
Counting inventory does not work for goal criteria


kostmo:
Added test cases in #859 to demonstrate.

kostmo:
As demonstrated in #926, the following does not work as `condition` code:

    x <- has "tree"; not x

However, for multi-objective scenarios, #795 provides a workaround by defining a `NOT` prerequisite.

xsebek:
I wonder if this is a minimal example - i.e. if `not false` works as a condition.

xsebek:
The problem is with the _kind of value_ returned from the hypothetical condition check:
![image](https://user-images.githubusercontent.com/44544735/210170210-f06e8230-6bca-4220-9060-dc14c425a705.png)

PR #859, opened 2022-11-12
Add tests for item possession in goal criteria

kostmo:
Towards #858

### DIFF
--- a/test/integration/Main.hs
+++ b/test/integration/Main.hs
@@ -47,6 +47,7 @@ import System.Environment (getEnvironment)
 import System.FilePath.Posix (takeExtension, (</>))
 import System.Timeout (timeout)
 import Test.Tasty (TestTree, defaultMain, testGroup)
+import Test.Tasty.ExpectedFailure (expectFailBecause)
 import Test.Tasty.HUnit (Assertion, assertBool, assertFailure, testCase)
 import Witch (into)
 
@@ -183,6 +184,12 @@ testScenarioSolution _ci _em =
         , testSolution Default "Testing/490-harvest"
         , testSolution Default "Testing/504-teleport-self"
         , testSolution Default "Testing/508-capability-subset"
+        , testGroup
+            "Possession criteria (#858)"
+            [ testSolution Default "Testing/858-inventory/858-possession-objective"
+            , expectFailBecause "Known bug #858" $
+                testSolution Default "Testing/858-inventory/858-counting-objective"
+            ]
         , testGroup
             "Require (#201)"
             [ testSolution Default "Testing/201-require/201-require-device"
--- a/data/scenarios/Testing/00-ORDER.txt
+++ b/data/scenarios/Testing/00-ORDER.txt
@@ -13,4 +13,5 @@
 378-objectives.yaml
 684-swap.yaml
 699-movement-fail
+858-inventory
 710-multi-robot.yaml
--- a/data/scenarios/Testing/858-inventory/00-ORDER.txt
+++ b/data/scenarios/Testing/858-inventory/00-ORDER.txt
@@ -1,0 +1,2 @@
+858-possession-objective.yaml
+858-counting-objective.yaml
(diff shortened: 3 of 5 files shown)


data/scenarios/Testing/380-counting-objective.yaml
```diff
@@ -0,0 +1,36 @@
...
+  with the "count" function instead.
+creative: false
+objectives:
+  - goal:
+      - |
+        Get 1 tree.
+    condition: |-
+      as base {
+        itemCount <- count "tree";
+        return $ itemCount > 0;
```
kostmo, line 15: This does not succeed even after `grab`-bing the `tree`.

test/integration/Main.hs
```diff
@@ -183,6 +183,11 @@ testScenarioSolution _ci _em =
         , testSolution Default "Testing/490-harvest"
         , testSolution Default "Testing/504-teleport-self"
         , testSolution Default "Testing/508-capability-subset"
+        , testGroup
+            "Possession criteria (#858)"
+            [ testSolution Default "Testing/379-possession-objective"
+            , testSolution Default "Testing/380-counting-objective"
```
xsebek, line 189: @kostmo you need to write that you expect the test to fail:
```suggestion
            , expectFailBecause "Known bug #858" $
                testSolution Default "Testing/380-counting-objective"
```

xsebek (approved):

data/scenarios/Testing/380-counting-objective.yaml
```diff
@@ -0,0 +1,36 @@
+version: 1
+name: Evaluate count of an item
+description: |
+  The goal criteria is exactly the same as
+  379-possession-objective but evaluated
```
xsebek, line 5: @kostmo you know that we number the tests based on their issues, right? :raised_eyebrow: 

For example `394-build-drill.yaml` even has a link to `https://github.com/swarm-game/swarm/issues/394`.

These numbers are not issues:
- #380
- #379

data/scenarios/Testing/380-counting-objective.yaml
byorgey: I've been poking at this a bit.  Curiously, I've found that it does work correctly if the test is wrapped in a `try` block, like this:
```
      as base {
        try {
          itemCount <- count "tree";
          return $ itemCount > 0
        } { return false };
      };
```
In theory this should not make a difference.

byorgey: The runtime code for `has` and `count` is almost identical.  The one big difference between them which seems like it might be relevant is that `has` does not require any capabilities to use, whereas `count` requires a `counter`.  But in theory the condition checking robot should be able to use `count` whether `base` has a `counter` or not (and indeed, if I add a `counter` installed on `base` it does not change anything).

xsebek: Hmm, maybe base can not count or something?

byorgey: I really want to figure out what exceptions are being thrown but I'm not sure how to export that information. @xsebek, any ideas?

xsebek: @byorgey I just checked by printing exceptions from the hypothetical check and _there are none_.

I just added this:
```Haskell
sendIO $ putStr "ERROR: "
sendIO $ print m
```
here:
https://github.com/swarm-game/swarm/blob/e73a2849e39bee48a8875bc53fba1a654285d074/src/Swarm/Game/Step.hs#L137-L146

Weirdly this is also a transformation that fixes the condition:
```
t <- time;
if (t < 0) {return true} {
    as base {
        itemCount <- count "tree";
        return $ itemCount > 0;
    };
}
```

xsebek: Looking at debug logs (see the [debug branch](https://github.com/swarm-game/swarm/tree/debug-count)), the count is correct:
image

Issue #98, opened 2021-09-28
Array type


byorgey:
The language should also have a type for (size-indexed?) multidimensional arrays.  e.g. you could imagine some kind of high-powered sensor device that would return a 2D array of the robot's surroundings, allowing it to do things like pathfinding.

There's a lot of language design to be done here; this issue is something of a placeholder.  Should they be mutable? Immutable? What kinds of operations should they support? Should the type of an array reflect its dimension? Its size along each dimension? etc.?

xsebek:
We already have immutable fixed-size arrays. The trick to getting an array of size *n* is to use an *n+1* tuple:
- base case:
  ```haskell
  def get1 : (a * b) -> a
      = fst
  end
  def set1 : x -> (a * b) -> (x * b) 
      = \x.\t. x,snd t
  end
  ```
- inductive case (imagine *n+1* in place of *2* and missing type annotations):
  ``` haskell
  def get2 : (a * b * c) -> b
     = \t. let in1 = snd t in fst in1
  end
  def set2 : x -> (a * b * c) -> (a * x * c) 
      = \x.\t. let in1 = snd t in fst t,set1 x in1
  end
  ```
- after that we can get all *n* by induction:
  ```haskell
  def get3 : (a * b * c * d) -> c
      = \t. let in1 = snd t in get2 in1
  end
  def set3 : x -> (a * b * c * d) -> (a * b * x * d) 
      = \x.\t. let in1 = snd t in fst t,set2 x in1
  end
  ```
- so to get an array with 1,2,3 do:
  ```haskell
  > (1,2,3,())
  (1, (2, (3, ()))) : int * int * int * ()
  > get3 (1,2,3,())
  3 : int
  > set3 42 (1,2,3,())
  (1, (2, (42, ()))) : int * int * int * ()
  ```

OK, I admit this does not look great, even though we could auto-generate all the definitions with Bash.
I think the way to go, to make this more usable would be to add `type` synonyms, like we have `def`, but I can not imagine how hard or easy that would be.
Thoughts, @byorgey?

byorgey:
Hi @xsebek , yes, good observation!  This is definitely a nice way to get yourself fixed-sized, immutable arrays if all you have is pairs, and if players want to do this, by all means they should!

However, this would be unsatisfactory as the "blessed" way to have arrays, for several reasons:
[...]

Issue #218, opened 2021-10-12
Ideas for encounters with alien actors/robots


fryguybob:
I thought I should add an issue to put ideas about encountering alien robots.  Some dimensions to the idea:

- Hostile vs Peaceful
  - Attack vs Trade
  - Resources only available from aliens
  - Security exploits and reprogramming
  - Helpful reprogramming (help stranded robots find their way home)
- Decoding new languages
  - Languages aliens are using to communicate
  - Languages aliens use to program robots
  - Random language generation?

Is there a way to make this kind of idea "modular"?

fryguybob:
> * What do you mean by "modular" in this context?

It was more of a feeling and I was hoping that others would flesh that out 😜.  Here are some examples:

- Sometimes modularity is discovered later.  For example, there are a lot of magical commands, but what if they were just programs built on a lower level API.  You don't discover this until you build a disassembler, first to help you make a debugger, but then later you try disassembling what you thought was magic and it turns out is just a program (with lots of MagicHash of course).
- There is an idea in the game The Witness that is very modular, but I won't spoil that if people haven't seen it.
- I guess maybe what I fell is, if aliens were only "story" it wouldn't be enough.  There has to be something about the game that you learn or gain via that story.

byorgey:
It seems that no one has figured out the secret yet (there are clues in some of the item descriptions), but it's relevant for thinking about this issue, so I'll put it here for anyone assiduous enough to read old issue comments: the "alien planet" you have landed on is actually Earth, and the "aliens" are humans.  You, the protagonist of the game, are actually what humans would call an alien.  This is an idea I had a while ago.  Understanding this flipped perspective shouldn't be necessary to play the game, but I'm kind of intending for there to be subtle clues throughout the game so that perceptive players will be able to figure it out.

byorgey:
Maybe the alien robots can transmit messages to you in some kind of code.

byorgey:
*Communication* seems like it will be a central idea underlying any kind of interaction with aliens (or their technology).  I put a few vague ideas in #737.

kostmo:
One idea for a challenge/scenario: the "natives" have built large stone monuments that illustrate some kind of pattern.  E.g. rows of rocks.  To demonstrate your intelligence/prove your worth to them, you "extend the pattern" by building one more row in some designated area next to it.

Perhaps this culminates in a "generate the next prime number" challenge, to borrow from the plot of "Contact".

xsebek:
[...]

Issue #112, opened 2021-09-29
Add some structures to the world


byorgey:
Minecraft has villages; why can't Swarm have some randomly-generated structures?  Rivers, mazes, wheat fields...?  This depends on #29 , but once that's done we should be able to generate entire structures while generating a tile.  Generating structures that straddle tiles, or that take up multiple tiles, would be trickier, but probably also possible.

* Just as in Minecraft, we want the same structures to generate every time when the world starts from a given seed.
    * For example, this could mean somehow using some kind of deterministic noise function based on the seed to decide what kind of structures to generate and where, and then drawing them into the generated tile.
    * I'm also curious to play with wave function collapse (see https://www.youtube.com/watch?v=nkIMTc1wPaI and https://github.com/i-am-tom/world-building-in-haskell/blob/master/src/Generator/WaveFunctionCollapse.hs) as a way to generate local structures.  The idea would be to thread a PRNG through the collapse algorithm, seeded based on the world seed, to get deterministic results.

polux:
> mazes

I'd love it if the map would contain structure that require to solve some algorithmic puzzles in order to reach a reward (some rare resource for instance).

byorgey:
> > mazes
> 
> I'd love it if the map would contain structure that require to solve some algorithmic puzzles in order to reach a reward (some rare resource for instance).

Right, that's what I had in mind when I said mazes.  But with programmable entities I'm sure we could come up with other algorithmic challenges as well. Like maybe there is a pile of numbers (see #116 ) and you have to put them in sorted order to make a door unlock?

Anrock:
What if those structures are remnants/ruins of some other civilization/previous colonists and miraculously (not so in case of previous colonists) they have a data port (#116) that's compatible with player robots? 

So instead of pile of numbers on the floor which you have to sort there are doors / computers / etc with data ports and you can read / write to them to get challenge inputs and put challenge response? I guess attaching a script to an entity with data port (you can even put into yaml files instead of hardcoding into game) will be sorta easier than spawning number piles.

And for players interacting with a door or computer with data port is more obvious then sorting pile of numbers, IMO.

byorgey:
> there are doors / computers / etc with data ports and you can read / write to them to get challenge inputs and put challenge response?

That's a cool idea too!

Anrock:
Yeah. With data ports it's possible to port basically any puzzle/challenge. We can even incorporate some puzzles from EXA Punks since swarm robots can do anything EXAs can :)

[...]

PR #861, opened 2022-11-13
Add generic construction and elimination functions for `text` type

byorgey:
Following the discussion on #847.  In theory, any functions involving `text` can be implemented using `toChar`, `++`, `charAt`, and `chars`.

Note that building a length-n `text` value by repeatedly using `++` with singletons created via `toChar` takes $O(n^2)$ time.  This could be relevant if e.g. implementing something like `mapText` to convert every character to uppercase.  Hopefully this will be infrequent and not too bad for small `text` values.  For more performant creation of large `text` values we can wait for #98.


data/entities.yaml
```diff
@@ -326,12 +326,31 @@
...
+    index in a (0-indexed) `text` value, represented as the numeric
+    value of a Unicode
+    codepoint.
+  - |
+    `mktext : int -> (int -> int) -> text` constructs a text value
+    from a length and an indexing function. That is, `chars (mktext n
+    f) == n`, and `char i (mktext n f) == f i` as long as `0 <= i < n`.
+  - |
+    `split : int -> text -> text * text` splits a `text` value into
+    two pieces, one before the given index and one after.
```
byorgey, line 353: The `string` description is now extremely long.  I wonder if we want to consider making some other device(s) to provide some of these functions instead of having them all provided by `string`.

src/Swarm/Language/Typecheck.hs
```diff
@@ -527,6 +527,8 @@ inferConst c = case c of
   Concat -> [tyQ| text -> text -> text |]
   Chars -> [tyQ| text -> int |]
   Split -> [tyQ| int -> text -> (text * text) |]
+  Char -> [tyQ| int -> text -> int |]
+  MkText -> [tyQ| int -> (int -> int) -> text |]
```
xsebek, line 531: Hmm, I can not imagine actually using `mkText` - creating the whole text just based on length and indices sounds like it will require a rather complicated index-to-char function.

What I would really like to use though is the `singleton` so the special version:
```
\c. mkText 1 (\_.c)
```

I can see the theoretical appeal of the proposed functions, but I worry they will not be easy for players to use.

On the other hand, a `singleton` (or `toChar`) fills the space between `++` and `charAt`, so that you can roundtrip without outside references like ASCII or Unicode table.

test/unit/TestEval.hs
```diff
@@ -238,6 +239,23 @@ testEval g =
...
+            $ \i (NonEmpty s) ->
+              let i' = i `mod` length s
+               in T.concat ["char ", from @String (show i'), " ", tquote s]
+                    `evaluatesToP` VInt (fromIntegral (ord (s !! i')))
+        , testCase
+            "mkText 0"
+            ("mkText 0 (\\x. 97)" `evaluatesTo` VText "")
+        , testCase
+            "mkText 5"
+            ("mkText 5 (\\x. x+97)" `evaluatesTo` VText "abcde")
```
xsebek, line 253: I would prefer using `let ai = charAt 0 "a" in` than magic constants. Though hopefully, Unicode is the last encoding we will ever need. 😄

data/entities.yaml
xsebek: Agreed, I am not sure what would allow you to convert text to numbers though.

Maybe some rotating discs to allude to Caesar's cypher? 🤔

src/Swarm/Game/Step.hs
```diff
@@ -1484,6 +1494,18 @@ execConst c vs s k = do
...
+      [VInt i, VText t]
+        | i < 0 || i >= fromIntegral (T.length t) ->
+          raise Char ["Index", prettyValue (VInt i), "out of bounds for length", from @String $ show (T.length t)]
+        | otherwise -> return $ Out (VInt . fromIntegral . ord . T.index t . fromIntegral $ i) s k
+      _ -> badConst
+    MkText -> case vs of
+      [VInt n, f]
+        | n < 0 -> raise MkText ["Length must be nonnegative:", prettyValue (VInt n)]
+        | n == 0 -> return $ Out (VText "") s k
+        | otherwise -> return $ Out (VInt (n - 1)) s (FApp f : FMkText f (n - 1) [] : k)
```
xsebek, line 1507: I am not sure I understand what is going on here. Is this supposed to be more efficient than using `++` from Swarm?

test/unit/TestEval.hs
byorgey: Good idea. :heavy_check_mark:

src/Swarm/Game/Step.hs
byorgey: Yes, it is more efficient, but I agree at the very least I would need to write some comments explaining how this works.  Though as I mentioned elsewhere I am now thinking of getting rid of this for the moment --- though we will probably need something similar to implement the build function for arrays (#98).

data/entities.yaml
byorgey: Good idea, how about a https://en.wikipedia.org/wiki/Secret_decoder_ring ? :smile:

byorgey: The question is where you would ever get a decoder ring in the first place.  My idea is that this would be the "alien" character encoding (i.e. Unicode) so there would be no way for you to craft it from scratch, you would have to learn/obtain it somewhere first.  Though once you have one you would of course be able to copy it and make more.  Maybe there could be some kind of "temple" structure (#112) where you can find one, or maybe you could get one through trade (#218).

xsebek (approved):

PR #867, opened 2022-11-15
Allow retoggling robot visibility before 2 seconds

kostmo:
As an enhancement to #761, a second `Meta+h` keypress before the 2 seconds expire should immediately unhide the robots.

This is handy for the impatient/nearsighted among us, where toggling in quick succession makes the robot easier to find.

### DIFF
--- a/src/Swarm/TUI/Controller.hs
+++ b/src/Swarm/TUI/Controller.hs
@@ -264,13 +264,15 @@ handleMainEvent ev = do
     MetaChar 'h' -> do
       t <- liftIO $ getTime Monotonic
       h <- use $ uiState . uiHideRobotsUntil
-      if h >= t
-        then -- ignore repeated keypresses
-          continueWithoutRedraw
-        else -- hide for two seconds
-        do
-          uiState . uiHideRobotsUntil .= t + TimeSpec 2 0
-          invalidateCacheEntry WorldCache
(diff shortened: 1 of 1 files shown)


xsebek:
@kostmo that is the opposite of how it is supposed to work now. 🤨 

It is written so that you can keep robots hidden while holding/pressing the keys and not overload your CPU.

If you are impatient, why don’t you decrease the timer to one second? 😄

xsebek:
Or are you trying to make robots manually blink? 🤔

A blink attribute is much more efficient than _rendering the visible world every time you press key_. 😉

kostmo:
> are you trying to make robots manually blink?

Yes.  The two-second timeout is a good default, but from a UX perspective I think it's nice to have the option to make the robots flash on and off exactly when you press the key.  I think it's analogous to when people wiggle the mouse to find their cursor on the screen; the instant feedback confirms that it is the user's action causing the effect, not some other, potentially unrelated process.

xsebek:
@kostmo in that case you want a blinking attribute. I had a PR for it, but closed it, the branch should still be here.

With robot hiding, the feedback is that the keyboard shortcut hint is highlighted.

Why don’t you create a separate shortcut for temporary robot blinking? They are different things, not “enhancements”. 🙂

PR #869, opened 2022-11-17
More informative world cursor display

kostmo:
There is plenty of room in the lower-right corner of the World Display pane, so we may as well elaborate on the cursor-location info.

Here are screenshots where the clicked-on tile has one (terrain), two (entity + terrain), and three (robot + entity + terrain) elements, respectively:

Screenshot from 2022-11-16 23-24-33

Screenshot from 2022-11-16 23-24-24

Screenshot from 2022-11-16 23-23-56

### DIFF
--- a/src/Swarm/TUI/View.hs
+++ b/src/Swarm/TUI/View.hs
@@ -317,8 +317,8 @@ drawGameUI s =
   addCursorPos = case s ^. uiState . uiWorldCursor of
     Nothing -> id
     Just coord ->
-      let worlCursorInfo = drawWorldCursorInfo (s ^. uiState . uiShowRobots) (s ^. gameState) coord
-       in bottomLabels . leftLabel ?~ padLeftRight 1 worlCursorInfo
+      let worldCursorInfo = drawWorldCursorInfo (s ^. gameState) coord
+       in bottomLabels . leftLabel ?~ padLeftRight 1 worldCursorInfo
   -- Add clock display in top right of the world view if focused robot
   -- has a clock installed
   addClock = topLabels . rightLabel ?~ padLeftRight 1 (drawClockDisplay $ s ^. gameState)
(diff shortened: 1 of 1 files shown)


src/Swarm/TUI/View.hs
```diff
@@ -865,19 +887,13 @@ drawWorld showRobots g =
...
 
--- | Get the 'Display' for a specific location, by combining the
---   'Display's for the terrain, entity, and robots at the location.
-displayLoc :: Bool -> GameState -> W.Coords -> Display
-displayLoc showRobots g coords =
-  sconcat $ terrain NE.:| entity <> robots
- where
-  terrain = terrainMap M.! toEnum (W.lookupTerrain coords (g ^. world))
-  entity = maybeToList (displayForEntity <$> W.lookupEntity coords (g ^. world))
-  robots =
```
kostmo, line 876: These three lines were extracted to top-level functions so they can be re-used in my new cursor rendering logic.

src/Swarm/TUI/View.hs
```diff
@@ -317,7 +317,7 @@ drawGameUI s =
   addCursorPos = case s ^. uiState . uiWorldCursor of
     Nothing -> id
     Just coord ->
-      let worlCursorInfo = drawWorldCursorInfo (s ^. uiState . uiShowRobots) (s ^. gameState) coord
```
kostmo, line 320: Don't need this function argument anymore, because we won't hide the robot in the cursor display if the user pressed `Meta+h`.

src/Swarm/TUI/View.hs
```diff
@@ -326,9 +326,31 @@ drawGameUI s =
...
+        [ from $ show x
+        , from $ show $ y * (-1)
+        ]
+
+  tileMembers = terrain : mapMaybe merge [entity, robot]
+  tileMemberWidgets =
+    map (padRight $ Pad 1) $
+      concat $
+        reverse $
+          zipWith f tileMembers ["at", "on", "with"]
```
kostmo, line 345: The prepositions are ordered backwards here so that `zipWith` will only pick off just `at`, or  `on` plus `at`, if we only have one or two members on the tile.

byorgey (approved):
Looks good!

PR #873, opened 2022-11-21
world editor prototype

kostmo:
Towards #558
I was motivated to build this after finding that editing scenario maps directly in the YAML file is rather constraining.

## What I've implemented so far
* A small, collapsible panel to the left of the REPL containing World Editing status/operations.  Enter world-editing mode with CTRL+e to show the panel.
    * This works only in `--cheat` mode
* Terrain selection
    * A "picker"/"eye dropper" middle-click mechanism to select a terrain style to draw.
    * A pop-up selector to choose between the 5 different types of terrain.
* Drawing terrain with the left mouse button
* Saving a rectangular section of the world map (terrain only) to a file with CTRL+s
* Code organization
    * The complete state of the World Editor, including "painted overlays" of terrain, is contained within the `uiWorldEditor` field of `UIState` record.
    * The bulk of the World Editor functionality shall be in new modules
    * Some refactoring of `Controller.hs` and `View.hs` to extract functions utilized by the World Editor (towards #707)

## Vision

* The audience for this tooling is strictly envisioned to be Scenario authors.
    * Though, if we eventually allow swarm-lang to program the UI, there may be some common code to extract.
* The World Editor is intended to be compatible with a workflow of editing maps in text form within YAML scenario files.

# Demos
## Round-trip with random world

    stack run -- --scenario creative --seed 0 --cheat

Then Ctrl+e, tab down to the Save button, hit Enter to save the map
In another tab run:

    stack run -- --scenario mymap.yaml

Toggle between tabs to compare, observe the derived map is an identical 41x21 subset.

### DIFF
--- a/src/Swarm/Game/Scenario/Cell.hs
+++ b/src/Swarm/Game/Scenario/Cell.hs
@@ -41,14 +41,16 @@ data PCell e = Cell
 --   and optionally an entity and robot.
 type Cell = PCell Entity
 
+mkPCellJson :: ToJSON b => (a -> b) -> PCell a -> Value
+mkPCellJson modifier x = toJSON $
+  catMaybes
+    [ Just . toJSON . getTerrainWord $ cellTerrain x
+    , toJSON . modifier <$> cellEntity x
+    , listToMaybe []
+    ]
+
 instance ToJSON Cell where
-  toJSON x =
-    toJSON $
-      catMaybes
-        [ Just $ toJSON $ getTerrainWord $ cellTerrain x
-        , toJSON . (^. entityName) <$> cellEntity x
-        , listToMaybe []
-        ]
+  toJSON = mkPCellJson $ view entityName
 
 -- | Parse a tuple such as @[grass, rock, base]@ into a 'Cell'.  The
 --   entity and robot, if present, are immediately looked up and
@@ -83,10 +85,4 @@ type CellPaintDisplay = PCell EntityFacade
 
 -- Note: This instance is used only for the purpose of WorldPalette
 instance ToJSON CellPaintDisplay where
-  toJSON x =
-    toJSON $
-      catMaybes
-        [ Just $ toJSON $ getTerrainWord $ cellTerrain x
-        , toJSON <$> cellEntity x
-        , listToMaybe []
-        ]
+  toJSON = mkPCellJson id
--- a/src/Swarm/Game/Scenario/Cell.hs
+++ b/src/Swarm/Game/Scenario/Cell.hs
@@ -82,6 +82,9 @@ instance FromJSONE (EntityMap, RobotMap) Cell where
 -- World editor
 ------------------------------------------------------------
 
+-- | Stateless cells used for the World Editor.
+-- These cells contain the bare minimum display information
+-- for rendering.
 type CellPaintDisplay = PCell EntityFacade
 
 -- Note: This instance is used only for the purpose of WorldPalette
--- a/src/Swarm/Game/Scenario/Cell.hs
+++ b/src/Swarm/Game/Scenario/Cell.hs
@@ -42,12 +42,13 @@ data PCell e = Cell
 type Cell = PCell Entity
 
 mkPCellJson :: ToJSON b => (a -> b) -> PCell a -> Value
-mkPCellJson modifier x = toJSON $
-  catMaybes
-    [ Just . toJSON . getTerrainWord $ cellTerrain x
-    , toJSON . modifier <$> cellEntity x
-    , listToMaybe []
-    ]
+mkPCellJson modifier x =
+  toJSON $
+    catMaybes
+      [ Just . toJSON . getTerrainWord $ cellTerrain x
+      , toJSON . modifier <$> cellEntity x
+      , listToMaybe []
+      ]
 
 instance ToJSON Cell where
   toJSON = mkPCellJson $ view entityName
--- a/src/Swarm/Game/Scenario/Cell.hs
+++ b/src/Swarm/Game/Scenario/Cell.hs
@@ -41,6 +41,7 @@ data PCell e = Cell
 --   and optionally an entity and robot.
 type Cell = PCell Entity
 
+-- | Re-usable serialization for variants of "PCell"
 mkPCellJson :: ToJSON b => (a -> b) -> PCell a -> Value
 mkPCellJson modifier x =
   toJSON $
--- a/src/Swarm/Game/Scenario/Cell.hs
+++ b/src/Swarm/Game/Scenario/Cell.hs
@@ -6,15 +6,18 @@
 module Swarm.Game.Scenario.Cell (
   PCell (..),
   Cell,
+  CellPaintDisplay,
 ) where
 
-import Control.Lens hiding (from, (<.>))
+import Control.Lens hiding (from, (.=), (<.>))
 import Control.Monad (when)
 import Control.Monad.Extra (mapMaybeM)
+import Data.Maybe (catMaybes, listToMaybe)
 import Data.Text (Text)
 import Data.Vector qualified as V
 import Data.Yaml as Y
 import Swarm.Game.Entity
+import Swarm.Game.Scenario.EntityFacade
 import Swarm.Game.Scenario.RobotLookup
 import Swarm.Game.Terrain
 import Swarm.Util.Yaml
@@ -37,6 +40,15 @@ data PCell e = Cell
 -- | A single cell in a world map, which contains a terrain value,
 --   and optionally an entity and robot.
 type Cell = PCell Entity
+
+instance ToJSON Cell where
+  toJSON x =
+    toJSON $
+      catMaybes
+        [ Just $ toJSON $ getTerrainWord $ cellTerrain x
+        , toJSON . (^. entityName) <$> cellEntity x
+        , listToMaybe []
+        ]
 
 -- | Parse a tuple such as @[grass, rock, base]@ into a 'Cell'.  The
 --   entity and robot, if present, are immediately looked up and
@@ -62,3 +74,19 @@ instance FromJSONE (EntityMap, RobotMap) Cell where
     robs <- mapMaybeM name2rob (drop 2 tup)
 
     return $ Cell terr ent robs
+
+------------------------------------------------------------
+-- World editor
+------------------------------------------------------------
+
+type CellPaintDisplay = PCell EntityFacade
+
+-- Note: This instance is used only for the purpose of WorldPalette
+instance ToJSON CellPaintDisplay where
+  toJSON x =
+    toJSON $
+      catMaybes
+        [ Just $ toJSON $ getTerrainWord $ cellTerrain x
+        , toJSON <$> cellEntity x
+        , listToMaybe []
+        ]
--- a/src/Swarm/Game/Scenario/EntityFacade.hs
+++ b/src/Swarm/Game/Scenario/EntityFacade.hs
@@ -1,5 +1,11 @@
 {-# LANGUAGE DerivingVia #-}
 
+-- | Stand-in type for an "Entity" for purposes
+-- that do not require carrying around the entire state
+-- of an Entity.
+--
+-- Useful for simplified serialization, debugging,
+-- and equality checking, particularly for the World Editor.
 module Swarm.Game.Scenario.EntityFacade where
 
 import Control.Lens hiding (from, (.=), (<.>))
@@ -10,6 +16,9 @@ import Swarm.Game.Entity qualified as E
 
 type EntityName = Text
 
+-- | This datatype is a lightweight stand-in for the
+-- full-fledged "Entity" type without the baggage of all
+-- of its other fields.
 data EntityFacade = EntityFacade EntityName Display
   deriving (Eq)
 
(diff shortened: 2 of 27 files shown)


xsebek:
@kostmo looks nice, but what do you mean by "unambiguous mapping in the scenario file"? A palette is a JSON object, so its keys are unique.

kostmo:
I have found myself splitting some of the larger modules to get this PR to work.  Perhaps I should spread this work across two PRs: the first will have no World Editor code in it---it will just be moving pre-existing code into different modules (i.e. toward #707).  This should make the subsequent PR (adding World Editor functionality) easier to review.

If I do that, the splits on their own may not make a ton of sense unless viewed from the context of this current PR.  In other words, take a look at the "final" PR to see the direction I'm going with those splits.

src/Swarm/Game/Scenario/WorldDescription.hs
```diff
@@ -53,9 +63,108 @@ instance FromJSONE (EntityMap, RobotMap) WorldDescription where
...
+
+instance ToJSON WorldDescriptionPaint where
+  toJSON w =
+    object
+      [ "default" .= defaultTerrain w
+      , "offset" .= offsetOrigin w
+      , "palette" .= Y.toJSON paletteKeymap
+      , "upperleft" .= ul w
+      , "map" .= Y.toJSON mapText
+      ]
```
kostmo, line 98: Here's my custom JSON serializer

src/Swarm/Game/Scenario/EntityFacade.hs
```diff
@@ -0,0 +1,24 @@
...
+
+import Control.Lens hiding (from, (.=), (<.>))
+import Data.Text (Text)
+import Data.Yaml as Y
+import Swarm.Game.Display (Display)
+import Swarm.Game.Entity qualified as E
+
+type EntityName = Text
+
+data EntityFacadeData = EntityFacadeData EntityName Display
```
xsebek, line 13: Why not just name it `EntityFacade` as the module name suggests?

src/Swarm/Game/Scenario/Cell.hs
```diff
@@ -20,15 +20,27 @@ import Swarm.Util.Yaml
...
   , cellRobots :: [IndexedTRobot]
   }
   deriving (Eq, Show)
 
+-- | A single cell in a world map, which contains a terrain value,
+--   and optionally an entity and robot.
+type Cell = ProtoCell Entity
+
+-- Note: This is used only for the purpose of WorldPalette
+instance ToJSON Cell where
```
xsebek, line 35: Since it appears naturally in the world palette it seems OK to me.

This is like having a warning on gardening tools that they are only supposed to be used safely in the garden.

src/Swarm/Game/Scenario/Cell.hs
```diff
@@ -20,15 +20,27 @@ import Swarm.Util.Yaml
...
+  toJSON x =
+    toJSON $
+      catMaybes
```
xsebek, line 38: Why not:
```Haskell
toJSON = toJSON . fmap mkPaint
```

src/Swarm/Game/Scenario/WorldDescription.hs
```diff
@@ -53,9 +63,108 @@ instance FromJSONE (EntityMap, RobotMap) WorldDescription where
...
+  M.fromList $ concatMap (map genTuple) cellGrid
+ where
+  genTuple c =
+    (toKey terrainEfd, terrainEfd)
+   where
+    terrainEfd = cellToTerrainEntityNamePair c
+
+-- | Note that display characters are not unique
+-- across different entities! However, the palette KeyMap
+-- as a conveyance serves to dedupe them.
```
xsebek, line 118: Could this function be split to first create the updated palette?

src/Swarm/Game/Scenario/WorldDescription.hs
```diff
@@ -53,9 +63,108 @@ instance FromJSONE (EntityMap, RobotMap) WorldDescription where
...
+  unassignedCells :: M.Map TerrainEntityNamePair TerrainEntityFacadePair
+  unassignedCells =
+    M.withoutKeys entityCells $
+      Set.fromList $
+        map (toKey . snd) preassignments
+
+  unassignedCharacters :: Set.Set Char
+  unassignedCharacters =
+    -- TODO: Make this pool larger?
+    Set.difference (Set.fromList ['a' .. 'z']) $
```
xsebek, line 144: I would suggest filtering all the `Char` values based on their Unicode category.

Letters, numbers, symbols and marks should give us a big enough pool. 🙂

src/Swarm/Game/Scenario/EntityFacade.hs
xsebek: Also, do you need the whole `Display`? I would expect only needing a `Char`.

xsebek (commented):
Cool, I think I understand the reasoning better after reading through this PR. 👍 

Can't wait to use it to create new scenarios. 🎨 👨‍🎨

src/Swarm/Game/Scenario/WorldDescription.hs
kostmo: Yes, I was thinking about that.  And `Data.Char` has functions that allow filtering by category.  But what should the code look like that generates the initial list to be filtered?

src/Swarm/TUI/Editor/EditorView.hs
```diff
@@ -0,0 +1,172 @@
...
+    cornerCoords <- case we ^. editingBounds . boundsSelectionStep of
+      LowerRightPending cornerCoords -> Just cornerCoords
+      _ -> Nothing
+    pure $ EU.isOutsideTopLeftCorner cornerCoords coords
+
+drawLabeledEntitySwatch :: EntityFacadeData -> Widget Name
+drawLabeledEntitySwatch (EntityFacadeData eName eDisplay) =
+  tile <+> txt eName
+ where
+  tile = padRight (Pad 1) $ renderDisplay eDisplay
```
kostmo, line 148: Here's where I use the whole `Display` rather than just the `Char`

src/Swarm/Game/Scenario/EntityFacade.hs
kostmo: Turns out I need to use the `Display` [here](https://github.com/swarm-game/swarm/pull/873/files#r1053872827).

src/Swarm/TUI/Model/StateUpdate.hs
```diff
@@ -137,3 +145,20 @@ scenarioToUIState siPair u = do
...
+ where
+  myWorld = scenario ^. scenarioWorld
+  Location left top = ul myWorld
+
+  setNewBounds maybeOldBounds =
+    if mapHeight /= 0 && mapWidth /= 0
+      then Just newBounds
+      else maybeOldBounds
+
+  -- TODO Note inversion of horizontal and vertical coordinates
```
kostmo, line 159: use `locToCoords` or `coordsToLoc` here

src/Swarm/TUI/Editor/EditorView.hs
```diff
@@ -0,0 +1,172 @@
...
+        [ str "Area:"
+        , areaContent
+        ]
+
+  renderBounds (W.Coords (x1, y1), W.Coords (x2, y2)) =
+    str $
+      L.intercalate
+        " @ "
+        [ rectSize
+        , show (y1, x1) -- Note the inverted coords!
```
kostmo, line 100: use `locToCoords` or `coordsToLoc`

src/Swarm/TUI/Editor/Util.hs
```diff
@@ -0,0 +1,95 @@
...
+  Maybe (W.Coords, W.Coords) ->
+  W.World Int Entity ->
+  [[CellPaintDisplay]]
+getEditedMapRectangle _ Nothing _ = []
+getEditedMapRectangle worldEditor (Just (tl, br)) w =
+  map renderRow [yTop .. yBottom]
+ where
+  -- We swap the horizontal and vertical coordinate.
+  -- TODO If this is necessary, then what is mouseLocToWorldCoords for?
+  toWorldCoords (W.Coords (mx, my)) = W.Coords (my, mx)
```
kostmo, line 79: Replace this with `locToCoords` or `coordsToLoc`

src/Swarm/TUI/View.hs
```diff
@@ -843,51 +862,7 @@ drawWorld showRobots g =
       let w = ctx ^. availWidthL
           h = ctx ^. availHeightL
           ixs = range (viewingRegion g (fromIntegral w, fromIntegral h))
-      render . vBox . map hBox . chunksOf w . map (drawLoc showRobots g) $ ixs
-
--- | Render the 'Display' for a specific location.
-drawLoc :: Bool -> GameState -> W.Coords -> Widget Name
```
kostmo, line 849: Move these functions in another PR

src/Swarm/Game/Scenario/Cell.hs
kostmo: hmm, that doesn't typecheck?

byorgey:
How's this going?  Did you use it to create the map in #1136 ?

kostmo:
> How's this going? Did you use it to create the map in #1136 ?

I actually did #1136 the traditional way.  I can't claim that the world editor is super useful in its current state.

However, although the feature is incomplete, I believe there is value in merging it as-is.  It is non-invasive to the game UI and it makes for a decent foundation to resume work on a better editor UX when time allows.  And it has a bunch of utility functions that I could imagine re-using for other features.

Keeping on top of the merge conflicts from `main` is a chore, and if #1138 were developed in parallel it will be a major effort to keep rebasing.

byorgey:
@kostmo sorry it's taken me a while to get to it, but I'm now slowly making my way through this PR.  I have some suggestions, but ultimately I think merging it as a feature toggle makes a lot of sense.

byorgey (commented):
Sorry it's taken me so long to look at this.  In general I agree that we should try to merge this, perhaps at first in such a way that it's disabled by default.  I think my biggest meta-comment is that this is a lot of code with very few comments, which made it difficult to review.  I lost steam somewhere in `TUI.Editor.Controller` --- it was a lot of work to try to figure out what all the code was for.  I'm not so much looking for comments that explain what the code *does*, but comments explaining things like purpose, design choices, where the code fits into the big picture, etc.

src/Swarm/TUI/Controller.hs
```diff
@@ -421,6 +434,14 @@ handleModalEvent = \case
     Brick.zoom (uiState . uiModal . _Just . modalDialog) (handleDialogEvent ev)
     modal <- preuse $ uiState . uiModal . _Just . modalType
     case modal of
+      Just TerrainPaletteModal -> do
+        lw <- use $ uiState . uiWorldEditor . terrainList
+        newList <- refreshList lw
+        uiState . uiWorldEditor . terrainList .= newList
```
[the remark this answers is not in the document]
kostmo: Simplified with `Brick.zoom`

src/Swarm/Game/Scenario/WorldDescription.hs
```diff
@@ -66,3 +74,112 @@ paintMap pal = traverse (traverse toCell . into @String) . T.lines
   toCell c = case KeyMap.lookup (Key.fromString [c]) (unPalette pal) of
     Nothing -> fail $ "Char not in world palette: " ++ show c
     Just cell -> return cell
+
+------------------------------------------------------------
+-- World editor
+------------------------------------------------------------
+
+type TerrainEntityNamePair = (TerrainType, Maybe EntityName)
```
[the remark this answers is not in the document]
kostmo: Replaced with a parameterized type alias `TerrainWith a`

src/Swarm/Game/Scenario/WorldDescription.hs
```diff
@@ -66,3 +74,112 @@ paintMap pal = traverse (traverse toCell . into @String) . T.lines
...
+      ]
+   where
+    cellGrid = area w
+    suggestedPalette = palette w
+    (mapText, paletteKeymap) = prepForJson suggestedPalette cellGrid
+
+toKey :: TerrainEntityFacadePair -> TerrainEntityNamePair
+toKey = fmap $ fmap (\(EntityFacade eName _display) -> eName)
+
+getUniquePairs :: [[CellPaintDisplay]] -> M.Map TerrainEntityNamePair TerrainEntityFacadePair
```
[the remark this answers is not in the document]
kostmo: I have now added a comment to this (renamed) function.  Is this rationale OK?  Maybe there's a better way:
> -- | We want to identify all of the unique (terrain, entity facade) pairs.
-- However, "EntityFacade" includes a "Display" record, which contains more
-- fields than desirable for use as a unique key.
-- Therefore, we extract just the entity name for use in a
-- (terrain, entity name) key, and couple it with the original
-- (terrain, entity facade) pair in a Map.

src/Swarm/Game/Scenario/Cell.hs
```diff
@@ -62,3 +74,19 @@ instance FromJSONE (EntityMap, RobotMap) Cell where
     robs <- mapMaybeM name2rob (drop 2 tup)
 
     return $ Cell terr ent robs
+
+------------------------------------------------------------
+-- World editor
+------------------------------------------------------------
+
+type CellPaintDisplay = PCell EntityFacade
```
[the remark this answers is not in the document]
kostmo: Added comments

byorgey (approved):
:rocket:

PR #875, opened 2022-11-21
Make code reformatting command a script in the repo

kostmo:


xsebek (commented):
Is this a good idea? Editor integration or git hooks seem like better solutions.

Are there any other scripts that we would want to have?

kostmo:
> Is this a good idea? Editor integration or git hooks seem like better solutions.

I don't believe this script is mutually-exclusive with a git hook.  In fact, a git hook could invoke the `reformat-code.sh` script instead of copying the the `find` + `fourmolu` pipeline invocation inline.

> Are there any other scripts that we would want to have?

Yes, I could imagine having a script to stuff a `.sw` file inline into a Scenario `.yaml` file, or a script to automate cutting a new `swarm` release, etc.  Generally `scripts/` is just a useful directory to have.

CONTRIBUTING.md
```diff
@@ -223,7 +223,7 @@ You can run the formatter from the shell:
 ```bash
 cabal install fourmolu-0.4.0.0
 cd path/to/the/root/of/swarm/repo
-find src/ app/ test/ -name "*.hs" | xargs fourmolu --mode=inplace --cabal-default-extensions
```
xsebek, line 226: Lets have both here and mention that there is a script for convenience.

It’s only a oneliner and it does not change often.

Many shells and plugins make reusing previous commands easy, so it’s possible some developers (like me) will prefer the explicit command.

xsebek (approved):

xsebek:
> I could imagine having a script to stuff a .sw file inline into a Scenario .yaml file

@kostmo standardizing workarounds that make it harder to maintain source files _is precisely what I want to avoid._ 😆 But the other examples make sense. 👍

PR #876, opened 2022-11-21
use nested sum type to simplify panel enumeration

kostmo:
This trick facilitates use of `listEnums` and more concise pattern matching.

Towards #873.

### DIFF
--- a/src/Swarm/TUI/Controller.hs
+++ b/src/Swarm/TUI/Controller.hs
@@ -49,6 +49,7 @@ import Control.Carrier.State.Lazy qualified as Fused
 import Control.Lens
 import Control.Lens.Extras (is)
 import Control.Monad.Except
+import Control.Monad.Extra (whenJust)
 import Control.Monad.State
 import Data.Bits
 import Data.Either (isRight)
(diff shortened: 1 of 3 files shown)


src/Swarm/TUI/Controller.hs
```diff
@@ -303,23 +304,24 @@ handleMainEvent ev = do
...
-      setFocus $ case n of
+      flip whenJust setFocus $ case n of
         -- Adapt click event origin to their right panel.
         -- For the REPL and the World view, using 'Brick.Widgets.Core.clickable' correctly set the origin.
         -- However this does not seems to work for the robot and info panel.
         -- Thus we force the destination focus here.
-        InventoryList -> RobotPanel
-        InventoryListItem _ -> RobotPanel
-        InfoViewport -> InfoPanel
-        _ -> n
```
kostmo, line 314: Probably need to fix this; currently focusing doesn't work properly when clicking on the REPL or World pane.

kostmo: Now fixed.

src/Swarm/TUI/Controller.hs
```diff
@@ -292,34 +293,37 @@ handleMainEvent ev = do
       | s ^. uiState . uiCheatMode -> gameState . creativeMode %= not
     MouseDown n _ _ mouseLoc ->
       case n of
-        WorldPanel -> do
+        FocusablePanel WorldPanel -> do
           mouseCoordsM <- Brick.zoom gameState (mouseLocToWorldCoords mouseLoc)
           uiState . uiWorldCursor .= mouseCoordsM
-        REPLInput -> do
-          setFocus REPLPanel
```
kostmo, line 299: This is moved to line `313`.

src/Swarm/TUI/Model.hs
```diff
@@ -235,6 +234,12 @@ data Name
     RobotPanel
   | -- | The info panel on the bottom left.
     InfoPanel
+  deriving (Eq, Ord, Show, Read, Bounded, Enum)
+
+-- | 'Name' represents names to uniquely identify various components
+--   of the UI, such as forms, panels, caches, extents, and lists.
+data Name
+  = FocusablePanel PanelFocusable
```
byorgey, line 242: Why are the names of these two things backwards from each other?  There is no technical reason they can't have the same name and I would never be able to keep straight which is which.

byorgey (approved):
Looks good.  I assume the purpose is that now we can add new focusable panels without having to touch quite as many places in the code, since e.g. in some places you can just say `listEnum` and it will pick up all the focusable panels?

kostmo:
> I assume the purpose is that now we can add new focusable panels without having to touch quite as many places in the code, since e.g. in some places you can just say `listEnum` and it will pick up all the focusable panels?

Yes, exactly.  I add a new panel in #873.

src/Swarm/TUI/Model.hs
kostmo: Ah you're right, since types and data constructors are different namespaces. I've consolidated on `FocusablePanel`.

src/Swarm/TUI/Controller.hs
```diff
@@ -292,34 +293,37 @@ handleMainEvent ev = do
...
       case focusGetCurrent fring of
-        Just REPLPanel -> handleREPLEvent ev
-        Just WorldPanel -> handleWorldEvent ev
-        Just RobotPanel -> handleRobotPanelEvent ev
-        Just InfoPanel -> handleInfoPanelEvent infoScroll ev
+        Just (FocusablePanel x) -> ($ ev) $ case x of
+          REPLPanel -> handleREPLEvent
+          WorldPanel -> handleWorldEvent
+          RobotPanel -> handleRobotPanelEvent
+          InfoPanel -> handleInfoPanelEvent infoScroll
```
kostmo, line 326: Here's one more benefit of a dedicated sum type for Panels: we can benefit from exhuastiveness checking from the compiler here.  The catch-all `_` case no longer allows us to forget about a `Panel`, it instead handles all the non-panel `Name`s.

PR #879, opened 2022-12-05
Decompose model module

kostmo:
Towards #707.

Halves the size of `Model.hs` by splitting into smaller modules.

There are no functional changes in this PR---only code relocation.


byorgey (approved):
How did you decide which things to leave in `Swarm.TUI.Model` and which things to split out?  It seems somewhat arbitrary to me.  Should we split out everything into submodules and just have `Swarm.TUI.Model` re-export everything?

kostmo:
> How did you decide which things to leave in `Swarm.TUI.Model`

My criteria was simply size of the related content.  If there was "not much" code for a particular theme, I didn't bother to pull it out.  So `Model.hs` is still an assortment of unrelated but small things.

I'm not sure there's much utility in having `Swarm.TUI.Model` re-export everything; all of the model-centric code remains underneath the `Swarm.TUI.Model*` hierarchy, so it's findable when one is casually browsing for it, and it's equally easy to find a specific `data` definition in VS Code (e.g. CTRL+SHIFT+f "data ModalType"), independent of the module arrangement.

PR #880, opened 2022-12-05
Split Scenario.hs

kostmo:
Partially towards #707, although `Scenario.hs` is not that large.  Primarily this refactoring is in anticipation of #873 (see [comment](https://github.com/swarm-game/swarm/pull/873#issuecomment-1336348773)).

Just like #879, there are no functional changes in this PR---only code relocation.

### DIFF
--- a/src/Swarm/Game/Scenario.hs
+++ b/src/Swarm/Game/Scenario.hs
@@ -56,25 +56,18 @@ import Control.Algebra (Has)
 import Control.Carrier.Lift (Lift, sendIO)
 import Control.Carrier.Throw.Either (Throw, throwError)
 import Control.Lens hiding (from, (<.>))
-import Control.Monad (filterM, when)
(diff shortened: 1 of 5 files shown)


src/Swarm/Game/Scenario/Cells.hs
```diff
@@ -0,0 +1,55 @@
+{-# LANGUAGE DerivingVia #-}
+{-# LANGUAGE OverloadedStrings #-}
+
+module Swarm.Game.Scenario.Cells (
```
byorgey, line 4: I think we typically use singular nouns for module names (e.g. `Swarm.Game.Entity`, `Swarm.Game.Recipe`, `Swarm.Language.Context`, etc.).  I would prefer this to be named `Swarm.Game.Scenario.Cell`.

byorgey (approved):
Looks good, though I think we should rename `Cells` to `Cell`.

PR #881, opened 2022-12-05
Extract utils from Controller and View

kostmo:
Just like #880, this refactoring is in anticipation of #873 (see [comment](https://github.com/swarm-game/swarm/pull/873#issuecomment-1336348773)) for ease of review, and there are no functional changes in this PR.

Also towards #707.


src/Swarm/TUI/Controller/ControllerUtils.hs
```diff
@@ -0,0 +1,24 @@
+{-# LANGUAGE PatternSynonyms #-}
+
+module Swarm.TUI.Controller.ControllerUtils where
```
byorgey, line 3: Why not just name this `Swarm.TUI.Controller.Utils`?

byorgey (approved):
Why not `Controller.Utils` and `View.Utils`?  I'm not sure what the benefit is of repeating words as in `Controller.ControllerUtils`.

src/Swarm/TUI/Controller/ControllerUtils.hs
kostmo: > Why not `Controller.Utils` and `View.Utils`? I'm not sure what the benefit is of repeating words as in `Controller.ControllerUtils`.

In previous projects that I've worked in, having multiple modules of the same name throughout the hierarchy caused issues for tooling.  But VS Code and our Swarm development workflow doesn't seem to have an issue with it, so I'm happy to name them all `Utils`.

Issue #883, opened 2022-12-06
Way for a robot to lose movement ability


kostmo:
**Is your feature request related to a problem? Please describe.**

Currently, once an inventory item is `install`ed, onto a robot, that robot will have the provided abilities forever.  This limits the possibilities for contriving interesting scenarios with limited resources.

In particular, having a robot lose its ability to move would be an interesting play mechanic.

This could be related to all of the following:
* taking damage (#89)
* exhausting fuel (#550), or
* getting stuck in a swamp (#205)

Or, a robot may voluntarily give up its "treads" to cannibalize them for parts.

**Describe the solution you'd like**
Provide a command named `uninstall` or `unequip` that restores the installed item back into the inventory.

    unequip : text -> cmd unit

**Describe alternatives you've considered**
Provide a command called `disable` that destroys an installed device (does not restore it back to inventory) and requires the (e.g.) `aggressor` capability.

    disable : robot -> text -> cmd unit

byorgey:
I like the idea of an `uninstall`/`unequip` command.  I especially like the idea of having *e.g.* just one or two `treads` and having to share them between multiple robots.

Though you could definitely get yourself stuck using it, *e.g.* imagine that you `uninstall` your `3D printer` and then `give` it to another robot.  Now there's no way to get it back because *e.g.* you can't `build` another robot to go `salvage` it for you.  But maybe that is OK!

I was going to say it should be named `uninstall` in parallel with `install`.  But then I realized, perhaps "install" is something you do to another robot, whereas "equip" is something you do to yourself?  I don't know, I think players would still be confused by having `install` but `unequip`.  I don't think there's any need to also provide an `equip : text -> cmd unit` command since you can always `install self`.

kostmo:
> Though you could definitely get yourself stuck using it

I think that "getting stuck" should be normalized as a *possibility* in the game, especially since we have a dedicated button to restart the scenario.  Unlike some other games where it might be quite frustrating to "lose progress" involving feats of precise manual timing, our progress is stored in code and can be recalled from the REPL history.

> I was going to say it should be named `uninstall` in parallel with `install`

I think it makes sense to have `equip`/`unequip` have a type signature of `text -> cmd unit` whereas `install`/`uninstall` have `robot -> text -> cmd unit`.

Though TBH I've only ever used `install` on `self`.  And the "proximity" requirements of performing `install` on another robot aren't clearly documented (see #884).

byorgey:
[...]

Issue #550, opened 2022-07-06
Resource depletion by movement (e.g. fuel, water spills) 


byorgey:
Aside from having to build a boat, currently water is extremely easy/boring to get. A random idea I just had that might make collecting water a bit more interesting: what if any water in the inventory spills when a robot moves? And also when placing water on the ground it just spills.  So you couldn't just send a robot to grab some water and bring it back.

- You could build a pipeline of robots to pass the water along, bucket brigade style. (See https://github.com/byorgey/swarm-defs/blob/main/pipe.sw )
- Or you could have robots sit on top of the water in a boat, programmed to `make` things that require water.
- Maybe there could also be some sort of "tank" device which can be installed so that water no longer spills when moving.  But that also seems kind of boring, it would just be yet one more device you need to build.

kostmo:
How about ice that melts as you transport it?  Would make for fun allusion to The Mosquito Coast when they haul an ice block up a mountain to impress the natives.

kostmo:
Could this be a way to evaluate (or provide more immersive justification) a shortest path (#608) solution?  Provide a known amount of fuel (equal to the a priori-known shortest path length)

byorgey:
I like that idea! The difficulty I can see is that *finding* a shortest path takes a lot more fuel than *following* a shortest path.  So I'm not sure how to evaluate/justify a shortest path solution that way.

Hmm, but maybe... how about if there's a flower in the middle of a maze, and you're supposed to water it, but you only have an amount of water equal to the length of a shortest path plus 1 (assuming one unit of water spills per move).  So to solve the challenge you would have to first *find and mark* a shortest path (without carrying any water) then finally send a robot to traverse the found path carrying the water.

xsebek:
So to make this concrete, is the proposal that the count of liquid in inventory would decrease while moving?

To me that sounds a bit surprising, I would just expect it to leak all the time when it is not in a container.
Realistically I would expect robots to spill the water while giving it to others, sort of like a tax on the transfer.

To be fair, this is easy to implement, just remove one of each liquid inventory per `Move`.

---

I like the idea that the robot crafting with water would need to stand on it (in a boat). I guess the rule would go something like:

> If you have a grabber and scanner installed and are standing on a grabbable entity that is part of the recipe, then the recipe requirements are reduced by one of the said entity or infinity if the entity has the `infinite` property.

Implementing this rule only requires adding a bit of code to the `Make` step.

---

[...]

Issue #545, opened 2022-07-04
Highlight commands in descriptions


xsebek:
I want the commands in descriptions (entities, goal) to be highlighted instead of surrounded with ticks.

For shorter descriptions, it does not matter, but with longer ones like the goal or grabber device, I find it hard to read.

This would also be a good opportunity to use backticks to disambiguate them from normal ticks that English uses and be compatible with Markdown syntax.

As a small code refactoring, we could call the attribute "highlight" instead of using the robot attribute. :sweat_smile:

*update, 15 May 23*: Collecting some related links here for easy access: #309, #574, https://github.com/jtdaugherty/brick/issues/400, #1106

byorgey:
Related: #309 .

xsebek:
Just to be clear about the higlight attribute, there is a `highlightAttr` that means cyan foreground and `robotAttr` that means bold white foreground, which is used as bold for some table headers. I think it is funny, but not very readable. :sweat_smile:

byorgey:
Yeah, we should add a new `bold` attribute or something.  Agree we should not use `robot` attribute for so many things!

byorgey:
Note, the difficulty with applying attributes to things in descriptions is that we have to give text to a function that does the text wrapping and constructs a widget.  So we have no opportunity to apply formatting to individual words.  This might require some kind of upstream PR to `brick` to give us the functionality we need.

xsebek:
Well at least the triple backtick examples should be doable with separate widgets.

I wonder if we could compose widgets so that they would still look like text and create our `markdownTxt` function to do it where relevant. That might be useful enough to publish as an open-source library. :smile:

byorgey:
> Well at least the triple backtick examples should be doable with separate widgets.

Maybe?  I am not 100% sure how easy that will be.



> I wonder if we could compose widgets so that they would still look like text and create our `markdownTxt` function to do it where relevant.

I'm not sure I understand what you mean.

xsebek:
I thought we could layout parts of the text to look like one text.

[...]

PR #885, opened 2022-12-06
A "bucket brigade" of robots

kostmo:
Demo with:

    stack run -- --scenario scenarios/Challenges/bucket-brigade.yaml --autoplay

The idea of this scenario will be to form a "bucket brigade", where, in lieu of mobility, robots form a chain to pass items along back to the base.

There needs to exist one set of "treads" so that the robots may distribute themselves into a line.  In this scenario, a single `treads` is shared and passed between each of the robots.

However, once the robots have arranged themselves, all of the robots still have "treads" equipped.
So to contrive a circumstance where the "bucket brigade" is necessary, we need some way to inhibit the robots' movement (see #883).

### DIFF
--- a/data/scenarios/Challenges/_bucket-brigade/brigade.sw
+++ b/data/scenarios/Challenges/_bucket-brigade/brigade.sw
@@ -1,0 +1,94 @@
+def doN = \n. \f. if (n > 0) {f; doN (n - 1) f} {}; end;
+
+def forever = \c. c; forever c; end;
+
+def pollUntilHas = \item.
+    itemInInventory <- has item;
+    if itemInInventory {} {
+        pollUntilHas item;
+    };
+    end;
+
+def pollUntilCount = \targetCount. \item.
+    itemCount <- count item;
+    if (itemCount < targetCount) {
+        pollUntilCount targetCount item;
+    } {};
+    end;
+
+def waitToGiveUntilHas = \item. \recipient.
+    pollUntilHas item;
+    give recipient item;
+    end;
+
+/** The "depth" is the number of spaces between
+the root robot and the target item.
+It is one minus the distance to the target item.
+*/
+def recurseUntilDepth = \depth.
+    let item = "coal lump" in
+    child <- build {
+
+        // NOTE: Treads are auto-installed via the implicit "require"
+        move;
+        unequip "treads";
+
+        pollUntilCount depth "repro kit";
+
+        // Use the recipe from "repro kit"
+        make "3D printer";
+
+        equip "solar panel";
+        equip "grabber";
+        equip "3D printer";
+        equip "dictionary";
+        equip "branch predictor";
+        equip "comparator";
+        equip "calculator";
+        equip "lambda";
+        equip "strange loop";
+        equip "string";
+        equip "clock";
+        equip "counter";
+        equip "mirror";
+        equip "scanner";
+        equip "logger";
+        equip "net";
+
+        if (depth > 1) {
+            recurseUntilDepth $ depth - 1;
+            forever $ waitToGiveUntilHas item parent;
+        } {
+            make "bucketwheel excavator";
+            forever (
+                drill forward;
+                give parent item;
+            );
+        };
+    };
+
+    let kitGiftCount = depth in
+    // say $ "Going to give child " ++ (format kitGiftCount) ++ " repro kits.";
+    doN kitGiftCount $ give child "repro kit";
+
+    end;
+
+def makeBriquette =
+    let briquette = "coal briquette" in
+    pollUntilCount 100 "coal lump";
+    make briquette;
+    briquetteAlreadyPlaced <- ishere briquette;
+    if briquetteAlreadyPlaced {} {
+        place briquette;
+    };
+    end;
+
+def go =
+    // Use the recipe from "repro kit" to unpack the right
+    // set of items to be taken from the inventory with the implicit "require":
+    make "3D printer";
+    recurseUntilDepth 8;
+    forever makeBriquette;
+    end;
+
+go;
--- a/data/scenarios/Challenges/bucket-brigade.yaml
+++ b/data/scenarios/Challenges/bucket-brigade.yaml
@@ -1,0 +1,213 @@
+version: 1
+name: Bucket Brigade
+author: Karl Ostmo
+description: |
+  Convey resources with limited mobility
+creative: false
+robots:
+  - name: base
+    display:
+      char: 'Ω'
+      attr: robot
+    heavy: true
+    dir: [0, 1]
+    devices:
+      - 3D printer
+      - ADT calculator
+      - branch predictor
+      - clock
+      - comparator
+      - counter
+      - dictionary
+      - grabber
+      - hearing aid
+      - lambda
+      - logger
+      - mirror
+      - net
+      - scanner
+      - strange loop
+      - string
+      - workbench
+      - briquette press
+    inventory:
+      - [10, repro kit]
+      - [1, treads]
+  - name: hauler
+    system: true
+    dir: [-1, 0]
+    devices:
+      - furnace
+    display:
+      invisible: false
+      char: '='
+      attr: robot
+    program: |
+      run "scenarios/Challenges/_bucket-brigade/hauler.sw"
+  - name: powerplant
+    system: true
+    dir: [1, 0]
+    display:
+      invisible: false
+      char: '▒'
+      attr: gold
+    program: |
+      run "scenarios/Challenges/_bucket-brigade/powerplant.sw"
+objectives:
+  - goal:
+      - Deliver a "coal lump" to the base.
+      - |
+        To excavate coal from the "lignite mine" (M), a robot needs to
+        `drill` while in posession of a "bucketwheel excavator".
+      - |
+        To assemble the excavator, you'll need to repurpose
+        some "treads".
+        Unfortunately, you have only one set of "treads".
+        You'll have to make do...
+    condition: |
+      as base {has "coal lump"}
+  - goal:
+      - Your base got some coal!
+      - Now fashion more of it into a "coal briquette", and place it
+        so that the hauler (=) can pick it up.
+      - The hauler will periodically stop at the base
+        to check if a load is available.
+    condition: |
+      hauler <- robotnamed "hauler";
+      as hauler {has "coal briquette"}
+solution: |
+  run "scenarios/Challenges/_bucket-brigade/brigade.sw"
+entities:
+  - name: bucketwheel excavator
+    display:
+      char: 'x'
+    description:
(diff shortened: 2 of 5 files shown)


byorgey:
Some other ways I can imagine to force the use of a bucket brigade:

1. There are no treads available, but you can `build` a robot in a certain direction so that it appears on a cell next to you instead of on the same cell.
2. The thing you are trying to transport to base "spills" if you try to move while carrying it (see #550), so the only way to move it is to just `give` it to the next in the chain.

data/scenarios/Challenges/bucket-brigade.yaml
```diff
@@ -0,0 +1,154 @@
...
+      - [10, repro kit]
+      - [1, treads]
+  - name: minerobot
+    system: true
+objectives:
+  - goal:
+      - Get some coal to the base.
+      - |
+        To excavate the coal, a robot needs to `drill` while
+        in posession of a "bucketwheel excavator".
```
byorgey, line 39: I think you should explain in the description that the `M` is the coal mine.

data/scenarios/Challenges/bucket-brigade.yaml
```diff
@@ -0,0 +1,154 @@
...
+    properties: [known, portable]
+  - name: repro kit
+    display:
+      char: 'k'
+    description:
+      - Kit that can be unpacked into everything a robot needs to reproduce.
+    properties: [known, portable]
+recipes:
+  - in:
+      - [1, repro kit]
```
byorgey, line 85: I like the idea of the `repro kit`.

data/scenarios/Challenges/bucket-brigade.yaml
```diff
@@ -0,0 +1,154 @@
...
+    description:
+      - A powerful machine designed to remove overburden in mining operations.
+    properties: [known, portable]
+    capabilities: [drill]
+  - name: lignite mine
+    display:
+      char: 'M'
+    description:
+      - Requires heavy equipment to remove the overburden.
+    properties: [known, unwalkable]
```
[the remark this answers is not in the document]
kostmo: Will make it walkable.

data/scenarios/Challenges/bucket-brigade.yaml
```diff
@@ -0,0 +1,154 @@
...
+      char: 'k'
+    description:
+      - Kit that can be unpacked into everything a robot needs to reproduce.
+    properties: [known, portable]
+recipes:
+  - in:
+      - [1, repro kit]
+    out:
+      - [1, solar panel]
+      - [1, treads]
```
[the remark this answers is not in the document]
kostmo: Fixed.  And I verified that upon completion, none of the robots possess treads:

    curl -s http://localhost:5357/robots | jq -C '.[]._installedDevices.counts[][1] | {name: .[1].name, count: .[0]}' | less -R

data/scenarios/Challenges/_bucket-brigade/brigade.sw
```diff
@@ -0,0 +1,75 @@
...
+
+/** The "depth" is the number of spaces between
+the root robot and the target item.
+It is one minus the distance to the target item.
+*/
+def recurseUntilDepth = \depth.
+    let item = "coal" in
+    child <- build {
+
+        // NOTE: Treads are auto-installed via the implicit "require"
```
kostmo, line 30: There are two things that made this solution tricky to implement.

1. The implicit `require` behavior steals things out of my inventory and equips them on the robot being built.  I'm not sure the benefit of such behavior outweighs the "surprise".  At least, perhaps, we should be able to disable that behavior?
2. Due to the above behavior, my `build` command was failing.  I narrowed the failure of my program down to the `build` command via `printf`-style debugging with the `say` command.  However, even then I was not sure **why** `build` was failing.  One cannot use the `try` command to retrieve the error message (and `say` it) of the actual failure that the `try` caught.  I considered using `upload`, assuming that the error message is part of the log.  But the error itself caused my child robot's program to halt, precluding an `upload` step after encountering the error.

In the end I cheated by using the web API to find out what the actual error was:

        curl -s http://localhost:5357/robots | jq -C '.[]._robotLog[] | select( ._leSaid == "ErrorTrace" )' | less -R

It told me that the `build` command failed because I didn't have `treads` in my inventory (and this was because they were `equipped`).  So the solution in this case was to remember to `unequip` the treads right after using them to `move`.

byorgey: Yes, the implicit `require` behavior of `build` is intended to make things easier in most situations, but it could make sense to have some sort of `raw_build` command which does not do any requirements analysis for you.

I debugged my solution by starting with the `--cheat` flag, then when something went wrong I switched into Creative mode so I could build a robot to go `salvage` all the dead robots and bring me back their logs.  That worked pretty well.

byorgey:
Here is my solution, although I made a few modifications to the scenario which I hope you think are reasonable:

- I added a `mirror` to the `base` devices (there is a `mirror` in the `repro kit` but none installed on the base)
- I added an `ADT calculator` to the `repro kit` (I used it in my solution since I checked if the output of `scan down` was equal to `inr "lignite mine"`; yes, I know one could use `ishere` instead, but it seems reasonable to give people an `ADT calculator`)
- I added a `lodestone` to the `repro kit` as well, since I ended up using `not` in my solution.  Again, yes, there's probably a way to do it without using `not`, but it seems like a reasonable thing to allow.

Anyway, given those changes, here's my solution:
```
def fmap : (a -> b) -> cmd a -> cmd b = \f. \x.
  a <- x; return (f a)
end

  b <- test; if b then else
end

def forever : {cmd a} -> cmd b = \c.
  force c; forever c
end

def while : cmd bool -> {cmd a} -> cmd unit = \test. \body.
end

def until = \test. while (fmap not test) end

def waitUntil = \test. until test {wait 1} end

def pass = \thing. forever { waitUntil (has thing); give parent thing } end

def bb =
  log "hi!";
  h <- scan down;
  if (h == inr "lignite mine") {
    make "bucketwheel excavator";
    forever { drill down; give parent "coal" }
  } {
    make "solar panel";  // unpack repro kit
    child <- build {
      require "scanner"; require "ADT calculator"; require "comparator";
      require "branch predictor"; require "workbench"; require "3D printer";
      require "grabber"; require "mirror"; require "lodestone";
      require "metal drill";
      log "waiting for repro kits...";
      wait 10;
      move;
      log "moved, unequipping treads...";
      unequip "treads";
      log "reproducing...";
      bb
    };
    while (has "repro kit") { give child "repro kit" };
    if (self == base) {} { pass "coal" }
  }
end
```

#540 makes this very annoying, since it was hard to figure out where I needed to insert `require` statements.  In theory those should not be necessary.

data/scenarios/Challenges/bucket-brigade.yaml
```diff
@@ -0,0 +1,155 @@
...
+      - [1, metal drill]
+      - [1, welder]
+  - in:
+      - [1, treads]
+    out:
+      - [1, bucketwheel excavator]
+  - in:
+      - [1, lignite mine]
+    out:
+      - [1, coal]
```
byorgey, line 115: I see that the `lignite mine` is destroyed when you mine it, and thus no additional `coal` is generated.  I was actually kind of disappointed that after I solved the scenario, when I selected "Keep playing" I didn't get to see my coal count continue to go up.  I actually wondered whether I had done something wrong, until I looked at how the scenario is implemented.

data/scenarios/Challenges/bucket-brigade.yaml
```diff
@@ -0,0 +1,155 @@
...
+    inventory:
+      - [10, repro kit]
+      - [1, treads]
+  - name: minerobot
+    system: true
+objectives:
+  - goal:
+      - Deliver some "coal" to the base.
+      - |
+        To excavate "coal" from the "mine" (M), a robot needs to
```
byorgey, line 39:
```suggestion
        To excavate "coal" from the "lignite mine" (M), a robot needs to
```

If you're going to put it in quotes it needs to match the name of the entity.  One of the bugs with my solution at first was that I was looking for a `mine` instead of a `lignite mine`, and I eventually had to send a robot to `scan` it just so I could figure out what it was called.

kostmo: Agreed.  It would be cool if we could have some designated markup for entity names that we can typecheck/validate against the list of known entities for that scenario (c.f. #545).

data/scenarios/Challenges/bucket-brigade.yaml
kostmo: Looking at other `drill` recipes, I take it that adding `[1, lignite mine]` to the `out` list will prevent it from being destroyed?  I haven't looked into the code regarding that behavior, but there must be special casing that makes `coal` go into your inventory while the `lignite mine` stays on the ground.

byorgey: Yes, in particular, all the `portable` outputs of the recipe go in your inventory, and any other output (of which there must be at most one) stays on the ground.

byorgey: This is also how drilling a `mountain` can turn it into *e.g.* a `mountain tunnel` --- if that recipe is chosen the `mountain` is destroyed but then the non-`portable` `mountain tunnel` is placed in the cell instead.

kostmo: I suppose there could be an interesting mechanic on the `in` side of the recipe as well---perhaps some consumable material like fuel or welding rods could be taken from your inventory as you drill some non-portable object.

byorgey: Yes, that should work, I think.

data/scenarios/Challenges/_bucket-brigade/brigade.sw
kostmo: #921 921

data/scenarios/Challenges/bucket-brigade.yaml
```diff
@@ -0,0 +1,211 @@
...
+        To assemble the excavator, you'll need to repurpose
+        some "treads".
+        Unfortunately, you have only one set of "treads".
+        You'll have to make do...
+    condition: |
+      as base {has "coal lump"}
+  - goal:
+      - Let the hauler (=) pick up a "coal briquette".
+      - The hauler will periodically stop at the base
+        to check for a load.
```
kostmo, line 72: I decided to add a bit more crafting to the scenario to make use of large quantities of coal, plus a small reward for selecting "Keep playing" after achieving both goals.

byorgey: I love it!

byorgey (approved):

PR #887, opened 2022-12-07
implement Equip/Unequip

kostmo:
See [comment](https://github.com/swarm-game/swarm/issues/883#issuecomment-1340347523) on #883

### DIFF
--- a/src/Swarm/Game/Step.hs
+++ b/src/Swarm/Game/Step.hs
@@ -836,6 +836,25 @@ execConst c vs s k = do
 
         return $ Out VUnit s k
       _ -> badConst
+    Equip -> case vs of
+      [VText itemName] -> do
+        item <- ensureItem itemName "equip"
+        myID <- use robotID
+        focusedID <- use focusedRobotID
+        installSelf item $ focusedID == myID
+        return $ Out VUnit s k
+      _ -> badConst
+    Unequip -> case vs of
+      [VText itemName] -> do
+        item <- ensureEquipped itemName
+        myID <- use robotID
+        focusedID <- use focusedRobotID
+        installedDevices %= delete item
+        robotInventory %= insert item
+        -- Flag the UI for a redraw if we are currently showing our inventory
+        when (focusedID == myID) flagRedraw
+        return $ Out VUnit s k
+      _ -> badConst
     Install -> case vs of
       [VRobot otherID, VText itemName] -> do
         -- Make sure the other robot exists and is close
@@ -848,15 +867,7 @@ execConst c vs s k = do
         case otherID == myID of
           -- We have to special case installing something on ourselves
           -- for the same reason as Give.
-          True -> do
-            -- Don't do anything if the robot already has the device.
-            already <- use (installedDevices . to (`E.contains` item))
-            unless already $ do
-              installedDevices %= insert item
-              robotInventory %= delete item
-
-              -- Flag the UI for a redraw if we are currently showing our inventory
-              when (focusedID == myID) flagRedraw
+          True -> installSelf item $ focusedID == myID
           False -> do
             let otherDevices = robotMap . at otherID . _Just . installedDevices
             already <- use $ pre (otherDevices . to (`E.contains` item))
@@ -1502,6 +1513,21 @@ execConst c vs s k = do
       let msg = "The operator '$' should only be a syntactic sugar and removed in elaboration:\n"
        in throwError . Fatal $ msg <> badConstMsg
  where
+  installSelf ::
+    (HasRobotStepState sig m, Has (Lift IO) sig m) =>
+    Entity ->
+    Bool ->
+    m ()
+  installSelf item viewingSelf = do
+    -- Don't do anything if the robot already has the device.
+    already <- use (installedDevices . to (`E.contains` item))
+    unless already $ do
+      installedDevices %= insert item
+      robotInventory %= delete item
+
+      -- Flag the UI for a redraw if we are currently showing our inventory
+      when viewingSelf flagRedraw
+
   badConst :: HasRobotStepState sig m => m a
   badConst = throwError $ Fatal badConstMsg
 
@@ -1526,6 +1552,13 @@ execConst c vs s k = do
     when (isCardinal d) $ hasCapabilityFor COrient (TDir d)
     let nextLoc = loc ^+^ applyTurn d (orient ? zero)
     (nextLoc,) <$> entityAt nextLoc
+
+  ensureEquipped :: HasRobotStepState sig m => Text -> m Entity
+  ensureEquipped itemName = do
+    inst <- use installedDevices
+    listToMaybe (lookupByName itemName inst)
+      `isJustOrFail` ["You don't have a", indefinite itemName, "equipped."]
+
   ensureItem :: HasRobotStepState sig m => Text -> Text -> m Entity
   ensureItem itemName action = do
     -- First, make sure we know about the entity.
--- a/src/Swarm/Language/Syntax.hs
+++ b/src/Swarm/Language/Syntax.hs
@@ -223,6 +223,10 @@ data Const
     Give
   | -- | Install a device on a robot.
     Install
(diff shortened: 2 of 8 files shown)


src/Swarm/Game/Step.hs
```diff
@@ -1512,6 +1533,13 @@ execConst c vs s k = do
     when (isCardinal d) $ hasCapabilityFor COrient (TDir d)
     let nextLoc = loc ^+^ applyTurn d (orient ? zero)
     (nextLoc,) <$> entityAt nextLoc
+
+  ensureEquipped :: HasRobotStepState sig m => Text -> m Entity
+  ensureEquipped itemName = do
+    inst <- use installedDevices
+    asum (map (listToMaybe . lookupByName itemName) [inst])
```
kostmo, line 1540: This line should be simplified

src/Swarm/Game/Entity.hs
```diff
@@ -604,6 +605,16 @@ deleteAll e (Inventory cs byN h) =
  where
   n = (fst <$> IM.lookup (e ^. entityHash) cs) ? 0
 
+-- | Delete entry for a certain entity from an inventory.
+deleteEntry :: Entity -> Inventory -> Inventory
+deleteEntry e (Inventory cs byN h) =
+  Inventory
+    (IM.delete (e ^. entityHash) cs)
+    (M.delete (T.toLower $ e ^. entityName) byN)
+    (h - (n + 1) * (e ^. entityHash))
```
kostmo, line 614: This formula is guesswork.  Can I actually just replace the whole line with `0`?

src/Swarm/Game/Entity.hs
```diff
@@ -604,6 +605,16 @@ deleteAll e (Inventory cs byN h) =
  where
   n = (fst <$> IM.lookup (e ^. entityHash) cs) ? 0
 
+-- | Delete entry for a certain entity from an inventory.
+deleteEntry :: Entity -> Inventory -> Inventory
```
kostmo, line 609: Adding this function was necessary because `deleteAll` keeps the entry around at `0` count, so it still shows up in the `Installed` list in the left pane.

src/Swarm/Game/Step.hs
```diff
@@ -1488,6 +1499,16 @@ execConst c vs s k = do
       let msg = "The operator '$' should only be a syntactic sugar and removed in elaboration:\n"
        in throwError . Fatal $ msg <> badConstMsg
  where
+  installSelf item focusedID myID = do
```
kostmo, line 1502: This implementation is now shared between `install` and `equip`.

kostmo:
Tested with `stack run -- --scenario classic` to exercise both `equip` and `unequip`, using `compass` and `turn west`.  After unequipping the compass, it doesn't let me turn by cardinal directions anymore.

src/Swarm/Game/Step.hs
```diff
@@ -1488,6 +1499,17 @@ execConst c vs s k = do
       let msg = "The operator '$' should only be a syntactic sugar and removed in elaboration:\n"
        in throwError . Fatal $ msg <> badConstMsg
  where
+  installSelf :: (HasRobotStepState sig m, Has (Lift IO) sig m) => Entity -> RID -> RID -> m ()
+  installSelf item focusedID myID = do
```
byorgey, line 1503: Seems a bit strange to take `focusedID` and `myID` as parameters when all we're going to do is compare them to see if they are equal.  Why not just take a `Bool` instead, called `viewingSelf` or something like that?

src/Swarm/Game/Step.hs
```diff
@@ -1512,6 +1534,13 @@ execConst c vs s k = do
     when (isCardinal d) $ hasCapabilityFor COrient (TDir d)
     let nextLoc = loc ^+^ applyTurn d (orient ? zero)
     (nextLoc,) <$> entityAt nextLoc
+
+  ensureEquipped :: HasRobotStepState sig m => Text -> m Entity
+  ensureEquipped itemName = do
+    inst <- use installedDevices
+    listToMaybe (lookupByName itemName inst)
+      `isJustOrFail` ["What is", indefinite itemName <> "?"]
```
byorgey, line 1542: This is nitpicky, but elsewhere I think we have reserved an error message like `What is a tree?` for situations where the robot doesn't even know about an item.  In this case the item is not equipped, but we might still have some in our inventory, or we might know about it even if we don't have any right now.  If we know about the item but don't have one currently equipped I would expect a message more like `You don't have a tree equipped.`

src/Swarm/TUI/Model.hs
```diff
@@ -889,22 +889,26 @@ populateInventoryList (Just r) = do
   sortOptions <- use uiInventorySort
   let mkInvEntry (n, e) = InventoryEntry n e
       mkInstEntry (_, e) = InstalledEntry e
-      itemList mk label =
+      itemList isInventoryDisplay mk label =
```
byorgey, line 892: I had to scratch my head over this change for a minute.  But let me see if I understand: the point is that when we have zero of an item, we only display it if we are in general displaying items with a count of zero *and* it's in the *inventory*.  In other words, we never display devices with a count of zero in the `installedDevices`.  But that situation never happened before since there was no way to remove devices from `installedDevices`; now, with `unequip`, there is.  Did I get that right?

byorgey (commented):
This is looking good.  We also need to update the help text for an appropriate device to mention `equip` and `unequip`.  Since you put these commands under `CGive`, that device would be `grabber`.  However, the poor `grabber` is already very overloaded with providing four different commands.  It's also unclear why a `grabber` would be needed to `equip` and `unequip` on yourself (I imagine the `grabber` arm doing stuff *outside* the robot).  We might consider (1) creating a new capability, `CEquip`, which is (2) provided by a new device.

kostmo:
> Since you put these commands under `CGive`

It's actually using the `CInstall` capability, not `CGive`.

src/Swarm/TUI/Model.hs
kostmo: Exactly!

src/Swarm/Game/Step.hs
kostmo: fixed

src/Swarm/Game/Step.hs
kostmo: fixed

byorgey:
> > Since you put these commands under `CGive`
> 
> It's actually using the `CInstall` capability, not `CGive`.
> 

Oh, ok, but that capability is still provided by a `grabber`.

src/Swarm/Language/Syntax.hs
```diff
@@ -542,6 +546,8 @@ constInfo c = case c of
       ["The current location has to be empty for this to work."]
   Give -> command 2 short "Give an item to another robot nearby."
   Install -> command 2 short "Install a device from inventory on a robot."
```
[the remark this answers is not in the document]
byorgey: Do we also want to change the `install` tutorial to an `equip` tutorial?  Even in the tutorial, it tells you to install something on yourself...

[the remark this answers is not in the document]
kostmo: Yes, but let's address that in #907.

kostmo:
> We might consider (1) creating a new capability, CEquip, which is (2) provided by a new device.

How about a device called a "welder" that provides `CEquip`?  Let's use it as a placeholder and change it later if we like.  Though I think it has a couple good points going for it:
* conveys a proper sense of semi-permanence/solidity to the "attachment" of items that one equips
* assuming robots are metal, it is applicable regardless of the "material" of the thing being equipped, with metal brackets etc. being implicit in the attachment process.

byorgey:
> How about a device called a "welder" that provides `CEquip`?

I like it!  It's kinda like the `plasma cutter` we used to have for `salvage`, but less destructive. :smile:

byorgey (approved):
Looks good!  Eventually we'll want a recipe for `welder`.

PR #899, opened 2022-12-11
show devices that enable commands

### LINKED ISSUE #788: Include known device(s) that provide commands in command help dialog
The commands dialog (F4) is great, but there's currently no way to remind yourself what device(s) are needed for a particular command other than looking through your inventory and reading descriptions, or simply trying the command and seeing what error message results.  I imagine this could be overwhelming and frustrating for new players.

I'd like the commands dialog to have another column which shows device(s) you know about which provide each command.  We certainly know this information internally, so in theory it shouldn't be hard to display it to the player.  Perhaps something like this (but with color of course):
```
    not : bool -> bool                     o lodestone
     if : ∀ a. bool -> {a} -> {a} -> a     y branch predictor
harvest : cmd text                         ≤ harvester
   grab : cmd text                         < grabber, ≤ harvester
```

kostmo:
fixes #788


kostmo:
Screenshot from 2022-12-10 23-56-49

byorgey (approved):
Looks fantastic.  I like that it only shows you devices you actually have or know about.

PR #900, opened 2022-12-12
remove log command

### LINKED ISSUE #896: Logging suggested in "Install" tutorial is not observable
The "Install" tutorial suggests the player to

    build {log "Hello World!"}

First of all, since the goal condition check runs immediately, the player would have no opportunity to see the log output (see #758).
However, that is not the real problem; the user may choose to "Keep playing" from the "Congratulations" dialog for an opportunity to see the updated log.  But even then it does not appear!

What the user would need to do instead is:

    build {log "Hello World!"; upload base}

If they then select "Keep playing", they would be able to see "Hello World!" in the log.

Remove the `log "Hello World!"` suggestion from this tutorial.

Add `upload base` to the suggested `build` command after `log` (and also add a `scanner` to the `base` inventory).

However, this would obligate more explanation for why `upload` is required, and that the user wouldn't see it unless they select "Keep playing".

kostmo:
fixes #896


byorgey (approved):

kostmo:
It turns out that `build {move}` fails because we don't have `treads` in the inventory.  How about just `build {}`?  The `move` wouldn't be visible anyway unless the user selected "Keep playing".

byorgey:
Oh, of course.  `build {}` should be good.

PR #901, opened 2022-12-12
Consolidate more UI code into UI.hs

kostmo:
Also fixes module pluralization according to [this comment](https://github.com/swarm-game/swarm/pull/880#discussion_r1045306013).

This is a no-op refactoring, only involving renaming/code relocations.

Towards #707.


byorgey (approved):
Looks good.  And the singular module name thing is not a hard and fast rule.  `Swarm.Language.Types` is an obvious exception. (We could consider renaming it too, but I don't see a compelling reason.)

PR #903, opened 2022-12-12
Parameterize scenario datatypes

kostmo:
Towards #873

## Motivation

The `Entity` datatype has a fair amount of state and extra fields that are not relevant to serializing a scenario (in particular, a world map) to a file.

In this PR, the `Cell` record is renamed to `PCell` (representing "Parameterized Cell"), replacing `Entity` with a type parameter.  This permits the introduction of a trimmed-down `Entity` type for the purpose of serialization.  This lightweight type can also be a reliable `Map` key.

A type alias `Cell` is created with the original `Entity` type as a parameter.

The `WorldDescription` and `WorldPalette` are likewise parameterized on the `Cell` type.

### DIFF
--- a/src/Swarm/Game/Scenario/Cell.hs
+++ b/src/Swarm/Game/Scenario/Cell.hs
@@ -2,7 +2,8 @@
 {-# LANGUAGE OverloadedStrings #-}
 
 module Swarm.Game.Scenario.Cell (
-  Cell (..),
+  PCell (..),
+  Cell,
 ) where
 
 import Control.Lens hiding (from, (<.>))
@@ -22,12 +23,18 @@ import Swarm.Util.Yaml
 
 -- | A single cell in a world map, which contains a terrain value,
 --   and optionally an entity and robot.
-data Cell = Cell
+--   It is parameterized on the Entity type to facilitate less
+--   stateful versions of the Entity type in rendering scenario data.
+data PCell e = Cell
   { cellTerrain :: TerrainType
-  , cellEntity :: Maybe Entity
+  , cellEntity :: Maybe e
   , cellRobots :: [IndexedTRobot]
   }
   deriving (Eq, Show)
+
+-- | A single cell in a world map, which contains a terrain value,
+--   and optionally an entity and robot.
+type Cell = PCell Entity
 
 -- | Parse a tuple such as @[grass, rock, base]@ into a 'Cell'.  The
 --   entity and robot, if present, are immediately looked up and
--- a/src/Swarm/Game/Scenario/WorldDescription.hs
+++ b/src/Swarm/Game/Scenario/WorldDescription.hs
@@ -21,22 +21,26 @@ import Witch (into)
 ------------------------------------------------------------
 
 -- | A world palette maps characters to 'Cell' values.
-newtype WorldPalette = WorldPalette
-  {unPalette :: KeyMap Cell}
+newtype WorldPalette e = WorldPalette
+  {unPalette :: KeyMap (PCell e)}
   deriving (Eq, Show)
 
-instance FromJSONE (EntityMap, RobotMap) WorldPalette where
+instance FromJSONE (EntityMap, RobotMap) (WorldPalette Entity) where
   parseJSONE = withObjectE "palette" $ fmap WorldPalette . mapM parseJSONE
 
 -- | A description of a world parsed from a YAML file.
-data WorldDescription = WorldDescription
-  { defaultTerrain :: Maybe Cell
+-- This type is parameterized to accommodate Cells that
+-- utilize a less stateful Entity type.
+data PWorldDescription e = WorldDescription
+  { defaultTerrain :: Maybe (PCell e)
   , offsetOrigin :: Bool
-  , palette :: WorldPalette
+  , palette :: WorldPalette e
   , ul :: Location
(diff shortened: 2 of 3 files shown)


src/Swarm/Game/Scenario/Cell.hs
```diff
@@ -22,13 +23,19 @@ import Swarm.Util.Yaml
 
 -- | A single cell in a world map, which contains a terrain value,
 --   and optionally an entity and robot.
-data Cell = Cell
+--   It is parameterized on the Entity type to facilitate less
+--   stateful versions of the Entity type in rendering scenario data.
+data PCell e = Cell
   { cellTerrain :: TerrainType
-  , cellEntity :: Maybe Entity
+  , cellEntity :: Maybe e
```
xsebek, line 30: So the idea is that instead of writing something like:
```Haskell
worldMapToText :: WorldPalette -> [[Cell]] -> Text
```
we would instead do it in two steps:
```
simplifyEntityForMe :: Entity -> SimplerEntity
simpleWorldMapToText :: WorldPalette (PCell SimplerEntity) -> Text
```
Is this because you want to have `ToJSON` (not `ToJSONE`) instance? 🤔

xsebek: @kostmo if this is about `ToJSON` we discussed with @byorgey, that you could simply use a custom tuple:
```Haskell
newtype Env e v = Env (e, v)
instance ToJSON (Env E Val) -- maybe we could codify this pattern as ToJSONE => ToJSON (Env, Val)
```
which is the same as
```Haskell
simplify :: E -> Val -> SimpleVal
instance ToJSON SimpleVal
```

kostmo: > Is this because you want to have ToJSON (not ToJSONE) instance?

I haven't quite wrapped my head around how `FromJSONE` works (or how `ToJSONE` would).

For the purpose of the World Editor, we just need to emit a fragment of the scenario file (either to a file or via the web API ).

I'm not sure how easy it is to read in its current state, but [here's what I'm trying to do in the World Editor PR](https://github.com/swarm-game/swarm/pull/873/files#r1048985734).

xsebek: After reading through your PR, I think it makes sense to keep it simple. 👍 

The idea with `FromJSONE` is that you need the environment to choose between parsing alternatives.

But with `ToJSON` no tricks are necessary, so I was incorrect in thinking that we would ever need it.

src/Swarm/Game/Scenario/WorldDescription.hs
```diff
@@ -22,23 +22,27 @@ import Witch (into)
...
+-- This type is parameterized to accommodate Cells that
+-- utilize a less stateful Entity type.
+data PWorldDescription c = WorldDescription
+  { defaultTerrain :: Maybe c
   , offsetOrigin :: Bool
-  , palette :: WorldPalette
+  , palette :: WorldPalette c
   , ul :: V2 Int64
-  , area :: [[Cell]]
+  , area :: [[c]]
```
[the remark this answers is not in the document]
kostmo: Fixed.

src/Swarm/Game/State.hs
```diff
@@ -144,6 +144,7 @@ import Swarm.Game.Recipe (
   reqRecipeMap,
  )
 import Swarm.Game.Robot
+import Swarm.Game.Scenario.WorldDescription
```
byorgey, line 147: Why do we need this new import?

src/Swarm/Game/Scenario/WorldDescription.hs
```diff
@@ -52,7 +56,7 @@ instance FromJSONE (EntityMap, RobotMap) WorldDescription where
 --   string into a nested list of 'Cell' values by looking up each
 --   character in the palette, failing if any character in the raw map
 --   is not contained in the palette.
-paintMap :: MonadFail m => WorldPalette -> Text -> m [[Cell]]
+paintMap :: MonadFail m => WorldPalette Cell -> Text -> m [[Cell]]
```
byorgey, line 59: Can we make this polymorphic, `WorldPalette c -> Text -> m [[c]]` or something like that?

src/Swarm/Game/Scenario/WorldDescription.hs
```diff
@@ -21,23 +21,27 @@ import Witch (into)
 ------------------------------------------------------------
 
 -- | A world palette maps characters to 'Cell' values.
-newtype WorldPalette = WorldPalette
-  {unPalette :: KeyMap Cell}
+newtype WorldPalette c = WorldPalette
+  {unPalette :: KeyMap c}
```
byorgey, line 25: Would it be better to have `newtype WorldPalette e = WorldPalette { unPalette :: KeyMap (PCell e) }`?  I am honestly not sure.  Or maybe we should just double down on the generality and call it `Palette` instead of `WorldPalette` --- it maps characters to anything.

byorgey (approved):

src/Swarm/Game/State.hs
kostmo: Huh, I would have expected a compiler warning about an unused import.  Anyway, I've removed it and it still works.

PR #911, opened 2022-12-14
Use custom types for immediate update

xsebek:
- make `WorldUpdate` and `RobotUpdate` serializable
- part of #50

### DIFF
--- a/src/Swarm/Game/CESK.hs
+++ b/src/Swarm/Game/CESK.hs
@@ -91,22 +91,23 @@ module Swarm.Game.CESK (
 
 import Control.Lens.Combinators (pattern Empty)
 import Data.Aeson (FromJSON, ToJSON)
-import Data.Aeson qualified
 import Data.IntMap.Strict (IntMap)
 import Data.IntMap.Strict qualified as IM
 import Data.List (intercalate)
+import Data.Text qualified as T
 import GHC.Generics (Generic)
-import Swarm.Game.Entity (Entity, Inventory)
+import Swarm.Game.Entity (Count, Entity)
 import Swarm.Game.Exception
 import Swarm.Game.Value as V
-import Swarm.Game.World (World)
 import Swarm.Language.Context
 import Swarm.Language.Pipeline
 import Swarm.Language.Pretty
 import Swarm.Language.Requirement (ReqCtx)
 import Swarm.Language.Syntax
+import Swarm.Language.Syntax qualified as Syntax
 import Swarm.Language.Types
 import Witch (from)
+import Swarm.Util.Location (Location)
 
 ------------------------------------------------------------
 -- Frames and continuations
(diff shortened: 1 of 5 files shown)


src/Swarm/Game/CESK.hs
```diff
@@ -350,39 +352,22 @@ prettyFrame (FDef x) = "def " ++ from x ++ " = _"
...
-  }
-
-newtype RobotUpdate = RobotUpdate
-  { robotUpdateInventory :: Inventory -> Inventory
+data WorldUpdate = ReplaceEntity
+  { updatedLoc :: V2 Int64
+  , originalEntity :: Entity
+  , newEntity :: Maybe Entity
   }
+  deriving (Eq, Ord, Show, Generic, FromJSON, ToJSON)
```
xsebek, line 368: The default instances do not follow our style, so I would like to customize it a bit and test it before merging.

src/Swarm/Game/CESK.hs
```diff
@@ -350,39 +352,22 @@ prettyFrame (FDef x) = "def " ++ from x ++ " = _"
 prettyFrame FExec = "exec _"
 prettyFrame (FBind Nothing t _) = "_ ; " ++ prettyString t
 prettyFrame (FBind (Just x) t _) = from x ++ " <- _ ; " ++ prettyString t
-prettyFrame FImmediate {} = "(_ : cmd a)"
+prettyFrame (FImmediate c w r) = unwords ["immediate@" <> T.unpack (Syntax.syntax $ Syntax.constInfo c), show w, show r]
 prettyFrame (FUpdate loc) = "store@" ++ show loc ++ "(_)"
 prettyFrame FFinishAtomic = "finishAtomic"
 
 --------------------------------------------------------------
 -- Wrappers for functions in FImmediate
```
byorgey, line 360: We should get rid of this comment too, `FImmediate` no longer contains functions.

src/Swarm/Game/CESK.hs
```diff
@@ -164,7 +166,7 @@ data Frame
     --   if there is one, to the output of the first command).
     FBind (Maybe Var) Term Env
   | -- | Apply specific updates to the world and current robot.
-    FImmediate WorldUpdate RobotUpdate
+    FImmediate Const [WorldUpdate] [RobotUpdate]
```
byorgey, line 169: Is the `Const` just for better error messages?  So we know what command was ultimately responsible for generating this `FImmediate` frame?

byorgey (approved):
Looks good.  Not only does this make it possible to serialize `WorldUpdate` and `RobotUpdate`, I think the code is also much nicer now!

src/Swarm/Game/CESK.hs
xsebek: Exactly, `FImmediate` was definitely created by some command which we use in the error message.

I will note this in the docstring. 👍

src/Swarm/Game/CESK.hs
xsebek: Indeed, I am considering moving them to Robot and World respectively anyway. 🙂

src/Swarm/Game/CESK.hs
xsebek: Scratch that, doing it here would blow up the scope and I want to clean up first.

xsebek:
Happy New Year to Swarm! :partying_face: @byorgey @kostmo

PR #914, opened 2022-12-14
Restyle main

kostmo:


kostmo:
@byorgey, is your local version of `fourmolu` updated (to `0.9.0.0`)?

byorgey (approved):
@kostmo , looks like it wasn't, thanks for the reminder!

Issue #606, opened 2022-07-28
Let world coordinates wrap at an arbitrary power of 2


byorgey:
This issue is still open for the idea of being able to specify the power of two at which world coordinates wrap.  *i.e.* add a field to the `world` in a scenario `.yaml` file like `size: n` which has a natural number specifying the power of two, i.e. `size: 32` would be the default, but you could also specify smaller values.  There can also be a Boolean field `wrap` which specifies whether the world should wrap around or just stop at the edges.

Below is the original issue description for switching world coordinates to `Int32`, which has now been done.

----

Right now we use `Int64` for world coordinates.  However, this is a little ridiculous:
- Even travelling at 1024 units per second, it would take a robot 285 million years of real-world time to travel all the way to the "edge" of the world (or rather, the place where the world wraps around).
- If each cell represents 1 square meter, this would be a planet about 42 billion times the size of Jupiter. 

On the other hand, with `Int32` for coordinates instead:
- It would take a robot travelling at 1024 cells/s about 24 days to reach the edge of the world.
- At 1m / cell, it would represent a planet about 9 times the size of Jupiter.

This is obviously still plenty big enough that no one will ever run out of space, but seems more reasonable/realistic.  ~~And if we say that each cell is 100m x 100m (not that it really matters), the planet would be about the right size (and would mean that features like forests and lakes etc. are kind of the right size too).~~ (Oops, ignore that last comment, I went the wrong way --- it should be 1 cm / cell, in which case lakes and forests are *definitely* not the right size. :laughing: )

xsebek:
Could we add some (`new`)`type Point`? I am not sure if we use `V2` for anything else.

byorgey:
Yes, making a `newtype Loc = Loc (V2 Int32)` is a good idea.  I don't think we use `V2` for anything else but just on principle.

xsebek:
Generalising this idea a bit, we could make the world coordinates "wrap around" an arbitrary number.

For example, we could have a small world that is only 128x128 cells large. Not just adding a wall, but an actual finite world.

The funny thing would be that it is a torus, so if you go east, you will come back from the west, arcade style! :space_invader:

It would be cool to create the city Sigil from [Planescape: Torment](https://en.wikipedia.org/wiki/Planescape:_Torment). :smile: The city is inside the torus above the insanely high mountain:
![Sigil](https://upload.wikimedia.org/wikipedia/en/0/02/Pgp_007_01.jpg)

[...]

PR #919, opened 2022-12-15
Add new `Location` and `Heading` types, and change coordinates from 64 to 32 bits

byorgey:
Towards #606 .  

- Rename `Location` from the parser (representing source locations) to `SrcLoc`
- Rename `Loc` in `CESK` (representing the index of a memory cell) to `Addr`
- Add type synonyms `Location = Point V2 Int32` and `Heading = V2 Int32`


src/Swarm/Game/State.hs
```diff
@@ -351,19 +351,19 @@ robotMap :: Lens' GameState (IntMap Robot)
...
 robotsAtLocation loc gs =
   mapMaybe (`IM.lookup` (gs ^. robotMap))
     . maybe [] IS.toList
     . M.lookup loc
     . view robotsByLocation
     $ gs
 
--- | Get robots in manhattan distastance from location.
-robotsInArea :: V2 Int64 -> Int64 -> GameState -> [Robot]
+-- | Get all the robots within a given Manhattan distastance from a location.
```
kostmo, line 365: "distastance"

src/Swarm/Util/Location.hs
```diff
@@ -0,0 +1,54 @@
...
+import Data.Aeson (FromJSONKey, ToJSONKey)
+import Data.Int (Int32)
+import Data.Yaml (FromJSON (parseJSON), ToJSON (toJSON))
+import Linear (V2 (..))
+import Linear.Affine (Affine (..), Point (..), origin)
+
+-- | A Location is a pair of (x,y) coordinates, both up to 32 bits.
+--   The positive x-axis points east and the positive y-axis points
+--   north.
+type Location = Point V2 Int32
```
kostmo, line 25: Maybe add some documentation here about why `Coords` is a different type than `Location`, and precisely how the conventions differ.

kostmo (approved):

src/Swarm/Util/Location.hs
byorgey: Good idea! :heavy_check_mark:

PR #920, opened 2022-12-16
Meeting other robots

### LINKED ISSUE #306: Commands to meet other robots

### LINKED ISSUE #931: Rename `robot` type to be more inclusive
*Internally* to the game implementation, anything that runs a program is a "robot".  However, to the *player*, there are robots (things they build and program), and there can also be other kind of active things like aliens, animals, ... In particular I'm thinking about the `meet` command which will return a `robot` value (see #920), but you should be able to use it to `meet` aliens, animals, etc.  So I'm wondering if we want to change the name of the `robot` type to something more generic and inclusive, and I'm open to suggestions what we should change it to.  `agent`? `actor`?  `creature`? `being`?

byorgey:
Closes #306. Closes #931.

- Renamed `robot` type to `actor` in anticipation of meeting other things besides robots.
- `meet : cmd (unit + actor)` returns an arbitrary actor within Manhattan distance 1, if any.
- `meetAll : (b -> actor -> cmd b) -> b -> cmd b` will run on every nearby actor.
- Added `antenna` device to provide the commands.
- Added "make a friend" challenge that requires the use of `meet`.


data/scenarios/Challenges/_friend/cat.sw
```diff
@@ -0,0 +1,70 @@
...
+  case fishGiver
+    (\_. say "meow??"; n <- count "fish"; cat n)
+    (\r. follow cat r)
+end
+
+def cat = \fishCount.
+  n <- count "fish";
+  if (n > fishCount)
+  { say "yum!";
+    if (n >= 3) { love cat } { cat n }
```
kostmo, line 64: Here's my experience report for solving this :)

* Was initially concerned that I'd have to write a program to locate the cat at some random displacement, but soon figured that, upon starting over, just making a mad dash straight to the cat is reproducible.
* Unfortunately I was tainted by seeing the number 3 in the code review.  I tried anyway just feeding one fish before dashing back to my base and got the "meow??".  I think that's a good enough hint.
* I again immediately dashed away after feeding 3 fish and got a `meow??`.  I figured maybe the cat needed some "bonding time" or maybe I didn't stick around long enough for them see who left the fish :).   So I retried a few times with increasing `wait` times until it worked.

data/scenarios/Challenges/friend.yaml
```diff
@@ -0,0 +1,92 @@
...
+      baseLoc <- as base {whereami};
+      return (catLoc == baseLoc)
+solution: |
+  run "scenarios/Challenges/_friend/friend-solution.sw"
+entities:
+  - name: fish
+    display:
+      char: 'f'
+    description:
+      - A smelly fish. Doesn't seem very useful.
```
kostmo, line 74: How about this:
```suggestion
      - A smelly fish. Rather unappetizing to a robot.
```
or this:
```suggestion
      - A smelly fish. Robots have no use for such things.
```

As I was considering phrasings, it occurs to me that we might want a style guideline for entity descriptions and such?  In particular, that they should be written in "third person" or "second person" perspective, but not "first person".

kostmo (approved):
I like it!

data/scenarios/Challenges/friend.yaml
byorgey: Sure, some style guidelines would be good.  I definitely agree about no "first person" for starters.

data/scenarios/Challenges/_friend/cat.sw
byorgey: Thanks for the experience report!  Yeah, the needed "bonding time" is because the cat repeatedly picks a random amount of time to wait before it (1) tries to move and (2) notices if it has more fish.  I could probably try to change it so it notices fish immediately.  Do you think that would be better, or it's OK the way it is?

kostmo: I think it would be better to notice the fish immediately.

kostmo: If there is still an issue with getting the fish to be immediately acknowledged, I think a more detailed hint---perhaps a Garfield-style internal monologue, like "Meow?? (who gave that fish?)"---would help.

byorgey: OK, I changed it so the cat is constantly polling to see if it has more fish than before.  This means you usually don't have to wait for the cat to follow you, although in my testing it is possible, if you give the fish and then *immediately* run away, to get far enough away before it runs `meet` than you get the `meow??`.  But other than that small race condition it starts following you right away.

PR #926, opened 2022-12-19
nonposession unit test

kostmo:
The `858-possession-objective.yaml` unit test scenario demonstrates that `has "tree"` works as a goal criteria.
However, `x <- has "tree"; not x` does not work!

Could be related to #858?


byorgey (approved):
Weird.  Definitely seems like it could be related to #858.  Hopefully the additional data point will help get to the bottom of this.

PR #937, opened 2022-12-29
Make goal text NonEmpty

kostmo:
Here's how the 20 scenarios with missing goals were identified:

    find data/scenarios -name "*.yaml" -printf "%p\t" -exec yq '.objectives | map(has("goal")) | all' {} \; | grep false | cut -f1

### DIFF
--- a/data/scenarios/Testing/201-require/533-reprogram-simple.yaml
+++ b/data/scenarios/Testing/201-require/533-reprogram-simple.yaml
@@ -5,7 +5,9 @@ description: |
   the target robot doesn't have.
   https://github.com/swarm-game/swarm/pulls/533
 objectives:
-  - condition: |
+  - goal:
+      - Reprogram
+    condition: |
       try {
         base_boats <- as base {count "boat"};
         base_solar <- as base {count "solar panel"};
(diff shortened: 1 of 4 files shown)


src/Swarm/Game/ScenarioInfo.hs
```diff
@@ -347,6 +347,7 @@ loadScenarioItem em path = do
   case isDir of
     True -> SICollection collectionName <$> loadScenarioDir em path
     False -> do
+      sendIO $ putStrLn $ unwords ["Loading path:", path]
```
byorgey, line 350: Seems like an unrelated debugging change?

byorgey (approved):
Seems reasonable to me.  But what's the motivation?

data/scenarios/Testing/201-require/533-reprogram-simple.yaml
```diff
@@ -5,7 +5,9 @@ description: |
   the target robot doesn't have.
   https://github.com/swarm-game/swarm/pulls/533
 objectives:
-  - condition: |
+  - goal:
+      - Reprogram
```
byorgey, line 9: Adding "fake" objective descriptions seems like an opportunity to better document the test suite.

xsebek: Exactly, I sometimes inspect the tests from main menu and having a goal would be helpful.

xsebek (commented):
I also wander what your motivation is, but requiring a goal description (even for tests) sounds reasonable to me. 🙂

PR #927, opened 2022-12-19
Boolean expressions of objective prerequisites

kostmo:
Closes #795

image

Screenshot from 2022-12-31 22-58-59
image

## Tasks
- [x] Vendor a subset of the `boolexpr` package
- [x] New display logic for parallel goals
- [x] Add a check for whether winning is impossible (due to a "not" prerequisite being achieved)
- [x] Add a "Lose" dialog analogous to the "Win" dialog
- [x] Validate no dependency cycles in prerequisites specs
- [x] Add tests for negative validation result from scenario parsing
- [x] Web API to inspect status of incomplete, completed, optional, and prerequisite objectives
- [x] Update all of the multi-goal scenarios to use prerequisites

## For follow-up PRs:
- [ ] Upload new GHC8-compatible version of `boolexpr` to Hackage (perhaps [take over the package](https://wiki.haskell.org/Taking_over_a_package)?)
- [ ] Add prerequisites and its logical operators to JSON schema
- [ ] Reverse topological sort of goals before evaluation, with a test that demonstrates necessity
- [ ] Automatically skip over the header rows when navigating the Goals list
- [ ] Add indicators for optional and hidden goals
- [ ] New Challenge for strategizing optional and mutually-exclusive goals

PR #940, opened 2023-01-01
Fix hypothetical result of condition

### LINKED ISSUE #858: Counting inventory does not work for goal criteria

xsebek:
- handle `VResult` in hypothetical
- log unexpected result values
- closes #858 

When a bind occurred in hypothetical evaluation, then the `Value` was a `VResult` that we silently discarded.
Now it is handled and all other values are logged.


kostmo (approved):
#927 touches this area of code, so I'll be careful to preserve this when I rebase.

byorgey (approved):
That's it!  I knew it must have something to do with `VResult`...

PR #943, opened 2023-01-01
Create swarm subdirectories as necessary

### LINKED ISSUE #941: Crash after completing the first step of the tutorial
After building from source, completing the "say Ready" tutorial step crashes the game with the error
`swarm: YamlException "could not open file for write: /home/rizoma/.local/share/swarm/saves/Tutorials_backstory.yaml"`

**To reproduce** 
Install from source with `stack` on Pop Os, launch and open the tutorial. Close the introduction message and type 'say "Ready!"'. After the congratulation message, the first two options will crash the game. Additionally, the menu will not reappear if you select the third option and press Ctrl-g.

**Expected behavior**
The tutorial should progress to the next step

**Additional context**
During the `stack` setup I had to stop two times to install two linkers with `apt`, I might still need some runtime dependencies that I don't know about

xsebek:
- fixes #941


src/Swarm/Util.hs
```diff
@@ -259,7 +259,7 @@ getDataFileNameSafe name = do
 
 dataNotFound :: FilePath -> IO Text
 dataNotFound f = do
-  d <- getSwarmDataPath False
+  d <- getSwarmDataPath False ""
```
kostmo, line 262: I take it that the empty string represents "no subdirectory", i.e. the "current directory"?  Perhaps a `Maybe` would convey the intentions better.

kostmo (approved):

src/Swarm/Util.hs
xsebek: It's how path works - `a//./` is the same as `a`.

A good solution might be to use a different function or make a proper ADT type for our sources. 🤔 

We should refactor this code and document it better, but let’s do that in a new PR.

byorgey (approved):

byorgey:
@xsebek  what's the status of this?  Were you going to do more testing or can it be merged?

xsebek:
@byorgey it definitely needs to be tested, because it is wrong! :sweat_smile: 
```
swarm: /root/.local/share/swarm/history: openFile: inappropriate type (Is a directory)
```
```
swarm: YamlException "could not open file for write: /root/.local/share/swarm/achievements"
```

xsebek:
Now it works in a fresh `ubuntu:bionic` container:
image

```
root@swarm-run:/# tree -L 2 ~/.local/share/swarm
/root/.local/share/swarm
|-- achievements
|-- data
|   |-- about.txt
|   |-- adjectives.txt
|   |-- entities.yaml
|   |-- logo.txt
|   |-- names.txt
|   |-- recipes.yaml
|   |-- scenarios
|   `-- schema
|-- history
`-- saves
    |-- Tutorials_backstory.yaml
    `-- Tutorials_move.yaml

4 directories, 10 files
```

PR #944, opened 2023-01-02
Wolf, Goat, Cabbage scenario

kostmo:
A challenge scenario that makes use of the NOT prerequisites to simplify evaluation of the win condition.  These prerequisites ensure that **at no point** did the player ever perform some illegal action.

## Demo

    ./scripts/play.sh --scenario data/scenarios/Challenges/wolf-goat-cabbage.yaml --autoplay

## Playing
image

## Losing
image

### DIFF
--- a/data/scenarios/Challenges/00-ORDER.txt
+++ b/data/scenarios/Challenges/00-ORDER.txt
@@ -5,6 +5,7 @@ word-search.yaml
 ice-cream.yaml
 hanoi.yaml
 bucket-brigade.yaml
+wolf-goat-cabbage.yaml
 friend.yaml
 Mazes
 Ranching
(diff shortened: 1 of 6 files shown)


byorgey (approved):

byorgey:
I was thinking about this scenario and I like that it can tell you when it's no longer winnable (could Towers of Hanoi be redesigned to make use of that feature?).  But I thought it might be nice if you get a custom losing message, *e.g.* "too bad! the wolf ate the sheep!".  I wonder what would be needed to make that happen.  Maybe each objective can have a `lose_message` which is shown whenever that objective is the "cause" of the failed scenario?  I'm not really sure how to determine which objective is the "cause" though.

xsebek:
@byorgey we could let a system robot show a popup notification.

It would mean two of them get shown after each other, which is not ideal.

---

Since the conditions that caused the failure must have been met, how about showing “failure” reason from all of them? In this case the only one met will mention wolf and sheep. Generally we would try to produce something like parser errors - “I was expecting X or Y”.

kostmo:
An objective can only "fail" if it has a "prerequisites" expression that contains a negated goal reference.  The BoolExpr library allows us to query for and list all of the negated references in an expression.  So we could just match that list against the "completed" objectives, and display as a summary in the "Condolences" popup.

PR #945, opened 2023-01-02
Add achievement for losing

kostmo:
## Test procedure

1. Start game

       ./scripts/play.sh
3. Observe empty achievements
4. Select "New Game" -> "Challenges" -> "Lake crossing"
5. `turn east; move`

image

image


byorgey (approved):
so sad

PR #948, opened 2023-01-02
Better CESK machine pretty-printing

### LINKED ISSUE #644: Better pretty-printing for CESK machine states
Our current pretty-printing for CESK machine states is kind of terrible, for at least two reasons:

- It just prints out the frames in a list, rather than printing them in a way that shows the natural nesting of contexts
- It does not take precedence etc. into account.

Instead, we should print out CESK machine states in a way that looks almost like valid terms, with some extra decorations to denote what we are currently focusing on, and some way to display special frames that don't correspond directly to terms.  So instead of printing out frames one after another, we print them out *nested*.

I hope to add some more concrete examples later.

This would be a first step toward #5 .

byorgey:
Closes #644.  Currently this only helps for debugging, but a first step towards #5.

Some kind-of-randomly-selected before/after examples:

BEFORE:
```
▶ say
  [_ "Ready!" | exec _]
◀ say
  [_ "Ready!" | exec _]
▶ "Ready!"
  [say _ | exec _]
◀ "Ready!"
  [say _ | exec _]
◀ say "Ready!"
  [exec _]
◀ ()
  []
```

AFTER:
```
E· (▶say◀ "Ready!")
E· (◀say▶ "Ready!")
E· (say ▶"Ready!"◀)
E· (say ◀"Ready!"▶)
E· ◀say "Ready!"▶
◀()▶
```

BEFORE:
```
▶ c
  [exec _ | b <- _ ; if b {noop} {force until c} | _ ∪ <Env> | _ ; grab | _ ∪ <Env> | loadEnv]
◀ as <a1> {has "Win"}
  [exec _ | b <- _ ; if b {noop} {force until c} | _ ∪ <Env> | _ ; grab | _ ∪ <Env> | loadEnv]
◀ false
  [b <- _ ; if b {noop} {force until c} | _ ∪ <Env> | _ ; grab | _ ∪ <Env> | loadEnv]
▶ if b {noop} {force until c}
  [exec _ | _ ∪ <Env> | _ ; grab | _ ∪ <Env> | loadEnv]
▶ if b {noop}
  [_ {force until c} | exec _ | _ ∪ <Env> | _ ; grab | _ ∪ <Env> | loadEnv]
▶ if b
  [_ {noop} | _ {force until c} | exec _ | _ ∪ <Env> | _ ; grab | _ ∪ <Env> | loadEnv]
```

AFTER:
```
L· ∪· ((b <- E· ▶c◀ ; if b {noop} {force until c}) ; grab)
L· ∪· ((b <- E· ◀as <a1> {has "Win"}▶ ; if b {noop} {force until c}) ; grab)
L· ∪· ((b <- ◀false▶ ; if b {noop} {force until c}) ; grab)
L· ∪· (∪· E· ▶if b {noop} {force until c}◀ ; grab)
L· ∪· (∪· E· (▶if b {noop}◀ {force until c}) ; grab)
L· ∪· (∪· E· (▶if b◀ {noop} {force until c}) ; grab)
```


src/Swarm/Game/CESK.hs
```diff
@@ -317,54 +310,60 @@ resetBlackholes (Store n m) = Store n (IM.map resetBlackhole m)
...
+prettyFrame (FLet x t _) (_, inner) = (11, hsep ["let", pretty x, "=", inner, "in", ppr t])
+prettyFrame (FTry v) (p, inner) = (10, "try" <+> pparens (p < 11) inner <+> prettyPrec 11 (valueToTerm v))
+prettyFrame (FUnionEnv _) inner = prettyPrefix "∪·" inner
+prettyFrame (FLoadEnv _ _) inner = prettyPrefix "L·" inner
+prettyFrame (FDef x) (_, inner) = (11, "def" <+> pretty x <+> "=" <+> inner <+> "end")
+prettyFrame FExec inner = prettyPrefix "E·" inner
+prettyFrame (FBind Nothing t _) (p, inner) = (0, pparens (p < 1) inner <+> ";" <+> ppr t)
+prettyFrame (FBind (Just x) t _) (p, inner) = (0, hsep [pretty x, "<-", pparens (p < 1) inner, ";", ppr t])
+prettyFrame FDiscardEnv inner = prettyPrefix "D·" inner
+prettyFrame (FImmediate _ _ _) inner = prettyPrefix "I·" inner
```
xsebek, line 356: Isn't this omitting too much? Maybe this could at least mention the original `Const` similarly to how `FUpdate` shows its memory cell parameter. :thinking:  

Also, HLint suggests `FImmediate {}`, but I would prefer either using the parameters or naming them (`_worldUps`,...) so it's easier to see what is left out.

src/Swarm/Game/CESK.hs
```diff
@@ -317,54 +310,60 @@ resetBlackholes (Store n m) = Store n (IM.map resetBlackhole m)
...
+prettyFrame (FUpdate loc) inner = prettyPrefix ("S@" <> pretty loc) inner
```
xsebek, line 357: To me `loc` is confusing - we have actual in-world locations updated in `FImmediate`, but here it means "memory cell address". The type alias used for it is `Addr` which to me would read easier as a parameter name too.

xsebek (approved):
This looks amazing! I really wish we could open some debugging window in the game and see all the robot frame updates during the last tick. :star_struck:

src/Swarm/Game/CESK.hs
byorgey: Good idea.  I've added the original `Const` to the output.  Not really sure of a good format to show the world update / robot update info but open to suggestions.

src/Swarm/Game/CESK.hs
byorgey: Oh, good point, I renamed a bunch of stuff recently but didn't update this I guess.

PR #949, opened 2023-01-03
Nested sum type for directions

kostmo:
Code simplification, improved type safety (see [this comment](https://github.com/swarm-game/swarm/pull/876#discussion_r1029843033)).

This technique may also simplify the implementation of `CReverse` (see #950) if we choose to do so.

### DIFF
--- a/src/Swarm/Game/Display.hs
+++ b/src/Swarm/Game/Display.hs
@@ -44,9 +44,9 @@ import Data.Map qualified as M
 import Data.Maybe (fromMaybe, isJust)
 import Data.Yaml
 import GHC.Generics (Generic)
-import Swarm.Language.Syntax (Direction (..))
+import Swarm.Language.Syntax (AbsoluteDir (..), Direction (..))
 import Swarm.TUI.Attr (entityAttr, robotAttr, worldPrefix)
-import Swarm.Util (maxOn, (?))
+import Swarm.Util (maxOn)
 import Swarm.Util.Yaml (FromJSONE (..), With (runE), getE, liftE, withObjectE)
 
 -- | Display priority.  Entities with higher priority will be drawn on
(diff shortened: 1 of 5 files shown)


src/Swarm/Game/Display.hs
```diff
@@ -128,9 +128,11 @@ instance ToJSON Display where
...
 -- | Look up the character that should be used for a display.
 displayChar :: Display -> Char
-displayChar disp = case disp ^. curOrientation of
-  Nothing -> disp ^. defaultChar
-  Just dir -> M.lookup dir (disp ^. orientationMap) ? (disp ^. defaultChar)
+displayChar disp = fromMaybe (disp ^. defaultChar) $ do
+  dir <- disp ^. curOrientation
+  case dir of
+    DAbsolute d -> M.lookup d (disp ^. orientationMap)
+    _ -> Nothing
```
byorgey, line 135:
```suggestion
  DAbsolute d <- disp ^. curOrientation
  M.lookup d (disp ^. orientationMap)
```

I think this ought to work --- pattern-matching the result of a bind results in a call to `fail`, which in the case of the `Maybe` monad would yield `Nothing`, exactly what is wanted here.  I don't have a strong preference though, pick whichever you think is clearer.

src/Swarm/Language/Syntax.hs
```diff
@@ -93,17 +96,23 @@ import Witch.From (from)
...
 
+data RelativeDir = DLeft | DRight | DBack | DForward | DDown
+  deriving (Eq, Ord, Show, Read, Generic, Data, Hashable, ToJSON, FromJSON, Enum, Bounded)
+
+-- | The type of directions. Used /e.g./ to indicate which way a robot
+--   will turn.
+data Direction = DAbsolute AbsoluteDir | DRelative RelativeDir
+  deriving (Eq, Ord, Show, Read, Generic, Data, Hashable, ToJSON, FromJSON)
+
 data DirInfo = DirInfo
```
byorgey, line 116: Now that we distinguish absolute and relative directions at the type level, can we just get rid of the `dirAbs` field in the `DirInfo` record?  We know it will be `Just` exactly for the `DAbsolute` directions, and we have the `toHeading` function, so the `dirAbs` field seems superfluous.

byorgey (approved):
Looks great, other than one more simplification that I think we can make (getting rid of `dirAbs`).

PR #955, opened 2023-01-04
New `heading` command

byorgey:
Define a new command `heading : cmd dir` which returns the robot's current heading (or `down` if it doesn't have one, or if the heading is some weird vector).

My immediate motivation is that I want to be able to define `excursion : cmd unit -> cmd unit` which executes the given command and then returns to the same location and orientation as before.  The `heading` command is one of the last missing pieces of machinery necessary to be able to implement this.

### DIFF
--- a/editors/emacs/swarm-mode.el
+++ b/editors/emacs/swarm-mode.el
@@ -77,6 +77,7 @@
                "create"
                "time"
                "whereami"
+               "heading"
                "blocked"
                "scan"
                "upload"
--- a/editors/vscode/syntaxes/swarm.tmLanguage.json
+++ b/editors/vscode/syntaxes/swarm.tmLanguage.json
@@ -58,7 +58,7 @@
 				},
 				{
 				"name": "keyword.other",
-				"match": "\\b(?i)(self|parent|base|if|inl|inr|case|fst|snd|force|undefined|fail|not|format|chars|split|charat|tochar|noop|wait|selfdestruct|move|turn|grab|harvest|place|give|install|equip|unequip|make|has|installed|count|drill|build|salvage|reprogram|say|listen|log|view|appear|create|time|whereami|blocked|scan|upload|ishere|meet|meetall|whoami|setname|random|run|return|try|swap|atomic|teleport|as|robotnamed|robotnumbered|knows)\\b"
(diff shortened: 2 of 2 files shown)


kostmo (approved):
Makes sense.

src/Swarm/Language/Syntax.hs
```diff
@@ -104,6 +107,9 @@ instance ToJSONKey AbsoluteDir where
 instance FromJSONKey AbsoluteDir where
   fromJSONKey = genericFromJSONKey defaultJSONKeyOptions
 
+-- | A relative direction is one which is defined with respect to the
```
xsebek, line 110: I like this definition, it makes sense to have `DDown` relative this way. 👍

src/Swarm/Game/Step.hs
```diff
@@ -1003,6 +1003,17 @@ execConst c vs s k = do
...
+      mh <- use robotOrientation
+      -- In general, (1) entities might not have an orientation, and
+      -- (2) even if they do, orientation is a general vector, which
+      -- might not correspond to a cardinal direction.  We could make
+      -- 'heading' return a 'maybe dir' i.e. 'unit + dir', or return a
+      -- vector of type 'int * int', but those would both be annoying
+      -- for players in the vast majority of cases.  We rather choose
+      -- to just return the direction 'down' in any case where we don't
+      -- otherwise have anything reasonable to return.
+      return $ Out (VDir (fromMaybe (DRelative DDown) $ mh >>= toDirection)) s k
```
xsebek, line 1016: Why not use `maybe`?

xsebek (approved):
LGTM! 👍

src/Swarm/Language/Syntax.hs
byorgey: Yes, I was confused about `DDown` so I added these comments after I figured it out. :smile:

src/Swarm/Game/Step.hs
byorgey: How would you use it, exactly?

xsebek: Never mind, I thought this was `maybe (DRelative Down) toDirection mh` but I see now its two `Maybe`s. 👍

Issue #397, opened 2022-06-14
Wrong device reported missing


byorgey:
**Describe the bug**
In certain situations with devices that don't exist, an error is generated saying that some other device is missing (even though it is not).

**To Reproduce**

- Start a classic mode game with `--cheat`
- Switch into creative mode and `create "branch predictor"`
- Switch back out of creative mode
- Type `build {move; if (1 > 2) {} {}}`
    - As expected we get an error saying no device provides the `compare` capability.
- Now type `build {move; turn left; if (1 > 2) {} {}}`
    - All we have done is add `turn left`; we would expect to get the same error.  Instead we get an error like `You do not have the devices required for: 'move; turn left; if (1 > 2) {noop} {noop}'  please obtain:   - branch predictor`
    - But we do have a branch predictor, so this error is bogus.  Not sure yet whether the capability checking or error reporting is to blame.

**Additional context**
First noticed this in https://github.com/swarm-game/swarm/pull/373#issuecomment-1154533582 .  Note that once #395 is merged then we *will* have a `comparator` device so the above example won't work any more.

byorgey:
An example that still shows this problem now is `build {move; teleport base (2,3)}` vs `build {move; turn left; teleport base (2,3)}`.  The first gives an appropriate error, whereas the second says we are missing `treads`.

byorgey:
I've traced this as far as seeing that the `Incapable` exception given to `formatExn` is already bogus.  `formatExn` seems to be doing its job properly, the problem is with wherever the `Incapable` exception is generated (probably the `checkRequirements` function).

byorgey:
The example with `teleport` no longer works since we get an error about needing an ADT calculator instead.  However, the problem still exists and can be illustrated by the difference between `build {move; x <- whereami}` (correct) vs `build {move; turn left; x <- whereami}` (incorrect).

byorgey:
Strangely both `move` and `turn left` are required in the examples here.  Omitting either one leads to the correct output.  Including a single one of those commands twice (e.g. `move; move`) also leads to correct output.  I've tried mixing other commands and the only thing so far that seems to trigger the incorrect output is to have *both* a call to `move` and a call to `turn left` before invoking a command without a device, such as `whereami`.

byorgey:
I think the problem has to do with the fact that in this code:

https://github.com/swarm-game/swarm/blob/main/src/Swarm/Game/Step.hs#L1600-L1606

[...]

PR #956, opened 2023-01-04
Add `GPS receiver` device to provide `senseloc` capability

byorgey:
Add a new `GPS receiver` device which enables the `senseloc` capability (and hence the `whereami` command).  Towards #26 .

- My immediate motivation is that I want to be able to define `excursion : cmd unit -> cmd unit` which executes the given command and then returns to the same location and orientation as before.  Being able to use the `whereami` command is one of the last missing pieces of machinery necessary to be able to do this in classic mode (the other is a `heading` command, see #955).
- The proposed recipe for `GPS receiver` is `antenna + circuit + clock + compass`, which is a somewhat difficult recipe (`antenna` requires `silver` which requires a `deep mine`; a `clock` requires `quartz` + a bunch of `iron gears`, etc.).  One might wonder whether we should make the recipe easier since finding out where you are seems like a kind of fundamental operation.  However, consider that in order to even be able to make use of the result of `whereami` you need at least (1) an `ADT calculator` to deal with the pair (which transitively requires `typewriter` -> `circuit` -> `silicon` -> `quartz`) (2) probably things like `comparator` and `calculator` to do anything useful with the coordinates.  By the time you have those things you can definitely already build `circuit` + `clock` + `compass`, and you're probably not that far away from getting some `silver` for an `antenna`.  Also, in practice it's an interesting/fun constraint to build up machinery that has to work entirely based on *relative* position without being able to find out your absolute coordinates.
- For some reason this is causing `Testing/508-capability-subset` to fail.  I think perhaps it is due to #397 ?  I will investigate. *EDIT*: unfortunately, that wasn't it!

### DIFF
--- a/data/scenarios/Testing/956-GPS.yaml
+++ b/data/scenarios/Testing/956-GPS.yaml
@@ -1,0 +1,42 @@
+version: 1
+name: Test GPS receiver device
+description: |
+  Test the `whereami` command and GPS receiver device.
+  https://github.com/swarm-game/swarm/issues/956
+objectives:
+  - condition: |
+      loc <- as base {whereami};
+      return $ loc == (0,0)
+solution: |
+  def x = \n. \c. if (n == 0) {} {c ; x (n-1) c} end;
+  loc <- whereami;
+  turn east; x (-fst loc) move;
+  turn south; x (snd loc) move;
+robots:
+  - name: base
+    dir: [0,1]
+    devices:
+      - dictionary
+      - GPS receiver
+      - treads
+      - compass
+      - ADT calculator
+      - strange loop
+      - comparator
+      - calculator
+      - branch predictor
+      - logger
+world:
+  default: [blank]
+  palette:
+    '^': [grass, null, base]
+    '.': [grass]
+  upperleft: [-6, 4]
+  map: |
+    ^..........
+    ...........
+    ...........
+    ...........
+    ...........
+    ...........
+    ...........
--- a/data/scenarios/Testing/956-GPS.yaml
+++ b/data/scenarios/Testing/956-GPS.yaml
@@ -7,15 +7,25 @@ objectives:
   - condition: |
       loc <- as base {whereami};
       return $ loc == (0,0)
+    goal:
+      - |
+        The goal is to move back to the origin after being
+        unceremoniously teleported to a random location.  The only way
+        to do this is to use a GPS receiver device to learn the
+        current location and use the result to decide how to move.
 solution: |
   def x = \n. \c. if (n == 0) {} {c ; x (n-1) c} end;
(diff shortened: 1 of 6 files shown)


xsebek:
What is going on? 😄 
```
Testing/508-capability-subset:                                                                       FAIL (0.28s)
[374](https://github.com/swarm-game/swarm/actions/runs/3839118001/jobs/6536452772#step:22:375)
        test/integration/Main.hs:263:
[375](https://github.com/swarm-game/swarm/actions/runs/3839118001/jobs/6536452772#step:22:376)
        Player exception: Fatal error: BASE: I could not build my robot! What happened?!
[376](https://github.com/swarm-game/swarm/actions/runs/3839118001/jobs/6536452772#step:22:377)
        
[377](https://github.com/swarm-game/swarm/actions/runs/3839118001/jobs/6536452772#step:22:378)
        Use -p '/Regression tests.Testing\/508-capability-subset/' to rerun this test only.

**Comment by byorgey** [comment]
Unfortunately fixing #397 didn't resolve the failing test. The `build` is failing with an error saying "You do not have required devices, please obtain: GPS receiver", but I don't understand why.  The `GPS` and `Tardis` entities defined in the scenario both provide the `senseloc` capability so it shouldn't matter that we don't have a `GPS receiver`.  But it's like it's not seeing those.

**Comment by byorgey** [comment]
doh!

https://github.com/swarm-game/swarm/blob/629cff3b703541fac332b83973dcc9c4bf8a2fc3/src/Swarm/Game/Entity.hs#L295-L296

Hint: combining two `Map Capability [Entity]` values with `<>` doesn't exactly do what we want. :facepalm:

**Comment by byorgey** [comment]
Well, finally got the tests to pass. Even though #397 wasn't the cause I'm glad I thought it was, because it gave me the push to finally get that sorted. :grin:

File: data/scenarios/Testing/956-GPS.yaml [hunk 1]
```diff
@@ -0,0 +1,42 @@
+version: 1
+name: Test GPS receiver device
+description: |
+  Test the `whereami` command and GPS receiver device.
+  https://github.com/swarm-game/swarm/issues/956
+objectives:
```
kostmo, line 6: Add a `goal` here that explains what the solution is doing to meet the `condition`

kostmo (approved):

data/scenarios/Testing/956-GPS.yaml
byorgey: Oh, good idea.  While I was at it I also thought of a way to make the test much better --- what do you think now?

kostmo: Yes, using `random` is great.

kostmo (approved):

PR #959, opened 2023-01-05
Fix capability checking, and refactor/add lots of comments

### LINKED ISSUE #397: Wrong device reported missing

byorgey:
Fixes #397.  The only way I could understand this in order to fix it was to totally refactor the code and add lots of comments as I went.  I feel like this is some of the most difficult code to wrap one's head around in the codebase.  Hopefully now it's a bit easier to understand (though still not easy, I imagine).


src/Swarm/Game/Step.hs
```diff
@@ -1712,65 +1712,85 @@ execConst c vs s k = do
...
         deviceOK d = parentInventory `E.contains` d || childDevices `E.contains` d
 
-        -- take a pair of device sets providing capabilities that is
-        -- split into (AVAIL,MISSING) and if there are some available
-        -- ignore missing because we only need them for error message
-        ignoreOK ([], miss) = ([], miss)
-        ignoreOK (ds, _miss) = (ds, [])
-
-        (deviceSets, missingDeviceSets) =
-          Lens.over both (nubOrd . map S.fromList) . unzip $
```
byorgey, line 1747: The bug was in the fact that we did a `nubOrd` here --- which could change the length of the list...

src/Swarm/Game/Step.hs
```diff
@@ -1712,65 +1712,85 @@ execConst c vs s k = do
...
-            map (ignoreOK . L.partition deviceOK) capDevices
-
-        formatDevices = T.intercalate " or " . map (^. entityName) . S.toList
-        -- capabilities not provided by any device in inventory
-        missingCaps = S.fromList . map fst . filter (null . snd) $ zip caps deviceSets
```
byorgey, line 1752: ... but here we `zip`ped that possibly-shorter list with `caps`, which started out having the same length.  But with repeated things that got nubbed away, the two lists were no longer aligned, which sometimes caused the wrong devices to be suggested for an unrelated missing capability.

xsebek (approved):
The detailed comments are great! :+1:

We should really split the `Step` module, it just keeps growing. :sweat_smile: 

EDIT:
> Honestly not sure how to write automated tests for this, any ideas?

I would add a test scenario with several entities providing the same capabilities that would trigger the wrong behaviour in the original version.

byorgey:
> We should really split the `Step` module, it just keeps growing. :sweat_smile: 

For sure!

> > Honestly not sure how to write automated tests for this, any ideas?
> 
> I would add a test scenario with several entities providing the same capabilities that would trigger the wrong behaviour in the original version.

Sure, good idea.

byorgey:
Added a test case and confirmed it fails on current `main` but passes with this fix!

data/scenarios/Testing/397-wrong-missing.yaml
```diff
@@ -0,0 +1,42 @@
...
+  device to be missing.
+  https://github.com/swarm-game/swarm/issues/397
+objectives:
+  - condition: |
+      as base { ishere "rock" }
+    goal:
+    - |
+      This is a dummy condition that just ensures the base has
+      finished trying to run the problematic `build` command.  The
+      scenario *should* generate an error message; what really care
```
xsebek, line 16:
```suggestion
      scenario *should* generate an error message; what we really care
```

data/scenarios/Testing/397-wrong-missing.yaml
```diff
@@ -0,0 +1,42 @@
...
+      about is whether the generated error message is correct, which
+      is checked in test/integration/Main.hs .
+solution: |
+  build {require 1 "rock"; wait 4; place "rock"};
```
xsebek, line 20: I guess you could just check `t <- time; return (t==5)` if that is the intent. :slightly_smiling_face:

test/integration/Main.hs
```diff
@@ -216,17 +219,27 @@ testScenarioSolution _ci _em =
         , testSolution Default "Testing/710-multi-robot"
         , testSolution Default "Testing/920-meet"
         , testSolution Default "Testing/955-heading"
+        , testSolution' Default "Testing/397-wrong-missing" AllowBadErrors $ \g -> do
+            let msgs =
+                  (g ^. messageQueue . to seqToTexts)
+                    <> (g ^.. robotMap . traverse . robotLog . to seqToTexts . traverse)
```
xsebek, line 225: Those are some impressive optics. :nerd_face:

byorgey: Haha, thanks, though I don't think it's that bad: `^..` means to list all the targets of a `Traversal` instead of just projecting out a single value.  `traverse` just gets all the sub-things.  So this means "get the robot map, then get all the robots out of it, and for each one get its log, convert it to a list of texts..." The last `traverse` is just so we get a single list of texts instead of a list of lists.

kostmo:
> To see whether the fix worked you have to actually inspect the text of the error message.

I've found that an effective strategy for this kind of issue is to define a global sum type for all possible errors, and adopt this as a standard for errors across the codebase.  Any relevant information at the error site is collected in the fields of this sum type's constructors, which is separately used by some toplevel error-message-rendering function to produce the human-readable text.

Then the test suite can pattern match on the error enum for circumstances where a certain error is expected to be produced.

byorgey:
Yes, that's a nice strategy, and we already do that to some extent with our `Exn` type.  It might be a nice idea to preserve structured data such as `Exn` as long as possible, for example, storing `Exn` values directly in robot logs rather than storing a serialized `Text` version, so we can retrieve those `Exn` values intact as part of a test suite or whatever.  That would definitely make the test suite code I wrote a bit nicer.  However, in this case the main thing I was complaining about wasn't so much that we have to look at the *text* of the error message but simply that we have to look at it *at all*.

PR #962, opened 2023-01-05
Fix `EntityMap` merging

### LINKED ISSUE #961: Custom entities with capabilities are not used to provide the capability
**Describe the bug**
If you create a custom entity in a scenario that provides a capability, that custom entity will never be installed to provide that capability, unless there are no built-in entities that provide that capability.

This was discovered in #956, where adding a built-in entity to provide the `senseloc` capability broke an unrelated test case with custom entities providing `senseloc`.

I know how to fix this and will do so shortly --- just creating an issue for it so we have a good documentation trail.

byorgey:
Fixes #961.  Merging the `entitiesByCap` maps simply with `(<>)` (which is the same as `M.union`) was not correct: map union is left-biased, so if both maps had entities keyed to the same capability, the entities from the right-hand map would simply be thrown away.  Instead, we must merge them with `M.unionWith (<>)` which will append the two `Entity` lists for a given capability.

### DIFF
--- a/data/scenarios/Testing/00-ORDER.txt
+++ b/data/scenarios/Testing/00-ORDER.txt
@@ -18,3 +18,4 @@
 920-meet.yaml
 955-heading.yaml
 397-wrong-missing.yaml
+961-custom-capabilities.yaml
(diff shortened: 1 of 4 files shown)


xsebek (approved):
Nice find, that `<>` looked perfectly valid to me. :+1:

PR #965, opened 2023-01-05
Merge new requirements context with existing when running from REPL

### LINKED ISSUE #964: `build` doesn't install devices required by definitions
**Describe the bug**
When `build`ing a robot, it seems that devices required by the robot's program aren't installed if they are required by a definition (rather than directly).

**To Reproduce**
- Start a classic mode game
- `def m1 = move end`
- `build {m1}`

**Expected behavior**
The constructed robot should have `treads` installed, and should move 1 step forward.

byorgey:
Fixes #964.  Unfortunately I don't know of any way to make a test for this.  Part of the reason we went so long without noticing is that it only specifically affected the code path for running terms from the REPL, and tests can never run that code.

### DIFF
--- a/src/Swarm/TUI/Controller.hs
+++ b/src/Swarm/TUI/Controller.hs
@@ -836,7 +836,7 @@ handleREPLEventTyping = \case
             -- `reprogram` at runtime.  See the discussion at
             -- https://github.com/swarm-game/swarm/pull/827 for more
             -- details.
-            . (gameState . baseRobot . robotContext . defReqs .~ reqCtx)
+            . (gameState . baseRobot . robotContext . defReqs <>~ reqCtx)
             -- Set up the robot's CESK machine to evaluate/execute the
             -- given term, being sure to initialize the CESK machine
             -- environment and store from the top-level context.


src/Swarm/TUI/Controller.hs
```diff
@@ -836,7 +836,7 @@ handleREPLEventTyping = \case
             -- `reprogram` at runtime.  See the discussion at
             -- https://github.com/swarm-game/swarm/pull/827 for more
             -- details.
-            . (gameState . baseRobot . robotContext . defReqs .~ reqCtx)
+            . (gameState . baseRobot . robotContext . defReqs <>~ reqCtx)
```
xsebek, line 839: > We simply dump the entire `reqCtx` into the robot's context

Should the comment say _add_ instead of _dump_? :slightly_smiling_face: To me, dumping could mean overwrite which I guess it did.

xsebek (approved):

xsebek:
@byorgey could you also change the line in integration tests?

Right now it works, but if we ever touch requirements while loading scenarios (e.g. adding starting definitions) then they will be silently overwritten.

PR #971, opened 2023-01-07
scenario with enemies

### LINKED ISSUE #1146: Nethack homage: Amulet of Yoneda
In [NetHack](https://www.nethack.org/) ([wiki](https://nethackwiki.com/wiki/Amulet_of_Yendor)):
> The ultimate goal of a game of NetHack is to find the **Amulet of Yendor** 

We should make obtaining the "Amulet of [Yoneda](https://en.wikipedia.org/wiki/Yoneda_lemma)" a goal of a custom scenario, or maybe even make it the toplevel goal of the game?

kostmo:
Uses the `meet` command to simulate a "hitbox" for enemy encounters.

Closes #1146

image

### DIFF
--- a/data/scenarios/Challenges/_blender/solution.sw
+++ b/data/scenarios/Challenges/_blender/solution.sw
@@ -1,0 +1,64 @@
+def doN = \n. \f. if (n > 0) {f; doN (n - 1) f} {}; end;
+
+def slowM = move; wait 2; end;
+
+def navigatePath =
+    doN 9 move;
+    turn left;
+    wait 13; // Get this timing right
+    doN 12 slowM;
+    turn right;
+    doN 8 move;
+    turn right;
+
+    wait 0; // Get this timing right
+    doN 12 slowM;
+    turn left;
+    doN 8 move;
+
+    turn left;
+    wait 0; // Get this timing right
+    doN 12 slowM;
+    turn left;
+    move;
+
+    k <- grab;
+    equip k;
+
+    doN 5 move;
+    drill forward;
+
+    doN 2 move;
+    turn left;
+
+    wait 0; // Get this timing right
+    doN 4 move;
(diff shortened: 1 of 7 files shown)


data/scenarios/Challenges/_blender/solution.sw
```diff
@@ -0,0 +1,64 @@
...
+    turn left;
+
+    wait 0; // Get this timing right
+    doN 4 move;
+    turn right;
+    drill forward;
+    doN 2 move;
+
+    // Close the door behind us!
+    drill back;
```
byorgey, line 41: This doesn't seem to be required --- the patrolling robots don't seem to go through unlocked doors even if it's the direction they would normally turn.  When I was solving the challenge I didn't think to lock the door behind me and I didn't have an issue.  I like this though --- I think we should fix the patrolling robots so that they *do* go through unlocked doors, it makes it more fun/interesting.

data/scenarios/Challenges/blender.yaml
```diff
@@ -0,0 +1,226 @@
...
+            // lowest index at which a robot does not exist
+            return false;
+          };
+          end;
+
+      anyHasMetBase 1;
+robots:
+  - name: base
+    dir: [1, 0]
+    devices:
```
byorgey, line 67: Can we get a `compass` please?  I solved it using a key handler + making a driving mode so I could do it manually, but I had to use the version you evidently prefer, of having up mean `move` and left/right mean `turn`.  However, my brain finds that profoundly confusing and I had to restart a few times since I messed up the turning.  I prefer having e.g. up mean `turn north; move`, left mean `turn west; move` and so on.

byorgey (approved):
Neat!  There seems to be some duplicated code but I guess that's something we need proper imports etc. to fix.

PR #972, opened 2023-01-07
swarm-lang IDE hover

### LINKED ISSUE #124: Implement language server OnHover handler
Once source locs are attached to the Type info, we can implement the hover handler to display the type of the current expression. Here is an example of such handler: [Dhall.LSP.Backend.Typing](https://github.com/dhall-lang/dhall-haskell/blob/1366883c2891d1bec411aef4616cfc7d177478a4/dhall-lsp-server/src/Dhall/LSP/Backend/Typing.hs#L32)

kostmo:
Closes #124

To test:
1. `stack install`
2. Close and re-open VS Code
3. Open a `.sw` file and mouseover some text.

### DIFF
--- a/src/Swarm/Language/LSP.hs
+++ b/src/Swarm/Language/LSP.hs
@@ -20,9 +20,11 @@ import Data.Text (Text)
 import Data.Text.IO qualified as Text
 import Language.LSP.Diagnostics
 import Language.LSP.Server
+import Language.LSP.Types (Hover (Hover))
 import Language.LSP.Types qualified as J
 import Language.LSP.Types.Lens qualified as J
 import Language.LSP.VFS
+import Swarm.Language.LSP.Hover qualified as H
 import Swarm.Language.Parse
 import Swarm.Language.Pipeline
 import System.IO (stderr)
@@ -111,4 +113,14 @@ handlers =
           Just vf@(VirtualFile _ version _rope) -> do
             validateSwarmCode doc (Just $ fromIntegral version) (virtualFileText vf)
           _ -> debug $ "No virtual file found for: " <> from (show msg)
+    , requestHandler J.STextDocumentHover $ \req responder -> do
+        debug $ "I am in hover handler: " <> from (show req)
+        let doc = req ^. J.params . J.textDocument . J.uri . to J.toNormalizedUri
+            pos = req ^. J.params . J.position
+        mdoc <- getVirtualFile doc
+        let maybeHover = do
+              vf <- mdoc
+              markdownText <- H.showHoverInfo doc Nothing pos vf
+              return $ Hover (J.HoverContents $ J.MarkupContent J.MkMarkdown markdownText) Nothing
+        responder $ Right maybeHover
     ]
--- a/src/Swarm/Language/LSP.hs
+++ b/src/Swarm/Language/LSP.hs
@@ -114,13 +114,12 @@ handlers =
             validateSwarmCode doc (Just $ fromIntegral version) (virtualFileText vf)
           _ -> debug $ "No virtual file found for: " <> from (show msg)
     , requestHandler J.STextDocumentHover $ \req responder -> do
-        debug $ "I am in hover handler: " <> from (show req)
         let doc = req ^. J.params . J.textDocument . J.uri . to J.toNormalizedUri
             pos = req ^. J.params . J.position
         mdoc <- getVirtualFile doc
         let maybeHover = do
               vf <- mdoc
-              markdownText <- H.showHoverInfo doc Nothing pos vf
-              return $ Hover (J.HoverContents $ J.MarkupContent J.MkMarkdown markdownText) Nothing
+              (markdownText, maybeRange) <- H.showHoverInfo doc Nothing pos vf
+              return $ Hover (J.HoverContents $ J.MarkupContent J.MkMarkdown markdownText) maybeRange
         responder $ Right maybeHover
     ]
--- a/src/Swarm/Language/LSP/Hover.hs
+++ b/src/Swarm/Language/LSP/Hover.hs
@@ -2,62 +2,84 @@
 
 module Swarm.Language.LSP.Hover where
 
+import Data.Graph
 import Data.Text (Text)
+import Data.Text qualified as T
 import Data.Text.Utf16.Rope qualified as R
 import Language.LSP.Types qualified as J
 import Language.LSP.VFS
 import Swarm.Language.Parse (readTerm')
 import Swarm.Language.Syntax
+import Swarm.Util qualified as U
 
 withinBound :: Int -> SrcLoc -> Bool
 withinBound pos (SrcLoc s e) = pos >= s && pos < e
 withinBound _ NoLoc = False
+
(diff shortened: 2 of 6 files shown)


src/Swarm/Language/LSP/Hover.hs
```diff
@@ -0,0 +1,93 @@
...
+  Int ->
+  -- | next element to inspect
+  Syntax ->
+  Term
+descend t pos (Syntax l1 t1) =
+  if withinBound pos l1
+    then digAt t1 pos
+    else t
+
+descend2 ::
```
kostmo, line 46: Is there a nicer way to express this, that does not require defining `descend` and `descend2`?

src/Swarm/Language/LSP/Hover.hs
```diff
@@ -0,0 +1,93 @@
...
+  SPair s1 s2 -> descend2 t pos s1 s2
+  SDef _ _ _ s -> descend t pos s
+  SBind _ s1 s2 -> descend2 t pos s1 s2
+  SDelay _ s -> descend t pos s
+  x -> x
+
+explain :: Term -> Text
+explain = \case
+  TUnit -> "The unit value."
+  TConst c -> briefDoc $ constDoc $ constInfo c
```
kostmo, line 75: Here I utilize the programmatic documentation of `TConst`.  For all of the other terms, I copied and pasted the Haddocks.

src/Swarm/Language/LSP.hs
```diff
@@ -111,4 +113,14 @@ handlers =
           Just vf@(VirtualFile _ version _rope) -> do
             validateSwarmCode doc (Just $ fromIntegral version) (virtualFileText vf)
           _ -> debug $ "No virtual file found for: " <> from (show msg)
+    , requestHandler J.STextDocumentHover $ \req responder -> do
+        debug $ "I am in hover handler: " <> from (show req)
```
kostmo, line 117: Remove

src/Swarm/Language/LSP/Hover.hs
```diff
@@ -0,0 +1,109 @@
...
+withinBound _ NoLoc = False
+
+showHoverInfo ::
+  J.NormalizedUri ->
+  J.TextDocumentVersion ->
+  J.Position ->
+  VirtualFile ->
+  Maybe Text
+showHoverInfo _ _ (J.Position myLine myCol) vf@(VirtualFile _ _ myRope) =
+  astSize
```
kostmo, line 23: Rename

src/Swarm/Language/LSP.hs
```diff
@@ -111,4 +113,14 @@ handlers =
...
+        let doc = req ^. J.params . J.textDocument . J.uri . to J.toNormalizedUri
+            pos = req ^. J.params . J.position
+        mdoc <- getVirtualFile doc
+        let maybeHover = do
+              vf <- mdoc
+              markdownText <- H.showHoverInfo doc Nothing pos vf
+              return $ Hover (J.HoverContents $ J.MarkupContent J.MkMarkdown markdownText) Nothing
```
kostmo, line 124: Supply the Syntax SrcLoc as the Range here (using rope to convert back to line/col)

byorgey (approved):
Looks nice!  I can confirm that this also works in emacs.

How hard would it be to also recurse into type annotations (on `def` or `let`) and generate a tooltip about the specific type one is hovering on?  Right now, if I hover over a type it just says something about `def`.  Feel free to leave that to another PR though.

src/Swarm/Language/LSP/Hover.hs
```diff
@@ -0,0 +1,167 @@
...
+  TText {} -> pure $ pureDoc "A text literal."
+  TAntiText {} -> pure $ pureDoc "An antiquoted Haskell variable name of type Text."
+  TBool {} -> pure $ pureDoc "A Boolean literal."
+  TRobot {} -> pure $ pureDoc "A robot reference.  These never show up in surface syntax, but are here so we can factor pretty-printing for Values through pretty-printing for Terms."
+  TRef {} -> pure $ pureDoc "A memory reference.  These likewise never show up in surface syntax but are here to facilitate pretty-printing."
+  TRequireDevice {} -> pure $ pureDoc "Require a specific device to be installed."
+  TRequire {} -> pure $ pureDoc "Require a certain number of an entity."
+  TVar x -> pure $ pureDoc $ "var: " <> U.bquote x
+  SLam {} -> pure $ pureDoc "A lambda expression, with or without a type annotation on the binder."
+  SApp (Syntax _ t1) (Syntax _ t2) ->
```
kostmo, line 150: I'm now rendering function application as a tree:

f

byorgey: This seems like overkill to me.  It's a lot of information and it's not immediately clear how the nested bullet points correspond to the syntax.  What's the intended audience for these hover tips?

kostmo: My original motivation was that when hovering over operators, all it would say about them was "function application".  So I wanted to see if I could get a bit more detail out of it, e.g. for `==`, now it at least mentions equality:
Screenshot from 2023-01-07 12-36-55

However, for functions of multiple arguments it does get messy.  So this is still experimental.  I would kind of like to commit the progress on this tree rendering and see if I can wrangle some more useful output from it.

src/Swarm/Language/LSP/Hover.hs
```diff
@@ -0,0 +1,167 @@
...
+          [ "A"
+          , (if isRecursive then "" else "non-") <> "recursive"
+          , "let expression"
+          , if null maybeTypeAnnotation then "without" else "with"
+          , "a type annotation on the variable."
+          ]
+  SPair {} -> pure $ pureDoc "A pair."
+  SDef {} -> pure $ pureDoc "A (recursive) definition command, which binds a variable to a value in subsequent commands."
+  SBind {} -> pure $ pureDoc "A monadic bind for commands, of the form `c1 ; c2` or `x <- c1; c2`."
+  SDelay {} -> pure $ pureDoc "Delay evaluation of a term, written `{...}`.  Swarm is an eager language, but in some cases (e.g. for `if` statements and recursive bindings) we need to delay evaluation.  The counterpart to `{...}` is `force`, where `force {t} = t`. Note that 'Force' is just a constant, whereas 'SDelay' has to be a special syntactic form so its argument can get special treatment during evaluation."
```
byorgey, line 152: This description probably needs to be cut down a lot.

src/Swarm/Language/LSP/Hover.hs
kostmo: Yes, I like that better.

byorgey (approved):

kostmo:
> Right now, if I hover over a type it just says something about `def`. Feel free to leave that to another PR though.

Do you suppose we can merge as is now?  There's certainly more iteration to be done, but I think it's at a decent milestone such that others (or me, sometime later) can continue with refinements.

byorgey:
> Do you suppose we can merge as is now? There's certainly more iteration to be done, but I think it's at a decent milestone such that others (or me, sometime later) can continue with refinements.

Yes, I think that is reasonable.

Issue #866, opened 2022-11-15
Code golf scoring


kostmo:
It would be fun to keep track of the "best" solution in terms of code size.

The code size might be determined by summing all of the REPL executions plus all transitive inclusions of code (`include` or `run`).

To discourage short variable names, it may be best to use "symbol count" rather than "character count" as the metric.

byorgey:
> symbol count

Or number of AST nodes.

byorgey:
Personally, I enjoy trying to optimize character count, but tastes vary.

kostmo:
> trying to optimize character count

That can be fun, but I think there's more challenge in crafting a parsimonious program structure than shrinking variable names.  I recognize appeal of finding out in which circumstances whitespace or semicolons are optional as well.

Perhaps we can track both AST node count and character count as metrics?

byorgey:
> Perhaps we can track both AST node count and character count as metrics?

Sure, why not? :smile: I agree, both are fun for different reasons.  One is about finding elegant program structures, as you say, and one is about pushing a language to its limits.

xsebek:
This sounds fun and should be easy to add to current best score. 👍

I think it needs an “automated mode” where you can not cancel your solution and start writing in REPL. 🤔 
Once that computation finishes you have either won or _failed_ (new state) and can leave or continue experimenting, now with REPL.

byorgey:
> I think it needs an “automated mode” where you can not cancel your solution and start writing in REPL.

I don't think I understand - what's the motivation?

xsebek:
@byorgey well if you could continue in REPL, then an empty solution would be always shortest and then just finish it in REPL.

byorgey:
I thought the idea was to count all things typed at the REPL plus anything run.

byorgey:
Ah, I think I see your point now.  It's also harder to share short solutions if they require typing multiple things at the REPL.  I agree, to keep things simple and consistent we should probably just score a single `.sw` file (+ transitive imports) when counting length.

kostmo:
Perhaps for now, one shall only be qualified for code golf scoring when using the `--run` CLI option.  Any repl commands entered before the goal is reached shall disqualify the attempt.

xsebek:
Maybe we could allow either `—run` or one REPL entry. 🤔

Then we could have states NoEntry, EntryWithLength and MoreEntries. If you finish the game with second state, it’s considered for shortest solution.

kostmo:
[...]

PR #974, opened 2023-01-07
Record best code size

kostmo:
towards #866

NOTE: #1116 should be merged first so that the schema change of save files is less disruptive.

## Examples

Different criteria can have their own best score:
image

All criteria share the same single best score:
image

## Behavior notes

* As currently designed, the code size will only be scored if the the player has specified their code **before** the scenario begins.  Furthermore, any input into the REPL will invalidate code size scoring for the current game.
* Because of this, the only way to score code so far is with a command-line argument of `--run` or `--autoplay`.  However, #1010  shall implement code size scoring when a scenario is launched from the UI.
* In the "best scores" display, if multiple "best score" criteria were all from the same game, they will be consolidated.  If all criteria are for the same game, the criteria labels will be omitted.
* The code size metrics will not be displayed if a "best score" was not obtained via `--run`.

### `run` command

Currently, the code entailed in a `run "somescript.sw"` command is not transitively included, so using `run` make the code size score meaningless.

### Unit tests

Run the subset of unit tests:

    scripts/run-tests.sh --test-arguments '--pattern "Tests.Precedence"'

### Manual integration tests

First, reset the score:

    rm -f ~/.local/share/swarm/saves/Tutorials_grab.yaml

Saving the following to `grab-soln.sw`:
```
move;
move; grab; turn back; move; turn back; move;
move; grab; turn back; move; turn back; move;
move; grab; turn back; move; turn back; move;
move; grab; turn back; move; turn back; move;
move; grab; turn back; move; turn back; move;
move; grab;
```
Run as follows:

    scripts/play.sh --scenario Tutorials/grab.yaml --run grab-soln.sw

This should establish a record for code size.

Then, play the Grab tutorial and immediately paste and run this in the REPL:

    move; move; grab; move; grab; move; grab; move; grab; move; grab; move; grab;

This solution is faster in terms of time, but should not displace the code-length record, since no code length should be recorded from a REPL solution.


src/Swarm/Language/Syntax.hs
```diff
@@ -965,3 +966,17 @@ mapFree1 x f = fvT %~ (\t -> if t == TVar x then f t else t)
 
 lowShow :: Show a => a -> Text
 lowShow a = toLower (from (show a))
+
+-- | Note that the size constants for the recursive constructors
+-- are somewhat arbitrary; perhaps we should take into account
```
byorgey, line 971: Why arbitrary constants?  Why not just count 1 for each constructor?

kostmo: Sure, that's consistent at least.

src/Swarm/TUI/View.hs
```diff
@@ -222,11 +224,7 @@ drawNewGameMenuUI (l :| ls) =
 
   drawDescription :: ScenarioItem -> Widget Name
   drawDescription (SICollection _ _) = txtWrap " "
-  drawDescription (SISingle (s, si)) = do
-    let oneBest = si ^. scenarioBestTime == si ^. scenarioBestTicks
-    let bestRealTime = if oneBest then "best:" else "best real time:"
-    let noSame = if oneBest then const Nothing else Just
-    let lastText = let la = "last:" in padRight (Pad $ T.length bestRealTime - T.length la) (txt la)
+  drawDescription (SISingle (s, si)) =
     vBox . catMaybes $
```
kostmo, line 228: The padding logic is getting hairy here; it would probably be cleaner to convert this to a Table widget.

src/Swarm/Game/ScenarioInfo.hs
```diff
@@ -153,22 +160,37 @@ scenarioBestTime :: Lens' ScenarioInfo ScenarioStatus
...
+updateScenarioInfoOnQuit
+  csd
+  z
+  ticks
+  completed
+  (ScenarioInfo p s bTime bTicks _prevCodeSize) = case s of
+    InProgress start _ _ ->
+      let el = (diffUTCTime `on` zonedTimeToUTC) z start
+          cur = (if completed then Complete else InProgress) start el ticks
+          -- TODO Offload this logic to a Semigroup instance of ScenarioStatus
```
kostmo, line 187: Would a `Semigroup` instance be a cleaner way to encode the "precedence" behavior?

src/Swarm/Game/Scenario/Scoring/CodeSize.hs
```diff
@@ -0,0 +1,30 @@
...
+
+charCount :: SrcLoc -> Int
+charCount NoLoc = 0
+charCount (SrcLoc start end) = end - start
+
+codeSizeFromDeterminator :: CodeSizeDeterminators -> Maybe ScenarioCodeMetrics
+codeSizeFromDeterminator (CodeSizeDeterminators maybeInitialCode usedRepl) = do
+  guard $ not usedRepl
+  ProcessedTerm (Module (Syntax' srcLoc term _) _) _ _ <- maybeInitialCode
+  return $ ScenarioCodeMetrics (charCount srcLoc) (measureAstSize term)
```
kostmo, line 30: I believe that both of these computations are correct, since it at least works when measuring the code size for `--autoplay` with inline `solution` code.

src/Swarm/Game/ScenarioInfo.hs
```diff
@@ -153,22 +160,37 @@ scenarioBestTime :: Lens' ScenarioInfo ScenarioStatus
...
+  Integer ->
+  Bool ->
+  ScenarioInfo ->
+  ScenarioInfo
+updateScenarioInfoOnQuit
+  csd
+  z
+  ticks
+  completed
+  (ScenarioInfo p s bTime bTicks _prevCodeSize) = case s of
```
kostmo, line 183: TODO: need to take `_prevCodeSize` into account and use the smaller of the two.

kostmo:
Hey @byorgey and @xsebek, would like to get feedback on the direction I'm taking this.

xsebek:
I will get to review it in the next few days, but re:
> In the first version [the preuse line](https://github.com/swarm-game/swarm/blob/ff2bf577b43851188e696eaef20c54a02ec681a4/src/Swarm/TUI/Controller.hs#L452) yields Nothing, thus bypassing the save. I guess it is because the scenario path must match the "normalized" version, which strips the leading data/scenarios/ or scenarios/.

This is a bug, I consciously coded the name resolution to normalize paths and symlinks. I think it worked at some point, but I would not swear it. :smile: We should have an integration test that all paths we might use to scenario resolve to the same. @kostmo kindly open an Issue, it should not be necessary for this feature. :slightly_smiling_face:

src/Swarm/Game/ScenarioInfo.hs
byorgey: Yes, that seems plausible to me although I am not 100% clear on the details.

src/Swarm/Language/Syntax.hs
```diff
@@ -1111,3 +1112,24 @@ freeVarsV = freeVarsT . (\f -> \case TVar x -> TVar <$> f x; t -> pure t)
...
+-- | Each constructor is a assigned a value of 1, plus
+-- any recursive syntax it entails.
+--
+-- Note that in this case statement, all of the constructors
+-- with nested Syntax are given explicit
+-- cases, whereas a catchall for non-recursive
+-- constructors is placed at the end.
+-- This catch-all is "dangerous" if we ever
+-- add more recursive terms to the language; the
+-- compiler won't warn us about missing cases.
```
byorgey, line 1125: I'd rather just list all the other cases explicitly.  Putting this comment here doesn't solve anything.  If my past experience is any guide, if we leave the catch-all case, we WILL add another AST node in the future, we WILL forget to update this code, and it WILL eventually bite us.

byorgey: On second thought, instead we should just do this one simple future-proof trick:
```
import Data.Lens.Plated (universe)

measureAstSize :: Data a => Term' a -> Int
measureAstSize = length . universe
```
:tada: 

`universe` generates a list of all recursive subterms of the same type, which is exactly what we want to count here.  You should check that this gives the same results as your explicit version was giving, but I think it should.

src/Swarm/Game/ScenarioInfo.hs
byorgey: Sounds like a `Semigroup` on `ScenarioCodeMetrics`.

src/Swarm/Game/Scenario/Scoring/CodeSize.hs
byorgey: Looks good to me.

byorgey (commented):
I think this is heading in a good direction.

src/Swarm/Language/Syntax.hs
kostmo: > `measureAstSize = length . universe`

That is slick.  I've come up with something similar, pushed in the latest commit: count the nodes of the `Tree` generated using code from #1014.  Does there happen to exist any Lens magic to automatically generate this tree?

byorgey: There sure does, see my comment on #1014 .

src/Swarm/Language/Syntax.hs
```diff
@@ -1113,23 +1115,33 @@ freeVarsV = freeVarsT . (\f -> \case TVar x -> TVar <$> f x; t -> pure t)
 mapFreeS :: Var -> (Syntax' ty -> Syntax' ty) -> Syntax' ty -> Syntax' ty
 mapFreeS x f = freeVarsS %~ (\t -> case t ^. sTerm of TVar y | y == x -> f t; _ -> t)
 
+syntaxAsTree :: Syntax' a -> Tree (Syntax' a)
+syntaxAsTree s0@(Syntax' _ t _) = case t of
```
byorgey, line 1119: I'll put the comment here too, to keep things in one place:
```
import Control.Lens.Plated (para)

syntaxAsTree :: Data a => Syntax' a -> Tree (Syntax' a)
syntaxAsTree = para Node
```

kostmo: :exploding_head:

kostmo:
Ready for another look, @byorgey and @xsebek.  I've updated the main PR description with screenshots.

src/Swarm/Game/Scenario/Status.hs
```diff
@@ -0,0 +1,235 @@
...
+  rearrangeTuples = map (snd . NE.head &&& NE.map fst) . groupByStartTime
+
+-- * High-level status of scenario play
+
+-- | A "ScenarioStatus" stores the status of a scenario along with
+--   appropriate metadata: "NotStarted", or "Played".
+--   The "Played" status has two sub-states: "Attempted" or "Completed".
+data ScenarioStatus
+  = NotStarted
+  | Played ProgressMetric BestRecords
```
byorgey, line 168: This makes for a nice test case to discuss our general approach to breaking changes (there has been some discussion of this before though I cannot find it at the moment).   The constructors of the `ScenarioStatus` type changed, and therefore the JSON format for saving `ScenarioStatus` values has changed. This means that if you have played any scenarios previously, and then update to a version of Swarm with this PR merged, it will break because it will complain that it cannot parse the scenario info files, and the only way to fix it is to blow away `~/.local/share/swarm/saves`.  This is annoying both because users will not know what they need to do, and also because they will necessarily lose all their saved progress.

Since we are just in alpha mode, I am not opposed to breaking things.  However, eventually, once we make a 1.0 release we will want to be more careful about backwards compatibility.  This could be a good opportunity to practice/experiment.

A concrete proposal is to adopt the policy that version x.n can always read save files written by version x.n and x.(n-1) (but it will only write them using version x.n format).  So you can upgrade just by running the new version and your old files will be converted.  In this case we could put in some special code to parse the old scenario info file format and turn it into the new format.  We might want to add version numbers like we already did for scenario description files.

xsebek: Well the “saves” are very alpha, but in the future we could utilize named optional fields or custom FromJSON instances to parse previous versions.

At minimum we could produce a nice failure when encountering the old style save. @kostmo if you want, this should be easy with custom instance and `fail` with message on alternative (`<|>`).

src/Swarm/Game/Scenario/Scoring/Metrics.hs
```diff
@@ -0,0 +1,55 @@
...
+
+scenarioOptions :: Options
+scenarioOptions =
+  defaultOptions
+    { fieldLabelModifier = map toLower . drop (length "_scenario")
+    }
+
+data DurationMetrics = DurationMetrics
+  { _scenarioElapsed :: NominalDiffTime
+  -- ^ Time elapsed until quitting the scenario.
```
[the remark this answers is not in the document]
xsebek: Ups, should have changed that. 😅

@kostmo would you be so kind and update it since you are touching this code please? 🙂

src/Swarm/Game/Scenario/Status.hs
kostmo: > it will break because it will complain that it cannot parse the scenario info files, and the only way to fix it is to blow away `~/.local/share/swarm/saves`.

This is precisely why I turned loading failure from a **fatal** error into a **warning** in #1116.  But, it is true, the saved progress will still be gone.

kostmo: > A concrete proposal is to adopt the policy that version x.n can always read save files written by version x.n and x.(n-1) (but it will only write them using version x.n format). So you can upgrade just by running the new version and your old files will be converted.

I like this strategy, as there is a definite protocol for purging the tech debt of legacy format-parsing code.

For now, though, let's enjoy the privilege of an alpha version and ruthlessly refactor, unburdened by tech debt!  Perhaps the beta release is the time to commission support for legacy format parsing.

kostmo: > At minimum we could produce a nice failure when encountering the old style save.

Filed #1124 for this.

src/Swarm/Game/Scenario/Scoring/Metrics.hs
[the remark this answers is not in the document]
kostmo: :+1:

src/Swarm/Game/Scenario/Status.hs
```diff
@@ -0,0 +1,235 @@
...
+
+makeLensesWith (lensRules & generateSignatures .~ False) ''BestRecords
+
+-- | The best status of the scenario, measured in real world time.
+scenarioBestByTime :: Lens' BestRecords ProgressMetric
+
+-- | The best status of the scenario, measured in game ticks.
+scenarioBestByTicks :: Lens' BestRecords ProgressMetric
+
+-- | The best code size of the scenario, measured both in character count and AST size.
```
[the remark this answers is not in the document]
kostmo: :+1:

kostmo:
Feedback addressed.

byorgey (approved):
:rocket:

src/Swarm/TUI/Controller.hs
```diff
@@ -454,7 +455,15 @@ saveScenarioInfoOnFinish = do
...
-        currentScenarioInfo %= updateScenarioInfoOnFinish t ts won
+
+        replHist <- use $ uiState . uiREPL . replHistory
+        let determinator = CodeSizeDeterminators initialCode $ replHist ^. replHasExecutedManualInput
+        currentScenarioInfo
+          %= updateScenarioInfoOnFinish
+            determinator
+            t
+            ts
+            won
```
xsebek, line 466: What a funny way to indent, I take it you put it on a new line and Fourmolu did the rest?

src/Swarm/Language/Syntax.hs
```diff
@@ -1027,3 +1030,13 @@ freeVarsV = freeVarsT . (\f -> \case TVar x -> TVar <$> f x; t -> pure t)
...
 mapFreeS x f = freeVarsS %~ (\t -> case t ^. sTerm of TVar y | y == x -> f t; _ -> t)
+
+-- | Transform the AST into a Tree datatype.
+-- Useful for pretty-printing (e.g. via "Data.Tree.drawTree").
+asTree :: Data a => Syntax' a -> Tree (Syntax' a)
+asTree = para Node
+
+-- | Each constructor is a assigned a value of 1, plus
+-- any recursive syntax it entails.
+measureAstSize :: Data a => Syntax' a -> Int
```
xsebek, line 1041: Could we have a doctest that shows two solution code sizes?

I would like to see that this really measures the constructors that I hope it measures.

src/Swarm/Game/Scenario/Status.hs
```diff
@@ -0,0 +1,235 @@
...
+type ProgressMetric = Metric ProgressStats
+
+-- * High scores by various criteria
+
+data BestRecords = BestRecords
+  { _scenarioBestByTime :: ProgressMetric
+  , _scenarioBestByTicks :: ProgressMetric
+  , _scenarioBestByCharCount :: ProgressMetric
+  , _scenarioBestByAstSize :: ProgressMetric
+  }
```
xsebek, line 78: Uh, can we have a `Map BestByCriteria ProgressStats` instead?

We can wait to reach the magic number 7, but this is already more than I would be comfortable touching.

src/Swarm/Game/Scenario/Scoring/Progress.hs
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+instance FromJSON Progress where
+  parseJSON = genericParseJSON untaggedJsonOptions
+
+instance ToJSON Progress where
+  toJSON = genericToJSON untaggedJsonOptions
+
+data Metric a = Metric Progress a
+  deriving (Eq, Ord, Show, Read, Generic, FromJSON, ToJSON)
```
xsebek, line 28: I like the short modules, but it took me a while to find the `Metric` type hidden in `Progress` instead of say `Metrics`.

@kostmo please add module docs and export list. If you find it hard to write one because the functions and types don't make sense in a module with that name, maybe consider a different name or partition.

There may not be technical reasons for partitioning the module, but some humans (I hear) like to give meaning to such nonsense so their little brains can handle numerous modules more easily.

src/Swarm/Game/Scenario/Status.hs
```diff
@@ -0,0 +1,235 @@
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE TemplateHaskell #-}
+{-# OPTIONS_GHC -Wno-orphans #-}
+
+module Swarm.Game.Scenario.Status where
```
xsebek, line 5: If I wrote modules `Scoring.X`, `Scoring.Y` and `Scoring.Z`, I might also write module `Scoring` to reexport the functions which are not internal to the implementation.

This looks like that module. Why is it named `Status` when it deals a lot with scores? :raised_eyebrow: 

If I had to guess what would be in a `Status` module, I would consider the `Objective` state tracking. But I do not see that here. This makes me confused.

xsebek (changes requested):
The code looks functionally OK to me, but I have not tested it manually yet.

I have a few nitpick which might not be worth addressing nor writing an Issue for.

I also have a few questions which could be answered by nice module documentation.
The only module with module level documentation is `Test.Scoring` though. :facepalm:

src/Swarm/Game/Scenario/Status.hs
xsebek: **Why is `ScenarioInfo` not in the `ScenarioInfo` module?!**

@kostmo please treat modules like a helpful tool, not an annoying constraint to bend and break.

byorgey: > please treat modules like a helpful tool, not an annoying constraint to bend and break.

@xsebek I can understand you're frustrated, but I'm not sure how helpful this kind of comment is (attributing certain motivations or attitudes to others).  Let's try to stick to making concrete suggestions on how to improve.

xsebek: Also, you forgot the lenses in the `ScenarioInfo` module:
```Haskell
makePrisms ''ScenarioItem
makePrisms ''ScenarioStatus
```
They might not be used much, so you might be able to delete them.

Honestly, I am surprised TH for generating instances even works in a different module, I have never seen it done that way.

kostmo: Appreciate the reminder about appropriate communication, @byorgey.

I'll make another pass at module organization/docs.

src/Swarm/TUI/Model/Repl.hs
```diff
@@ -111,6 +114,22 @@ replIndex :: Lens' REPLHistory Int
...
+-- | Note: Instead of adding a dedicated field to the REPLHistory record,
+-- an early attempt entailed checking for:
+--
+--    _replIndex > _replStart
+--
+-- However, executing an initial script causes
+-- a "REPLOutput" to be appended to the REPL history,
+-- which increments the replIndex, and thus makes
+-- the Index greater than the Start even though
+-- the player has input not commands into the REPL.
```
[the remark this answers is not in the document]
kostmo: That is one alternative.  Will add a comment about this possibility in the code.

src/Swarm/TUI/Controller.hs
kostmo: Yes, it seems that putting the argument list on its own line causes Fourmolu to give a new lines to each individual argument.

But placing the newline just before `updateScenarioInfoOnFinish` results in a stable Fourmolu indentation on a single extra line.  I'll do that.

src/Swarm/Game/Scenario/Status.hs
kostmo: Created #1132

src/Swarm/Game/Scenario/Status.hs
kostmo: > lenses in the `ScenarioInfo` module

Tried moving them but got some odd "Not in scope" errors.  Can follow up on this in a later PR.

src/Swarm/Language/Syntax.hs
kostmo: Added several unit tests.

src/Swarm/Game/Scenario/Scoring/Progress.hs
kostmo: Reorganized the modules.

src/Swarm/TUI/View.hs
```diff
@@ -221,32 +206,127 @@ drawNewGameMenuUI (l :| ls) =
...
+        . pairToList
+        . fmap (\x -> T.singleton $ if x == 0 then ',' else ' ')
+    elaboratedCriteria =
+      if hasDistinctByCriteria
+        then
+          map mkCriteriaRow $
+            flip zip [(0 :: Int) ..] $
+              NE.toList $
+                NE.reverse $
+                  NE.map describeCriteria criteria
```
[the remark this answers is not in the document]
kostmo: Changed to dots.

src/Swarm/Util.hs
```diff
@@ -257,6 +258,10 @@ quote t = T.concat ["\"", t, "\""]
 bquote :: Text -> Text
 bquote t = T.concat ["`", t, "`"]
 
+-- | Surround some text in parentheses.
+parens :: Text -> Text
+parens t = T.concat ["(", t, ")"]
```
kostmo, line 263: `parens` rendering function

PR #978, opened 2023-01-07
strip `VResult` before storing value in `REPLDone`

### LINKED ISSUE #977: Fatal error when using `itN` resulting from a bind
**Describe the bug**
Sometimes, using one of the auto-generated `itN` variables can trigger a fatal "bad application of `execConst`" error, because the `itN` contains a `VResult` which the `execConst` code is not expecting to see.

**To Reproduce**
- Start a creative game.
- `create "tree"`
- `def mk = r <- build {}; return r end`
- `mk`
- `give it0 "tree"`

**Expected behavior**
This ought to work.  Probably we just need to strip `VResult` before storing the value of an `itN` variable.

**Screenshots**
itN-fatal

byorgey:
Fixes #977.


src/Swarm/Game/Step.hs
```diff
@@ -139,15 +139,14 @@ gameTick = do
...
-      case v of
+      case stripVResult <$> v of
         -- Log exceptions in the message queue so we can check for them in tests
         Left exn -> do
           let h = hypotheticalRobot (Out VUnit emptyStore []) 0
           em <- use entityMap
           m <- evalState @Robot h $ createLogEntry ErrorTrace (formatExn em exn)
           emitMessage m
         Right (VBool res) -> when res markWin
-        Right (VResult (VBool res) _env) -> when res markWin
```
kostmo, line 150: #927 continues to accrue merge conflicts in [this part of the code](https://github.com/swarm-game/swarm/pull/927#discussion_r1061820199).

src/Swarm/Game/Value.hs
```diff
@@ -87,6 +88,11 @@ data Value where
   VRef :: Int -> Value
   deriving (Eq, Show, Generic, FromJSON, ToJSON)
 
+-- | Ensure that a value is not wrapped in 'VResult'.
+stripVResult :: Value -> Value
+stripVResult (VResult v _) = v
```
xsebek, line 93: Just to make sure, there is no way a `VResult` could contain a `VResult`, right?

src/Swarm/Game/Step.hs
xsebek: @kostmo if you are OK with letting me push to your branch, I can fix it for you. :wink: 

We have a historic number of open PRs now, so I will try to close mine and review and help where I can.

Afterwards, we should make a release, there is a lot of cool stuff to share. :+1:

xsebek (approved):

src/Swarm/Game/Value.hs
byorgey: Uhh... well, it's not *supposed* to, but it's a good question.  We might as well make `stripVResult` recursive, just in case.

PR #954, opened 2023-01-04
One file per achievement record

kostmo:
Closes #953

## Testing

First, earn a couple of achievements by playing the game.  Then, stuff some garbage into the `achievement` directory and rerun the game:

```
kostmo@dell-ubuntu:~/github/swarm$ touch ~/.local/share/swarm/achievement/foo
kostmo@dell-ubuntu:~/github/swarm$ mkdir ~/.local/share/swarm/achievement/bar
kostmo@dell-ubuntu:~/github/swarm$ stack run
Aeson exception:
Error in $: parsing Swarm.TUI.Model.Achievement.Attainment.Attainment(Attainment) failed, expected Object, but encountered Null
Entry "/home/kostmo/.local/share/swarm/achievement/bar" is not a file!
```

Observe that the game starts fine and is able to display our original 2 achievements.

PR #982, opened 2023-01-08
Extend logs for warnings

xsebek:
* split logs into a separate module
* make some logs warnings and fatal (to be continued...)
* prepare failure logging infrastructure for TUI

This should be merged before #954 so it is not squashed.

### DIFF
--- a/src/Swarm/Game/Log.hs
+++ b/src/Swarm/Game/Log.hs
@@ -1,0 +1,74 @@
+{-# LANGUAGE QuasiQuotes #-}
+{-# LANGUAGE TemplateHaskell #-}
+
+-- |
+-- Module      :  Swarm.Game.Log
+-- Copyright   :  Ondřej Šebek
+-- Maintainer  :  ondras98@icloud.com
+--
+-- SPDX-License-Identifier: BSD-3-Clause
+--
+-- A data type to represent in-game logs by robots.
+--
+-- Because of the use of system robots, we sometimes
+-- want to use special kinds of logs that will be
+-- shown to the player.
+--
+-- TODO: Currently we abuse this system for system
+-- logs, which is fun, but we should eventually make
+-- a dedicated `SystemLogEntry` type for 'RuntimeState'
+-- message queue.
+module Swarm.Game.Log (
+  LogSource (..),
+  ErrorLevel (..),
+
+  -- * Robot log entries
+  LogEntry (..),
+  leText,
+  leSaid,
+  leRobotName,
+  leTime,
+  leLocation,
+  leRobotID,
+) where
+
+import Control.Lens hiding (contains)
+import Data.Aeson (FromJSON, ToJSON)
+import Data.Text (Text)
+import GHC.Generics (Generic)
+import Swarm.Util.Location (Location)
+
+-- | Severity of the error - critical errors are bugs
+--   and should be reported as Issues.
+data ErrorLevel = Debug | Warning | Error | Critical
+  deriving (Show, Eq, Ord, Generic, FromJSON, ToJSON)
+
+-- | Source of the robot log.
+data LogSource
+  = -- | Log produced by 'Swarm.Language.Syntax.Say'
+    Said
+  | -- | Log produced by 'Swarm.Language.Syntax.Log'
+    Logged
+  | -- | Log produced by an exception or system.
+    ErrorTrace ErrorLevel
+  deriving (Show, Eq, Ord, Generic, FromJSON, ToJSON)
+
+-- | An entry in a robot's log.
+data LogEntry = LogEntry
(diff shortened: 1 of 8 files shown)


src/Swarm/TUI/Controller.hs
```diff
@@ -109,8 +109,8 @@ handleEvent = \case
   AppEvent (UpstreamVersion ev) -> do
     let logReleaseEvent l e = runtimeState . eventLog %= logEvent l ("Release", -7) (T.pack $ show e)
     case ev of
-      Left e@(FailedReleaseQuery _e) -> logReleaseEvent ErrorTrace e
-      Left e -> logReleaseEvent Said e
```
xsebek, line 113: @kostmo @byorgey FYI: this used to create nice yellow warning logs by abusing the fact that _messages said by the robot with ID `-7` happen to be in yellow._ :sweat_smile:

src/Swarm/Game/Log.hs
```diff
@@ -0,0 +1,74 @@
...
+    ErrorTrace ErrorLevel
+  deriving (Show, Eq, Ord, Generic, FromJSON, ToJSON)
+
+-- | An entry in a robot's log.
+data LogEntry = LogEntry
+  { _leTime :: Integer
+  -- ^ The time at which the entry was created.
+  --   Note that this is the first field we sort on.
+  , _leSaid :: LogSource
+  -- ^ Whether this log records a said message.
```
byorgey, line 62: Maybe we should change the name and comment of this field (`_leSource`, perhaps?). The current name and comment make it sound like it is a `Bool`.

src/Swarm/Game/Log.hs
```diff
@@ -0,0 +1,74 @@
...
+  , _leRobotName :: Text
+  -- ^ The name of the robot that generated the entry.
+  , _leRobotID :: Int
+  -- ^ The ID of the robot that generated the entry.
+  , _leLocation :: Location
+  -- ^ Location of the robot at log entry creation.
+  , _leText :: Text
```
byorgey, line 69: At https://github.com/swarm-game/swarm/pull/959#issuecomment-1372626765 we discussed maybe storing something more structured here than just a `Text`, which I think is worth exploring.  But we should probably leave that to a different PR.

src/Swarm/TUI/Model/Failure.hs
```diff
@@ -0,0 +1,51 @@
+{-# LANGUAGE OverloadedStrings #-}
+
+-- |
+-- Module      :  Swarm.Game.Log
```
byorgey, line 4: Fix this module name

src/Swarm/TUI/Model/Failure.hs
```diff
@@ -0,0 +1,51 @@
...
+-- Copyright   :  Ondřej Šebek
+-- Maintainer  :  ondras98@icloud.com
+--
+-- SPDX-License-Identifier: BSD-3-Clause
+--
+-- A data type to represent sustem failures.
```
byorgey, line 10:
```suggestion
-- A data type to represent system failures.
```

byorgey (approved):
Looks good!

src/Swarm/Game/Log.hs
xsebek: I would like to split a separate type for system logs which should also use structured errors.

I think that will make the design clear, now it's all tangled and adding anything to it would be a mistake IMO.

src/Swarm/TUI/Model/Failure.hs
xsebek:
```suggestion
-- Module      :  Swarm.Game.Failure
```

byorgey: But isn't it `TUI.Model.Failure`?

src/Swarm/Game/Log.hs
xsebek: I am sure it was a `Bool` at some point. :slightly_smiling_face: Now it's fixed.

src/Swarm/Game/Log.hs
```diff
@@ -0,0 +1,74 @@
...
+    Said
+  | -- | Log produced by 'Swarm.Language.Syntax.Log'
+    Logged
+  | -- | Log produced by an exception or system.
+    ErrorTrace ErrorLevel
+  deriving (Show, Eq, Ord, Generic, FromJSON, ToJSON)
+
+-- | An entry in a robot's log.
+data LogEntry = LogEntry
+  { _leTime :: Integer
```
kostmo, line 58: These fields sound [French](https://youtu.be/kCpjgl2baLs?t=45) :stuck_out_tongue_winking_eye:

PR #983, opened 2023-01-08
warn of unused variables

### LINKED ISSUE #863: Warn of unused variables in LSP
I had written code as follows:
```
def hasItemInDirection = \item. \direction.
  x <- scan direction;
  case x (\_. return false;) (\y.
    return $ y == "tree";
  );
  end;
```

I took a while to discover why my program was not working---the `item` parameter was unused.

* Display unused variable warnings
* Differentiation between *warnings* and *errors* in the LSP

kostmo:
Closes #863.

Implements unused variable checking for `Let`, `Lambda`, and `Bind`.

Also:
* Now distinguishes between parsing errors and warnings via red vs. yellow squigglies in VS Code.
* Added unit tests

Screenshot from 2023-01-09 13-08-28

### DIFF
--- a/src/Swarm/Language/LSP/VarUsage.hs
+++ b/src/Swarm/Language/LSP/VarUsage.hs
@@ -1,0 +1,108 @@
+{-# LANGUAGE OverloadedStrings #-}
+
+module Swarm.Language.LSP.VarUsage where
+
+import Control.Monad (guard)
+import Data.List.NonEmpty (NonEmpty (..))
+import Data.Map (Map)
+import Data.Map qualified as M
+import Data.Set (Set)
+import Data.Set qualified as S
+import Data.Text (Text)
+import Data.Text qualified as T
+import Language.LSP.Types qualified as J
+import Swarm.Language.Parse qualified as P
+import Swarm.Language.Syntax
+import Swarm.Util qualified as U
+
+data BindingType
+  = Lambda
+  | Let
+  | Bind
+  deriving (Eq, Show)
+
+data VarUsage = VarUsage LocVar BindingType
+
+type BindingSites = Map Var (NonEmpty SrcLoc)
+
+data Usage = Usage
+  { usages :: Set LocVar
+  -- ^ Variable references
+  , problems :: [VarUsage]
+  -- ^ Variable declarations without any references
+  }
+
+instance Semigroup Usage where
+  Usage y1 z1 <> Usage y2 z2 =
+    Usage
+      (y1 <> y2)
+      (z1 <> z2)
+
+instance Monoid Usage where
+  mempty = Usage mempty mempty
+
+toErrPos :: Text -> VarUsage -> Maybe (J.Range, Text)
+toErrPos code (VarUsage (LV loc v) scope) = do
+  -- A leading underscore will suppress the unused variable warning
+  guard $ not $ "_" `T.isPrefixOf` v
+  rangePair <- case loc of
+    SrcLoc s e -> Just (s, e)
+    _ -> Nothing
+  let (start, end) = P.getLocRange code rangePair
+      ((startLine, startCol), (endLine, endCol)) = (minusOne start, minusOne end)
+      range =
+        J.Range
+          (J.Position (fromIntegral startLine) (fromIntegral startCol))
+          (J.Position (fromIntegral endLine) (fromIntegral endCol))
+  return (range, txt)
+ where
+  txt =
+    T.unwords
+      [ "Unused variable"
+      , U.quote v
+      , "in"
+      , T.pack $ show scope
+      , "expression"
+      ]
+  minusOne (x, y) = (x - 1, y - 1)
+
+-- | Descends the syntax tree rooted at a variable declaration,
+-- accumulating variable references.
+-- Generates a "problem" if an associated variable reference
+-- is not encountered in the subtree for this declaration.
+checkOccurrences ::
+  BindingSites ->
+  LocVar ->
+  BindingType ->
+  [Syntax] ->
+  Usage
+checkOccurrences bindings lv@(LV loc v) declType childSyntaxes =
+  Usage childUsages $ missing <> deeperMissing
+ where
+  deeperBindings = M.insertWith (<>) v (pure loc) bindings
+  Usage childUsages deeperMissing = mconcat $ map (getUsage deeperBindings) childSyntaxes
+  missing = [VarUsage lv declType | lv `S.notMember` childUsages]
+
+-- | Build up the bindings map as a function argument as
+-- we descend into the syntax tree.
+-- Aggregates unused bindings as we return from each layer.
+getUsage ::
+  BindingSites ->
+  Syntax ->
+  Usage
+getUsage bindings (Syntax _pos t) = case t of
+  TVar v -> Usage myUsages mempty
+   where
+    myUsages = case M.lookup v bindings of
+      Nothing -> mempty
+      Just (loc :| _) -> S.singleton $ LV loc v
+  SLam v _ s -> checkOccurrences bindings v Lambda [s]
+  SApp s1 s2 -> getUsage bindings s1 <> getUsage bindings s2
+  SLet _ v _ s1 s2 -> getUsage bindings s1 <> checkOccurrences bindings v Let [s2]
+  SPair s1 s2 -> getUsage bindings s1 <> getUsage bindings s2
+  SDef _ _v _ s -> getUsage bindings s
+  SBind maybeVar s1 s2 -> case maybeVar of
+    Just v -> checkOccurrences bindings v Bind [s1, s2]
+    Nothing -> getUsage bindings s1 <> getUsage bindings s2
+  SDelay _ s -> getUsage bindings s
+  _ -> mempty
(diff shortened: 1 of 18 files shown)


src/Swarm/Language/LSP/VarUsage.hs
```diff
@@ -0,0 +1,156 @@
...
+
+type UpScopes = Map Var [(UpwardDecl, Syntax)]
+
+-- | This positioning is customized per language construct.
+-- E.g. the "Syntax" node for @let@ includes the literal string \"let\"
+-- as well as an equals sign.
+--
+-- TODO: Perhaps if the language parser represented
+-- the "Var" as a \"located\" "Syntax" node as well, we would not need to
+-- reverse-engineer bits of the grammar here.
```
byorgey, line 40: This should definitely be possible.  I'm surprised the parser doesn't already do that.

byorgey (commented):
I think this whole thing can be simplified quite a bit.
1. If the parser does not already give us a `SrcLoc` for each variable, it absolutely should.  We should fix that rather than having a convoluted and error-prone `toErrPos` function.
2. All the stuff about downward vs upward scopes seems much too complicated, and possibly incorrect.  `def` and `bind` do not really scope "up" in any way that I can make sense of, they just scope over binds nested further to the right.  Note also that it is not allowed to have a `def` anywhere except at the topmost level, for example it is not allowed to have `(def a = ... end; def b = ... end); ...` so we don't have to worry about getting defined names from a "nested" bind like in that example.

Note that we already have a traversal `fvT` defined in `Swarm.Language.Syntax` which will fetch all the free variables in a term.  However, we need something slightly more general here since we need all the free variable *occurrences*, *with* their `SrcLoc`.  Actually in the `annotate-ast` branch I had to generalize `fvT` to traverse over the `Syntax` nodes instead of `Term`, which would be perfect here.  Let me see if I can extract something useful out of that as a separate PR.

If we had a way to traverse over free variables with their associated `SrcLoc`, then in theory this should be as easy as just listing them and turning each one into an unused var warning.

**EDIT**: sorry, I was confused.  It's not free variables we want (i.e. variable occurrences with no binder), but the opposite (variable bindings with no occurrences).

kostmo:
> `def` and `bind` do not really scope "up" in any way that I can make sense of, they just scope over binds nested further to the right.

In my first pass at implementing this, I got the sense that Def and Bind were somehow inherently different than Let and Lambda in the way they would need to be checked, though the "upward" scope idea is half-baked.  In reading your comment, maybe "rightward" scope is what I was thinking.

src/Swarm/Language/LSP/VarUsage.hs
kostmo: Yes, let's replace the `Var` in these constructors with a `LocVar` type.  Will you do so in another PR?

It's almost tempting to parameterize the `Syntax` type over what is now the `term` field, so we can stick `Var` there instead.

byorgey:
> I got the sense that Def and Bind were somehow inherently different than Let and Lambda in the way they would need to be checked

Of course the scoping rules are a bit different, but fundamentally they should not need to be handled in a different way.

I am imagining something like this (pseudocode):
```
data VarBinding = VB Var SrcLoc

-- | Takes the set of currently in-scope variables and a term, and returns the set of all variable occurrences in the term.
varOccs :: Map Var VarBinding -> Syntax -> Set VarBinding
varOccs inScope (Syntax l t) = case t of
  TUnit -> S.empty
  ...
  TVar x -> inScope ! x  -- in theory this lookup can't fail but in practice we want to handle it more safely than (!)
  SLam x xloc _ s -> varOccs (M.insert x (VB x xLoc) inScope) s
  SApp s1 s2 -> varOccs inScope s1 `S.union` varOccs inScope s2
  ...
  SBind _ (Syntax _ (SDef ...)) s2 -> ...  need a special case for this so we can include the defined name into inScope when recursing on s2
  ... etc.
```

src/Swarm/Language/LSP/VarUsage.hs
byorgey: Sure, I'll see about making another PR with `LocVar` filled in by the parser.

byorgey:
@kostmo let me know whether/what kind of additional help you'd like with this.

byorgey:
Chatting with @xsebek in IRC helped me realize I was wrong about nested `def` not being allowed.  It *is* allowed, and in fact *must be* allowed if we want importing to work. **HOWEVER**, I would argue that we *should not* warn about unused `def`s, precisely because `def` defines *global* names which can scope past the end of the expression/file containing them.  It is going to be very common to define a `.sw` file full of `def`s, where some of the defined names are used in others, but many of them are simply there so that you can import the `.sw` file and use them at the REPL.  Hence a normal `.sw` file is going to be full of "unused" names defined via `def` and it would be really weird and annoying to warn about them.  Just like how Haskell does not warn you about "unused" top-level definitions (*unless* you make an explicit export list; then it will warn you about names which are not used in the module and also not exported).

`let`, bind, and lambdas, on the other hand, all define names in a local scope that is easy to determine syntactically.  It should be possible to handle them all in essentially the same way, using an approach like the pseudocode I wrote in a comment above.

For more context about the difference between `def` vs everything else, see also my comment at https://github.com/swarm-game/swarm/issues/997#issuecomment-1377479041 .

kostmo:
@byorgey this is now working for `let`, `lambda`, and `bind` and unit tests have been added.

src/Swarm/Language/LSP/VarUsage.hs
```diff
@@ -0,0 +1,112 @@
...
+  Usage
+getUsage bindingSites s0@(Syntax _pos t) = case t of
+  TVar v -> Usage myUsages mempty
+   where
+    myUsages = case M.lookup v bindingSites of
+      Nothing -> mempty
+      Just _ -> S.singleton v
+  SLam v _ s -> checkOccurrences bindingSites s0 v Lambda [s]
+  SApp s1 s2 -> getUsage bindingSites s1 <> getUsage bindingSites s2
+  SLet _ v _ s1 s2 -> checkOccurrences bindingSites s0 v Let [s1, s2]
```
byorgey, line 105: This is a bit tricky.  We do want to recurse into `s1` in order to look for occurrences of other variables; however, if `v` occurs in `s1` it should *not* count as a use of `v`.  That's because it would correspond to a recursive variable definition, which is fine, but using a variable to define itself recursively should not count as a use.  For example, we should get a warning for `fac` here, but not a warning for `one`:
```
let one = 1 in let fac = \n. if (n == 0) {one} {n * fac (n-1)} in 17
```
I guess what we want is just
```
getUsage bindingSites s1 <> checkOccurrences bindingSites s0 v Let [s2]
```

src/Swarm/Language/LSP/VarUsage.hs
```diff
@@ -0,0 +1,112 @@
...
+  | Let
+  | Bind
+  deriving (Eq, Show)
+
+data VarUsage = VarUsage LocVar BindingType
+
+type BindingSites = Map Var (BindingType, Syntax)
+
+data Usage = Usage
+  { usages :: Set Var
```
byorgey, line 28: It's not quite correct to just have a `Set Var` for `usages`, since variables are not uniquely determined by their name alone, but rather by their name + location of their binding site.  I think this should be a `Set LocVar` instead.

Concretely, if you try something like this:
```
def y = \x. \x. x + 1 end
```
no warning is generated, since the usage of `x` makes the `checkOccurrences` function think everything is OK for both binders. But in fact, there *should* be a warning, since the first `x` is not used.

src/Swarm/Language/LSP/VarUsage.hs
```diff
@@ -0,0 +1,112 @@
...
+    myUsages = case M.lookup v bindingSites of
+      Nothing -> mempty
+      Just _ -> S.singleton v
+  SLam v _ s -> checkOccurrences bindingSites s0 v Lambda [s]
+  SApp s1 s2 -> getUsage bindingSites s1 <> getUsage bindingSites s2
+  SLet _ v _ s1 s2 -> checkOccurrences bindingSites s0 v Let [s1, s2]
+  SPair s1 s2 -> getUsage bindingSites s1 <> getUsage bindingSites s2
+  SDef _ _v _ s -> getUsage bindingSites s
+  SBind maybeVar s1 s2 -> case maybeVar of
+    Just v -> checkOccurrences bindingSites s0 v Bind [s1, s2]
```
byorgey, line 109: Note this is fine as is: it does not have the same issue as `SLet`, because binds cannot be recursive.

byorgey (commented):
This is looking really nice!  Just a few remaining issues + some new edge cases for the test suite.

src/Swarm/Language/LSP/VarUsage.hs
kostmo: Got it.  Now handled.  Add unit tests, including for a shadowed variable that has usage between the first and second binding site.

src/Swarm/Language/LSP/VarUsage.hs
kostmo: Makes sense.  Added a unit test for this.  The provided code snippet works.

kostmo:
All feedback addressed.  Added 4 more tests.

byorgey (approved):
Looks great!

kostmo:
@Mergifyio refresh

PR #986, opened 2023-01-08
Improve OnHover for Def/Let, App and Const

### LINKED ISSUE #985: Unfold function application in LSP Hover
I would like to see all the function parameters in hover in this format:

```markdown
<!--- f a b c -->
Function application of:
* `f: Bool -> Bool -> Bool -> Bool`
with parameters:
* `a: Bool`
* `b: Bool`
* `c: Bool`
```

The required function is:
```Haskell
unfoldApps :: Term -> NonEmpty Term
unfoldApps trm = NonEmpty.reverse . flip NonEmpty.unfoldr trm $ \case
  SApp (Syntax _ t1) (Syntax _ t2) -> (t2, Just t1)
  t -> (t, Nothing)
```

Not sure if it already exists somewhere in the codebase.

---

Honestly, I am already working on this, but I am making an Issue for transparency. :wink:

xsebek:
* unfold function application and tuples
* add type signatures to syntax using #990 
* show hovered variable definition name using #993
* fix indenting for multiline text

- closes #985

### DIFF
--- a/src/Swarm/Language/LSP/Hover.hs
+++ b/src/Swarm/Language/LSP/Hover.hs
@@ -1,18 +1,38 @@
 {-# LANGUAGE OverloadedStrings #-}
 
-module Swarm.Language.LSP.Hover where
+module Swarm.Language.LSP.Hover (
+  showHoverInfo,
+
+  -- * Documentation rendering
+  DocLine (..),
+  renderDoc,
+  treeToMarkdown,
+
+  -- * Finding source location
+  narrowToPosition,
+
+  -- * Explaining source position
+  Length (..),
+  explain,
+) where
 
 import Control.Applicative ((<|>))
 import Control.Monad (guard)
 import Data.Graph
+import Data.List.NonEmpty (NonEmpty (..))
 import Data.Maybe (fromMaybe)
 import Data.Text (Text)
 import Data.Text qualified as T
 import Data.Text.Utf16.Rope qualified as R
 import Language.LSP.Types qualified as J
 import Language.LSP.VFS
+import Swarm.Language.Context as Ctx
 import Swarm.Language.Parse (readTerm')
+import Swarm.Language.Pipeline (ProcessedTerm (..), processParsedTerm)
+import Swarm.Language.Pretty (prettyText)
 import Swarm.Language.Syntax
+import Swarm.Language.Typecheck (inferConst)
+import Swarm.Language.Types (Polytype, TModule, moduleCtx)
 import Swarm.Util qualified as U
 
 withinBound :: Int -> SrcLoc -> Bool
@@ -37,9 +57,12 @@ showHoverInfo _ _ p vf@(VirtualFile _ _ myRope) =
   case readTerm' content of
     Left _ -> Nothing
     Right Nothing -> Nothing
-    Right (Just stx) -> Just (treeToMarkdown 0 $ explain term, finalPos)
+    Right (Just stx@(Syntax l t)) -> Just (treeToMarkdown 0 $ explain Long mMod term, finalPos)
      where
-      Syntax sloc term = narrowToPosition stx $ fromIntegral absolutePos
+      (betterTerm, mMod) = case processParsedTerm stx of
+        Left _e -> (t, Nothing)
+        Right (ProcessedTerm betterT modul _req _reqCtx) -> (betterT, Just modul)
+      Syntax sloc term = narrowToPosition (Syntax l betterTerm) $ fromIntegral absolutePos
       finalPos = do
         (s, e) <- case sloc of
           SrcLoc s e -> Just (s, e)
@@ -103,10 +126,9 @@ data DocLine a = DocLine
 renderDoc :: DocLine Text -> Text
 renderDoc (DocLine d t)
   | d == 0 = t
-  | d == 1 = "* " <> t
-  | otherwise = indent (d - 1) <> "* " <> t
+  | otherwise = T.drop 2 . indent (max 0 (4 * (d - 1)) + 2) $ "* " <> t
  where
-  indent x = T.replicate (4 * x) " "
+  indent x = T.unlines . map (T.replicate x " " <>) . T.lines
 
 instance Functor DocLine where
   fmap f (DocLine d x) = DocLine d $ f x
@@ -118,10 +140,21 @@ treeToMarkdown :: Int -> Tree (DocLine Text) -> Text
 treeToMarkdown d (Node (DocLine _n t) children) =
   T.unlines $ renderDoc (DocLine d t) : map (treeToMarkdown $ d + 1) children
 
-explain :: Term -> Tree (DocLine Text)
-explain = \case
+-- | Top level explanations can use fancy markdown syntax, but list items
+--   have to be more conservative and ideally shorter.
+data Length = Brief | Long
+  deriving (Eq, Show)
+
+explain :: Length -> Maybe TModule -> Term -> Tree (DocLine Text)
+explain len mMod trm = case trm of
   TUnit -> pure $ pureDoc "The unit value."
-  TConst c -> pure $ pureDoc $ briefDoc $ constDoc $ constInfo c
+  TConst c ->
+    pure . pureDoc $
+      typeSignature
+        len
+        (prettyText c)
+        (Just $ inferConst c)
+        (briefDoc $ constDoc $ constInfo c)
   TDir {} -> pure $ pureDoc "A direction literal."
   TInt {} -> pure $ pureDoc "An integer literal."
   TAntiInt {} -> pure $ pureDoc "An antiquoted Haskell variable name of type Integer."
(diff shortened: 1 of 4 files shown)


src/Swarm/Language/LSP/Hover.hs
```diff
@@ -130,23 +163,67 @@ explain = \case
...
+-- * `*: int -> int -> int`
+--   Multiply the given integer values.
+-- * with parameters:
+--     * An integer literal.
+--     * An integer literal.
+explainFunction :: Length -> Maybe TModule -> Term -> Tree (DocLine Text)
+explainFunction len mMod t =
+  case unfoldApps t of
+    (TConst Force :| [innerT]) -> explain len mMod innerT
+    (TConst Force :| f : params) -> explainF f params
```
xsebek, line 194: This is a fascinating problem - the internally inserted `Force` is shown instead of the inner terms.

If we kept the locations, this workaround might be unnecessary.

image

src/Swarm/Language/LSP/Hover.hs
```diff
@@ -176,17 +176,6 @@ explain len mMod trm = case trm of
 --
 -- Note that 'Force' is often inserted internally, so
 -- if it shows up here we drop it.
---
--- >>> tprint = error . ("This is fine:\n" <>) . T.unpack
```
xsebek: On second thought, I could have dropped this hack instead and figured out what the problem is.

src/Swarm/Language/LSP/Hover.hs
```diff
@@ -36,9 +56,12 @@ showHoverInfo _ _ p vf@(VirtualFile _ _ myRope) =
   case readTerm' content of
     Left _ -> Nothing
     Right Nothing -> Nothing
-    Right (Just stx) -> Just (treeToMarkdown 0 $ explain term, finalPos)
+    Right (Just stx@(Syntax l t)) -> Just (treeToMarkdown 0 $ explain Long mMod term, finalPos)
      where
-      Syntax sloc term = narrowToPosition stx $ fromIntegral absolutePos
+      (betterTerm, mMod) = case processParsedTerm stx of
+        Left _e -> (t, Nothing)
+        Right (ProcessedTerm betterT modul _req _reqCtx) -> (betterT, Just modul)
```
byorgey, line 63: See my comment below re: `force`.  I think we *don't* want to use `betterT` here, but always explain things in terms of the user's original parsed syntax `stx`.  Of course we still do want to use the output of typechecking (here, `modul`).

src/Swarm/Language/LSP/Hover.hs
```diff
@@ -101,10 +124,9 @@ data DocLine a = DocLine
 renderDoc :: DocLine Text -> Text
 renderDoc (DocLine d t)
   | d == 0 = t
-  | d == 1 = "* " <> t
-  | otherwise = indent (d - 1) <> "* " <> t
+  | otherwise = T.drop 2 . indent (max 0 (4 * (d - 1)) + 2) $ "* " <> t
```
byorgey, line 127: Is the `max 0` necessary?  If `d > 0` then `4 * (d - 1) >= 0`.

byorgey: Also, I don't understand why we're adding 2 to the indentation and then doing `T.drop 2`.  Don't those cancel each other out?

src/Swarm/Language/LSP/Hover.hs
```diff
@@ -116,10 +138,21 @@ treeToMarkdown :: Int -> Tree (DocLine Text) -> Text
...
   TUnit -> pure $ pureDoc "The unit value."
-  TConst c -> pure $ pureDoc $ briefDoc $ constDoc $ constInfo c
+  TConst c ->
+    pure . pureDoc $
+      typeSignature
+        len
+        (prettyText c)
+        (Just $ inferConst c)
+        (briefDoc $ constDoc $ constInfo c)
   TDir {} -> pure $ pureDoc "A direction literal."
```
byorgey, line 156: While we're at it, could we display something like `left : dir` rather than `A direction literal`?

src/Swarm/Language/LSP/Hover.hs
```diff
@@ -130,23 +163,56 @@ explain = \case
...
-        T.unwords
-          [ "A"
-          , (if isRecursive then "" else "non-") <> "recursive"
-          , "let expression"
-          , if null maybeTypeAnnotation then "without" else "with"
-          , "a type annotation on the variable."
-          ]
+  TVar v -> pure $ pureDoc $ "`" <> v <> ": " <> maybe "?" prettyText (Ctx.lookup v . moduleCtx =<< mMod) <> "`"
+  SLam v _mType _syn -> pure . pureDoc $ "A lambda expression binding the variable " <> U.bquote v <> "."
+  SApp _ _ -> explainFunction len mMod trm
```
byorgey, line 168:
```suggestion
  SApp {} -> explainFunction len mMod trm
```

src/Swarm/Language/LSP/Hover.hs
```diff
@@ -130,23 +163,56 @@ explain = \case
...
+explainDefinition len isDef mMod isRecursive var maybeTypeAnnotation =
+  pureDoc $
+    typeSignature len var (maybeTypeAnnotation <|> mTyp) $
+      T.unwords
+        [ "A"
+        , (if isRecursive then "" else "non-") <> "recursive"
+        , if isDef then "definition" else "let"
+        , "expression"
+        , if null maybeTypeAnnotation then "without" else "with"
+        , "a type annotation on the variable."
```
byorgey, line 205:
```suggestion
        , "a type annotation."
```

Saying "on the variable" seems kind of superfluous to me.

src/Swarm/Language/LSP/Hover.hs
xsebek: I suspect force uses pattern matching that gets rid of the source position.

My hope was that the inferred types would be stored in the AST (def/let/…) but that is not the case.

@byorgey is there any useful (in this context) change in the syntax, or should I just use the original?

byorgey (approved):

src/Swarm/Language/LSP/Hover.hs
xsebek: oh, now I scrolled to this comment. 😄 Will do.

src/Swarm/Language/LSP/Hover.hs
xsebek: Huh, I suppose the maximum is redundant.

The indent and drop cancel only for first line of text.
The later lines will be indented two spaces extra.

src/Swarm/Language/LSP/Hover.hs
xsebek: I would show both. 🙂

src/Swarm/Language/LSP/Hover.hs
byorgey: Really? Isn't `T.drop 2 . indent (n + 2)` equivalent to `indent n`?

xsebek: No:
```Haskell
>>> T.putStrLn $ indent 4 "* abc\ndef"
    * abc
    def
>>> T.putStrLn $ T.drop 2 $ indent (4 + 2) "* abc\ndef"
    * abc
      def
```

xsebek:
I have rebased on the AST with types (#991) in the branch [](https://github.com/swarm-game/swarm/tree/hover-types).

@byorgey and @kostmo feel free to test it. :slightly_smiling_face: 

The results look quite nice:

## Local variables
|let|def|
|--|--|
|image|image|

|bind|lambda|
|--|--|
|image|image|

## Tuples

|Tuple|Individual element (needed fix)|
|--|--|
|Screenshot from 2023-01-13 20-56-56|Screenshot from 2023-01-13 20-57-03|



## Literal types

All literals and constants get a type header.

|Different generic and specific|Same types|
|--|--|
|image|image|

|Text|Direction|
|--|--|
|image|image|

## Delay

|image|
|--|

src/Swarm/Language/LSP/Hover.hs
```diff
@@ -130,23 +163,56 @@ explain = \case
...
-    pure $
-      pureDoc $
-        T.unwords
-          [ "A"
-          , (if isRecursive then "" else "non-") <> "recursive"
-          , "let expression"
-          , if null maybeTypeAnnotation then "without" else "with"
-          , "a type annotation on the variable."
-          ]
+  TVar v -> pure $ pureDoc $ "`" <> v <> ": " <> maybe "?" prettyText (Ctx.lookup v . moduleCtx =<< mMod) <> "`"
```
kostmo, line 166: Use `bquote`:
https://github.com/swarm-game/swarm/blob/b8bc2869ba15fe7842aa520ffab05616d1fdbdd5/src/Swarm/Util.hs#L412-L414

src/Swarm/Language/LSP/Hover.hs
```diff
@@ -130,23 +163,56 @@ explain = \case
...
+        , "a type annotation on the variable."
+        ]
+ where
+  mTyp = Ctx.lookup var . moduleCtx =<< mMod
+
+typeSignature :: Length -> Var -> Maybe Polytype -> Text -> Text
+typeSignature len v typ body =
+  T.unlines $
+    case len of
+      Brief -> ["`" <> short <> "`", body]
```
kostmo, line 214: `bquote`

kostmo (approved):

xsebek:
@kostmo please take a look at this again, I have changed quite a lot after rebasing on the AST types. :wink:

kostmo (approved):
Very nice! :ship:

byorgey (approved):
Nice!!

The formatting for function application explanations looks weird in emacs:

func-app-hover

Not sure what's going on there.  But I don't think it should hold up merging this.

xsebek:
@byorgey it seems emacs has a different take on markdown. :confused:  It should be easy to tweak by adding newlines, but I don't have emacs set up. :sweat_smile:

byorgey:
@xsebek no worries, maybe I can poke at it a bit.

xsebek:
It seems emacs can render triple backtick code in lists, but if it turns out to be an issue, I can restore my original code.

At some point, I had an extra parameter for "depth" that used single backticks in lists. It did not have nice syntax highlighting but it might render better for emacs.

It was making the code a bit messy, so it's best avoided if possible. :sweat_smile:

Issue #681, opened 2022-09-12
Variables in a local monadic binder escape to outer scopes


byorgey:
_Figured out what's causing the bug.  Basically the `b <- ...` in the definition of `until` is overriding the definition of `b` in `hanoi`, which absolutely should not happen._

_Originally posted by @byorgey in https://github.com/swarm-game/swarm/issues/669#issuecomment-1243051295_

To reproduce, put the following in a `.sw` file and `--run` it in Creative mode:
```
def f = a <- scan down end;
let a = 2 in f; return (a+1)
```
The binding of the variable `a` from the definition of `f` is escaping into the outer scope from which `f` is called, so that `a` is now bound to `inl ()` instead of `2`, which obviously causes a runtime crash when the `+` operator is applied to a non-number.

*Edited to add*: we should keep the below discussion around for reference, but see #1087 for my latest thinking on the issue.

byorgey:
I was trying to do a `git bisect` but it seems like this bug has been around for a long, long time.  For example it seems to be present already in https://github.com/swarm-game/swarm/commit/008d4000a542734055be13456b6252d915af6817 , which is the commit that introduced `--run` (it gets more annoying to test before that).

xsebek:
Yikes, good job figuring this one out @byorgey. 👍 

Maybe we could make the example a unit test, to make it easier to debug/fix? 🙂 

I don’t know how the binder machinery works, so this is just a guess on my part, but could we fix this by renaming variables internally?

For example the inner variable could be named `f::a[0]` and the outer one would be `_::a[0]`. 🤔

byorgey:
> Maybe we could make the example a unit test, to make it easier to debug/fix? slightly_smiling_face

Yes, good idea.
 
> I don’t know how the binder machinery works, so this is just a guess on my part, but could we fix this by renaming variables internally?

I mean, that's a reasonable idea, but I think it would just be a band-aid, and would probably lead to other bugs.  In theory, we should be able to fix it simply by managing environments in such a way that the dynamic semantics (i.e. which values flow where) matches the static semantics (the scopes etc. enforced by the type checker).

xsebek:
The worst thing about this is that it has horrible error messages:
![image](https://user-images.githubusercontent.com/44544735/190256292-65e64907-e7da-486a-8e8c-22b1ae628007.png)
```
def repeat = \c. c; repeat c end;
repeat ( c <- scan down; case c (\_. say "Hi") (\_. return ()) )
```

[...]

PR #991, opened 2023-01-09
Annotate ASTs with types at every node

### LINKED ISSUE #990: Annotate AST nodes with inferred types
While performing type inference, we recurse through the AST, inferring a type for each node.  Currently we throw away all this intermediate information and return only the top-level inferred type for an entire AST.  However, there are many situations where having this intermediate type information would be helpful, for example, to help with things like LSP hover hints, or to apply type-specific elaboration.

Rather than simply returning a type, the type inference engine should return a new AST annotated with a type at every node.  We already have a `SrcLoc` at every node via the `Syntax` wrapper type, so this would just require adding a new type field to `Syntax`.

I have started on implementing this in the https://github.com/swarm-game/swarm/tree/annotate-ast branch.  Rewriting the elaboration code is proving tricky but if I bang my head against it a bit more I think I will get it.

byorgey:
- Closes #990

Values of type `Syntax` are as before: parsed syntax, with each node annotated with `SrcLoc`.

Values of type `Syntax' Polytype`, however, have each node annotated with *both* a `SrcLoc` *and* a `Polytype`.  (`Syntax` is really just a synonym for `Syntax' ()`.)

Type inference takes a `Syntax` and outputs a `TModule`, which now contains a `Syntax' Polytype`, in other words, a new version of the AST where every node has been annotated with the inferred type of the subterm rooted there.

---

Why is this useful?
1. It will enable us to do type-specific elaboration/rewriting.  For example I think this will allow us to solve #681 , because we only want to apply a rewrite to variables with a command type.
2. It makes type information for any specific subterm easily available.  For example I hope we will be able to use this to enhance the `OnHover` LSP handler, e.g. to show the type of the term under the mouse.

I imagine the code changes might look kind of intimidating but I don't think it's really that bad once you understand what is going on, so I'm happy to answer any questions or explain anything.


src/Swarm/TUI/Controller.hs
```diff
@@ -1106,10 +1106,8 @@ handleRobotPanelEvent = \case
 makeEntity :: Entity -> EventM Name AppState ()
 makeEntity e = do
   s <- get
-  let mkTy = PolyUnit
-      mkReq = R.singletonCap CMake
-      mkProg = TApp (TConst Make) (TText (e ^. entityName))
-      mkPT = ProcessedTerm mkProg (Module mkTy empty) mkReq empty
+  let name = e ^. entityName
+      mkPT = [tmQ| make $str:name |]
```
xsebek, line 1110: Nice use of TH! :+1:

src/Swarm/Language/Syntax.hs
```diff
@@ -774,9 +785,20 @@ mkOp c s1@(Syntax l1 _) s2@(Syntax l2 _) = Syntax newLoc newTerm
...
+  , _sType :: ty
+  }
+  deriving (Eq, Show, Functor, Foldable, Traversable, Data, Generic, FromJSON, ToJSON)
+
+type Syntax = Syntax' ()
+
+pattern Syntax :: SrcLoc -> Term -> Syntax
+pattern Syntax l t = Syntax' l t ()
+
+{-# COMPLETE Syntax #-}
```
kostmo, line 801: What does `COMPLETE` do?

byorgey: It specifies a set of constructors/pattern synonyms which should be considered "complete" for the purposes of pattern match coverage checking.  i.e. if you pattern-match on everything listed in the `COMPLETE` pragma then GHC will not warn about missing cases.  Using `COMPLETE` is common when creating pattern synonyms since there's no way in general to know whether an arbitrary set of pattern synonyms is going to cover all possible cases, so without the pragma in general GHC will generate lots of warnings about incomplete pattern matches.

https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/pragmas.html#complete-pragmas

src/Swarm/Language/Typecheck.hs
```diff
@@ -306,46 +322,54 @@ inferModule s@(Syntax _ t) = (`catchError` addLocToTypeErr s) $ case t of
...
-  TUnit -> return UTyUnit
-  TConst c -> instantiate . toU $ inferConst c
-  TDir _ -> return UTyDir
-  TInt _ -> return UTyInt
-  TAntiInt _ -> return UTyInt
-  TText _ -> return UTyText
-  TAntiText _ -> return UTyText
-  TBool _ -> return UTyBool
-  TRobot _ -> return UTyActor
+  TUnit -> return $ Syntax' l TUnit UTyUnit
```
xsebek, line 352: Maybe use `Syntax' l t` to make it clearer where `t` is changed? :slightly_smiling_face: 

Another alternative would be:
```Haskell
return $ s{sTerm = UTyUnit}
```

test/integration/Main.hs
```diff
@@ -248,7 +248,7 @@ testScenarioSolution _ci _em =
       Left x -> assertFailure $ unwords ["Failure in initGameStateForScenario:", T.unpack x]
       Right gs -> case gs ^. winSolution of
         Nothing -> assertFailure "No solution to test!"
-        Just sol@(ProcessedTerm _ _ _ reqCtx) -> do
+        Just sol@(ProcessedTerm _ _ reqCtx) -> do
```
xsebek, line 251: Not a single new test is needed? :open_mouth: At least a small unit test to showcase the usage would be nice.

xsebek (approved):
Looks good to me. :+1:  I will be happy to rebase #986 on this and test it out. :rocket:

src/Swarm/Game/CESK.hs
```diff
@@ -284,11 +286,15 @@ initMachine t e s = initMachine' t e s []
...
-    Empty -> In t e s (FExec : k)
-    _ -> In t e s (FExec : FLoadEnv ctx reqCtx : k)
-initMachine' (ProcessedTerm t _ _ _) e s k = In t e s k
+initMachine' (ProcessedTerm (Module t' ctx) _ reqCtx) e s k =
+  case t' ^. sType of
+    Forall _ (TyCmd _) ->
+      case ctx of
+        Empty -> In t e s (FExec : k)
+        _ -> In t e s (FExec : FLoadEnv ctx reqCtx : k)
+    _ -> In t e s k
```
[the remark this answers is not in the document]
byorgey: :heavy_check_mark:

src/Swarm/Language/Syntax.hs
```diff
@@ -758,97 +775,12 @@ constInfo c = case c of
...
-    TBind mv t1 t2 = SBind (LV NoLoc <$> mv) (STerm t1) (STerm t2)
-
--- | Match a TDelay without syntax
-pattern TDelay :: DelayType -> Term -> Term
-pattern TDelay m t = SDelay m (STerm t)
-
--- | COMPLETE pragma tells GHC using this set of pattern is complete for Term
-{-# COMPLETE TUnit, TConst, TDir, TInt, TAntiInt, TText, TAntiText, TBool, TRequireDevice, TRequire, TVar, TPair, TLam, TApp, TLet, TDef, TBind, TDelay #-}
+  lowShow :: Show a => a -> Text
+  lowShow a = toLower (from (show a))
```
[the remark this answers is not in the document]
byorgey: What do you mean?  Just the one, as far as I can tell.

src/Swarm/Language/Syntax.hs
```diff
@@ -898,87 +830,235 @@ data Term
...
     --   eager language, but in some cases (e.g. for @if@ statements
     --   and recursive bindings) we need to delay evaluation.  The
     --   counterpart to @{...}@ is @force@, where @force {t} = t@.
     --   Note that 'Force' is just a constant, whereas 'SDelay' has to
     --   be a special syntactic form so its argument can get special
     --   treatment during evaluation.
-    SDelay DelayType Syntax
-  deriving (Eq, Show, Data, Generic, FromJSON, ToJSON)
+    SDelay DelayType (Syntax' ty)
+  deriving (Eq, Show, Functor, Foldable, Traversable, Data, Generic, FromJSON, ToJSON)
```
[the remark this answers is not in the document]
byorgey: Added a comment explaining this in the source code.

src/Swarm/Language/Syntax.hs
```diff
@@ -898,87 +830,235 @@ data Term
...
+
+------------------------------------------------------------
+-- Free variable traversals
+------------------------------------------------------------
+
 -- | Traversal over those subterms of a term which represent free
 --   variables.
-fvT :: Traversal' Term Term
-fvT f = go S.empty
+fvSS :: forall ty. Traversal' (Syntax' ty) (Syntax' ty)
```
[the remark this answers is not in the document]
byorgey: :heavy_check_mark:

src/Swarm/Language/Syntax.hs
```diff
@@ -898,87 +830,235 @@ data Term
...
+
+fvST :: forall ty. Traversal' (Syntax' ty) (Term' ty)
+fvST = fvSS . sTerm
 
 -- | Traversal over the free variables of a term.  Note that if you
 --   want to get the set of all free variables, you can do so via
 --   @'Data.Set.Lens.setOf' 'fv'@.
-fv :: Traversal' Term Var
-fv = fvT . (\f -> \case TVar x -> TVar <$> f x; t -> pure t)
+fvSV :: Traversal' (Syntax' ty) Var
```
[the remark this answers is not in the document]
byorgey: :heavy_check_mark:

test/integration/Main.hs
byorgey: Oh, very good point!  I was so focused on the refactoring, and so elated when I finally got everything compiling and passing all the tests again, that I totally forgot about adding some tests. :sweat_smile:  I will add some.

src/Swarm/Language/Typecheck.hs
xsebek: @byorgey how about using that pattern arrow on left side to convert between types? If the conversion is easy, it will be inlined and we will get simple `t` on right side.

(just a thought while commuting, dont take it too seriously)

byorgey:
I've realized a big issue with the way this works: in particular it doesn't make sense to annotate every node of the AST with a `Polytype`; we should instead have a single `Polytype` for the entire thing at the top level, and annotate every node with a `Type`.  For example suppose we have `def wow = \x. format x ++ "!" end`.  The type of `wow` will correctly be inferred as `forall a. a -> text`, but if you hover over `x`, its type will be shown as `forall a. a` which does not make any sense.  It should just have type `a`, or perhaps some made-up type variable like `s0`.  I think I have an idea how to fix this...

xsebek:
@byorgey won't every definition and let binding have a `Polytype`?

I would not mind if the type was stored in the `SDef`/`SLet` instead of the original `Maybe Polytype`. :slightly_smiling_face:

byorgey:
> won't every definition and let binding have a Polytype?

Yes, they should.  Still trying to figure out a good way to annotate different things with `Type` or ` Polytype` as appropriate...

xsebek:
@byorgey could you merge this as is and improve on it in the main branch? :slightly_smiling_face: 

The types are morally wrong with the `forall`, but it is still better than nothing and we will not be using them to figure the binding sites of variables, so it is not practically a problem. (In fact, we could strip all `forall` and it would make sense.)

I would like to rebase and merge my OnHover rework and _the types added here are invaluable_.

Not that I am in a rush, I just don't want to have merge conflicts with @kostmo if he works on the LSP further. :sweat_smile:

byorgey:
You're right, I will do that. Can't let the perfect become the enemy of the good. : smile:

test/integration/Main.hs
byorgey: Added some basic tests.

byorgey:
As discussed above, we can merge this as-is now, and I've created #1042 for tracking the remaining issue.

test/unit/TestLanguagePipeline.hs
```diff
@@ -254,14 +257,25 @@ testLanguagePipeline =
...
-        --     (let s = getSyntax
-        --            [tmQ| def f : (int -> int) -> int -> cmd unit = \g. \x. |]
+        , testCase
+            "get all annotated variable types"
+            (let s = getSyntax
+                   [tmQ| def f : (int -> int) -> int -> text = \g. \x. format (g x) end |]
+
+                 isVar (TVar {}) = True
+                 isVar _ = False
+                 getVars = map (_sTerm &&& _sType) . filter (isVar . _sTerm) . universe
```
xsebek: > `universe`

:exploding_head:

xsebek (approved):

test/unit/TestLanguagePipeline.hs
byorgey: See, I put that in there on purpose to give people ideas. :smile:  https://hackage.haskell.org/package/lens-5.2/docs/Control-Lens-Plated.html has some pretty powerful tools for working with ASTs.

PR #995, opened 2023-01-10
Isolate work directory for 'stack test'

### LINKED ISSUE #936: Entire project is recompiled when switching from "stack build" to "stack test"
Are there any flags that can be passed to the `stack` invocation so that when alternating between `stack build` and `stack test`, it doesn't have to rebuild every module?

kostmo:
Closes #936.

With this change, `stack build` should be invoked as normal, but to run tests one should invoke the new script, which will place compilation artifacts in a separate directory so as to not stomp on the previous `stack build` artifacts.

## Demo

```
stack build
./scripts/run-tests.sh
stack build
```
All perform incremental compilation rather than from-scratch.


README.md
```diff
@@ -136,6 +136,8 @@ contribute to Swarm development, you can build from source.
 1. Go get a snack while `stack` downloads a Haskell compiler and
    all of Swarm's dependencies.
 
+NOTE: To run tests, please use the [`scripts/run-tests.sh` script](scripts/run-tests.sh) to avoid problems with incremental compilation in alternation with `stack build`.
```
xsebek, line 139: This sounds too generic.

Besides, random contributors do not need to worry about it.

It would make much more sense as a TIP for regular developers in DEVELOPING.md to speed up Stack until it solves its issue.

scripts/run-tests.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash -e
+
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+cd $SCRIPT_DIR/..
+
+# See https://github.com/swarm-game/swarm/issues/936
+STACK_WORK=.stack-work-test stack test
```
xsebek, line 7: Do you want to pass parameters here?

xsebek (commented):
Thanks for documenting the workaround @kostmo. 👍 

I leave judgment to @byorgey as I am happy with Cabal so far. 😄

README.md
kostmo: > in DEVELOPING.md

Sorry, which file is that?  I only see the one in the [VS Code plugin development directory](https://github.com/swarm-game/swarm/blob/main/editors/vscode/DEVELOPING.md).

scripts/run-tests.sh
kostmo: good idea, fixed

README.md
xsebek: My mistake, this was the section I am thinking about which already has some tips:
https://github.com/swarm-game/swarm/blob/main/CONTRIBUTING.md#development-workflow

kostmo: Got it, fixed.

CONTRIBUTING.md
```diff
@@ -203,6 +203,8 @@ check the `hie.yaml.stack` file. Pure Cabal should work fine with the defaults,
 but you might want to [add a GHC flag](https://haskell-language-server.readthedocs.io/en/latest/features.html)
 to build docs into `cabal.project.local`.
 
+**TIP:** To run tests, please use the [`scripts/run-tests.sh` script](scripts/run-tests.sh) to avoid problems with incremental compilation in alternation with `stack build`.
```
byorgey, line 206:
```suggestion
**TIP:** To run tests, you can use the [`scripts/run-tests.sh` script](scripts/run-tests.sh) to avoid problems with incremental compilation in alternation with `stack build`.
```

I suggest this slight rewording.  "Please use" makes it sound like something will go wrong if you don't, but of course everyone can use whatever development workflow they like.

byorgey (approved):

Issue #1000, opened 2023-01-10
Immediate freeze upon launching most scenarios


kostmo:
Starting scnenarios through the menu or through the command line with `--scenario` causes the program to freeze.  `pkill swarm` must be executed in the terminal to recover.  E.g. this one freezes:

    stack run -- --scenario data/scenarios/creative.yaml

It seems that not all scenarios freeze, however; these ones work:

    stack run -- --scenario data/scenarios/Tutorials/scan.yaml
    stack run -- --scenario data/scenarios/Challenges/2048.yaml

byorgey:
Issue 1000!  @kostmo is the winner! :tada:

byorgey:
Is this on the current `main` branch?  I can't reproduce this.

kostmo:
Observed at commit 68c992aa on the main branch.  Tried the following without success:

* `stack clean`
* `git clean -fdx`
* Reboot computer

It seems that the freeze is not absolute; if I wait long enough, the world UI will display, but will then still be unresponsive to keyboard input.

byorgey:
I still cannot reproduce.  Maybe some of your data/save/history files got corrupted or huge somehow, and swarm is taking a very long time trying to read/process it somehow?

kostmo:
In case it is relevant, here's what I encountered before the freeze issue:
```
kostmo@dell-ubuntu:~/github/swarm$ stack run
Warning: while processing Tutorials/00-ORDER.txt: files not listed in 00-ORDER.txt will be ignored
  - install.yaml
Failed to load scenarios: Aeson exception:
Error in $.solution: Could not parse term: 1: Unbound variable install
kostmo@dell-ubuntu:~/github/swarm$ find -name install.yaml
./.stack-work/install/x86_64-linux-tinfo6/3598b35d16175b2123168d7c37dbbc360ad9ba49ad215feda9c6c5d30c8c8528/9.2.5/share/x86_64-linux-ghc-9.2.5/swarm-0.2.0.0/scenarios/Tutorials/install.yaml
kostmo@dell-ubuntu:~/github/swarm$ git clean -fdx
Removing .hie/
Removing .stack-work/
Removing stack.yaml.lock
kostmo@dell-ubuntu:~/github/swarm$ find -name install.yaml
kostmo@dell-ubuntu:~/github/swarm$
```

I have since nuked the `.local/share/swarm` directory, but am still hitting the freeze issue.

kostmo:
[...]

PR #1001, opened 2023-01-11
Update to GHC 9.4 and `brick` 1.5

kostmo:
A couple upgrades:

- Upgrade to Brick 1.5, which has a new nicer dialog API, with names to identify buttons instead of positional indices: https://hackage.haskell.org/package/brick-1.5/changelog
- Update stack resolver to a nightly one, with GHC 9.4.  This fixes? mitigates? #1000, at least, no one has yet observed the issue under GHC 9.4 that we know of.


swarm.cabal
```diff
@@ -147,7 +147,7 @@ library
     build-depends:    base                          >= 4.14 && < 4.18,
                       aeson                         >= 2 && < 2.2,
                       array                         >= 0.5.4 && < 0.6,
-                      brick                         >= 1.0 && < 1.5,
+                      brick                         >= 1.0 && < 1.6,
```
byorgey, line 150: Should this be `brick >= 1.5 && < 1.6`?  I assume it will not compile with earlier versions of `brick`, right?

byorgey:
@kostmo do you have a sense of what would be required to finish this?  Can I help?  I'd love to get this merged since #1000 is very annoying.  It's easy to work around for more experienced developers but I'd really like to be able to continue telling people to `stack run` and have it just work.

swarm.cabal
kostmo: correct

kostmo:
> Can I help?

I'll happily hand this over to you if you want to finish it.  Mostly I think the work entails creating and assigning new members of the `Name` type to dialog buttons.

byorgey:
OK, I'll take a look.

src/Swarm/TUI/Model/Name.hs
```diff
@@ -16,8 +16,17 @@ data GoalWidget
   | GoalSummary
   deriving (Eq, Ord, Show, Read, Bounded, Enum)
 
+-- | Clickable buttons in modal dilaogs.
```
kostmo, line 19: "dialogs"

xsebek:
@byorgey should we merge it before release?

xsebek (approved):
LGTM! 👍

PR #1004, opened 2023-01-14
Ensure unique solution for word search

kostmo:
Make the game a little more challenging by first ensuring there are no solutions, then inserting exactly one solution at a random location.

Also improved the use of `random` in the solution generator; now it is called at most once per cell, rather than recursing until a desirable value is found.

### DIFF
--- a/data/scenarios/Challenges/_word-search/create-puzzle.sw
+++ b/data/scenarios/Challenges/_word-search/create-puzzle.sw
@@ -26,12 +26,11 @@ def whichOrdinal = \str.
     end;
 
 /**
-Calls "whichOrdinal" on the cell to the north.
 Returns -1 if not a recognized letter.
 */
 def getAdjacentOrdinal = \d.
-    foo <- scan d;
-    str <- case foo (\_. return "") (\s. return s);
+    maybeEntity <- scan d;
+    str <- case maybeEntity (\_. return "") (\s. return s);
     whichOrdinal str;
     end;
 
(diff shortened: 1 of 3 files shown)


byorgey (commented):
Seems to work great, except that my solution is not accepted!  That is, my solution seems to work correctly and marks the three letters, but I do not get a box pop up telling me that I've won.  Here's my solution:

```
  b <- test; if b then else
end

def orC : cmd bool -> cmd bool -> cmd bool = \t1. \t2.
  b1 <- t1;
  b2 <- t2;
  return (b1 || b2)
end

def until = \test. \body.
  ifC test {} {force body ; until test body}
end

def x3 = \c. c; c; c end

def w = \c.
  ifC (ishere "capital W") {
    turn back; x3 (drill down; move)
  } {c}
end

def o = \c.
  ifC (ishere "capital O") {try {move; w c} {c}} {c}
end

def c =
  ifC (orC isempty blocked) {
    ifC isempty {
      turn back; move;
    } {
      turn back
    };

    try {
      until (isempty) {move}; turn back; move
    } { turn back };
    turn right; move; turn left;
    ifC (isempty) {
      turn left; move; c
    }
    {c}
  } {
    ifC (ishere "capital C") {move; o c} {move; c}
  }
end

def sol =
  move; move; turn left; move; move; move; move; turn right; c
end
```

kostmo:
I've coded a workaround for what seems to be a bug: #1013.

byorgey (approved):
Works great!

PR #1008, opened 2023-01-15
Ice Cream Shop

kostmo:
A challenge of dealing with large-ish numbers.

Screenshot from 2023-01-14 18-01-53

### DIFF
--- a/data/scenarios/Challenges/ice-cream.yaml
+++ b/data/scenarios/Challenges/ice-cream.yaml
@@ -104,14 +104,14 @@ entities:
       attr: 'gold'
     properties: [portable]
     description:
-      - Bawk bawk
+      - Organic from local, free-range, meticulously pampered chickens.
   - name: milk
     display:
       char: 'm'
       attr: 'silver'
     properties: [portable]
     description:
-      - Moo
+      - Pasteurized, homogenized, commoditized, then metabolized.
   - name: ice
     display:
       char: 'i'
@@ -125,7 +125,7 @@ entities:
       attr: 'snow'
     properties: [portable]
     description:
-      - Basis for most desserts
+      - Basis of most desserts
   - name: vanilla
     display:
       char: 'v'
--- a/data/scenarios/Challenges/ice-cream.yaml
+++ b/data/scenarios/Challenges/ice-cream.yaml
@@ -11,18 +11,20 @@ objectives:
         Congratulations on the grand opening of your new ice cream shop.
         You have advertised: "All you can eat, for 1 bitcoin!"
       - |
-        Here comes your first customer. They look hungry!
+        A customer is approaching the "Counter". They look hungry!
+      - "..."
+      - |
+        Customer: "`give` me a cone, and then I'll tell you how many scoops I want."
+      - "..."
       - |
         Oh dear, you've forgotten to stock your shop with a "calculator".
         Let's hope this order is simple.
-      - |
-        Customer: "Give me a cone, and then I'll tell you how many scoops I want."
     condition: |
       customer <- robotnamed "customer";
       as customer {has "cone"};
   - goal:
       - |
-        Give the customer exactly what they've requested.
+        `give` the customer exactly what they've requested.
       - |
         Perhaps there is something else around the shop that can assist with
         preparing the order...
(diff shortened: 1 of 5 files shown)


data/scenarios/Challenges/ice-cream.yaml
```diff
@@ -0,0 +1,242 @@
...
+      - A traditional ice cream garnish
+  - name: Counter
+    display:
+      char: '▒'
+      attr: 'sand'
+    properties: [portable, known]
+    capabilities: [count]
+    description:
+      - Where customers are served.
+      - Also... it can count things.
```
byorgey, line 93: :laughing:  Unfortunately I never noticed the `Counter`.  I just manually wrote a function `x2288` which repeats its argument 2288 times (by defining `x64 = \c. x32 c; x32 c` and so on).

byorgey (approved):
This is fantastic. :smile:

data/scenarios/Challenges/ice-cream.yaml
byorgey: Suggestion: what about equipping the base with a `scanner`?  It was a bit annoying wandering around trying to `grab` everything to see what it was.

xsebek: @kostmo I would suggest that the customer wait outside while we prepare the order.

That way the counter will be visible. :slightly_smiling_face:

data/scenarios/Challenges/ice-cream.yaml
```diff
@@ -0,0 +1,242 @@
...
+  - name: briefcase
+    display:
+      char: 'B'
+      attr: rubber
+    capabilities: [sum, prod]
+    properties: [portable]
+    description:
+      - Standard business accoutrement.
+      - Particularly useful for `meet`-ings.
+      - Enables the `case` expression.
```
xsebek, line 145: Oh, of course, I heard about the counter from @byorgey here, so I did not expect anything else and used `meetAll`. :smile:

data/scenarios/Challenges/ice-cream.yaml
```diff
@@ -0,0 +1,242 @@
...
+      - workbench
+      - grabber
+      - dictionary
+      - lodestone
+      - lambda
+      - logger
+      - welder
+      - hearing aid
+      - net
+      - strange loop
```
xsebek, line 52: Do we really need all this? In case of an error, the inventory would scroll to the logger, so it's hard to see the acquired items. That may have been another reason why I missed the briefcase. :sweat_smile:

data/scenarios/Challenges/ice-cream.yaml
```diff
@@ -0,0 +1,242 @@
...
+      - Particularly useful for `meet`-ings.
+      - Enables the `case` expression.
+  - name: awning1
+    display:
+      char: '▒'
+      attr: flower
+    description:
+      - Decoration
+    properties: [known]
+  - name: awning2
```
xsebek, line 153: What does this mean? :raised_eyebrow:

xsebek (approved):
Nice challenge, it took me a while to figure out how to use the counter and `meetAll`. :+1:

data/scenarios/Challenges/ice-cream.yaml
kostmo: I'm using two decorative entities to emulate the classic two-color [dessert parlor awnings](https://www.google.com/search?q=awning+ice+cream+parlor&tbm=isch).

data/scenarios/Challenges/ice-cream.yaml
kostmo: > Suggestion: what about equipping the base with a `scanner`? It was a bit annoying wandering around trying to `grab` everything to see what it was.

To be fair, it's best that you grab all of those items anyway :)

data/scenarios/Challenges/ice-cream.yaml
kostmo: The only two that are not essential are `lodestone` and `net`.  I've removed those, but added a `scanner` per @byorgey's [comment](https://github.com/swarm-game/swarm/pull/1008#discussion_r1070611027).

PR #1009, opened 2023-01-15
stack exec invocation script with --fast compilation

kostmo:
`stack` invocation with `--fast` to work around #1000

Example invocation:

    ./scripts/play.sh --scenario data/scenarios/Challenges/Ranching/gated-paddock.yaml --autoplay

### DIFF
--- a/scripts/play.sh
+++ b/scripts/play.sh
@@ -1,0 +1,7 @@
+#!/bin/bash -ex
+
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+cd $SCRIPT_DIR/..
+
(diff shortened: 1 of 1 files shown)


scripts/play.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash -ex
+
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+cd $SCRIPT_DIR/..
+
+# See https://github.com/swarm-game/swarm/issues/1000#issuecomment-1378632269
+stack build --fast && stack exec swarm -- $@
```
xsebek, line 7: I would prefer a note explaining simply what this does so I can read it if I am on a train and the internet is down.

xsebek (approved):

PR #1010, opened 2023-01-15
Scenario launch options selection

### LINKED ISSUE #358: Allow player to choose a specific seed through the UI
Currently, a specific seed can be chosen on the command line, *e.g.*
```
stack run -- -c 00-classic -s 0
```
But if you choose *e.g.* classic mode via the "New Game" menu, you get a random seed with no way to override it.  There should be some (optional) mechanism by which the user can specify a seed if they wish (e.g. for "set seed"-style speedruns, to play the same world as their friend, etc.).

### LINKED ISSUE #866: Code golf scoring

kostmo:
Closes #358 and closes #866.

Allows specification of a seed value and/or the path of a script to run.  Specifying a script to run in advance allows eligibility for code size scoring.

Some effort was invested into integrating the Brick `FileBrowser` widget and discovering its idiosyncrasies.  This paves the way for more applications of `FileBrowser` within Swarm.

## Usage

From the scenario selection menu, press the `o` key to pop up a dialog for launch options.

Screenshot from 2023-06-06 01-38-25

Any manually-selected initial-script or seed are persisted to disk and will pre-populate the launch configuration dialog upon the next play.  If a certain scenario is subsequently launched the normal way (i.e. by pressing `Enter` instead of `o`), then this clears the saved script path/seed, and the next pop-up of the launch configuration dialog will not see its fields pre-populated.

## Warning: Save format changed

This PR changes the `ScenarioStatus` datatype, and therefore game status/progress saved previously to this PR will not be recognized.  See https://github.com/swarm-game/swarm/pull/974#discussion_r1111335102 for discussion about this situation.

### DIFF
--- a/src/Swarm/TUI/Launch/Controller.hs
+++ b/src/Swarm/TUI/Launch/Controller.hs
@@ -23,13 +23,13 @@ import Swarm.TUI.Model.StateUpdate
 import Swarm.TUI.Model.UI
 import Swarm.Util (listEnums)
 
-cacheValidatedInputs :: EventM Name AppState ()
+cacheValidatedInputs :: EventM Name LaunchOptions ()
 cacheValidatedInputs = do
-  launchControls <- use $ uiState . uiLaunchConfig . controls
+  launchControls <- use controls
   parsedParams <- liftIO $ parseWidgetParams launchControls
-  uiState . uiLaunchConfig . editingParams .= parsedParams
+  editingParams .= parsedParams
 
-  currentRing <- use $ uiState . uiLaunchConfig . controls . scenarioConfigFocusRing
+  currentRing <- use $ controls . scenarioConfigFocusRing
 
   let eitherLaunchParams = toValidatedParams parsedParams
       modifyRingMembers = case eitherLaunchParams of
@@ -38,7 +38,7 @@ cacheValidatedInputs = do
       maybeCurrentFocus = focusGetCurrent currentRing
       refocusRing = maybe id focusSetCurrent maybeCurrentFocus
 
-  uiState . uiLaunchConfig . controls . scenarioConfigFocusRing .= refocusRing (makeFocusRingWith $ modifyRingMembers listEnums)
+  controls . scenarioConfigFocusRing .= refocusRing (makeFocusRingWith $ modifyRingMembers listEnums)
 
 -- | If the FileBrowser is in "search mode", then we allow
 -- more of the key events to pass through. Otherwise,
@@ -95,8 +95,8 @@ handleFBEvent ev = do
         closeModal
     _ -> return ()
  where
-  closeModal = do
-    uiState . uiLaunchConfig . controls . fileBrowser . fbIsDisplayed .= False
+  closeModal = Brick.zoom (uiState . uiLaunchConfig) $ do
+    controls . fileBrowser . fbIsDisplayed .= False
     cacheValidatedInputs
 
 handleLaunchOptionsEvent ::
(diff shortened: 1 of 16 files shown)


src/Swarm/TUI/View.hs
```diff
@@ -165,35 +172,52 @@ drawLogo = centerLayer . vBox . map (hBox . T.foldr (\c ws -> drawThing c : ws)
...
+                    $ l
+                ]
+            , padLeft (Pad 5) (maybe (txt "") (drawDescription . snd) (BL.listSelectedElement l))
+            ]
+      , launchOptionsMessage
+      ]
+
+  launchOptionsMessage = case (displayedFor, snd <$> BL.listSelectedElement l) of
+    (Nothing, Just (SISingle _)) -> hCenter $ txt "Press 'o' for launch options"
+    _ -> emptyWidget
```
byorgey, line 206: It's very annoying that the entire menu shifts up and down by one line every time the `Press o for launch options` message appears and disappears.  If we change it to always take up a vertical line of space then the menu stays put:

```suggestion
    _ -> txt " "
```

Also, maybe this can say something like "Press 'o' for launch options, or 'Enter' to launch with defaults" ?

byorgey (approved):
Wow, amazing work!

Typing in the initial seed field takes a long time (i.e. it takes a second or two to update after every keypress), I wonder why?

src/Swarm/TUI/Launch/Controller.hs
```diff
@@ -0,0 +1,155 @@
...
+    _ -> return ()
+ where
+  closeModal = do
+    uiState . uiLaunchConfig . controls . fileBrowser . fbIsDisplayed .= False
+    cacheValidatedInputs
+
+handleLaunchOptionsEvent ::
+  ScenarioInfoPair ->
+  BrickEvent Name AppEvent ->
+  EventM Name AppState ()
```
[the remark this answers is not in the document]
kostmo: Can't do it everwhere because `startGameWithSeed` gets called, but I've made use of a few more `zooms` which does simplify the code.

src/Swarm/TUI/Launch/Controller.hs
```diff
@@ -0,0 +1,155 @@
...
+import Swarm.TUI.Model
+import Swarm.TUI.Model.Name
+import Swarm.TUI.Model.StateUpdate
+import Swarm.TUI.Model.UI
+import Swarm.Util (listEnums)
+
+cacheValidatedInputs :: EventM Name LaunchOptions ()
+cacheValidatedInputs = do
+  launchControls <- use controls
+  parsedParams <- liftIO $ parseWidgetParams launchControls
```
kostmo, line 29: > Typing in the initial seed field takes a long time (i.e. it takes a second or two to update after every keypress), I wonder why?

Here's why: the seed field and the script field are currently getting validated together upon every dialog event.  The script field is validated by parsing the code.  I'll look into separating the validation of each field.

src/Swarm/TUI/View.hs
byorgey: The message is updated, but it still has `emptyWidget`, so the menu still wiggles up and down depending on whether the message is displayed.  I'll open a new PR to fix it.

xsebek:
@kostmo I finally got around to play with this and its great! 👍 

Running custom scripts from UI feels like this was always supposed to be the way to do it. 🙂

PR #1018, opened 2023-01-17
swarm logo animation

kostmo:
./scripts/play.sh --scenario data/scenarios/Fun/logo-burst.yaml

image


byorgey (approved):
Hah, this is super fun!

PR #1024, opened 2023-01-22
Make `sand` a bit harder to get

### LINKED ISSUE #474: Sand should not be so easy to get
Right now, `sand` respawns immediately and infinitely.  This is not very realistic and it also makes it far too easy to get sand.

Perhaps we should make sand more like `rock` and `copper ore`: it does not regrow, but there is a way to produce more of it.  For example, perhaps drilling a `rock` produces some number of `sand`.

byorgey:
Closes #474.

- `sand` is no longer `infinite`
- `sand` can be obtained by drilling `rock`s.
- `rock`s, in turn, occasionally drop from mines when using a normal `drill` (but not a more efficient `metal drill`)


xsebek (approved):
Should the rock description change too?

PR #1026, opened 2023-01-22
Whack-a-mole scenario

kostmo:
image

### DIFF
--- a/data/scenarios/Challenges/_gopher/gopher.sw
+++ b/data/scenarios/Challenges/_gopher/gopher.sw
@@ -1,6 +1,9 @@
 def doN = \n. \f. if (n > 0) {f; doN (n - 1) f} {}; end;
 def abs = \n. if (n < 0) {-n} {n} end;
 
+def elif = \t. \then. \else. {if t then else} end
+def else = \t. t end
+
 def randSign = \x.
     opposite <- random 2;
     if (opposite == 1) {
@@ -72,33 +75,19 @@ def arrive = \fieldWidth. \fieldHeight.
 def getTauntStage = \startingAmount. \newCount.
     if ((newCount * 5) / startingAmount < 1) {
         return (0, "Hey, maybe we can work this out?");
-    } {
-        if ((newCount * 5) / startingAmount < 2) {
-            return (1, "I didn't hear no bell!");
-        } {
-            if ((newCount * 5) / startingAmount < 3) {
-                return (2, "Why don't you just give up?");
-            } {
-                if ((newCount * 5) / startingAmount < 4) {
-                    return (3, "Close one!");
-                } {
-                    if (newCount < startingAmount - 2) {
-                        return (4, "OK, no more Mr. Nice Gopher!");
-                    } {
-                        if (newCount < startingAmount - 1) {
-                            return (5, "Bet you can't do that again!");
-                        } {
-                            if (newCount < startingAmount) {
-                                return (6, "Beginner's luck!");
-                            } {
-                                return (7, "You'll never catch me!");
-                            };
-                        };
-                    };
-                };
-            };
-        };
-    };
+    } $ elif ((newCount * 5) / startingAmount < 2) {
+        return (1, "I didn't hear no bell!");
+    } $ elif ((newCount * 5) / startingAmount < 3) {
+        return (2, "Why don't you just give up?");
+    } $ elif ((newCount * 5) / startingAmount < 4) {
+        return (3, "Close one!");
+    } $ elif (newCount < startingAmount - 2) {
+        return (4, "OK, no more Mr. Nice Gopher!");
+    } $ elif (newCount < startingAmount - 1) {
+        return (5, "Bet you can't do that again!");
+    } $ elif (newCount < startingAmount) {
+        return (6, "Beginner's luck!");
+    } $ else {return (7, "You'll never catch me!");};
     end;
 
 def waitWhileHere = \e. \remainingTime.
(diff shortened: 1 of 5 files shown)


test/integration/Main.hs
```diff
@@ -162,6 +162,7 @@ testScenarioSolution _ci _em =
         , testSolution (Sec 5) "Challenges/2048"
         , testSolution (Sec 3) "Challenges/word-search"
         , testSolution (Sec 3) "Challenges/ice-cream"
+        , testSolution (Sec 90) "Challenges/gopher"
```
kostmo, line 165: As is, this test take 45 seconds on my computer to complete.  So I'm not sure whether we actually want this to be part of the test suite.

kostmo: I have now fixed this by adding a `wait 64` in the "polling" loop of the built robots.

data/scenarios/Challenges/_gopher/solution.sw
```diff
@@ -0,0 +1,179 @@
...
+            return $ x != "flower";
+        });
+
+    if shouldContinue {
+        if (n > 0) {
+            scanDirections $ n - 1;
+        } {
+            // Inserting this "wait" speeds up the
+            // integration test significantly (more than 3x)
+            wait 64;
```
kostmo, line 55: The integration test runs much faster with this.

byorgey (commented):
Nice!  My solution (written without looking at yours) is below.  Looks like they are functionally pretty similar.

I think we definitely need to fix the race condition before merging.  On the other hand I don't have a strong opinion about the `teleport`, I just wanted to hear your thought process and why you designed it that way.

```
def mod = \a. \b. (a - b*(a/b)) end

def forever = \c. c ; forever c end

def x4 = \c. c;c;c;c end

def drill_mound = \d.
  h <- scan d;
  if (h == inr "mound") { drill d } {}
end

def gopher_turret =
  forever (
    x4 ( turn left; drill_mound forward );
    drill_mound down
  )
end

def idoN_rec : int -> int -> (int -> cmd unit) -> cmd unit = \k. \n. \c.
  if (k == n) {} {c k; idoN_rec (k+1) n c}
end

def idoN : int -> (int -> cmd unit) -> cmd unit = idoN_rec 0 end

def doN : int -> cmd unit -> cmd unit = \n. \c. idoN n (\_. c) end

def place_turret = \row. \col.
  if (mod (2*row + col) 5 == 0) { build {gopher_turret}; return () } {}
end

def carpet_row = \r.
  idoN 28 (\c. place_turret r c; move);
  turn back; doN 28 move; turn left; move; turn left
end

def carpet = idoN 19 carpet_row end
```

data/scenarios/Challenges/_gopher/gopher.sw
```diff
@@ -0,0 +1,152 @@
...
+    let offsetXunsigned = (leadDist / 2) + offsetXrand in
+    let offsetYunsigned = leadDist - offsetXunsigned in
+
+    offsetX <- randSign offsetXunsigned;
+    offsetY <- randSign offsetYunsigned;
+
+    let startX = newDestinationX + offsetX in
+    let startY = newDestinationY + offsetY in
+
+    teleport self (startX, startY);
```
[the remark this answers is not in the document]
byorgey: Currently it also "burrows" through water, mountains, etc... which looks a bit strange.

data/scenarios/Challenges/_gopher/gopher.sw
```diff
@@ -0,0 +1,152 @@
...
+                    return (3, "Close one!");
+                } {
+                    if (newCount < startingAmount - 2) {
+                        return (4, "OK, no more Mr. Nice Gopher!");
+                    } {
+                        if (newCount < startingAmount - 1) {
+                            return (5, "Bet you can't do that again!");
+                        } {
+                            if (newCount < startingAmount) {
+                                return (6, "Beginner's luck!");
```
xsebek, line 92: This `if` nesting looks a bit strange, maybe you could use:
```
def elif = \t. \then. \else. {if t then else} end
def else = \t. t end
```

xsebek (commented):
Wow, this is a huge challenge, I am not sure I am ready to face that gopher. :smile:

data/scenarios/Challenges/_gopher/gopher.sw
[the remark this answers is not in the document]
kostmo: My original motivation for `teleport`-ing was to make it harder to predict where the gopher would go next.  However, I have now got rid of the `antenna`, which should prevent the player from "stalking" the gopher via `meet` as he tunnels (not to say that would be easy, either).

I think now I can forego the teleport and make the movement more natural.

data/scenarios/Challenges/_gopher/gopher.sw
```diff
@@ -0,0 +1,152 @@
...
+
+def waitWhileHere = \e. \remainingTime.
+    here <- isHere e;
+    if here {
+        wait 1;
+        if (remainingTime > 0) {
+            waitWhileHere e $ remainingTime - 1;
+        } {
+            // Pick up the "dropping"
+            grab;
```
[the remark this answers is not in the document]
kostmo: Fixed.

data/scenarios/Challenges/_gopher/gopher.sw
kostmo: fixed

kostmo:
I think this is ready now.

byorgey (approved):
Looks great!  I had to make a minor change to my solution since the type of `drill` changed but after that change it still worked, and I like the new non-teleporting gopher.

I especially like this challenge because it feels like it might just be possible if you type really fast but it quickly gets tedious/impossible and you start to wonder if there is a better way...

PR #1027, opened 2023-01-23
Device with recipe for 'appear' capability

kostmo:
Towards #26.

See [Groucho glasses](https://en.wikipedia.org/wiki/Groucho_glasses) 🥸


xsebek (approved):
I would not mind if the glasses required copper wire/glass, but it is nice heaving the recipe easy. 👍

PR #1028, opened 2023-01-23
device for setname

kostmo:
Towards #26


xsebek (approved):
Nice pun. 👍 

Can you add a recipe too please? Preferably not silver, since there is not much power in `setname` and it could be a bit dissapointing. 😅

Issue #950, opened 2023-01-03
Devices that Decouple Turn from Move capability


kostmo:
Originally raised in [this comment](https://github.com/swarm-game/swarm/issues/205#issuecomment-1284895726).

Currently, `treads` provide both `CTurn` and `CMove`.  This could be rationalized within swarm-world logic via treads providing "skid steering".  We should create a new device, `rollers`, that offer `CMove` capability but not `CTurn`.

Without the "turn" capability (`CTurn`), robots will only be able to travel forwards along a straight line from the base.  This would facilitate interesting puzzle challenges.

To extend this thought further, we could add a new capability `CReverse` that enables only a single direction change: `turn back`.  This would facilitate challenges where longitudinal, but not lateral, movement is allowed.  This proposed capability is related to the `COrient`.

byorgey:
Alternatively, what about adding a `reverse` command which moves backwards?  `treads` and `rollers` could both provide it.  That way we wouldn't have to go to the trouble of distinguishing directions.

xsebek:
I vaguely remember we created an Issue for challenge DSL - in this case, the base could be given the `reverse` command implemented like:
```yaml
base:
  commands_or_something: |
    def reverse = doWithoutChecks { turn back; } end
```

xsebek:
I am in favour of splitting `CTurn` from `CMove`, that seems like a logical distinction. :+1: 

Taking this further to `CReverse` feels wrong to me. For example, in Robot Karel's language, the basic task is to turn around using only `turn left`. If I wanted to create such a challenge in Swarm, taking the `CReverse` approach would lead me to create a `CLeft` capability.

Taking this much further I would prefer writing scenarios in scenario files using basic building blocks. Having to make changes in the Haskell codebase to express the scenario conditions ("you can not turn right",...) kind of limits it to Swarm maintainers.

byorgey:
Also related: #864 .  If we had that one could easily e.g. make a scenario with a `rollers` device and disallow crafting `treads`.  Though I'm not against adding a globally available `rollers` device.

Also related: #1524 ; using the envisioned ability to attach code to entities, it would be possible to e.g. create a custom `reverser` device that runs the code `turn back`.

byorgey:
Also also related: #1553.

xsebek:
@kostmo, can this be closed? Scenarios can now specify device with separate capabilities for each command.

kostmo:
[...]

PR #1029, opened 2023-01-23
device for turning

kostmo:
Toward #950 and #26

### DIFF
--- a/data/entities.yaml
+++ b/data/entities.yaml
@@ -667,6 +667,21 @@
   properties: [portable]
   capabilities: [random]
 
+- name: slew ring
+  display:
+    attr: device
+    char: 'o'
+  description:
+  - A low-friction, rotational roller bearing.
+  - Equipping a slew ring on a robot allows it to turn.
+  - The `turn` command takes a direction as an argument, which
+    can be either absolute (north, west, east, south) or relative
+    (left, right, forward, back, down).
+  - 'Example:'
+  - '  move; turn left; move; turn right'
+  capabilities: [turn]
+  properties: [portable]
+
 - name: treads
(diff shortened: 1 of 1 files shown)


xsebek (commented):
I like the concept (and shorter documentation), but I am not sure this should go in default entities.

`threads` still have turn, so how is this useful besides an academic “one-capability-one-device”?

This might as well be in a scenario and players would not have to waste resources on an extra device.

kostmo:
> This might as well be in a scenario and players would not have to waste resources on an extra device.

I would be fine with defining this per-scenario instead, but I don't quite understand this statement.  Why would making this a member of the default entities obligate players to waste resources on it?

xsebek:
Because they will discover it through recipes. The game does not say “this entity is only for scenarios dont worry about it” in any way.

Also if we add an achievement like “all devices installed” then it will literally be required.

byorgey:
Yeah, once you have the ingredients to make a particular entity, its name shows up in your Recipes.  But then the only way to find out what it is and what it does is to actually make one.  At that point you can finally read its description.  I think I agree with @xsebek. As a player, I assume that when the UI flashes "new recipe!" at me, it feels like it is giving me a hint that I might find it useful to construct this new thing.  Given that, I would find it annoying if after going to the trouble of making that thing, I find out that it actually gives me no new capabilities whatsoever.  It would feel sort of like the UI lied to me.

I think we should *either* just define this in whatever custom scenarios need it, *or* add this and also remove the `turn` capability from `treads`.  But that would just add more complexity to the early game and tutorials.

(Maybe we should change how entity discovery works, but that is a bigger conversation.)

PR #1031, opened 2023-01-23
DNA copying challenge

### LINKED ISSUE #932: Challenge scenario requiring the use of a stack
I was thinking the other day that it would be really cool to have [stacks](https://en.wikipedia.org/wiki/Stack_(abstract_data_type)) somehow.  Like a physical thing that you could push and pop entities on.  This would let you simulate [pushdown automata](https://en.wikipedia.org/wiki/Pushdown_automaton) and would probably allow for some cool automated production scenarios.

*However*, I don't necessarily want to just create a new device or built-in thing.  What I really want to think about is, what would you need to be able to program a stack yourself?  I'm thinking something like this:
- We might need recursive types (#154) to be able to store the stack itself, though you might be able to get away with a Church-encoded list (but popping would be expensive).
- You need some way to communicate (#94) at close range to be able to "invoke methods" on the stack robot.  *e.g.* I'm imagining to push, you `place` the item you want to push and then send a `push` message to the stack robot, which will pick up the item and store it.  To pop, you send a `pop` message, and it will place an item down.

Would be happy to hear about other approaches I haven't thought of.

kostmo:
A challenge that utilizes a stack.
Closes #932.

Demo:
```
scripts/play.sh --scenario data/scenarios/Challenges/dna.yaml --autoplay
```
Screenshot from 2024-10-13 17-06-06

### DIFF
--- a/data/scenarios/Challenges/00-ORDER.txt
+++ b/data/scenarios/Challenges/00-ORDER.txt
@@ -16,6 +16,7 @@ wave.yaml
 gallery.yaml
 wolf-goat-cabbage.yaml
 blender.yaml
+dna.yaml
 friend.yaml
 pack-tetrominoes.yaml
 dimsum.yaml
--- a/data/scenarios/Challenges/_dna/lab.sw
+++ b/data/scenarios/Challenges/_dna/lab.sw
@@ -1,0 +1,177 @@
+def doN = \n. \f. if (n > 0) {f; doN (n - 1) f} {}; end;
+
+def getBaseForNumber = \n.
+    if (n == 0) {
+        "guanine";
+    } {
+        if (n == 1) {
+            "cytosine";
+        } {
+            if (n == 2) {
+                "adenine";
+            } {
+                "thymine";
+            };
+        };
+    };
+    end;
+
+def getNumberForBase = \n.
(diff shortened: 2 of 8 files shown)


xsebek:
@kostmo could you finish this using the `let` binding workaround suggested in https://github.com/swarm-game/swarm/issues/1032#issuecomment-1402465755?

This challenge looks really good and the workaround is only one extra `let` that we can remove once the issue is fixed.

xsebek:
@kostmo, is this ready now that the bug has been fixed? 🙂

I think this was before `rec`, so the base might need a device for that.

byorgey (approved):
Love it!!  My solution can be seen here: https://github.com/byorgey/swarm-defs/blob/main/dna.sw

I note only that it was slightly confusing that e.g. the `siphon` says it enables the `drill` command, but in fact one must say `use "siphon"` rather than `drill`.  That's not something that can/should be solved in this PR but I thought it worth mentioning.

Once I had what I thought was sure to be a working solution, I enjoyed watching in horror as the mushrooms spread uncontrollably and ruined it, requiring me to switch the cloning order. :mushroom: :mushroom: :mushroom: :grin:

Issue #1032, opened 2023-01-23
Values referenced after a recursive call get overwritten


kostmo:
**Describe the bug**

The following program produces unexpected output in the log:
```
      def go = \n.
          if (n > 0) {
              idx <- random 4;
              log $ "In: " ++ format idx;

              go $ n - 1;

              log $ "Out: " ++ format idx;
              move;
          } {};
          end;

      go 4;
```
Output in log:
```
In: 1
In: 3
In: 0
In: 2
Out: 0
Out: 0
Out: 0
Out: 0
```

xsebek:
@byorgey this looks to me like a case of #681, am I right? :thinking:

byorgey:
Yep, almost certainly #681, although to be honest I don't quite understand why we don't see the final `2` being repeated, like this:
```
In: 1
In: 3
In: 0
In: 2
Out: 2
Out: 2
Out: 2
Out: 2
```
Most likely duplicate, but I'll leave this open for now so when we fix #681 we can make sure it fixes this too.

xsebek:
@kostmo better workaround: use a let binding
```
def go = \n.
    if (n > 0) {
        idx <- random 4;
        let indx = idx in
        log $ "In: " ++ format indx;

        go $ n - 1;

        log $ "Out: " ++ format indx;
        move;
    } {};
end;

go 4;
```

---

@byorgey when I tried the example code by @kostmo right now, it did repeat the last value.

Maybe we have a different version or @kostmo made a typo. :slightly_smiling_face:

byorgey:
After looking at #1033 , I am now 100% certain this is caused by #681.  It turns out that #1021 is also causing issues here which is why I was confused.

PR #1033, opened 2023-01-23
recursion bug demo

kostmo:
Two demonstrations for #1032 are added in this PR.

## Value overwriting demo

```
./scripts/play.sh --scenario scenarios/Testing/1032-recursion/1032-recursion-zeroed-output.yaml
```
Log output:
```
In: 1
In: 3
In: 0
In: 2
Out: 0
Out: 0
Out: 0
Out: 0
```

## Truncation demo
This program is **identical** to the one above, except that the final `move` command is omitted.  Omitting this seems to cause the program to stop producing output prematurely.
```
./scripts/play.sh --scenario scenarios/Testing/1032-recursion/1032-recursion-truncated-output.yaml
```
Log output:
```
In: 1
In: 3
In: 0
In: 2
Out: 0
Out: 0
```

### DIFF
--- a/data/scenarios/Testing/1032-recursion/1032-recursion-truncated-output.yaml
+++ b/data/scenarios/Testing/1032-recursion/1032-recursion-truncated-output.yaml
@@ -1,0 +1,42 @@
+version: 1
+name: Demo recursion bug
+author: Karl Ostmo
+description: |
+  Demo recursion bug
+creative: false
+seed: 1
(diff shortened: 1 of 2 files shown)


byorgey:
@kostmo, thanks!  This was helpful.  I think we can close this now that it's understood what the issues are (or else turn it into a failing test case for when we fix #681 (soon)).

PR #1045, opened 2023-01-25
Todo enforcement line number

kostmo:
As an extension to #514, include the line number in the grep output, and move the grep logic to a script, so that it can be used within VS Code and clicking on the file:line will open the editor at the right place.

vscode-demo

The GitHub action still works.  See this CI run: https://github.com/swarm-game/swarm/actions/runs/4001861308/jobs/6868527786


byorgey (approved):
Looks good to me.

xsebek:
@kostmo thanks for improving the GitHub action 👍

PR #1055, opened 2023-01-26
Focus ring for goals modal

### LINKED ISSUE #1054: No longer possible to scroll long goals
**Describe the bug**
When the goal dialog is open and the goal description text is too large for the dialog (either because there is a lot of text or the window is too small), a scrollbar will show on the right edge, but it is not possible to scroll it to read the rest of the description.

**To Reproduce**
The `conditionals` tutorial is a good one to exhibit this bug since it has a very long description.  Open the `conditionals` tutorial and then decrease the window size until the goal does not fit.

**Screenshots**
unscrollable-goal

**Additional context**
This was caused by the below code, which was added as part of #927 .  All modals used to fall through to the call to `handleInfoPanelEvent`, which is poorly named since really all it is doing is handling keys like arrows/PgUp/PgDown etc. in a scrollable panel.  However, now arrow keys are used to select a highlighted item in the list of objectives.

https://github.com/swarm-game/swarm/blob/bf8a9a66c5f69d3a61f3d4dc01029fdcc4282aaf/src/Swarm/TUI/Controller.hs#L408-L416

I am not sure of the best way to fix this.  At the very least, when there is no list of objectives to show, arrow keys etc. should work to scroll the goal.  But that's not a solution, since sometimes we might need to scroll around in the list of objectives **and** scroll the goal description text...

kostmo:
Fixes #1054

## Testing

    ./scripts/play.sh --scenario data/scenarios/Testing/795-prerequisite/795-prerequisite-mutually-exclusive.yaml

and

    ./scripts/play.sh --scenario data/scenarios/Tutorials/conditionals.yaml

### DIFF
--- a/src/Swarm/Game/Scenario/Objective/Presentation/Render.hs
+++ b/src/Swarm/Game/Scenario/Objective/Presentation/Render.hs
@@ -28,21 +28,20 @@ makeListWidget (GoalTracking _announcements categorizedObjs) =
 
 renderGoalsDisplay :: GoalDisplay -> Widget Name
 renderGoalsDisplay gd =
-  padAll 1 $
-    if hasMultiple
-      then
-        hBox
-          [ leftSide
-          , hLimitPercent 70 goalElaboration
-          ]
-      else goalElaboration
+  if hasMultiple
+    then
+      hBox
+        [ leftSide
+        , hLimitPercent 70 $ padLeft (Pad 2) goalElaboration
+        ]
+    else goalElaboration
  where
   hasMultiple = hasMultipleGoals $ gd ^. goalsContent
   lw = _listWidget gd
   fr = _focus gd
   leftSide =
     hLimitPercent 30 $
-      vBox
+      padAll 1 $ vBox
         [ hCenter $ str "Goals"
         , padAll 1 $
             vLimit 10 $
@@ -54,12 +53,13 @@ renderGoalsDisplay gd =
     (True, Just (GoalWidgets GoalSummary)) -> withAttr lightCyanAttr
     _ -> id
 
+  -- Note: An extra "padRight" is inserted to account for the vertical scrollbar,
+  -- whether or not it appears.
   goalElaboration =
     clickable (GoalWidgets GoalSummary) $
       maybeScroll ModalViewport $
-        padLeft (Pad 2) $
-          maybe emptyWidget (highlightIfFocused . singleGoalDetails . snd) $
-            BL.listSelectedElement lw
+        maybe emptyWidget (padAll 1 . padRight (Pad 1) . highlightIfFocused . singleGoalDetails . snd) $
+          BL.listSelectedElement lw
 
 getCompletionIcon :: Objective -> GoalStatus -> Widget Name
 getCompletionIcon obj = \case
--- a/src/Swarm/Game/Scenario/Objective/Presentation/Render.hs
+++ b/src/Swarm/Game/Scenario/Objective/Presentation/Render.hs
@@ -4,6 +4,7 @@
 module Swarm.Game.Scenario.Objective.Presentation.Render where
 
 import Brick hiding (Direction, Location)
+import Brick.Focus
 import Brick.Widgets.Center
 import Brick.Widgets.List qualified as BL
 import Control.Applicative ((<|>))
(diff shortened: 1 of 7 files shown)


src/Swarm/Game/Scenario/Objective/Presentation/Render.hs
```diff
@@ -37,20 +38,26 @@ renderGoalsDisplay gd =
...
         , padAll 1 $
             vLimit 10 $
-              BL.renderList (const drawGoalListItem) True lw
+              withFocusRing fr (BL.renderList drawGoalListItem) lw
         ]
-  goalsCount = sum . M.elems . M.map NE.length . goals $ gd ^. goalsContent
+
+  -- Adds very subtle coloring to indicate focus switch
+  highlightIfFocused = case focusGetCurrent fr of
+    Just (GoalWidgets GoalSummary) -> withAttr lightCyanAttr
```
byorgey, line 53: I think the light cyan is nice to indicate focus when there's an objective list and a goal description, but I find it annoying that all the goals are now colored light cyan, even when there's only a single objective.

src/Swarm/TUI/Controller.hs
```diff
@@ -406,11 +407,18 @@ handleModalEvent = \case
...
+        V.EvKey (V.KChar '\t') [] -> uiState . uiGoal . focus %= focusNext
+        _ -> do
+          focused <- use $ uiState . uiGoal . focus
+          case focusGetCurrent focused of
+            Just (GoalWidgets w) -> case w of
+              ObjectivesList -> do
+                lw <- use $ uiState . uiGoal . listWidget
+                newList <- refreshList lw
+                uiState . uiGoal . listWidget .= newList
+              GoalSummary -> handleInfoPanelEvent modalScroll (VtyEvent ev)
```
byorgey, line 420: The strange thing now is that the entire modal is one big scrollable widget, so when you focus on the description you can scroll both the description and the objective list together; and conversely when you are focused on the objective list it doesn't scroll, even if the currently highlighted objective is off the bottom of the modal.  I would think that (1) the objective list and the description should be made scrollable separately (maybe the objective list has a scrollbar on the left?), and (2) navigating through the objective list should cause it to scroll just like we do for inventory.  I don't quite remember how that works for the inventory but I don't think it's hard, it's mostly handled automatically by brick.

src/Swarm/TUI/Controller.hs
```diff
@@ -406,11 +407,18 @@ handleModalEvent = \case
     Brick.zoom (uiState . uiModal . _Just . modalDialog) (handleDialogEvent ev)
     modal <- preuse $ uiState . uiModal . _Just . modalType
     case modal of
-      Just GoalModal -> do
-        lw <- use $ uiState . uiGoal . listWidget
-        newList <- refreshList lw
-        uiState . uiGoal . listWidget .= newList
-      Just _ -> handleInfoPanelEvent modalScroll (VtyEvent ev)
+      Just GoalModal -> case ev of
+        V.EvKey (V.KChar '\t') [] -> uiState . uiGoal . focus %= focusNext
```
byorgey, line 411: It's strange that you can hit Tab to "switch focus" even when there is only a single goal and there is no objective list displayed.

src/Swarm/Game/Scenario/Objective/Presentation/Render.hs
kostmo: fixed

src/Swarm/TUI/Controller.hs
kostmo: fixed

src/Swarm/TUI/Controller.hs
kostmo: Fixed now.  List widgets do have scrolling built in (without scroll bars).

byorgey (approved):
Looks good to me!

We should also make the objectives clickable, just like inventory items are clickable, and while we're at it we can make it so clicking on the goal description will change focus as well.  But we can do that in a separate PR.

byorgey:
Hmm, actually, I just noticed that now the scrollbar overwrites the rightmost column of text:

scrollbar

Compare to the main branch:

scrollbar-main

In this PR the scrollbar has somehow moved over two columns to the left.  Looking at the code I'm not yet sure what causes this.

kostmo:
> scrollbar overwrites the rightmost column of text

Seems to be fixed after fiddling with the padding a bit.

PR #1065, opened 2023-01-31
Fix scenarios being skipped or repeated when choosing "keep playing"

### LINKED ISSUE #1064: Challenges get skipped and repeated after choosing "keep playing"
**Describe the bug**
This seems like #682 all over again, but slightly different.  After selecting "keep playing" on a scenario then progressing to the next, the following scenario will be skipped, and the next one will be repeated.

**To Reproduce**
1. Start the game and select `Tutorial`.
2. Complete the first scenario with `say "Ready!"` but select `Keep playing` from the dialog.
3. Now hit Ctrl-Q and go to `Next challenge`.
4. The crafting tutorial will appear (should be the `move` tutorial).
5. Complete the crafting tutorial and select `Next challenge`.
6. The crafting tutorial will appear again.

byorgey:
Introduce an invariant that the currently selected menu item is ALWAYS the same as the currently played scenario.   `quitGame` is now the ONLY place where the menu should be advanced.

Fixes #1064.

### DIFF
--- a/src/Swarm/TUI/Controller.hs
+++ b/src/Swarm/TUI/Controller.hs
@@ -759,7 +759,6 @@ doGoalUpdates = do
       gameState . winCondition .= WinConditions (Unwinnable True) x
       openModal LoseModal
 
-      uiState . uiMenu %= advanceMenu
       return True
     WinConditions (Won False) x -> do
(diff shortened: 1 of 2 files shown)


xsebek (approved):
Thanks for the comments, the new invariant seems much more reasonable to me. :+1:

Issue #934, opened 2022-12-28
How to manage "vendored" boolexpr dependency?


kostmo:
The [`boolexpr` package](https://hackage.haskell.org/package/boolexpr) needs [this PR](https://github.com/np/boolexpr/pull/2) to support modern GHC.  I was able to include the "boolexpr" package as a "[vendored dependency](https://stackoverflow.com/a/39643873/105137)" by copying in the code and [adding it as a local package in `stack.yaml`](https://github.com/swarm-game/swarm/blob/98b793e5289083a30def9faae2fda4d1bcad6e38/stack.yaml#L3).  This works with `stack build`.  However, the [CI build](https://github.com/swarm-game/swarm/actions/runs/3791622200/jobs/6447185772) doesn't seem to respect this local dependency.

---

**UPDATE:** The PR np/boolexpr#2 has been merged - thanks @np and @kostmo!

xsebek:
We used to do this in cabal-extra (I think was the name) with other libraries.

The problem is that this will not work for release. The built binary will have the released version of the dependency.

xsebek:
@kostmo I could not find your comment about whether you reached out to the `boolexpr` author via mail in #927.

I reached out to the author on the Keybase messaging platform and will update you with the results. :wink:

PR #1069, opened 2023-01-31
Split swarm libraries

### LINKED ISSUE #1043: Split Swarm components into libraries
I propose we split the sources into Cabal libraries.
 
## Practical problem: compilation

When I run tests after changing some `Swarm.Language` modules, every Swarm source file (60+) is recompiled.

However, we do not need to compile LSP or TUI or Web API, because we do not test them.

## Theoretical problem: dependencies

We mostly* avoided using TUI or Web dependencies in the code for language and game logic.

In theory, we could switch the dependencies for new ones or add another UI without changing the game or language code.

Left unchecked, the dependencies could slowly leak and make future experiments less viable.

At least it would be easier to see which dependencies are needed for what part of the Swarm application. 

# Solution

In my mind, Swarm code is split like this in order of dependency:
- `Swarm.Language`
  - `Swarm.Language.LSP` (move to its own directory?)
  - `Swarm.Game`
    - `Swarm.DocGen` (not really required in the final application)
    - `Swarm.TUI`
      - `Swarm.Web` 

`Swarm.App` is just a small module to neatly initialise everything in order.

There is also `Swarm.Version`, which has a failure type that we use in TUI to show it nicely.
We could split off the failure type and use the network call and Git polling only in App.

---

<sub>*) except for Brick `AttrName` in `Swarm.Game.Display`.

xsebek:
- split libraries to enforce dependency policy
  - add `-Wunused-packages` to keep the dependencies minimal
- closes #1043

---

This is how our dependencies look like and the libraries should enforce that we keep everything well ordered:

```sh
find src -name '*.hs' | xargs graphmod -q -p | dot -Tsvg -o swarm.svg
```

swarm

On a high level, the libraries look like this:
```mermaid
graph TD;
    swarm-language --> swarm-util;
    swarm-language-server --> swarm-util;
    swarm-language-server --> swarm-language;
    swarm-game --> swarm-util;
    swarm-game --> swarm-language;
    swarm-tui --> swarm-util;
    swarm-tui --> swarm-language;
    swarm-tui --> swarm-game;
    swarm-tui --> swarm-version;
    swarm-app --> swarm-util;
    swarm-app --> swarm-language;
    swarm-app --> swarm-game;
    swarm-app --> swarm-version;
    swarm-app --> swarm-tui;
    swarm-exe --> swarm-app;
    swarm-exe --> swarm-language;
    swarm-exe --> swarm-language-server;
    swarm-exe --> swarm-version;
    swarm-exe --> swarm-tui;
```


byorgey:
We also need to make sure to update `stack.yaml` as well as the release action.

src/swarm-language/Swarm/Language/Syntax.hs
```diff
@@ -145,6 +145,9 @@ data RelativeDir = DLeft | DRight | DBack | DForward | DDown
 data Direction = DAbsolute AbsoluteDir | DRelative RelativeDir
   deriving (Eq, Ord, Show, Read, Generic, Data, Hashable, ToJSON, FromJSON)
 
+-- | TODO: unify Heading without backward dependency?
+type Heading = V2 Int32
```
xsebek, line 149: @byorgey do you have any suggestions on how to resolve this?

Maybe we could split off a `Swarm.Language.Direction` which could export `Heading`. :thinking:

byorgey: I think we should just move all this DirInfo and heading stuff to `Swarm.Game.Direction` or something like that.  It is all about the meaning/semantics of aspects of the language, which really doesn't belong here.

byorgey: To expand on my previous comment, which was written hastily on a phone: stuff in `Swarm.Language.*` is about defining the *static* aspects of the language, e.g. the syntax and type system.  Stuff in `Swarm.Game.*` defines the *dynamic* aspects of the language, i.e. the *meaning* of things in the language and what they do/are at runtime (plus a lot of other game stuff that is inextricably tied to the language semantics).  So the `Direction` type definitely belongs here in `Swarm.Language.Syntax`. But all the stuff about converting `Direction`s to `Heading`s really should go in `Swarm.Game.*` somewhere.

xsebek: Done in #1103. :+1:

src/swarm-game/Swarm/Game/Util/Failure.hs
```diff
@@ -11,7 +11,7 @@
 --
 -- These failures are often not fatal and serve
 -- to create common infrastructure for logging.
-module Swarm.TUI.Model.Failure where
+module Swarm.Game.Util.Failure where
```
xsebek, line 14: @kostmo FYI I had to move the module back to the Game just to compile, but the change in #1116 is better (no `Util` directory and less code in the `Util` module itself) so I will rebase once merged. :+1: 

It seems that I am also susceptible to making silly `Util` modules when desperate. :sweat_smile: 

Splitting the libraries with Cabal has been a great learning experience for me, so I hope not to repeat my mistake from #982 where I put this module in TUI in the first place. :facepalm:

swarm.cabal
```diff
@@ -1,6 +1,6 @@
 cabal-version:      2.4
 name:               swarm
-version:            0.3.0.1
+version:            0.3.0.0
```
xsebek, line 3: - [ ] fix this copy-paste of the whole file
- [ ] remove unneeded dependencies

swarm.cabal
```diff
@@ -295,7 +625,7 @@ test-suite swarm-integration
 benchmark benchmark
   import:         stan-config, common, ghc2021-extensions
   main-is:        Benchmark.hs
-  hs-source-dirs: bench
+  hs-source-dirs: test/bench
```
xsebek, line 628: @byorgey I moved the `app` and `bench` directories to `src` and `test` respectively.

Sometimes I forgot to recompile the benchmark until the CI told me, so I hope seeing it with other tests will remind me about it. :slightly_smiling_face:

.github/workflows/haskell-ci.yml
```diff
@@ -179,8 +171,14 @@ jobs:
...
           if [ $((HCNUMVER >= 90400)) -ne 0 ] ; then hlint --version ; fi
+      - name: save cache (tools)
+        uses: actions/cache/save@v3
+        if: always()
+        with:
+          key: ${{ runner.os }}-${{ matrix.compiler }}-tools-0367592e
+          path: ~/.haskell-ci-tools
       - name: checkout
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
```
xsebek, line 190: Speaking of CI, it looks like regenerating this file will get rid of GitHub warnings about deprecated actions. :slightly_smiling_face: 

Let's hope it will all work nicely. :crossed_fingers:

swarm.cabal
```diff
@@ -149,70 +298,73 @@ library
...
+                      swarm-version,
+    hs-source-dirs:   src/swarm-user-interface
+    default-language: Haskell2010
+    default-extensions:
+      -- Avoid unexpected unevaluated thunk buildup
+      -- See discussion in #415
+      StrictData
+
+
+library
```
byorgey, line 341: Why doesn't this `library` have a name?

swarm.cabal
```diff
@@ -79,12 +80,113 @@ common ghc2021-extensions
       -- Not GHC2021, but until we get \cases we use \case a lot
       LambdaCase
 
-library
+
+library boolexpr
```
byorgey, line 84: What's our plan for `boolexpr`?  I don't think we can have an internal library called `boolexpr` when there is already a package on Hackage with that name.

byorgey (commented):
I still mostly think that this is a good idea but I am still a bit concerned about the maintenance burden of managing package versions, doing releases, etc.  What's our plan for updating the release action?

swarm.cabal
xsebek: It did not before, it is simply `swarm`. I am not 100% clear about whether we need this or if `swarm-app` would be better or maybe if it could even merge with `swarm-exe`.

I do not think we need to do anything about this right now.

xsebek:
@byorgey the release action can stay exactly as it is. All commands it uses stay the same, just as you can still use the same Stack/Cabal commands as before.

> I am still a bit concerned about the maintenance burden of managing package versions, doing releases

- package versions: once we thin out `swarm-util` the lowest library can set the package version range and the dependent libraries can remove them - this will result in two kinds of dependencies:
    - common to basically all (`base`)
    - specific to one package (`unification-fd`, `brick`)
- we can use tooling like `packdeps` or `cabal-bounds` to ease the pain
- we can now rely on `-Wunused-packages` to keep things minimal
- ultimately we will not make changes too often
- releases stay the same (write CHANGELOG entry, bump Cabal version, push git tag :beach_umbrella: :tropical_drink: )

byorgey:
OHHHH, I just realized what my confusion was.  I thought by doing this we were going to have to upload a bunch of separate packages to Hackage, like `swarm-language`, `swarm-game`, etc.  But we can still just upload the one `swarm` package to Hackage, which happens to have a bunch of internal `library` components... right?

If that's true, then none of my concerns are valid (e.g. version number management, release tooling, and what we will do with `boolexpr`)!

xsebek:
>  But we can still just upload the one swarm package to Hackage, which happens to have a bunch of internal library components... right?

First, yes. Second, I am not sure.

The Hackage release is still just `$CABAL v2-sdist` and we let `haskell-actions/hackage-publish@v1` handle it.

But _I think_ it will actually publish the libraries separately _and also publish `swarm`_. Think about it like the distinction between `optics-core` and `optics`. This means that I should probably use the `reexported-modules` field in `swarm.cabal`. It looks like Hackage displays the  reexported modules now:
https://hackage.haskell.org/package/optics

swarm.cabal
xsebek: @kostmo @byorgey let's discuss this in #934 :eyes:

byorgey:
> But _I think_ it will actually publish the libraries separately _and also publish `swarm`_. Think about it like the distinction between `optics-core` and `optics`.

But `optics-core` and `optics` each have their own separate `.cabal` file.  There is absolutely no way that publishing something with a single `.cabal` file could actually cause multiple separate packages to be uploaded to Hackage.  How would each one get a version number?  Would they all have the *same* `.cabal` file? etc.

byorgey:
> This means that I should probably use the `reexported-modules` field in `swarm.cabal`.

Yes, probably so.  We want someone who depends on the `swarm` package to be able to use any of the modules from any of our internal packages (possibly except for a few "internal/private" ones).

byorgey:
We should probably try running Haddock on this locally to make sure that it produces something reasonable.

xsebek:
@byorgey I tested it with Haddock and it showed all modules. :slightly_smiling_face: 

Feel free to test it yourself and with Stack. Otherwise, I will have to download Stack to test it myself. :smile:

byorgey:
I tried it with `stack`, and the first thing that happens is it prints this concerning message:
```
SubLibrary dependency is not supported, this will almost certainly fail
```
...and then it proceeds to build everything perfectly. :shrug:

PR #1076, opened 2023-02-01
Make panes collapsible

### LINKED ISSUE #813: Make panes collapsible
Panes like the inventory could be collapsable. The UI should leave space for future complexity.

_originally posted by @jrvieira in #206_

---

Note that collapsible REPL ties in with being able to send commands through the Web API:
- #625

## Solution

- add keyboard shortcuts to toggle the REPL and Inventory panels
  - when they are "off", keep only one line/column widget so that clicking it toggles the panel again

## Alternatives

We could also have "clickable" widgets (with new names) that hide the panels, but that could be harder or lead to more boilerplate.

We could have _more_ shortcuts to make the panels larger and smaller - once they reach a certain limit, they would collapse to the clickable widget that restores them to normal size. (In case you just pressed something and don't know how to undo it.)

ussgarci:
Add Ctrl-s shortcut to collapse/expand REPL panel

closes #813


xsebek:
Hi @ussgarci, great to see you again! 🙂 

If you need any help, please feel free to ask. 😉

src/Swarm/TUI/View.hs
```diff
@@ -588,6 +588,7 @@ helpWidget theSeed mport =
     , ("Ctrl-z", "decrease speed")
     , ("Ctrl-w", "increase speed")
     , ("Ctrl-q", "quit the current scenario")
+    , ("Ctrl-r", "collapse/expand REPL")
```
byorgey, line 591: Note we already use ctrl-r for reverse search in the repl, so we'll have to pick a different shortcut for this.

ussgarci: Thanks for the heads up!  :+1:

ussgarci: changed to "Ctrl-k"

byorgey:
I guess the panel is still surrounded by a border even when it has 0 size.  Maybe we can force it to have height 0 but full width?  Or maybe we can completely replace it (panel and all) with a horizontal line?

src/Swarm/TUI/View.hs
byorgey: Ctrl-k is in use also, it kills text to the end of the line (just like in emacs or on the command line).  Maybe Ctrl-SPC?  I don't know if that will work in terminals.  I would also suggest Ctrl-M (for Minimize) but I seem to recall that just generates a `Return` key event... Ctrl-n, perhaps?  Grasping at straws here...

src/Swarm/TUI/View.hs
```diff
@@ -334,6 +308,37 @@ drawGameUI s =
   fr = s ^. uiState . uiFocusRing
   moreTop = s ^. uiState . uiMoreInfoTop
   moreBot = s ^. uiState . uiMoreInfoBot
+  showREPL = s ^. uiState . uiShowREPL
+  rightPanel = if showREPL then worldPanel ++ replPanel else worldPanel
```
byorgey, line 312: This might be a bit nicer (assuming we make `replPanel` just a Widget and not a list):
```suggestion
  rightPanel = worldPanel ++ [ replPanel | showREPL ]
```

byorgey: I think completely hiding the REPL panel is reasonable.  We might also like to try replacing it with something like `hBorder` or `hBorderWithLabel`?  Not sure which would look better.

src/Swarm/TUI/View.hs
```diff
@@ -334,6 +308,37 @@ drawGameUI s =
...
+        fr
+        (FocusablePanel WorldPanel)
+        ( plainBorder
+            & bottomLabels . rightLabel ?~ padLeftRight 1 (drawTPS s)
+            & topLabels . leftLabel ?~ drawModalMenu s
+            & addCursorPos
+            & addClock
+        )
+        (drawWorld (s ^. uiState . uiShowRobots) (s ^. gameState))
+    , drawKeyMenu s
```
byorgey, line 325: We should also add a hint about the keybinding to hide the REPL to the top line of the key menu.

src/Swarm/TUI/Controller.hs
```diff
@@ -320,6 +320,8 @@ handleMainEvent ev = do
     -- toggle creative mode if in "cheat mode"
     ControlChar 'v'
       | s ^. uiState . uiCheatMode -> gameState . creativeMode %= not
+    -- toggle collapse/expand REPL
+    ControlChar 'k' -> uiState . uiShowREPL %= not
```
byorgey, line 324: As I mentioned in a comment elsewhere Ctrl-k is actually also taken (though it's not obvious from the Swarm source code because it's one of the default keybindings provided by the brick editor widget: https://hackage.haskell.org/package/brick-1.6/docs/Brick-Widgets-Edit.html ).

src/Swarm/TUI/View.hs
ussgarci: Hi @byorgey .  Here is the minimized REPL with a hBorder.

image

byorgey: @ussgarci I think that looks nice, though we probably shouldn't hard-code the color as cyan, in case we *e.g.* later allow the focus color to be customized.  Ideally, it will be either cyan or white (or whatever the focused/unfocused colors are) depending on whether the REPL is focused.  I like that you can still type in the REPL even when it is minimized so it makes sense to show whether it has focus.  Off the top of my head I don't remember how to do this but maybe you can figure out how we decide to draw the other windows depending on whether they are focused.

src/Swarm/TUI/Controller.hs
ussgarci: Ah, thank you.  I have changed it to Ctrl-s.  :crossed_fingers:

src/Swarm/TUI/View.hs
ussgarci: Thanks for the feedback.  I have removed the hard-coded cyan and am now indicating the focus properly.

src/Swarm/TUI/View.hs
```diff
@@ -334,6 +308,40 @@ drawGameUI s =
   fr = s ^. uiState . uiFocusRing
   moreTop = s ^. uiState . uiMoreInfoTop
   moreBot = s ^. uiState . uiMoreInfoBot
+  showREPL = s ^. uiState . uiShowREPL
+  rightPanel = if showREPL then worldPanel ++ replPanel else worldPanel ++ minimizedREPL
+  minimizedREPL = case focusGetCurrent fr of
+    (Just (FocusablePanel REPLPanel)) -> [separateBorders $ clickable (FocusablePanel REPLPanel) (forceAttr highlightAttr hBorder)]
+    _ -> [separateBorders $ clickable (FocusablePanel REPLPanel) hBorder]
```
byorgey, line 315: Nice touch making it still clickable.

src/Swarm/TUI/Controller.hs
```diff
@@ -320,6 +320,8 @@ handleMainEvent ev = do
     -- toggle creative mode if in "cheat mode"
     ControlChar 'v'
       | s ^. uiState . uiCheatMode -> gameState . creativeMode %= not
+    -- toggle collapse/expand REPL
+    ControlChar 's' -> uiState . uiShowREPL %= not
```
byorgey, line 324: We should add a call to `invalidateCacheEntry WorldCache` here, so it redraws the world instantly instead of waiting for the next time it decides to redraw.  The idea is that the world map is cached, so that we don't need to redraw it every frame if it hasn't changed.  This probably saves a bunch of cycles when nothing is going on, but it is also kind of a pain to deal with figuring out all the different places where the cache needs to be invalidated.

byorgey (approved):
This looks great!  The only other things that should probably happen before merging are (1) add the call to `invalidateCacheEntry WorldCache` as explained in a comment, and (2) fill out the description for the PR (the PR description is used as the commit description for the squashed commit created when the PR is merged).  Once you're happy with those things feel free to stick on the `merge me` label and the mergify bot will do the rest.

byorgey:
@ussgarci sorry if I was unclear, what I meant about the PR description is that it should say `closes #813` *somewhere* but it should also contain a sentence or two describing what change was made, e.g. "Add Ctrl-S shortcut for collapsing the REPL panel" or something like that.

byorgey:
It looks to me like the CI failures are spurious, once the PR description is updated I can restart those jobs and hopefully it will complete successfully.

ussgarci:
@byorgey I've updated the description.  Thank you!  :+1:

byorgey:
:tada:

PR #1078, opened 2023-02-01
allow `vty-5.38`

byorgey:
Tried with `cabal test --constraint='vty >= 5.38'` and all seemed to work fine.


byorgey:
The only other upper bound that's currently out of date is `mtl`, but some of our dependencies (*e.g.* `servant-server`) don't allow `mtl-2.3` so we can't build with that yet even if we allowed it.

xsebek (approved):

xsebek:
Cool, I did not know that cabal parameter. :+1:

PR #1080, opened 2023-02-04
Fix goal dependence for Ice Cream Shop

kostmo:
This scenario was written before prerequisites were merged.  This restores the goal dependency.


byorgey (approved):
Makes sense.  Are there any others like this, or did we get all the rest?

kostmo:
> Are there any others like this, or did we get all the rest?

This should be the only one.

PR #1081, opened 2023-02-04
Add tweezers for debugging

### LINKED ISSUE #5: Add a way to visualize a robot's currently executing program

xsebek:
- add `tweezers` device for debugging
- show robots CESK machine in `tweezers` description
  - add the view also to the pop-up of _entity with the capability_ - same for logs
- show the CESK machine in REPL when toggled (<kbd>Alt</kbd>+<kbd>D</kbd>)
- step through the CESK machine steps of the selected robot instead of ticks
- closes #5


xsebek:
Current view:

Screenshot from 2023-02-04 10-27-22

It's a bit redundant to have it in three places, but we will see what is most useful. :slightly_smiling_face: 

After this is done, I will try to take a stab at pretty printing in general so that long machines can be shown on multiple lines. Another improvement would be having a view of the current environment store.

src/Swarm/Game/State.hs
```diff
@@ -736,6 +744,7 @@ initGameState = do
   return $
     GameState
       { _creativeMode = False
+      , _gameStep = RobotStep LT
```
xsebek, line 747: @byorgey please give it a try - I set it to the robot step by default for testing. :slightly_smiling_face:

xsebek:
The code needs to be refactored and tweaked a little, but it works. :tm: For testing I used:
```
def circle = move; turn left; move; circle end
// to see how many steps other robots make
build {move; turn left; circle} 

def sum =\s.\i.\e. if (i>e) {s} {sum (s+i) (i+1) e} end
// pause and debug first
sum 0 1 10
```

The missing piece is turning on the single-step mode for a robot. :thinking: 

I will have to think about it a bit, but I am hoping to do it from UI in a style similar to a debugger.
Using the current `view`/focused robot approach is viable too, but I think it would be harder to switch between robots.

src/Swarm/Game/State.hs
byorgey: Hmm, is that what makes it so terribly slow?  I mean when you just let it run without being paused, for me it runs at about 0.5 ticks per second...

src/Swarm/Game/State.hs
```diff
@@ -347,6 +352,9 @@ let exclude = ['_viewCenter, '_focusedRobotID, '_viewCenterRule, '_activeRobots,
 -- | Is the user in creative mode (i.e. able to do anything without restriction)?
 creativeMode :: Lens' GameState Bool
 
+-- | Is the user in creative mode (i.e. able to do anything without restriction)?
```
byorgey, line 355: copy-pasto?

src/Swarm/Game/State.hs
```diff
@@ -269,11 +271,14 @@ makeLenses ''Notifications
 defaultRobotStepsPerTick :: Int
 defaultRobotStepsPerTick = 100
 
+data Step = WorldTick | RobotStep Ordering
```
byorgey, line 274: What is the `Ordering` for?

xsebek: Its to track whether we are running the focused robot (EQ) or robots before/after it.

Once the robots _after_ are finished the game tick is incremented.

src/Swarm/Game/State.hs
xsebek: Yes, maybe single stepping focused robot should not be allowed outside of paused mode. 😄

I did not add a switch to turn it on/off yet.

src/Swarm/Game/State.hs
byorgey: I see.  Please add a comment! :smile:

xsebek:
Actually, I changed my mind, using the focused robot is fine because `view` is a `Tangible Short` command.

Once `view` is run and the focus changes, we may need to run the robots between the previously focused robot and the newly focused robot. A type like this might help:
```Haskell
data SingleStep = SBefore | SSingle RID | SAfter RID
```

If `newRID > oldRID` then the `SAfter` means "after old before new".

One extra thing we have to handle is moving the view back to base in UI with <kbd>C</kbd> in the World panel.

---

Once I add a toggle for this mode*, this will be ready for review again.

*) This is another case where we have to finish stepping the robot and tick robots after it.

src/Swarm/Game/State.hs
xsebek: I added some documentation. :heavy_check_mark:

src/Swarm/Game/State.hs
xsebek: Well spotted, fixed. :heavy_check_mark:

src/Swarm/Game/Step.hs
```diff
@@ -122,24 +119,116 @@ gameTick = do
...
-          Nothing -> return ()
-        _otherREPLStatus -> return ()
-    Nothing -> return ()
+          Nothing -> pure ()
+        _otherREPLStatus -> pure ()
+    Nothing -> pure ()
 
-  -- Possibly update the view center.
-  modify recalcViewCenter
+  when ticked $ do
```
byorgey, line 126: Why do we only do this when `ticked` is true?  In practice, it seems like this means when single-stepping a robot that is executing `move`, I first see the robot move when the `move` command is executed, and then I have to hit Ctrl-O to step a few more times before the view center updates.  Is there a particular reason for doing it this way?

It's actually particularly confusing since the view center updates at the same time that all the other robots get a chance to move etc., which makes it hard to see what is happening.  I would rather have the view center update at the same time that the focused robot moves, then you can more easily see what all the other robots are doing.

Also, don't we want to check the win condition immediately when the currently focused robot does something, rather than waiting for the end of the tick?  Otherwise I imagine it would be confusing when single stepping your robot and it e.g. does something that achieves the objective but nothing happens until you do a few more steps.

src/Swarm/TUI/Controller.hs
```diff
@@ -636,8 +649,8 @@ updateAchievements = do
 --   etc.).
 runGameTick :: EventM Name AppState ()
 runGameTick = do
-  zoomGameState gameTick
-  updateAchievements
+  ticked <- zoomGameState gameTick
+  when ticked updateAchievements
```
byorgey, line 653: Why do this only when `ticked` is true?  Shouldn't we update the achievements the instant one is achieved, even if we're in the middle of stepping the focused robot?

src/Swarm/TUI/View.hs
```diff
@@ -1097,6 +1098,20 @@ drawRobotLog s =
...
       drawLogEntry (not allMe) e
 
+-- | Show the 'CESK' machine of focused robot. Puts a separator above.
+drawRobotMachine :: AppState -> Bool -> Widget Name
+drawRobotMachine s showName = case s ^. gameState . to focusedRobot of
+  Nothing -> machineLine "no selected robot"
+  Just r ->
+    vBox
+      [ machineLine $ r ^. robotName <> "#" <> r ^. robotID . to tshow
+      , txt $ r ^. machine . to prettyText
```
byorgey, line 1108: This doesn't wrap at all if the pretty-printed machine is very long, right?  How hard would it be to do that?  (Off the top of my head I forget.)

src/Swarm/Game/State.hs
```diff
@@ -269,11 +272,35 @@ makeLenses ''Notifications
...
+--     │              ┌────┐                │
+--     │              ▼    │                │
+-- ┌───┴───┐ step  ┌───────┴───┐ step  ┌────┴─────┐
+-- │SBefore├──────►│SSingle RID├──────►│SAfter RID│
+-- └───────┘       └───────────┘       └────┬─────┘
+--                      ▲                   │
+--                      └───────────────────┘
+--                         view RID > oldRID
+-- @
+data SingleStep = SBefore | SSingle RID | SAfter RID
```
byorgey, line 293: This is better but I am still having trouble understanding exactly what this means, I think it still needs a lot more comments.
- Does the `SingleStep` value tell us which robot(s) should be stepped *next*, or which ones we just finished stepping?
- Why do `SSingle` and `SAfter` contain an `RID` but `SBefore` does not?  What do those `RID` values mean?
- Why is it only safe to transition to `WorldTick` from `SBefore`?

byorgey (commented):
I've looked at almost everything but ran out of steam when trying to carefully read through the `singleStep` function.  I'll look at it soon.

src/Swarm/Game/Step.hs
xsebek: The view centre could be recalculated at every step, I will fix that.

Win condition **should not** be checked on every step because:
1. It is slow.
2. That is not how it works for normal time mode, defeating the purpose of debugging.

src/Swarm/TUI/Controller.hs
```diff
@@ -301,6 +306,12 @@ handleMainEvent ev = do
         do
           uiState . uiHideRobotsUntil .= t + TimeSpec 2 0
           invalidateCacheEntry WorldCache
+    -- debug focused robot
+    MetaChar 'd' | isPaused && hasDebug -> do
```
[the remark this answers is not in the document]
xsebek: It does not really work well with the game unpaused IMO. The focus jumps around too quickly and it is hard to make sense of it.

I will turn it off along with the single-step mode since they use the same toggle (at least for now).

src/Swarm/TUI/Controller.hs
xsebek: The same reasoning as with the objective evaluation, it takes time and should not produce different results.

Imagine there is a robot that goes after you that stops you from what you are trying to achieve. It must have its turn, otherwise, you could briefly win in single-step mode, but never in normal mode.

src/Swarm/TUI/View.hs
```diff
@@ -776,7 +777,8 @@ drawKeyMenu s =
       [ may goal (NoHighlight, "^g", "goal")
       , may cheat (NoHighlight, "^v", "creative")
       , Just (NoHighlight, "^p", if isPaused then "unpause" else "pause")
-      , Just (NoHighlight, "^o", "step")
+      , may isPaused (NoHighlight, "^o", "step")
+      , may (isPaused && hasDebug) (if s ^. uiState . uiShowDebug then Alert else NoHighlight, "M-d", "debug")
```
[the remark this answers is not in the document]
xsebek: I will split an Issue for a more customized debugging selection.

I do not see the value of debugging outside of the paused mode. Imagine viewing the Python debugger in real time, makes me dizzy just thinking about it. :dizzy_face: But if that is what the player/developer really meant (based on a more nuanced debugging selection) then sure, we can enable it. :slightly_smiling_face:

src/Swarm/TUI/View.hs
xsebek: Well... the REPL panel has a fixed size and this is part of it.

@byorgey do you mean that the REPL would have at least two lines, but would greedily fill the panel unless the Machine view takes it for itself?

I think we should improve the pretty printing of code in general so that it wraps nicely to multiple lines and we can know how much space is needed. If parts of the text are widgets (with colours) then we can also focus (with brick or manually) on the current expression, so it will always be in view.

xsebek:
> I've looked at almost everything but ran out of steam when trying to carefully read through the `singleStep` function. I'll look at it soon.

@byorgey I (hope I) fixed the issues you found, please take a look at the `singleStep` function when you have time. :wink: 

There are just enough small changes in all the branches that I do not see an easy way to refactor it, but maybe you will find a way. :slightly_smiling_face:

src/Swarm/Game/Step.hs
byorgey: Yes, I see, that makes sense.

src/Swarm/TUI/View.hs
byorgey: Good point, I forgot the REPL has a fixed size.  I don't know what I mean exactly, I just know that currently, if the CESK machine gets very big then it will just go off the right edge, and we should do something to improve it.

We should definitely improve the pretty printing of code in general, although that will not help with this particular issue since pretty-printing CESK machines and pretty-printing swarm-lang code use completely different code paths (for good reason, even though we try to use similar pretty-printed syntax as much as possible).

src/Swarm/Game/Step.hs
```diff
@@ -122,24 +119,114 @@ gameTick = do
...
+      ticks += 1
+      return True
+    SAfter rid | otherwise -> do
+      let postRID = snd . IS.split rid $ robotSet
+      singleStep SBefore rid postRID
+ where
+  h = hypotheticalRobot (Out VUnit emptyStore []) 0
+  debugLog txt = do
+    m <- evalState @Robot h $ createLogEntry (ErrorTrace Debug) txt
+    emitMessage m
```
byorgey, line 229: OK, I have carefully read through `singleStep` now.  I *think* it makes sense, but it is still very complicated and I am afraid it will end up being the source of many bugs with debugging mode.  I think we should still add a lot more inline comments.

byorgey (approved):
Overall this is very nice, and I suspect it will come in very handy while developing scenarios.  I think there's still more that could be improved but feel free to split things out into issues to deal with later.

src/Swarm/Game/Step.hs
xsebek: Bugs in debugging function could be annoying but at least will not bother players. 😄

I think we could unit test this to make sure it does the right thing.

Another idea I had was to run the entire integration test suite in single step mode which should produce the same result with raised timeouts.

byorgey: > Bugs in debugging function could be annoying but at least will not bother players.

Why not?  Players can use the debugger too. This is a programming game after all. :smile: 



> Another idea I had was to run the entire integration test suite in single step mode which should produce the same result with raised timeouts.

That sounds like a great idea!  Though probably we shouldn't do that as part of the normal CI test suite since it might be very slow.  But if the test suite passes in single-step mode then that would be a pretty good signal it is correct.  (Although it probably wouldn't exercise corner cases like e.g. running the `view` command so the focused robot changes in the middle of single-stepping.)

xsebek: @byorgey actually this was easy:

```diff
                   & baseRobot . robotContext . defReqs <>~ reqCtx
                   & baseRobot . machine .~ initMachine sol Ctx.empty emptyStore
+                  & gameStep .~ RobotStep SBefore
-          m <- timeout (time s) (snd <$> runStateT playUntilWin gs')
+          m <- timeout (-1) (snd <$> runStateT playUntilWin gs')
           case m of
```

On CI the normal tests take 52s and for me, they took 116.09s in single-step mode and all passed. :heavy_check_mark:

src/Swarm/Game/Step.hs
```diff
@@ -122,24 +119,114 @@ gameTick = do
...
+      mOldR <- uses robotMap (IM.lookup focRID)
+      case mOldR of
+        Nothing -> do
+          debugLog "The previously debugged robot does not exist!"
+          singleStep (SAfter rid) focRID postFoc
+        Just oldR -> do
+          newR <- tickRobotRec oldR
+          insertBackRobot focRID newR
+          when (newR ^. tickSteps == 0) $ gameStep .= RobotStep (SAfter focRID)
+          singleStep ss rid postFoc
```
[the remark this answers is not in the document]
xsebek: Done! It does not combine too much but at least highlights the differences. :slightly_smiling_face:

src/Swarm/Game/Step.hs
```diff
@@ -122,24 +119,114 @@ gameTick = do
...
+    ----------------------------------------------------------------------------
+    -- run robots from the beginning until focused robot
+    SBefore -> do
+      runRobotIDs preFoc
+      gameStep .= RobotStep (SSingle focRID)
+      -- also set ticks of focused robot
+      mOldR <- use $ robotMap . at focRID
+      steps <- use robotStepsPerTick
+      forM_ mOldR $ \r -> robotMap %= IM.insert focRID (r & tickSteps .~ steps)
+      return False
```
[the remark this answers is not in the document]
xsebek: Done!

src/Swarm/Game/Step.hs
```diff
@@ -122,24 +119,114 @@ gameTick = do
...
+ where
+  stepOneRobot rn rob = tickRobot rob >>= insertBackRobot rn
+
+singleStep :: (Has (State GameState) sig m, Has (Lift IO) sig m) => SingleStep -> RID -> IS.IntSet -> m Bool
+singleStep ss focRID robotSet = do
+  let (preFoc, focusedActive, postFoc) = IS.splitMember focRID robotSet
+  case ss of
+    ----------------------------------------------------------------------------
+    -- run robots from the beginning until focused robot
+    SBefore -> do
```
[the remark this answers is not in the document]
xsebek: Yes, I tried to use `SBefore | null preFoc` but that skipped robot initialisation. :sweat_smile: 

I went with your other suggestion.

PR #1085, opened 2023-02-05
Parse --run code instead of wrapping with run() command

kostmo:
This change supports accurate code size measurement for #974 by preventing the AST from being obfuscated behind a `run()` command when a program is supplied via the `--run` CLI argument.

For historical interest, the `--run` command was introduced in #312.


src/Swarm/Util/RunCode.hs
```diff
@@ -0,0 +1,28 @@
...
+import Data.Text.IO qualified as TIO
+import Swarm.Language.Pipeline (ProcessedTerm, processTermEither)
+
+data SolutionSource
+  = ScenarioSuggested
+  | PlayerAuthored
+
+data CodeToRun = CodeToRun SolutionSource ProcessedTerm
+
+getParsedInitialCode :: Maybe FilePath -> ExceptT Text IO (Maybe CodeToRun)
```
xsebek, line 16: This function can not return ScenarioSuggested solution.

Please do not create Util modules for things that belong somewhere.

When you want to parse a term, please put that function next to other functions that do so.

When you create a data type used in scenarios and initialising game state, please put it near game state.

xsebek (changes requested):
I completely agree with parsing the file first. 👍

Please just remove the nondescript Util module.

src/Swarm/Language/Pipeline.hs
```diff
@@ -50,14 +51,19 @@ data ProcessedTerm
...
+
 instance FromJSON ProcessedTerm where
-  parseJSON = withText "Term" tryProcess
-   where
-    tryProcess :: Text -> Y.Parser ProcessedTerm
-    tryProcess t = case processTerm t of
-      Left err -> fail $ "Could not parse term: " ++ from err
-      Right Nothing -> fail "Term was only whitespace"
-      Right (Just pt) -> return pt
+  parseJSON = withText "Term" $ tryProcess . processTermEither
```
byorgey, line 66:
```suggestion
  parseJSON = withText "Term" $ either fail return . processTermEither
```

src/Swarm/Util/RunCode.hs
byorgey: I agree, the `Util` namespace should be used only for things that get used from many unrelated places in the code.

byorgey (commented):
Makes sense to me, especially since we eventually want to get rid of the `run` command.

kostmo:
Comments now addressed.  Moved the new code into `State.hs`.  Also am now recording the SHA1 of the program text for use in #974.

src/Swarm/Game/State.hs
```diff
@@ -260,6 +261,32 @@ instance Monoid (Notifications a) where
...
+    PlayerAuthored String
+
+data CodeToRun = CodeToRun SolutionSource ProcessedTerm
+
+getParsedInitialCode :: Maybe FilePath -> ExceptT Text IO (Maybe CodeToRun)
+getParsedInitialCode toRun = case toRun of
+  Nothing -> return Nothing
+  Just filepath -> do
+    contents <- liftIO $ TIO.readFile filepath
+    pt@(ProcessedTerm (Module (Syntax' srcLoc _ _) _) _ _) <-
```
xsebek, line 276: Why not use a `let` with lenses if you need that `srcLoc`?

src/Swarm/Game/State.hs
```diff
@@ -260,6 +261,32 @@ instance Monoid (Notifications a) where
 
 makeLenses ''Notifications
 
+data SolutionSource
+  = ScenarioSuggested
+  | -- | Includes the SHA1 of the program text
+    PlayerAuthored String
```
xsebek, line 267: The `String` type certainly does not suggest a hash and the motivation for it is missing. :raised_eyebrow: 

Why do you need a hash of the original text?

xsebek (approved):
Looks good to me now. :slightly_smiling_face:

But please wait for @byorgey too, as he was the author of the original `--run` and knows the code around it better.

src/Swarm/Game/State.hs
kostmo: What would the lens expression look like?

src/Swarm/Game/State.hs
kostmo: Added a newtype wrapper and a bit of explanation.

Admittedly the idea is not fully fleshed out yet, but if we have some kind of public leaderboard, we could publish the "fingerprint" of someone's solution without actually exposing the program text.  Out-of-band, a "judge" could verify that a supplied program text does hash to the leaderboard score.

Or the Sha1 can just be proof to yourself of which file on your local disk you used to achieve that score.

src/Swarm/Game/State.hs
```diff
@@ -260,6 +262,36 @@ instance Monoid (Notifications a) where
...
+  Just filepath -> do
+    contents <- liftIO $ TIO.readFile filepath
+    pt@(ProcessedTerm (Module (Syntax' srcLoc _ _) _) _ _) <-
+      ExceptT $
+        return $
+          left T.pack $
+            processTermEither contents
+    let strippedText = stripSrc srcLoc contents
+        programBytestring = TL.encodeUtf8 $ TL.fromStrict strippedText
+        sha1Hash = showDigest $ sha1 programBytestring
```
byorgey, line 288: If we are going to record hashes, I think we should hash some canonical representation of a program rather than the original source. 
* For example, at a basic level, we could first format the code by round-tripping through parsing + pretty-printing before hashing, or we could directly hash the AST.  That way, modifying the whitespace in your program would not change the hash.
* To make things slightly more interesting we could also do something like canonically rename all variables, so that renaming something in your program would not change the hash either.

However, I would be fine with leaving this to a future PR.

byorgey (approved):

PR #1090, opened 2023-02-06
lower max bounds on yaml package

kostmo:
Observed a CI breakage seemingly due to new `yaml` package release:
https://github.com/swarm-game/swarm/actions/runs/4107732825/jobs/7087565682

Also tested in #1091.

`yaml` release PR for reference:
https://github.com/snoyberg/yaml/pull/206


byorgey (approved):
Sure, this is a good workaround for now.  To be clear, this is a problem with the `yaml` package itself.  See this issue: https://github.com/snoyberg/yaml/issues/207

PR #1100, opened 2023-02-09
Use internal attribute type in Display

xsebek:
This avoids using `Brick.AttrName` and we can remove the `-fno-warn-orphans` from `Attr.hs`.

As a bonus, we can now specify a terrain attribute (`terrain_stone`) in a scenario along with the `default` attribute.

- split off from #1069
- part of #1043


byorgey (approved):
LGTM!

kostmo (approved):

PR #1103, opened 2023-02-11
Move Heading and conversions to Location module

xsebek:
- remove the useless `DirInfo` type
- move the `Direction -> Heading` functions to the `Location` module
- split off from #1069
- part of #1043
- part of #707

### DIFF
--- a/src/Swarm/Game/Location.hs
+++ b/src/Swarm/Game/Location.hs
@@ -43,7 +43,7 @@ import Data.Int (Int32)
 import Data.Map (Map)
 import Data.Map qualified as M
 import Data.Yaml (FromJSON (parseJSON), ToJSON (toJSON))
-import Linear (Additive (..), V2 (..), perp)
+import Linear (Additive (..), V2 (..), negated, perp)
 import Linear.Affine (Affine (..), Point (..), origin)
 import Swarm.Language.Syntax (AbsoluteDir (..), Direction (..), RelativeDir (..), isCardinal)
 import Swarm.Util qualified as Util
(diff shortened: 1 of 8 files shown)


src/Swarm/Game/Location.hs
```diff
@@ -71,11 +90,68 @@ deriving instance FromJSON (V2 Int32)
...
+  DRelative e -> case e of
+    DLeft -> relative perp
+    DRight -> relative (fmap negate . perp)
+    DBack -> relative (fmap negate)
+    DDown -> relative (const down)
+    DForward -> relative id
+  DAbsolute e -> cardinal $ toHeading e
+ where
+  cardinal = const
+  relative = id
```
byorgey, line 134: What is the point of defining `relative = id`?  Why not just have `DLeft -> perp` and so on?  (Maybe this code was already like that but I didn't notice before.)

src/Swarm/Language/Syntax.hs
```diff
@@ -145,93 +133,21 @@ data RelativeDir = DLeft | DRight | DBack | DForward | DDown
...
- where
-  -- name is generate from Direction data constuctor
-  -- e.g. DLeft becomes "left"
-  directionSyntax = toLower . T.tail . from $ case d of
-    DAbsolute x -> show x
-    DRelative x -> show x
-
-  cardinal = DirInfo directionSyntax . const
-  relative = DirInfo directionSyntax
+-- | Direction name is generate from Direction data constuctor
```
byorgey, line 136:
```suggestion
-- | Direction name is generated from Direction data constuctor
```

byorgey (approved):

src/Swarm/Game/Location.hs
```diff
@@ -71,11 +90,68 @@ deriving instance FromJSON (V2 Int32)
...
+down = zero
+
+-- | The 'applyTurn' function gives the meaning of each 'Direction' by
+--   turning relative to the given heading or by turning to an absolute
+--   heading
+applyTurn :: Direction -> Heading -> Heading
+applyTurn d = case d of
+  DRelative e -> case e of
+    DLeft -> relative perp
+    DRight -> relative (fmap negate . perp)
```
kostmo, line 127: I've been meaning to replace `fmap negate` with just [`negated`](https://hackage.haskell.org/package/linear-1.22/docs/Linear-Vector.html#v:negated)

kostmo (approved):

src/Swarm/Game/Location.hs
xsebek: Neat, thanks! :+1:

src/Swarm/Game/Location.hs
xsebek: No good reason, I think we originally used it for the absolute/relative distinction, but that has disappeared with the direction constructors.

PR #1106, opened 2023-02-13
Parse markdown descriptions

### LINKED ISSUE #309: Check code examples in entity descriptions as part of CI
When making updates for #303 , we almost missed the fact that we needed to update some code examples in entity descriptions (@noahyor caught it).  When updating the entity descriptions, I caught several other things that needed to be updated which had been out of date for a while.

Code samples in entity descriptions ought to be automatically checked in our CI process (by parsing and typechecking them), just like code in the `examples/` directory is checked.

I am not quite sure how this would work.  There would need to be some special way to mark code examples in entity descriptions to make it possible to extract them during the CI process. Then we would have to process them specially when displaying entity descriptions.  Maybe it's as simple as inventing some special delimiters, then stripping those delimiters when displaying entity descriptions.

### LINKED ISSUE #545: Highlight commands in descriptions

xsebek:
* use CommonMark to parse simple markdown AST parametrised on inline/block code
* validate swarm code (`Document Text -> Document Syntax`)
* update descriptions to use markdown with following conventions:
  - `move` - valid swarm code (the easy to write default)
  - `wedge`{=entity} - for swarm entities
  - `unit`{=type} - for swarm types
  - `require <a> <b>`{=snippet} - raw snippets for invalid code
  - **Alt-G** - bold for keyboard shortcuts
- highlight code in brick widgets
- closes #309
- closes #545
- precedes #574
- precedes #1406
- precedes #1407


src/Data/Text/Markdown.hs
```diff
@@ -0,0 +1,121 @@
...
+  CMark.TEXT txt -> leaf $ LeafText txt
+  CMark.CODE txt -> leaf' $ LeafCode <$> parse txt
+  CMark.CODE_BLOCK i txt -> leaf' $ LeafCodeBlock (T.unpack i) <$> parse txt
+  _ -> Left $ "unsupported markdown node at " <> maybe "???" show p <> ":" <> show t
+ where
+  parse txt =
+    readTerm txt & \case
+      Left e -> Left (T.unpack e)
+      Right Nothing -> Left "empty code"
+      Right (Just s) -> case processParsedTerm s of
```
xsebek, line 71: We currently use inline code for types like `dir -> cmd unit`.

I think I can solve the type signatures as an alternative parse.

src/Data/Text/Markdown.hs
```diff
@@ -0,0 +1,121 @@
...
+-- TODO: use transformer kind of Error? (report all errors)
+fromGeneralNode :: CMark.Node -> Either String Node
+fromGeneralNode (CMark.Node p t cs) = case t of
+  CMark.DOCUMENT -> Node Document <$> mapM fromGeneralNode cs
+  CMark.PARAGRAPH -> Node Paragraph <$> mapM fromGeneralNode cs
+  CMark.SOFTBREAK -> leaf $ LeafText "\n" -- ?
+  CMark.LINEBREAK -> leaf $ LeafText "\n"
+  CMark.TEXT txt -> leaf $ LeafText txt
+  CMark.CODE txt -> leaf' $ LeafCode <$> parse txt
+  CMark.CODE_BLOCK i txt -> leaf' $ LeafCodeBlock (T.unpack i) <$> parse txt
```
xsebek, line 64: A small problem is that `logger` is a valid swarm code, but so is almost any word.

To check that our descriptions are up to date, we should not allow unbound variables.

To get reasonable highlighting, I think we can use _italics_ and **bold** text. Converting those from markdown to a stream of text will be a bit harder, but we would probably want to do it at some point anyway.

src/Data/Text/Markdown.hs
```diff
@@ -0,0 +1,121 @@
...
+fromText :: Text -> Node
+fromText = either error id . fromTextE
+
+fromTextM :: MonadFail m => Text -> m Node
+fromTextM = either fail pure . fromTextE
+
+fromTextE :: Text -> Either String Node
+fromTextE = fromGeneralNode . CMark.commonmarkToNode [CMark.optSourcePos] []
+
+-- TODO: use transformer kind of Error? (report all errors)
```
xsebek, line 55: @byorgey I noticed we do not depend on `transformers`, do you know of any type like [`Control.Applicative.Lift.Errors`](https://hackage.haskell.org/package/transformers-0.6.1.0/docs/Control-Applicative-Lift.html#t:Errors) that we could use?

It would be neat to show more errors at once where possible. :slightly_smiling_face:

src/Swarm/Game/Scenario/Objective.hs
```diff
@@ -113,7 +113,7 @@ objectiveAchievement :: Lens' Objective (Maybe AchievementInfo)
 instance FromJSON Objective where
   parseJSON = withObject "objective" $ \v ->
     Objective
-      <$> (fmap . map) reflow (v .:? "goal" .!= [])
+      <$> (v .:? "goal" .!= [])
```
xsebek, line 116: When I tried to use `Node` instead of `[Node]` no errors were shown in integration tests.

I think it is because of laziness - a list always evaluates to `Node Document`.

@byorgey how should I get around it? I suppose something like `deepseq` could work or I could make the entire `Node` strict. :thinking:

src/Swarm/TUI/View/Objective.hs
```diff
@@ -84,11 +85,11 @@ drawGoalListItem _isSelected e = case e of
...
   Goal gs obj -> getCompletionIcon obj gs <+> titleWidget
    where
-    textSource = obj ^. objectiveTeaser <|> obj ^. objectiveId <|> listToMaybe (obj ^. objectiveGoal)
+    textSource = obj ^. objectiveTeaser <|> obj ^. objectiveId <|> listToMaybe (Markdown.toText <$> obj ^. objectiveGoal)
     titleWidget = maybe (txt "?") withEllipsis textSource
 
 singleGoalDetails :: GoalEntry -> Widget Name
 singleGoalDetails = \case
-  Goal _gs obj -> displayParagraphs $ obj ^. objectiveGoal
+  Goal _gs obj -> displayParagraphs . map Markdown.toText $ obj ^. objectiveGoal
```
xsebek, line 93: This needs to change to properly support the whole markdown text (with code) instead of a list of paragraphs.

I suppose it might make sense to use `[[StreamNode]]` as paragraphs so that we can add padding.

xsebek (commented):
@byorgey its only in a MVP stage right now, but it already caught this inaccuracy:
```
    parse scenario "data/scenarios/Tutorials/requireinv.yaml":  FAIL
      test/integration/Main.hs:115:
      Aeson exception:
      Error in $.objectives[0].goal[0]: 1:8:
        |
      1 | require
        |        ^
      reserved word 'require' cannot be used as variable name
```

Because `require` is not a command we can not use it on its own. :smile:

src/Data/Text/Markdown.hs
kostmo: I vote to add `transformers` to our dependencies.  We depend on `mtl` which has a direct dependence on `transformers` anyway.

byorgey: Agreed, there's no reason we can't add a direct dependency on `transformers`.

src/Data/Text/Markdown.hs
```diff
@@ -0,0 +1,121 @@
...
+  deriving (Eq, Show)
+
+instance ToJSON Node where
+  toJSON = String . toText
+
+instance FromJSON Node where
+  parseJSON v =
+    withArray "markdown paragraphs" (foldMap parseJSON) v
+      <|> withText "markdown text" fromTextM v
+instance Semigroup Node where
```
byorgey, line 37: What's the thinking behind this `Semigroup` instance?

I *think* it's associative (I can't come up with an examples where associativity fails), do you have a proof?

src/Data/Text/Markdown.hs
```diff
@@ -0,0 +1,121 @@
...
+  Node Document cs1 <> Node Document cs2 = Node Document (cs1 <> cs2)
+  Node Document cs1 <> n2 = Node Document (cs1 <> [n2])
+  n1 <> Node Document cs2 = Node Document (n1 : cs2)
+  n1 <> n2 = Node Document [n1, n2]
+
+instance Monoid Node where
+  mempty = Node Document []
```
byorgey, line 44: This is not actually an identity element.  For example `LeafText "t" <> mempty = Node Document [LeafText "t"] /= LeafText "t"`.  I'm not sure how much this matters.  I suppose it is an identity if we consider `Node Document [x] = x`.

src/Data/Text/Markdown.hs
```diff
@@ -0,0 +1,121 @@
...
+
+fromTextE :: Text -> Either String Node
+fromTextE = fromGeneralNode . CMark.commonmarkToNode [CMark.optSourcePos] []
+
+-- TODO: use transformer kind of Error? (report all errors)
+fromGeneralNode :: CMark.Node -> Either String Node
+fromGeneralNode (CMark.Node p t cs) = case t of
+  CMark.DOCUMENT -> Node Document <$> mapM fromGeneralNode cs
+  CMark.PARAGRAPH -> Node Paragraph <$> mapM fromGeneralNode cs
+  CMark.SOFTBREAK -> leaf $ LeafText "\n" -- ?
```
byorgey, line 60: According to https://github.github.com/gfm/#soft-line-breaks it seems like a `SOFTBREAK` is what we get at the end of a physical line of parsed text which does not end with two or more spaces.  Typically we ignore such breaks and just reflow the text, so I would think we should not generate a `LeafText "\n"` here, unless I am misunderstanding something.

src/Data/Text/Markdown.hs
byorgey: Yeah, we can make a parallel to `readTerm` for types, like `readType = runParser (fully parseType)`.  If `readTerm` fails you can just try running `readType` too.

src/Swarm/Game/Scenario/Objective.hs
byorgey: I'm not sure I understand the problem.  What do you mean by "use `Node` instead of `[Node]`"?

src/Data/Text/Markdown.hs
xsebek: Thanks, @kostmo, I will try that. :+1:

src/Data/Text/Markdown.hs
xsebek: I think we might need to use a space instead of newline here.

src/Data/Text/Markdown.hs
xsebek: I think all parsed markdown will have document as the root node.

Maybe I won’t need the instance in the end, but it could be handy for preserving the paragraph lists syntax in scenarios.

src/Swarm/Game/Scenario/Objective.hs
xsebek: The whole description should be one document that we render.

Right now it is a list of documents, even though markdown supports paragraphs.

xsebek: Using `Node` only should parse the document as JSON array (see the ToJSON instance that uses Semigroup)

byorgey: OK, I see now why we want to use `Node` instead of `[Node]`, and why that should work.  And just to make sure I'm on the same page, right now if you build the code as is, we get a bunch of errors, as we should.  But you're saying that if you use `Node` instead of `[Node]` it doesn't generate any errors?  I don't understand what laziness would have to do with that.

xsebek: Maybe the error is not evaluated, but it may also be that I have my parser function wrong. :shrug: 

I will try to get back to this tomorrow.

xsebek:
Huzzah! 🥳 

Screenshot 2023-08-03 at 3 04 15

Turns out CommonMark has a neat extension that allows me to annotate inline code as entities/types/invalid code snippets.
Took me a while to rewrite, but at least I don't have to study a C library to figure out how it works. 😅 

@kostmo If it is OK, I would switch from CMark-GFM to CommonMark. AFAIK it was only used in one place and the change was simple one-liner.

src/Data/Text/Markdown.hs
xsebek: Actually, I may not want to do this here after all. Whenever I change code here, it impacts a lot of scenarios and they are reported per file already. At the end, Aeson does the parsing and error reporting, so there is not much I can do.

src/Swarm/Game/Scenario/Objective.hs
xsebek: I simplified the types and this problem went away. 🤷

src/Data/Text/Markdown.hs
xsebek: The types should make more sense now. 🙂 There is still some magic around nesting, but it stays at two levels only (document and paragraph).

src/Data/Text/Markdown.hs
xsebek: Note to self: check that spacing is preserved.

src/Swarm/Doc/Pedagogy.hs
```diff
@@ -157,7 +161,10 @@ generateIntroductionsSequence =
 loadScenarioCollection :: IO ScenarioCollection
 loadScenarioCollection = simpleErrorHandle $ do
   entities <- ExceptT loadEntities
-  (_, loadedScenarios) <- liftIO $ loadScenariosWithWarnings entities
+  (failures, loadedScenarios) <- liftIO $ loadScenariosWithWarnings entities
+  when (notNull failures) $
+    liftIO $
+      hPutStrLn stderr "Loading failures: " >> mapM_ (T.putStrLn . prettyFailure) failures
```
kostmo, line 167: Should the "Loading failures:" heading and the individual error lines both be sent to `stderr`?

xsebek: Yes, well spotted. 👍 The pedagogy output was pretty long, so I was thinking about redirecting `stdout` to `/dev/null` so I could see the failures only. But by then I fixed the failures. 😄

xsebek:
@byorgey @kostmo I rewrote the implementation from CMark-GFM to CommonMark and while it is a bit of a DIY project, I hope it will make reading descriptions a little easier. 🙂

For reference, the markdown highlighting conventions I went with are:
- `move` - valid swarm code (the easy to write default)
- `wedge`{=entity} - for swarm entities (I generated a regex to catch all standard ones, but they are not validated)
- `unit`{=type} - for swarm types (there are not so many and I do not validate them yet)
- `require <a> <b>`{=snippet} - raw snippets for invalid code
- **Alt-G** - bold for keyboard shortcuts

If we want to do something else with any of those cases, they should be easy to grep and replace.

data/entities.yaml
```diff
@@ -955,13 +955,13 @@
...
   - |
     A dictionary allows a robot to remember definitions and reuse them
-    later.  You can access this ability with either a `def` command,
+    later.  You can access this ability with either a `def`{=snippet} command,
     which creates a name for an expression or command that is
-    available from then on, or with a `let` expression, which names an
+    available from then on, or with a `let`{=snippet} expression, which names an
     expression or command locally within another expression.
   - '  def m2 : cmd unit = move; move end'
   - '  let x : int = 3 in x^2 + 2*x + 1'
```
xsebek, line 963: It seems some block expressions are not converted yet. 👀

src/Data/Text/Markdown.hs
```diff
@@ -0,0 +1,268 @@
...
+import Data.Text (Text)
+import Data.Text qualified as T
+import Data.Tuple.Extra (both, first)
+import Data.Vector (toList)
+import Data.Yaml
+import Swarm.Language.Module (moduleAST)
+import Swarm.Language.Parse (readTerm)
+import Swarm.Language.Pipeline (ProcessedTerm (..), processParsedTerm)
+import Swarm.Language.Pretty (prettyText, prettyTypeErrText)
+import Swarm.Language.Syntax (Syntax)
```
xsebek, line 43: Depending on Swarm in Data module is a hack. 🤷  Suggestions welcome for better placement or architecture.

src/Swarm/Doc/Pedagogy.hs
```diff
@@ -87,16 +90,16 @@ extractCommandUsages idx siPair@(s, _si) =
...
   goalTextParagraphs = concatMap (view objectiveGoal) $ view scenarioObjectives s
-  allWords = concatMap (T.words . T.toLower) goalTextParagraphs
-  getBackticked = T.stripPrefix "`" <=< T.stripSuffix "`"
-  backtickedWords = mapMaybe getBackticked allWords
-
-  commandConsts = filter isConsidered allConst
-  txtLookups = M.fromList $ map (syntax . constInfo &&& id) commandConsts
+  allCode = concatMap findCode goalTextParagraphs
+  filterConst :: Syntax -> [Const]
+  filterConst sx = mapMaybe toConst $ universe (sx ^. sTerm)
```
xsebek, line 98: @kostmo this is one of the little nice things a parsed code gives us. 🤩 

Hopefully `universe` works just as well as your original implementation, but please check that I did not break it. 😅

src/Swarm/Game/Failure.hs
```diff
@@ -12,12 +12,13 @@ import Data.Yaml (ParseException)
 
 data SystemFailure
   = AssetNotLoaded Asset FilePath LoadingFailure
+  deriving (Show)
```
xsebek, line 15: Took me a while to find how this is pretty printed, so I added `Show` instance while debugging.

Should the printing function be here or at least a link to it?

src/Swarm/TUI/View/Objective.hs
```diff
@@ -87,11 +88,11 @@ drawGoalListItem _isSelected e = case e of
...
   Goal gs obj -> getCompletionIcon obj gs <+> titleWidget
    where
-    textSource = obj ^. objectiveTeaser <|> obj ^. objectiveId <|> listToMaybe (obj ^. objectiveGoal)
+    textSource = obj ^. objectiveTeaser <|> obj ^. objectiveId <|> listToMaybe (Markdown.toText <$> obj ^. objectiveGoal)
     titleWidget = maybe (txt "?") (withEllipsis End) textSource
 
 singleGoalDetails :: GoalEntry -> Widget Name
 singleGoalDetails = \case
-  Goal _gs obj -> displayParagraphs $ obj ^. objectiveGoal
+  Goal _gs obj -> vBox . map (padBottom (Pad 1) . drawMarkdown) $ obj ^. objectiveGoal
```
xsebek, line 96: I had to inline `displayParagraphs` because it took text instead of widgets. Maybe this can be unified. 🤔

swarm.cabal
```diff
@@ -200,8 +201,9 @@ library
                       boolexpr                      >= 0.2 && < 0.3,
                       brick                         >= 1.5 && < 1.10,
                       bytestring                    >= 0.10 && < 0.12,
+                      commonmark,
+                      commonmark-extensions,
```
xsebek, line 205: - [ ] When this is done, I should add proper bounds.

src/Swarm/Game/Failure.hs
byorgey: Adding more documentation is always a good thing!

byorgey:
Nice! I'm working on a detailed review, hopefully finished later today.

data/scenarios/Tutorials/craft.yaml
```diff
@@ -9,12 +9,12 @@ description: |
...
       - Robots can use the `make` command to make things, as long as
-        they have a `workbench` and the proper ingredients.  For
+        they have a `workbench`{=entity} and the proper ingredients.  For
         example, `make "circuit"` will make a circuit.
       - Your base has a few trees in its inventory. Select them to see the
         available recipes.
-      - Your goal is to make a "branch predictor", so you will have to make
-        some "branch"es first.
+      - Your goal is to make a `branch predictor`{=entity}, so you will have to make
+        some `"branch"`es first.
```
byorgey, line 17: Why not
```suggestion
        some `branch`{=entity}es first.
```
?

src/Data/Text/Markdown.hs
```diff
@@ -0,0 +1,268 @@
...
+unStream = \case
+  TextNode a t -> (TextNode a, t)
+  CodeNode t -> (CodeNode, t)
+  RawNode a t -> (RawNode a, t)
+  ParagraphBreak -> error "Logic error: Paragraph break can not be unstreamed!"
+
+-- | Get chunks of nodes not exceeding length and broken at word boundary.
+--
+-- The split will end when no more nodes (then words) can fit or on 'ParagraphBreak'.
+chunksOf :: Int -> [StreamNode] -> [[StreamNode]]
```
byorgey, line 205: It is a shame we seem to be reimplementing some of the functionality of https://hackage.haskell.org/package/word%2Dwrap ...

byorgey (approved):

data/entities.yaml
```diff
@@ -343,7 +343,7 @@
   description:
   - Employs hot lead typesetting to arrange glyphs into a mold for printing.
   - |
-    An equipped `linotype` device enables the `format` command:
+    An equipped `linotype`{=entity} device enables the `format` command:
```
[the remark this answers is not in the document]
xsebek: Yes, using plain backticks will parse as (non-raw) code. In next stage we parse and validate all code as swarm code (`Document Text -> Document Syntax`).

I also just checked and while I get `Syntax Polytype` from validation, I discard that extra info with `void`.

data/entities.yaml
```diff
@@ -1363,7 +1363,7 @@
     installs a custom handler function that can be activated to
     respond to keyboard inputs typed at the REPL.
   - |
-    `key : text -> key` constructs values of type `key`, for
+    `key : text -> key` constructs values of type `key`{=entity}, for
```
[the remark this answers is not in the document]
xsebek: I think it should actually be just `key` as its valid swarm code, right?

data/scenarios/Challenges/blender.yaml
```diff
@@ -11,7 +11,7 @@ objectives:
         `grab` the Amulet of Yoneda from the northwest sanctum while
         timing your passage carefully to avoid Side Effects (X) on patrol.
       - |
-        To unlock a red door, `drill` it with the "door key" equipped.
+        To unlock a red door, `drill`{=entity} it with the "door key" equipped.
```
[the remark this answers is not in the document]
xsebek: Yes, I matched all cases with regex and did not check that it was used as verb. 👍

data/scenarios/Tutorials/craft.yaml
xsebek: Because `"branch"` is what the player should write and it is valid swarm expression.

I have considered displaying all entities with quotes, but that could be misleading if written without them.

Either way, with the `{=entity}` tag we can change our mind and replace all occurrences using regex. 🙂

data/entities.yaml
xsebek: I think I got most of them now. 🙂

src/Data/Text/Markdown.hs
xsebek: We are doing slightly more - we have annotated chunks of text, which we break while keeping the annotation.

Maybe this could be generalised as some kind of traversal?

src/Swarm/Game/Failure.hs
xsebek: Done!

src/Data/Text/Markdown.hs
xsebek: Yeah, our use case is specifically inspecting swarm code in descriptions, so generalising does not make that much sense.

byorgey:
:rocket:

byorgey:
Hmm, it doesn't seem to be working for me.  After pulling `main` and rebuilding, I get this:

markdown

xsebek:
@byorgey odd, it looks like if you used old exe with new data. 😕 It worked fir me on the move tutorial, but I did not check alk of them.

I will test it with autoplay and add a script to go through all of them.

byorgey:
Oh, I see, the goal dialog is formatted properly, but the formatting doesn't show up when you look at the description of an entity.

PR #1109, opened 2023-02-14
Allow scrolling the world map unless `scrollable: false` is set

### LINKED ISSUE #1108: Scrolling map in challenge scenarios
Several of my students complained that in the `lambda` tutorial, they cannot see the whole maze.  Currently, the only options are:
- Make the font size smaller
- Start in `--cheat` mode so you can scroll the world map
- Build a robot that makes it part of the way through the maze, bind it to a name, and use `view` to make the camera follow that robot.

Part of the problem is that the view is *centered* on the base which makes the maze off center.  Another issue is that although not being able to scroll the map makes good sense in classic mode, there's no reason you shouldn't be able to scroll the world map in a challenge scenario like this.

One simple solution would be to add a flag to scenario descriptions saying whether the map should be scrollable, and perhaps have it default to `true`.

Alternatively, or additionally, it might be nice to be able to specify the center of focus for a scenario separate from the base.

byorgey:
Adds a new boolean key `scrollable` to the world description in scenario files, with a default value of true.

Closes #1108.


byorgey:
I don't know what happened with the TODO checking. I got an email saying that it didn't run because it was empty, or something like that, and it seems not to be listed among the checks above.

kostmo (approved):

xsebek (approved):
Thanks for updating the JSON schema as well. 👍

xsebek:
@byorgey could you please paste the message here or forward me the email?

I will check why the TODO was not run. 🤨

byorgey:
@xsebek I can forward you the email, but I don't think I have your email address.  You can PM it to me on IRC or something.

PR #1110, opened 2023-02-16
add random "static" to `view` outside a certain range

### LINKED ISSUE #17: Give `view` a limited broadcast range

byorgey:
The basic principle here is that you cannot communicate with robots that are too far away.  We also want to uphold the principle that it is not possible to instantly communicate information over arbitrary distances.

Concretely, there is now an "inner radius" (16 by default) and an "outer radius" (64 by default).
* Inside of the inner radius, communication is perfect.
* Between the inner and outer radius, the `view` becomes increasingly covered with "static".
* Outside the outer radius, the view is completely static, *and* it is impossible to see a viewed robot's inventory.

In particular there should be absolutely no difference between trying to `view` a robot that is deceased vs a robot that is too far away.  A nice bonus feature is that if you are `view`ing a robot that is suddenly destroyed, e.g. by driving into the water, the entire view instantly turns to static, which looks cool.

The static is based on a hash of the offset from the view center to the cell + the current tick count / 16.  That means that the static pattern (1) doesn't shift when the viewed robot moves around, and (2) randomly changes every second or so.

[See this asciinema cast](https://asciinema.org/a/564988) for a demo of it in action.

Closes #17 .

- [x] update `antenna` description


src/Swarm/TUI/View/CellDisplay.hs
```diff
@@ -58,9 +79,47 @@ displayLoc showRobots g coords =
...
+addStatic :: GameState -> W.Coords -> Display -> Display
+addStatic g coords d = maybe d displayStatic (getStatic g coords)
 
-hidingMode :: GameState -> HideEntity
-hidingMode g
-  | g ^. creativeMode = HideNoEntity
-  | otherwise = maybe HideAllEntities HideEntityUnknownTo $ focusedRobot g
+-- | Draw black or white static.
+displayStatic :: Bool -> Display
+displayStatic = bool mempty (terrainMap M.! IceT)
```
kostmo, line 89: Cool effect.  It really ramps up the immersion into the "remote exploration" concept.

Take or leave this idea: once the `Far` range is reached, overlay a `[connection lost]` message in red color.

kostmo (approved):

src/Swarm/TUI/View/CellDisplay.hs
```diff
@@ -58,9 +79,47 @@ displayLoc showRobots g coords =
...
+-- | Add random "static" based on the distance to the robot being
+--   @view@ed.
+addStatic :: GameState -> W.Coords -> Display -> Display
+addStatic g coords d = maybe d displayStatic (getStatic g coords)
 
-hidingMode :: GameState -> HideEntity
-hidingMode g
-  | g ^. creativeMode = HideNoEntity
-  | otherwise = maybe HideAllEntities HideEntityUnknownTo $ focusedRobot g
+-- | Draw black or white static.
```
kostmo, line 87: I wonder what it would look like with a couple of shades of gray mixed in.  Typically static has a "grayish" appearance, due to the specks being so small and blurred together.  But these specks are large and chunky.  What we lack in spatial resolution in our TUI display, we could make up for in "color depth".

kostmo: For a more granular effect, in the long term we could explore using [half blocks](https://en.wikipedia.org/wiki/Block_Elements) (or even quadrant blocks).

src/Swarm/Game/State.hs
```diff
@@ -708,6 +718,64 @@ viewingRegion g (w, h) = (W.Coords (rmin, cmin), W.Coords (rmax, cmax))
...
+--         radii.
+--       - If the robot is beyond the maximum radius, it is 'Far'.
+--   * By default, the minimum radius is 16, and maximum is 64.
+--   * If the focused robot has an @antenna@ installed, it doubles
+--     both radii.
+--   * If the base has an @antenna@ installed, it also doubles both radii.
+focusedRange :: GameState -> Maybe RobotRange
+focusedRange g
+  | Nothing <- focusedRobot g = Nothing
+  | g ^. creativeMode || g ^. worldScrollable = Just Close
```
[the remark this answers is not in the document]
byorgey: Good idea! :heavy_check_mark:

src/Swarm/TUI/View/CellDisplay.hs
byorgey: Oh, very good idea, and good observation re: gray and spatial vs color resolution.  I remember talking about this with you before, thanks for pointing out the relevant concept again.  Indeed, I thought it needed more gray, so I tried changing  to use light & dark gray instead of black & white, but that looked bad (the gray blended in with any `stone` texture in the map).  But maybe using gray in combination with black and white, and/or using more granular blocks, could help make it look better.

byorgey: #1145

PR #1116, opened 2023-02-18
Plumb rich failure types

### LINKED ISSUE #1115: Adopt rich Failure types throughout asset loading code
Placeholder since I've started working on this.

kostmo:
closes #1115

This PR makes use of the rich types defined in #982 and converts many functions of `Has (Throw Text)` type to `ExceptT`.

It also logs a warning instead of crashing with terminal failure if "save" info cannot be loaded, which will be important for #974 where the save file schema changes.

Screenshot from 2023-02-18 13-22-16


src/Swarm/TUI/Model/Failure.hs
```diff
@@ -13,39 +13,26 @@
...
-import Data.Yaml (ParseException, prettyPrintParseException)
+import Data.Yaml (ParseException)
 
 data SystemFailure
-  = AssetNotLoaded Asset FilePath LoadingFailure
+  = AssetNotLoaded Asset PathLoadFailure
 
-data Asset = Achievement | Data | History | Save
+data PathLoadFailure
+  = PathLoadFailure FilePath LoadingFailure
```
xsebek, line 23: Why do you curry error constructor? :slightly_frowning_face: 

As far as I see, you do not use the `PathLoadFailure` for anything else. The `SystemFailure` in question is already named `AssetNotLoaded` so what do you get by introducing this extra step?

src/Swarm/Game/Entity.hs
```diff
@@ -107,7 +108,8 @@ import GHC.Generics (Generic)
 import Swarm.Game.Display
 import Swarm.Game.Location
 import Swarm.Language.Capability
-import Swarm.Util (binTuples, dataNotFound, getDataFileNameSafe, plural, reflow, (?))
+import Swarm.TUI.Model.FailureRender (prettyPathLoadingFailure)
```
xsebek, line 111: @kostmo please move the failure type (not widget render) that game needs to its directory.

If `Game` needs to load something and pretty print it in text, then it should be in its directory.

I understand it's less convenient, but at least we will not be lumping all errors into one big pile.

xsebek:
@kostmo thanks for working on this, once you are done I can rework logging to use these types. :+1:

src/Swarm/Util.hs
```diff
@@ -213,26 +215,31 @@ getDataDirSafe p = do
...
-      fe <- doesFileExist fp
-      return $ if fe then Just fp else Nothing
+  d <- ExceptT $ getDataDirSafe "."
+  let fp = d </> name
+  fe <- liftIO $ doesFileExist fp
+  ExceptT $
+    return $
+      if fe
+        then Right fp
+        else Left $ PathLoadFailure fp $ DoesNotExist File
```
kostmo, line 230: `PathLoadFailure` is used independently here.

src/Swarm/Game/Recipe.hs
```diff
@@ -139,17 +139,24 @@ instance FromJSONE EntityMap (Recipe Entity) where
...
-      res <- sendIO $ decodeFileEither @[Recipe Text] fileName
-      textRecipes <- res `isRightOr` (from @String @Text . prettyPrintParseException)
-      resolveRecipes em textRecipes
-        `isSuccessOr` (T.append "Unknown entities in recipe(s): " . T.intercalate ", ")
+loadRecipes ::
+  MonadIO m =>
+  EntityMap ->
+  ExceptT SystemFailure m [Recipe Entity]
+loadRecipes em = do
+  fileName <- withExceptT (AssetNotLoaded $ Data Recipes) $ getDataFileNameSafe f
```
kostmo, line 147: Here is where more context is available as to the `Asset` in question.

src/Swarm/TUI/Model/Failure.hs
kostmo: The `SystemFailure` type has an `Asset` member, and it was unclear [in certain low-level functions](https://github.com/swarm-game/swarm/pull/1116/files#r1111098032) which asset is relevant.  However, in the higher-level functions, the context becomes clear, and then I am able to use `withExceptT` to [supplement the error with the `Asset` argument](https://github.com/swarm-game/swarm/pull/1116/files#r1111100570).

src/Swarm/Game/Entity.hs
kostmo: I don't quite follow what the proper resolution here is; can you be more explicit? :)

Perhaps the convention will become more clear once #1043 is merged.

xsebek: @kostmo imagine that you could not import TUI from Game and write code accordingly. 😉 

If you decide to ignore it then hope I will not merge the library split faster than you! 😆

src/Swarm/Util.hs
xsebek: OK, since Util is the lowest library, you would have to move this failure type here.

Or do not use the type here.

_Everything_ now depends on it. Please try to imagine the dependency graph you are creating.

src/Swarm/Game/Entity.hs
kostmo: I have now moved/split some of the modules in a way that hopefully makes more sense for the dependency graph.  But, to my earlier point about #1043, I find this quite challenging to manage in my head without some kind of mechanical enforcement.  For me, at least, I think that maintaining optimal placement of the module files will only be practical once it is enforced/assisted by true package divisions.

For now, is the module arrangement sufficient to merge this PR, and we can perfect their placement later?

src/Swarm/Failure.hs
```diff
@@ -0,0 +1,36 @@
+-- |
+-- Module      :  Swarm.Game.Failure
```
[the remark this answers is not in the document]
kostmo: Case in point for #1120.  In fact I'm going to get a head start on that and remove it.

src/Swarm/Game/Entity.hs
```diff
@@ -363,12 +366,11 @@ instance ToJSON Entity where
 -- | Load entities from a data file called @entities.yaml@, producing
 --   either an 'EntityMap' or a pretty-printed parse error.
 loadEntities :: MonadIO m => m (Either Text EntityMap)
```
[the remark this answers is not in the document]
kostmo: Sure.  But TBH my original intent was just to plumb enough types to implement 843fdad ("Do not crash if scenario status cannot be loaded") so as to unblock #974.  So how about let's save this for a follow-up PR.

src/Swarm/Game/Recipe.hs
```diff
@@ -139,17 +139,24 @@ instance FromJSONE EntityMap (Recipe Entity) where
...
-    Nothing -> sendIO (dataNotFound f) >>= throwError
-    Just fileName -> do
-      res <- sendIO $ decodeFileEither @[Recipe Text] fileName
-      textRecipes <- res `isRightOr` (from @String @Text . prettyPrintParseException)
-      resolveRecipes em textRecipes
-        `isSuccessOr` (T.append "Unknown entities in recipe(s): " . T.intercalate ", ")
+loadRecipes ::
+  MonadIO m =>
+  EntityMap ->
+  ExceptT SystemFailure m [Recipe Entity]
```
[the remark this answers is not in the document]
kostmo: I hadn't really wrapped my head around the "capability style", so I fell back on `ExceptT` which I am familiar with.  Let's defer as a separate issue (#1122).

src/Swarm/Game/Scenario.hs
```diff
@@ -200,36 +201,40 @@ scenarioStepsPerTick :: Lens' Scenario (Maybe Int)
...
+  listToMaybe <$> liftIO (filterM doesFileExist candidates)
+ where
+  e2m = fmap eitherToMaybe . runExceptT
 
 -- | Load a scenario with a given name from disk, given an entity map
 --   to use.  This function is used if a specific scenario is
 --   requested on the command line.
 loadScenario ::
-  (Has (Lift IO) sig m, Has (Throw Text) sig m) =>
+  MonadIO m =>
```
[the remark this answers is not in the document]
kostmo: Sure.

src/Swarm/Game/ScenarioInfo.hs
```diff
@@ -244,34 +249,34 @@ orderFileName = "00-ORDER.txt"
 -- | Recursively load all scenarios from a particular directory, and also load
 --   the 00-ORDER file (if any) giving the order for the scenarios.
 loadScenarioDir ::
-  (Has (Lift IO) sig m, Has (Throw Text) sig m) =>
+  MonadIO m =>
```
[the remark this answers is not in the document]
kostmo: Simply ignorance :)

src/Swarm/TUI/Model/FailureRender.hs
```diff
@@ -0,0 +1,37 @@
+{-# LANGUAGE OverloadedStrings #-}
+
+module Swarm.TUI.Model.FailureRender where
```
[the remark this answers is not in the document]
kostmo: :+1:

src/Swarm/Game/State.hs
```diff
@@ -988,7 +997,12 @@ initGameStateForScenario ::
   Maybe FilePath ->
   ExceptT Text IO GameState
 initGameStateForScenario sceneName userSeed toRun = do
-  g <- initGameState
+  (warnings, g) <- initGameState
+  unless (null warnings) $
+    except $
+      Left $
+        T.unlines $
+          map prettyFailure warnings
```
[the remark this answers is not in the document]
kostmo: Done.  If not for `fourmolu` I would normally put several of the functions on the same line.  It looks like `fourmolu` puts `.` chains in a straight column rather than staggered indents as for `$` chains.

src/Swarm/Util/GameData.hs
```diff
@@ -0,0 +1,119 @@
+{-# LANGUAGE OverloadedStrings #-}
+
+-- Various utilities related to loading game data files.
+module Swarm.Util.GameData where
```
xsebek, line 4: @kostmo why do you want to have everything in `Util`?

It is great that you found what part of the util module is used in the Game (and TUI) and even included "game" in the module name.

Please finish the job and move it to Game. :wink:

src/Swarm/Failure/Render.hs
```diff
@@ -0,0 +1,37 @@
+{-# LANGUAGE OverloadedStrings #-}
+
+module Swarm.Failure.Render where
```
xsebek, line 3: @kostmo @byorgey why is this split from the `Failure` module? It depends on nothing else.

AFAIK we do not have any other modules that have data declaration and basic functions separately.

Look at `Exception.hs` - it has `data Exn` _and_ `formatExn` in the same module.

xsebek (changes requested):
@kostmo I like how you split the modules :+1: Please just move them as I commented.

Some of this will be easier once Cabal enforces it on split libraries, but doing that is _a lot of not-very-fun work_, so I would appreciate it if you made it easier for me.

I can not wait to have it done and get to more enjoyable things like rewriting the logging. :sweat_smile:

src/Swarm/Failure/Render.hs
byorgey: I was just commenting on a better place for the module to be, but I agree, it can/should probably just go in the `Failure` module itself.

src/Swarm/Game/Recipe.hs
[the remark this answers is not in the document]
byorgey: OK, sounds good.

src/Swarm/Game/ScenarioInfo.hs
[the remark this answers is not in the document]
byorgey: OK, no worries. :smile:   We'll fix it all up in #1122 .

src/Swarm/Game/State.hs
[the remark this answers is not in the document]
byorgey: Yes, I don't mind `f $ g $ h $ x` so much when it's on a single line (though I personally still prefer using `.` as much as possible) but indeed, I had in mind the fact that `fourmolu` insists on doing a hanging indent for each `$`, but lines up `.` chains in a single column.

byorgey (approved):
Looking good to me!

src/Swarm/Failure/Render.hs
kostmo: In this case I had intended to use quoting functions from `Swarm.Util` in the rendering code.  And up until 75669d7, `Swarm.Util` depended on types in the `Failure` module, causing a circular dependency.  So I did the work of splitting the modules, but later decided I didn't need the quoting functions after all.

However, after doing that splitting work, I'm not particularly inclined to merge `Swarm.Failure` and `Swarm.Failure.Render` back into a single module again---IMO, there is no downside to the types and the rendering functions being in separate modules, even if there is no technical requirement to be so.  I've found it to be a common step in the evolution of a module: auxiliary functions that had been self-contained with the data definitions eventually needing to pull in an external dependency, requiring the data to move to a separate "types" module to break a circular reference.  It may very well become necessary again for `Failure` rendering at some point.

src/Swarm/Util/GameData.hs
kostmo: Ok, moved to `Swarm.Game.ResourceLoading`.

kostmo:
@xsebek, moved the new modules under `Game`.  Good to go?

src/Swarm/Failure/Render.hs
xsebek: @kostmo that is inconsistent with the code base style though.

Rendering widgets - sure.

But pure functions that are critical to usability of the type (`ParseError` and `prettyParseError`) can absolutely go into the same module and they do everywhere else in Swarm.

Which language puts pure functions on data in different module? In C++ and Python, I have always seen the functions that are basically “methods” in the same module.

Regarding dependencies, your reasoning is backwards.  The Util module should simply not import anything else. It has to be on the bottom of dependency graph. We should try to move stuff _from_ it (as you did) instead of importing modules _in_ it.

xsebek: Finally from a practical standpoint this is premature and silly.

We have modules 2000+loc and argue here for splitting a 100loc module in a novel way, based on OOP best practise.

The future split could go another way though - split failures based on which part of game or UI produced it.
Whoever adds more code to failures in the future can decide which split fits better.

src/Swarm/TUI/Model/Failure.hs
xsebek: @kostmo I only see the `getData[Dir|File]Safe` functions. Why do this hard, when you can simply pass them the `Asset` type?
```diff
- left (AssetNotLoaded $ Data AppAsset)  <$> getDataDirSafe "."
+ getDataDirSafe AppAsset "."
```

This also makes more sense, because no other `Asset` is in the `data` directory.

xsebek (approved):
@kostmo I think this is good enough to go. :+1: 

Please read my comment on `PathLoadFailure` though - it seems to me you are unnecessarily making the types harder to use.

I can not understand the urge to create separate "Render" modules, but I guess I can live with one extra folder and a file. I will just have to sort the folders and files alphabetically and not "folders first" so it is less annoying. :unamused:

src/Swarm/Failure/Render.hs
xsebek: As a side note, the smart constructor use case is a good point, but _it does not apply here_.

We will be adding failure types and their pretty printing at the same time, not tweaking the pretty printing function. This is not about "rendering a widget" - I am not going to change the colour of exceptions tomorrow and recompile the whole code base because `Text` does not have a colour `Attr` that I could change.

What I am trying to point out is that failure data type and format are tightly linked, think of it like C++ [`e.what()`](https://en.cppreference.com/w/cpp/error/exception/what) method instead of `failureWidgetWithAttr`.

PR #1117, opened 2023-02-18
Fix TODO action ID

xsebek:
Previously the action did not run because of an error:
```
The workflow is not valid. .github/workflows/check_todos.yml (Line: 14, Col: 3):
  The identifier 'Enforce issue references' is invalid. IDs may only contain alphanumeric characters, '_', and '-'.
  IDs must start with a letter or '_' and and must be less than 100 characters.
```

- fixes #1101


.github/workflows/check_todos.yml
```diff
@@ -11,7 +11,8 @@ on:
     branches:
       - main
 jobs:
-  Enforce issue references:
+  enforce-issue-references:
+    name: Enforce issue references
```
xsebek, line 15: @kostmo you can see the error here:
- https://github.com/swarm-game/swarm/actions/workflows/check_todos.yml

If you use VS Code, the YAML extension should pick up the JSON Schema and show the error immediately:
|image|
|---|

byorgey (approved):
Thanks for tracking this down!

xsebek:
@byorgey no problem :wink:

I will fill out the GitHub survey and note that it would be nice to make the error more visible so that we do not forget to run some actions. :smile:

.github/workflows/check_todos.yml
kostmo: Cool, thanks for fixing.

PR #1118, opened 2023-02-18
Save time upon finishing scenario

### LINKED ISSUE #1111: "Best time" includes time elapsed while Congratulations dialog is displayed
It also includes the time elapsed while the initial Goal dialog is displayed.  Maybe it is OK to count the time spent reading modal dialogs against the total, but one would think that the timer should stop once the game is won.

In contrast, the "ticks" are only counted when the game is paused.  So there is somewhat of a discrepancy in that the "wall-clock" time includes all of the time when the game is paused.

xsebek:
- closes #1111

### DIFF
--- a/src/Swarm/Game/ScenarioInfo.hs
+++ b/src/Swarm/Game/ScenarioInfo.hs
@@ -24,7 +24,7 @@ module Swarm.Game.ScenarioInfo (
   scenarioStatus,
   scenarioBestTime,
   scenarioBestTicks,
-  updateScenarioInfoOnQuit,
+  updateScenarioInfoOnFinish,
   ScenarioInfoPair,
 
   -- * Scenario collection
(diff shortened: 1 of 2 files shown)


xsebek:
I tried removing the saves directory and running solution for the spiral maze on `main` and here:
```sh
rm -rf ~/.local/share/swarm/saves

S=scenarios/Challenges/Mazes/easy_spiral_maze.yaml
yq '.solution' $S > maze.sw  # so you can run "maze.sw" later ;)

git checkout main
cabal run swarm:swarm -- --scenario $S  # let it sit in the quit dialog for a minute

git checkout finish-scenario-time
cabal run swarm:swarm -- --scenario $S # let it sit in the quit dialog for a minute
```

byorgey (approved):
Makes sense!

PR #1135, opened 2023-02-27
hackman

kostmo:
# Demo
    scripts/play.sh --scenario Challenges/hackman.yaml --autoplay
image

Makes use of both #1058 and #1023.

### DIFF
--- a/data/scenarios/Challenges/_hackman/ghost.sw
+++ b/data/scenarios/Challenges/_hackman/ghost.sw
@@ -1,0 +1,156 @@
+/* Algorithm:
+Go forward until along any straight passage until reaching
+an intersection. Then randomly choose any available direction
+(except backwards).
+*/
+
+def isItemInDirection = \direction. \item.
+    x <- scan direction;
+    return $ case x (\_. false) (\y. y == item);
+    end;
+
+// A ghost is not blocked by a "gate" when it is leaving the "nursery".
+def canCrossGate =
+    isGate <- isItemInDirection forward "gate";
+    loc <- whereami;
+    let amCentered = loc == (0, 0) in
+    return $ isGate && amCentered;
+    end;
+
+def isBlockedAhead =
+    isBlockedByWall <- blocked;
+    ghostCanCrossGate <- canCrossGate;
+    return $ isBlockedByWall && not ghostCanCrossGate;
+    end;
+
+def checkLeftBlocked =
+    turn left;
+    isBlocked <- isBlockedAhead;
+    turn right;
+    return isBlocked;
+    end;
+
+def checkRightBlocked =
+    turn right;
+    isBlocked <- isBlockedAhead;
+    turn left;
+    return isBlocked;
+    end;
+
+def chooseDirection : cmd dir =
+    leftBlocked <- checkLeftBlocked;
+    rightBlocked <- checkRightBlocked;
+    forwardBlocked <- isBlockedAhead;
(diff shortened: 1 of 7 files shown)


data/scenarios/Challenges/_hackman/ghost.sw
```diff
@@ -0,0 +1,156 @@
...
+
+                                d <- random 3;
+                                if (d == 0) {
+                                    return left;
+                                } {
+                                    if (d == 1) {
+                                        return right;
+                                    } {
+                                        // go straight
+                                        return forward;
```
byorgey, line 108: It would be really nice to have lists in the language, wouldn't it?  Then we could rewrite this tedious case analysis to just pick randomly from a list of available directions...

data/scenarios/Challenges/hackman.yaml
```diff
@@ -0,0 +1,231 @@
...
+              return false;
+          }
+          end;
+
+      allHasPowerup 4;
+robots:
+  - name: base
+    display:
+      orientationMap:
+        east: 'e'
```
byorgey, line 88: Is there a reason not to use `c` and `ↄ` for east and west?

byorgey (approved):
Nice!  I wrote some code to follow and eat pellets + donuts as long as there are any adjacent.  But I had to manually move hackman between each string of pellets once it ate a bunch and ended up next to blank spaces.  I'm still wondering if there is a less manual way to do it.  What about equipping hackman with a `bitcoin`?  Then at the very least you could code it to wander around randomly until encountering more things to eat.

byorgey:
I'm proud of this solution I developed, which eventually manages to eat everything (though it takes a while):

```

  b <- test; if b then else
end

def while : cmd bool -> {cmd a} -> cmd unit = \test. \body.
end

def unless = \test. \then. \else. ifC test else then end

def until = \test. \body.
  ifC test {} {force body ; until test body}
end

def elsif = \p. \then. \else. {if p then else} end
def elsifC = \p. \then. \else. {ifC p then else} end

def forever : cmd a -> cmd b = \c.
  c; forever c
end

def inverseDir : dir -> dir = \d.
  if (d == left) {right} {if (d == right) {left} {d}}
end

def isBlocked : dir -> cmd bool = \d.
  turn d;
  b <- blocked;
  turn (inverseDir d);
  return b
end

def followLeft =
  unless (isBlocked left) {turn left}
  { unless (isBlocked forward) {}
    { unless (isBlocked right) {turn right} {turn back} } };
  move
end

def followRight =
  unless (isBlocked right) {turn right}
  { unless (isBlocked forward) {}
    { unless (isBlocked left) {turn left} {turn back} } };
  move
end

def isFood = \x. x == inr "pellet" || x == inr "donut" end

def isFoodTo = \d. res <- scan d; return (isFood res) end

def stepToFood =
  ifC (isFoodTo left) {turn left; move}
  $ elsifC (isFoodTo forward) {move}
  $ elsifC (isFoodTo right) {turn right; move}
  $ elsifC (isFoodTo back) {turn back; move} {}
end

def followFood = while (isFoodTo down) { grab; stepToFood } end

def eatAll =
  forever (
    until (isFoodTo down) { followRight; followLeft; followLeft; stepToFood };
    followFood
  )
end
```

data/scenarios/Challenges/hackman.yaml
kostmo: Agreed, these are probably better.

Issue #1007, opened 2023-01-14
"use" command as generalization of "drill"


kostmo:
*Issue formerly titled: "Recipes with custom drills do not work in creative mode"*

Creating this bug for visibility/findability later, as I've mostly figured out the problem and a workaround.

In `word-search.yaml`, for example, I define a custom entity `"highlighter"` that has `drill` capability.  I also define recipes that require this `"highlighter"`.  However if I try to play this scenario in creative mode, it will say "no way to make [...]" or something to that effect.

This is because in creative mode, the player is granted a `"drill"` entity automatically, and in fact it defaults to that drill when computing recipe inputs/outputs for the `drill` command, ignoring the user-defined `highlighter` recipe.

One workaround is, of course, to avoid Creative mode.  Another workaround is to eschew custom entities with `drill` capability, and instead incorporate the standard `"drill"` entity into one's recipe.

xsebek:
The drill is overall a bit hacky for which I am sorry. :smile: I don't think "highlighting with drill" makes much sense, but maybe we could solve that and the input/output problem simultaneously.

I would propose adding a `use` command (or `apply`?) that uses a given installed device on a nearby location:
```
use :: entity -> direction -> cmd unit
```

@byorgey would it have to be a syntax form to work with requirement analysis? Or could we make it work as long as the entity parameter is known at definition? :thinking: 

Maybe it's too cumbersome to use (pun intended), but I think it would work better with custom recipes and entities.

byorgey:
I like the idea of `use`.  Yes, it would have to be a special form (or we could make it a normal constant and just add a special checking rule to make sure it's applied to a text literal - that would give better error messages).

kostmo:
Would the `use` command only work with an `equipped` device?  Let's say I need a `"key"` to open a `"locked door"` (where a recipe converts the `"locked door"` to an `"unlocked door"`).  Shall we require the player to execute this sequence of commands?
```
equip "key";
use "key" forward;
```

On the bikeshedding front, I'm not necessarily opposed to `use` as the command name, but may I also submit `operate` for consideration?

byorgey:
I think having it only work with an `equipped` device makes sense.  However, note that getting something your example to work (doing an `equip` immediately followed by a `use`) would require a more careful job of requirements analysis than we currently do (see https://github.com/swarm-game/swarm/issues/921#issuecomment-1356403359).

xsebek:
[...]

PR #1136, opened 2023-02-27
adventure game scenario

kostmo:
This is my attempt at a longer "adventure game"-style quest.
The primary mechanic of this scenario is in bridging different regions of the map by crafting paths.
The quest is not entirely linear; there a few things that can be done in different order.

A few observations:
* I make heavy use of the "drill".  Having the `use` command described in #1007 would be nicer.
* Some recipes (e.g. using lava) might overlap in their ingredients, so to force a certain output you may have to temporarily drop (`place`) the ingredient for the undesired recipe.
* I start the player off with a few items in their inventory with zero quantity so that the associated recipes can be browsed.  For some of these items, especially the non-portable ones, it doesn't really make sense to be under a heading called "Inventory".  Perhaps something like "compendium" or "library" might make more sense for the title of that pane in the UI, since it represents things that you "know" about?
* I developed the "solution" code simultaneously with designing the map, so I've never played the scenario with "fresh eyes".  So I don't know how much of a "slog" it will feel like to a new player.  I imagine that playing it through for the first time may generate a few ideas about improving the swarm UI, perhaps regarding saving progress, or exporting REPL content to a `.sw` file.
    * Navigating through a twisty passage and encoding this as commands can be a bit tedious.  Using "pilot" mode is easier.  But you will have to exit creative mode again when it comes time to drill something (#1007).  Inputting "landmark" strings into the REPL at the beginning and end of a piloting sequence could make it easier to find within the REPL history for copy-pasting.
* I am quite interested in feedback about making the scenario more or less challenging, what clues might be helpful, potential lore, and stylistic improvements.
* To emphasize the aspect of swarm as a "programming game", I included lots of repetition.
* A large map like this is fertile ground for "easter eggs".  I've added a couple so far.
* Since `--autoplay` implies `--cheat`, the "hidden/optional" goals are displayed in the popup.  It would be better if they were ordered below the mandatory active goals, so that when sitting back and watching the `--autoplay`, the next "logical" goal text gets displayed by default instead of the optional ones.

image

Demo:

    scripts/play.sh --scenario Challenges/bridge-building.yaml --autoplay

### DIFF
--- a/data/scenarios/Challenges/00-ORDER.txt
+++ b/data/scenarios/Challenges/00-ORDER.txt
@@ -1,4 +1,5 @@
 chess_horse.yaml
+bridge-building.yaml
 teleport.yaml
 2048.yaml
 word-search.yaml
--- a/data/scenarios/Challenges/_bridge-building/flower-ring-check.sw
+++ b/data/scenarios/Challenges/_bridge-building/flower-ring-check.sw
@@ -1,0 +1,62 @@
+def doN = \n. \f. if (n > 0) {f; doN (n - 1) f} {}; end;
+
+
+def isItemInDirection = \direction. \item. 
+    x <- scan direction;
+    return $ case x (\_. false) (\y. y == item);
+    end;
+
+def isFlankedByItem = \item.
+    hasLeft <- isItemInDirection left item;
+    hasRight <- isItemInDirection right item;
+    return $ hasLeft && hasRight;
+    end;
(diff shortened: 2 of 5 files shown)


byorgey:
@kostmo This is amazing!! @noahyor and I had a lot of fun playing it together, we just finished.  Some more feedback/thoughts coming soon!

PR #1142, opened 2023-03-06
web api to parse, render, and run code

### LINKED ISSUE #625: Implement run command for the web service

kostmo:
Closes #625.

# Demo

## Running code

Start a creative game:

    scripts/play.sh --scenario data/scenarios/creative.yaml

In another terminal window:

    curl -H "Content-Type: text/plain; charset=utf-8" -i http://localhost:5357/code/run --data "move;"

Observe that the robot has moved, and that the command sent via `curl` has been added to the REPL history.

## Parsing and rendering code

Start swarm:

    scripts/play.sh

Run curl:

    curl --silent -H "Content-Type: text/plain; charset=utf-8"  -i http://localhost:5357/code/render --data @data/test/language-snippets/warnings/unused-vars/multiple-lambda-first-unused.sw

See output:
```
HTTP/1.1 200 OK
Transfer-Encoding: chunked
Date: Mon, 06 Mar 2023 10:16:12 GMT
Server: Warp/3.3.23
Content-Type: text/plain;charset=utf-8

def put = \x. \y. place y end
|
`- \x. \y. place y
   |
   `- \y. place y
      |
      `- place y
         |
         +- place
         |
         `- y
```


src/Swarm/Web.hs
```diff
@@ -93,6 +106,18 @@ mkApp appStateRef =
...
+  codeRenderHandler contents = do
+    return $ T.pack $ case processTermEither contents of
+      Right (ProcessedTerm (Module stx@(Syntax' _srcLoc _term _) _) _ _) ->
+        drawTree . fmap prettyString . para Node $ stx
+      Left x -> x
+
+  codeRunHandler contents = do
+    foo <- liftIO $ atomicModifyIORef' appStateRef $
+      swap . runState (runBaseWebCode contents)
+    return $ T.pack foo
```
kostmo, line 119: @byorgey, I got this to typecheck, but it doesn't seem to work.

To test this, I started a Classic game.  Executing `move;` in the REPL pane works. However, executing the following `curl` command has no effect on the game UI:

    curl -H "Content-Type: text/plain; charset=utf-8" -i http://localhost:5357/code/run --data "move;"

Any ideas?

src/Swarm/TUI/Controller.hs
```diff
@@ -937,6 +938,56 @@ handleREPLEventPiloting x = case x of
       & replPromptText .~ nt
       & replPromptType .~ CmdPrompt []
 
+runBaseWebCode :: MonadState AppState m => T.Text -> m String
```
kostmo, line 941: At first the return value was `m ()`, but I made it return a `String` for debugging purposes.

src/Swarm/Web.hs
byorgey: The `IORef` only stores a *copy* of the `AppState`, so modifying it is not going to affect the running game. See the code in https://github.com/swarm-game/swarm/blob/main/src/Swarm/App.hs and in particular how the `eventHandler` defined on line 87 always just reads the current game state and writes it into the `IORef`. 

Changing the `eventHandler` to somehow do two-way synchronization of the state sounds like a bad idea---since the web server and the game are running asynchronously in different threads, the state could get updated simultaneously in both places.  Probably what we should do instead is simply say that the copy of the `AppState` in the state monad is the single canonical state, and the `appStateRef` only ever contains a read-only copy.  If the web server wants to update the state it needs to send some appropriate event to the game's main event queue where it can be processed.  i.e. we can add a new `AppEvent` (see `Swarm.TUI.Model`) containing a `Text` value indicating that someone sent it to the `code/run` endpoint.

kostmo: Ah yes, that makes way more sense.  I'll admit it was satisfying to have the "type tetris" finally click into place with the `atomicModifyIORef` and `State` monad, so am mildly disappointed that it does absolutely nothing :smile:

src/Swarm/TUI/Controller.hs
```diff
@@ -949,6 +953,51 @@ handleREPLEventPiloting x = case x of
       & replPromptText .~ nt
       & replPromptType .~ CmdPrompt []
 
+runBaseWebCode :: MonadState AppState m => T.Text -> m ()
+runBaseWebCode uinput = do
+  s <- get
+  let topCtx = topContext s
+  unless (s ^. gameState . replWorking) $
```
kostmo, line 960: Alternatively, we could perform this `replWorking` check before we send the `RunWebCode` event over the channel, so that we could return a "rejected" status via the web API.

Or maybe we need to do the check in both places to avoid a race condition.

src/Swarm/Web.hs
```diff
@@ -99,15 +111,22 @@ docsBS =
  where
   intro = SD.DocIntro "Swarm Web API" ["All of the valid endpoints are documented below."]
 
-mkApp :: IORef AppState -> Servant.Server SwarmAPI
-mkApp appStateRef =
+mkApp ::
+  -- | Readonly
+  IORef AppState ->
```
byorgey, line 116: It would be nice to encode this invariant in the types.  I wonder if there's a standard way to create a value which gives read-only access to an `IORef`.  Of course we could create our own opaque newtype which wraps `IORef`, and export only a function to read it but no constructor.  But maybe something like that already exists in a standard library somewhere?

src/Swarm/TUI/Controller.hs
byorgey: That could be nice I suppose, I don't really care.  But yes, I think it would not be correct to check *only* before sending the `RunWebCode` event.  We have to check here in any case, because the base could start running in between sending and receiving/processing the event.

byorgey (approved):
Sorry it took me so long to look at this!

byorgey:
@kostmo are there any other PRs that you'd like feedback on at the moment?  I think I'm now caught up but I could very easily have missed some.

kostmo:
> any other PRs

#974 is blocked on a stale change request, maybe @xsebek can unblock it?

src/Swarm/Web.hs
kostmo: Couldn't find a standard library version, so I created `ReadableIORef`.

Issue #153, opened 2021-10-03
Add type synonyms (or just generalize to System Fω)


xsebek:
## Problem
I find the tuples not as usable when I can not give a name to my user type.
This would be even more frustrating with sum types (#46).

## Solution
I would like `type` similar to `def`:
```haskell
type usertypename
  = forall typeparam1 ... typeparamn.
  {{some construction using + and *}}
end
```

## Alternatives
Generate the boilerplate with Bash/Python to inline human-readable type definitions.

byorgey:
Agreed, this would be nice indeed.  As a nitpick, I think the syntax should be
```
type usertypename typeparam1 ... typeparamn = 
  {{some construction using + and *}}
end
```
because we are defining a parameterized type, not a synonym for a polymorphic type.  (The latter quickly gets us into rank-n polymorphism, which I am emphatically avoiding.)

byorgey:
Note this would also require adding kinds and kind checking.

xsebek:
> kinds and kind checking

Ouch, a `C-Project` it is then. :disappointed:

byorgey:
> > kinds and kind checking
> 
> Ouch, a `C-Project` it is then. 😞

I mean, we can get away without kind checking if we require type synonyms to always be fully applied, like in Haskell. However, the reason Haskell requires that (type families) doesn't apply here, and especially if we want things like `rec` we are going to want partially applied type synonyms.

`C-Project` is probably justified, but I actually don't think it will be all that bad.=)

byorgey:
So one potentially nice way to do this is to generalize the language to [System Fω](https://en.wikipedia.org/wiki/System_F-omega) and explicitly add `Type`, the type of types, and allow lambda expressions that send types to types.  In other words we could just write
```
def usertypename : Type -> Type -> Type = \a. \b. a + b * () end
```
or something like that.  This is conceptually nice since we no longer need a separate syntax for type definitions at all.  Of course it complicates type checking; I definitely haven't worked out the details yet, but at any rate it's worth considering further.

byorgey:
Possibly relevant: https://pure.tue.nl/ws/files/1931163/9511421.pdf

xsebek:
@byorgey maybe as a stepping stone, how about adding _not prametrized_ type synonyms first?

It would be useful for annotating code and would make the final task a bit easier by introducing the `type` syntax.

byorgey:
[...]

Issue #99, opened 2021-09-28
Redo type checker using "propagate types inward" trick


byorgey:
Peyton Jones *et al*, in section 5.4 of _[Practical type inference for arbitrary-rank types](https://www.microsoft.com/en-us/research/wp-content/uploads/2016/02/putting.pdf)_, describe an apparently well-known trick for improving type error messages: instead of an infer function, write a *check* function which takes an expected type; we can then recover `infer` by simply generating a fresh unification variable and passing that along as the "expected type".

I hope this may lay the groundwork for better error messages when type inference fails.

TristanCacqueray:
To add source location to the `TypeErr` message, I tried to thread a `TermLoc` data into the `TypeF` and ideed that seems rather tricky =)

My first attempt looks like this:

```haskell
data TermLoc = TermLoc {start :: Int, end :: Int}
  deriving (Eq, Show, Data)

-- Add TermLoc to the relevant terms
data Term
  = TVar TermLoc Var
  | TDef TermLoc Var (Maybe Polytype) Term
  | ...
```

which could be parsed like so:

```haskell
locatedTerm :: (TermLoc -> a -> Term) -> Parser a -> Parser Term
locatedTerm mkTerm parser = do
  start <- getOffset
  a <- parser
  end <- getOffset
  pure $ mkTerm (TermLoc start' end') a
```

so that we could add the location of `UnboundVar` and `DefNotTopLevel` type errors. This is a rather simple solution, could it also work for the "propagate types inward" trick?

byorgey:
> To add source location to the `TypeErr` message, I tried to thread a `TermLoc` data into the `TypeF` and ideed that seems rather tricky =)

Note that to add a `TermLoc` to types, instead of adding it to `TypeF` directly, it could probably be done something like so:
``` haskell
type Type = Fix (Compose ((,) TermLoc) TypeF)
```
with lots of utility functions and pattern synonyms to make to not painful to use.  Being able to easily decorate every node like this is one of the benefits of using a two-level type. Off the top of my head though I'm not sure how this would interact with unification; we would need to tell the library to not worry about `TermLoc`s that don't match. 

However, in any case I'm not sure adding `TermLoc`s to types like this makes sense.  Lots of `Type` values arise in ways that don't correspond to a particular place in the code.

[...]

PR #1148, opened 2023-03-08
Records

### LINKED ISSUE #1093: Records
Records were mentioned in #1087 as a way to represent imported modules.  They would also be a cool, handy feature on their own.  Records can make it very convenient to package and work with "objects".  For example, compare:
```
-- with records
f : [size : int, name : text, action : cmd unit] -> cmd unit
f r = if (r.size > 10 && r.name != "base") { r.action } {}

-- without records, ugh
f : int * text * (cmd unit) -> cmd unit
f r = if (fst r > 10 && fst (snd r) != "base") { snd (snd r) } {}
```

However, records have one very big downside: inferring record types requires some fancy type system features such as e.g. subtyping and [row polymorphism](https://en.wikipedia.org/wiki/Row_polymorphism).  A simple alternative could be to just give up on inferring record types and require type annotations (would probably require #10).

I do not know whether adding records would be worth it, or what the right tradeoffs would be if we did add them.  For now, just adding this issue as a placeholder for discussion.

byorgey:
Add record types to the language: record values are written like `[x = 3, y = "hi"]` and have types like `[x : int, y : text]`.  Empty and singleton records are allowed.  You can project a field out of a record using standard dot notation, like `r.x`.  If things named e.g. `x` and `y` are in scope, you can also write e.g. `[x, y]` as a shorthand for `[x=x, y=y]`.

Closes #1093 .

#153 would make this even nicer to use.

One reason this is significant is that record projection is our first language construct whose type cannot be inferred, because if we see something like `r.x` all we know about the type of `r` is that it is a record type with at least one field `x`, but we don't know how many other fields it might have.  Without some complex stuff like row polymorphism we can't deal with that, so we just punt and throw an error saying that we can't infer the type of a projection.  To make this usable we have to do a better job checking types, a la #99 . For example `def f : [x:int] -> int = \r. r.x end` would not have type checked before, since when checking the lambda we immediately switched into inference mode, and then encountered the record projection and threw up our hands.  Now we work harder to push the given function type down into the lambda so that we are still in checking mode when we get to `r.x` which makes it work.  But it is probably easy to write examples of other things where this doesn't work.  Eventually we will want to fully implement #99 ; in the meantime one can always add a type annotation (#1164) on the record to get around this problem.

Note, I was planning to add a `open e1 in e2` syntax, which would take a record expression `e1` and "open" it locally in `e2`, so all the fields would be in scope within `e2`.  For example, if we had  `r = [x = 3, y = 7]` then instead of writing `r.x + r.y` you could write `open r in x + y`.  This would be especially useful for imports, as in `open import foo.sw in ...`.  However, it turns out to be problematic: the only way to figure out the free variables in `open e1 in e2` is if you know the *type* of `e1`, so you know which names it binds in `e2`.  (In all other cases, bound names can be determined statically from the *syntax*.)  However, in our current codebase there is one place where we get the free variables of an untyped term: we decide at parse time whether definitions are recursive (and fill in a boolean to that effect) by checking whether the name of the thing being defined occurs free in its body.  One idea might be to either fill in this boolean later, after typechecking, or simply compute it on the fly when it is needed; currently this is slightly problematic because we need the info about whether a definition is recursive when doing capability checking, which is currently independent of typechecking.

I was also planning to add `export` keyword which creates a record with all names currently in scope --- this could be useful for creating modules.  However, I realized that very often you don't really want *all* in-scope names, so it's not that useful to have `export`.  Instead I added record punning so if you have several variables `x`, `y`, `z` in scope that you want to package into a record, you can just write `[x, y, z]` instead of `[x=x, y=y, z=z]`.  Though it could still be rather annoying if you wanted to make a module with tons of useful functions and had to list them all in a record at the end...

Originally I started adding records because I thought it would be a helpful way to organize modules and imports.  However, that would require having records that contain fields with polymorphic types.  I am not yet sure how that would play out.  It would essentially allow encoding arbitrary higher-rank types, so it sounds kind of scary.  In any case, I'm still glad I implemented records and I learned a lot, even if they can't be used for my original motivation.

I can't think of a way to make a scenario that requires the use of records.  Eventually once we have proper #94 we could make a scenario where you have to communicate with another robot and send it a value of some required type.  That would be a cool way to test the use of other language features like lambdas, too.


data/entities.yaml
```diff
@@ -1193,3 +1193,24 @@
...
+- name: victrola
+  display:
+    attr: device
+    char: 'Q'
+  description:
+  - |
+    A device for reading and writing data on circular platters made of
+    a soft plastic material, and also, for some reason, transmitting
+    the data as pressure waves through the air.  The stylus must be
+    made of an extremely hard and durable material.
```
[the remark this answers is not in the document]
kostmo: Or, how about a [quantum dot](https://en.wikipedia.org/wiki/Quantum_dot)?  Sounds sufficiently lofty for an end-game item?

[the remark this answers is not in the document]
byorgey: I still like `victrola`: it lets you "play (with) records".  And "records" in the audio sense are actually a nice analogue for records in the programming sense: both let you store multiple pieces of data in one object.  In contrast a dot lets you project out fields from a record, but why would it let you *create* records?

What if I just get rid of the description text that mentions "pressure waves through the air"?  I agree it's probably distracting.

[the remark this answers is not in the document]
byorgey: > the convenience of records might be considered to be their "dot syntax"

The convenience of records is the ability to put multiple pieces of data together into one, where the pieces of data are *named* instead of identified by position (as they would be with a normal product type).  If you still don't like `victrola`, maybe we could consider something related to (name)tags?  Just brainstorming here...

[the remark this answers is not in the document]
kostmo: I can live with it.  Given that projection is nonetheless an integral operation to record usage, I wonder if we can still work `quantum dot` in somehow.  How about as an ingredient to the `victrola` recipe?

Another consideration is generic vs. trademarked name of the record player.  I take it that `victrola` has become a somewhat genericized brand name, as "kleenex" has become for facial tissues.  There's a whole [Wikipedia section on terminology](https://en.wikipedia.org/wiki/Phonograph#Terminology), but if we want to aim for the least ambiguous, brand-neutral term for the device, maybe we would choose `turntable`.

data/entities.yaml
```diff
@@ -1193,3 +1193,23 @@
...
+
+- name: victrola
+  display:
+    attr: device
+    char: 'Q'
+  description:
+  - |
+    A device for reading and writing data on circular platters made of
+    a soft plastic material.  The stylus must be made of an extremely
+    hard and durable material.
```
kostmo, line 1205: Reminds me of [The Portable Phonograph](http://in-sightforum.com/wp-content/uploads/2019/01/The-Portable-Phonograph-L1.pdf) from high school:

> "I have been using thorns as needles," he said. "But tonight, because we have a musician
among us" - he bent his head to the young man, almost invisible in the shadow - "I will use
a steel needle. There are only three left."
>
> The two middle-aged men stared at him in speechless adoration. The one with the big
hands, who wanted to write, moved his lips, but the whisper was not audible.
>
> "Oh, don't," cried the young man, as if he were hurt. "The thorns will do beautifully."
>
> "No," the old man said. "I have become accustomed to the thorns, but they are not really
good. For you, my young friend, we will have good music tonight."
>
> "After all," he added generously, and beginning to wind the phonograph, which creaked,
"they can’t last forever."

src/Swarm/Language/Types.hs
```diff
@@ -129,8 +135,21 @@ data TypeF t
...
 
+-- | Unify two Maps by insisting they must have exactly the same keys,
+--   and if so, simply matching up corresponding values to be
+--   recursively unified.  There could be other reasonable
+--   implementations, but in our case we will use this for unifying
+--   record types, and we do not have any subtyping, so record types
+--   will only unify if they have exactly the same keys.
+instance Ord k => Unifiable (Map k) where
+  zipMatch m1 m2 = do
+    guard $ M.keysSet m1 == M.keysSet m2
```
kostmo, line 150: I'm a fan of using `on`, but up to you:
```suggestion
    guard $ ((==) `on` M.keysSet) m1 m2
```

src/Swarm/Game/Step.hs
```diff
@@ -2290,6 +2310,9 @@ compareValues v1 = case v1 of
     VPair v21 v22 ->
       (<>) <$> compareValues v11 v21 <*> compareValues v12 v22
     v2 -> incompatCmp v1 v2
+  VRcd m1 -> \case
+    VRcd m2 -> mconcat <$> zipWithM compareValues (M.elems m1) (M.elems m2)
```
kostmo, line 2314: Another opportunity for `on`:
```suggestion
    VRcd m2 -> mconcat <$> zipWithM (compareValues `on` M.elems) m1 m2
```

src/Swarm/Language/Types.hs
byorgey: Good idea, thanks!

data/entities.yaml
byorgey: I wasn't familiar with that story, thanks for the reference!  I just went and read the whole thing.  I wonder if we could somehow stick a reference to the story into the description...

data/entities.yaml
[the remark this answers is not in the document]
byorgey: > I wonder if we can still work `quantum dot` in somehow. How about as an ingredient to the `victrola` recipe?

Sounds fun to me.  What should be the recipe for a `quantum dot`?



> I take it that `victrola` has become a somewhat genericized brand name, as "kleenex" has become for facial tissues. There's a whole [Wikipedia section on terminology](https://en.wikipedia.org/wiki/Phonograph#Terminology), but if we want to aim for the least ambiguous, brand-neutral term for the device, maybe we would choose `turntable`.

Sure, I'm aware that `victrola` is a genericized brand name.  Why would it be better to use an unambiguous, brand-neutral term?  To me, `victrola` sounds a lot funnier than `turntable`.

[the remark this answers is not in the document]
byorgey: @kostmo I added a recipe for `quantum dot` (the Wikipedia page tells me one way to make quantum dots is with silicon with nonthermal plasma, which is created by passing an electric current through a gas; and the resulting quantum dots are often suspended in e.g. water; hence the recipe I came up with), and I added `quantum dot` to the `victrola` recipe and updated the description.  Let me know what you think.

data/entities.yaml
```diff
@@ -1193,3 +1193,33 @@
...
+- name: victrola
+  display:
+    attr: device
+    char: 'Q'
+  description:
+  - |
+    A device for reading and writing data on circular platters made of
+    a soft plastic material.  The stylus must be made of an extremely
+    hard and durable material, with a special tip that relies on
+    quantum effects to extract information stored at very high density.
```
kostmo, line 1206: There's a bit of "superlative overload" here.  How about:

```suggestion
    A device for reading and writing data on circular platters made of
    a soft plastic material.  The stylus must be made of a
    hard and durable material, with a special tip that relies on
    quantum effects to extract high-density information.
```

data/entities.yaml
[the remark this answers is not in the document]
kostmo: Looks good.

src/Swarm/Game/Step.hs
```diff
@@ -692,6 +692,26 @@ stepCESK cesk = case cesk of
...
+  -- to focus on evaluating the next one.
+  Out v s (FRcd e done x ((y, t) : rest) : k) ->
+    return $ In (fromMaybe (TVar y) t) e s (FRcd e ((x, v) : done) y rest : k)
+  -- Evaluate a record projection: evaluate the record and remember we
+  -- need to do the projection later.
+  In (TProj t x) e s k -> return $ In t e s (FProj x : k)
+  -- Do a record projection
+  Out v s (FProj x : k) -> case v of
+    VRcd m -> case M.lookup x m of
+      Nothing -> badMachineState s $ "Record projection for variable " <> x <> " that does not exist"
```
kostmo, line 712: To relieve us from "bookkeeping" regarding spaces:
```suggestion
      Nothing -> badMachineState s $ T.unwords ["Record projection for variable", x, "that does not exist"]
```

src/Swarm/Language/Parse.hs
```diff
@@ -217,8 +222,17 @@ parseTypeAtom =
...
+    <|> TyRcd <$> brackets (parseRecord (symbol ":" *> parseType))
     <|> parens parseType
 
+parseRecord :: Parser a -> Parser (Map Var a)
+parseRecord p = (parseBinding `sepBy` symbol ",") >>= fromListUnique
+ where
+  parseBinding = (,) <$> identifier <*> p
+  fromListUnique kvs = case findDup (map fst kvs) of
+    Nothing -> return $ Map.fromList kvs
+    Just x -> fail $ "duplicate field name '" ++ from @Text x ++ "' in record literal"
```
kostmo, line 234: Can eliminate more manual whitespacing:
```suggestion
    Just x -> fail $ T.unwords ["duplicate field name", squote $ from @Text x, "in record literal"]
```

kostmo (approved):

src/Swarm/Language/Parse.hs
byorgey: Indeed, thanks for the reminders!

byorgey:
How come this got merged even though the checks failed!?

byorgey:
Oh, I see, our mergify configuration only specifies that the 8.7, 9.0, and 9.2 builds have to succeed!  So currently the TODO check or the 9.4 build could fail and it would still merge.  Whoops.

src/Swarm/Util.hs
```diff
@@ -185,6 +198,15 @@ isIdentChar c = isAlphaNum c || c == '_' || c == '\''
 replaceLast :: Text -> Text -> Text
 replaceLast r t = T.append (T.dropWhileEnd isIdentChar t) r
 
+-- | Fail with a Text-based message, made out of phrases to be joined
+--   by spaces.
+failT :: MonadFail m => [Text] -> m a
+failT = fail . from @Text . T.unlines
```
kostmo, line 204: Should be `T.unwords`?

byorgey: Yes, it should!  See, this is exactly why I didn't want to merge it yet - I wanted you to look at the new code first!

PR #1155, opened 2023-03-09
Greenhouse

kostmo:
NOT READY FOR REVIEW

## Scenario demo

    scripts/play.sh --scenario data/scenarios/Challenges/Ranching/greenhouse.yaml --autoplay

## Scenario generation tool

    stack run gen-scenario


scripts/merge-yaml.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash -xe
+
+# Obtained from: https://stackoverflow.com/a/67036496/105137
+yq eval-all '. as $item ireduce ({}; . *+ $item)' $@
```
kostmo, line 4: Merges two `.yaml` files.

data/scenarios/Challenges/Ranching/_greenhouse/gen-scenario.sh
```diff
@@ -0,0 +1,26 @@
...
+
+# NOTE: If this command does not work within VS Code console, run this parent script
+# from an independent terminal app.
+# See https://stackoverflow.com/questions/75659711/vscode-yq-fails-with-permission-denied
+./scripts/merge-yaml.sh "$SUPPLEMENTAL_CONTENT_PATH $SCENARIO_TEMPLATE" > $SCENARIO_DESTINATION
+
+stack exec gen-flowers -- --data $FLOWERS_DATA --out $SCRIPT_DIR/planter.sw --codegen-template $SCRIPT_DIR/code/planter-template.sw.j2
+
+# Place the solution checking script inline into the YAML file as an optimization
+PROGRAM=$(cat $SCRIPT_DIR/solution-checker.sw | sed -e 's/[[:blank:]]\+$//') yq -i '.objectives[0].condition = strenv(PROGRAM) | .objectives[].condition style="literal"' $SCENARIO_DESTINATION
```
kostmo, line 24: Placing `.sw` code inline into `.yaml` file

data/scenarios/Challenges/Ranching/_greenhouse/flowers.json
```diff
@@ -0,0 +1,175 @@
+[
```
kostmo, line 1: flower data

byorgey:
> * I'm emulating a "case" statement with 40 cases via a chain of if/else statements.  I haven't measured this rigorously, but this seems to result in a slowdown proportional to the value of the case index.

That makes perfect sense since it will literally have to evaluate each `if` in sequence, so it will take linear time to find the right case.  We might consider adding a `switch` expression of some sort, but I'm not sure off the top of my head how it would work.

FYI, the `ginger` dependency causes problems for building with `stack`, since it has several restrictive upper bounds that are incompatible with our chosen stack resolver.  I can test with `cabal` for now.

byorgey:
I managed to complete this challenge but I found it confusing and tedious.  I am not sure whether there is supposed to be a less-tedious way to complete it; if so I have not figured it out yet.
- I found it tedious and annoying that the goal just says to leave the mushrooms but it turns out there are multiple types of mushrooms so I have to learn all their names and then hardcode them in a condition.
    - Relatedly, it is confusing when you fail the mushroom harvesting goal: the goal dialog says you have failed the flower curation goal, but the real reason is that you have completed the mushroom harvesting goal, which is really an anti-goal, but the UI does not really make this clear.
- I don't understand what I am supposed to do with a flower pouch.  Apparently if I make one I can then drill flowers to turn them into other flowers??  This makes no sense to me and I don't understand how it's supposed to help.
- I initially failed the "leave no trace" goal when I used `harvest`, but I quickly figured out that I was supposed to use `swap` with a `decoy`, which was fun.
- I wrote some code to traverse the entire garden plot, `scan` each cell, and only `swap` with a decoy if it wasn't a mushroom and wasn't something I already have.  That was rewarding and worked just like I wanted.
- Then I had 40 different flowers and I just manually planted them one by one in the planters.  That part was super tedious.

OK, so I just looked at the solution and I see now how it's possible to solve without tediously placing each flower manually.  It seems a little unmotivated though.  Why would a "flower pouch" let you turn certain flowers into certain other flowers?

byorgey:
OK, I have now read the design commentary and I understand better what you're going for here.  I think the core mechanic is an interesting idea, but I guess when I actually tried playing the scenario the core idea didn't occur to me and I just ended up having a frustrating experience.  I'm not sure how to improve the experience off the top of my head but I will keep thinking about it.

As for the generation code, I am torn.  On the one hand having it live in the same repo might be a good idea since then we are forced to ensure that it stays up-to-date with any changes we make to the language etc.  On the other hand, for the same reason, it creates an additional burden/barrier for anyone wanting to build from source (builds will take longer, or else we have to provide instructions for only building the main `swarm` component, etc.).  Putting it in a different repo under the `swarm-game` organization could be a good compromise.

kostmo:
> the core idea didn't occur to me and I just ended up having a frustrating experience

Thanks for persisting through the tedium to provide an authentic and valuable experience report.  On the one hand, it validates that the scenario mechanic is exactly as cumbersome as designed.  On the other hand, sorry for putting you through that!  The "shortcut" needs to be more discoverable.  Will think more on that.

> Putting it in a different repo under the `swarm-game` organization could be a good compromise.

Let's do this.  Would you please add me to the [people](https://github.com/orgs/swarm-game/people)?

byorgey:
> On the other hand, sorry for putting you through that! The "shortcut" needs to be more discoverable. Will think more on that.

I will think about it too!

> Would you please add me to the [people](https://github.com/orgs/swarm-game/people)?

Of course!  Done (I think).  I honestly don't know why some contributors ended up being "members" and some didn't.

kostmo:
>> a chain of if/else statements [...] seems to result in a slowdown proportional to the value of the case index.
>
> That makes perfect sense

I find it surprising that it consumes more "game ticks".  I could imagine it slowing down the overall "simulation" of the game, but is the evaluation of an `if` supposed to be "tangible", in that it makes one robot slower relative to another?

byorgey:
> is the evaluation of an `if` supposed to be "tangible", in that it makes one robot slower relative to another?

No, but there is a limit on how many CESK machine steps a robot can take in one game tick (currently 100).  This is so that you can't freeze the game by running very long computations without doing anything. But it also means that long computations will take up more ticks. There is a lot of overhead so I would not be surprised if evaluating a 40-case if takes many hundreds of CESK steps.

PR #1162, opened 2023-03-17
let-in syntax highlighting

kostmo:
Currently, the syntax highlighter matches any word prefixed with `in`, e.g. `inL` or `inner` is matched.  This PR adds a word boundary to the regex.


byorgey (approved):

xsebek:
We could add a few tests like this - there should be a guide to do a snapshot in DEVELOPING.md

I wonder if there are other word boundary issues.

Issue #10, opened 2021-09-20
Add type ascription syntax to the language


byorgey:
It would be nice to be able to annotate arbitrary expressions with type annotations, as in `(3 : int) + 5`.  It's not clear how necessary this is at the moment --- type inference, plus giving type annotations to top-level definitions, usually seems to be able to do the job.  But as we add more polymorphic things it may become increasingly helpful.  And in any event it can be helpful as a form of documentation.

Adding this would require

* adding an AST node to represent a type annotation
* adding a case for annotations to the parser
* adding a case to the pretty-printer
* adding a case for type inference: hitting a type annotation simply makes it call the `check` function.

byorgey:
Some examples of doing this in https://github.com/disco-lang/disco/ :

- https://github.com/disco-lang/disco/blob/3810177216dd3e3632404d60f6866d0ec9bb7922/src/Disco/Parser.hs#L513-L520
- In the pretty-printer, we treat it as having precedence level 1.

byorgey:
I still don't think I've ever wanted this.  So it's pretty low priority, but still good to have on principle (unless it somehow ends up significantly complicating parsing or something).

Issue #211, opened 2021-10-11
Escaping Skolem variables lead to incorrect types


byorgey:
[Ilya Smirnov left a comment](https://byorgey.wordpress.com/2021/09/08/implementing-hindley-milner-with-the-unification-fd-library/#comment-40016) on my blog post about implementing Hindley-Milner with the `unification-fd` library pointing out a bug.  Since the implementation of type inference for swarm-lang was taken more or less directly from that blog post, the bug exists in Swarm as well.  It's a bit obscure and unlikely to be tripped over by casual players, but it's a bug nonetheless.  Trying to execute
```
 let f = \y. (let f : a -> a = y in f 3) in f (\x.\y. x y) + 1
```
(in creative mode so arithmetic is allowed) results in a fatal error:
```
 fatal error: Bad application of execConst:
  ◀ let x = 3 in \y. x y + 1
  []
```
I will follow up with a fix shortly, but wanted to officially record the bug for completeness.

byorgey:
`let f = \y. (let f : a -> a = y in f 3) in f (\x.\y. x y) + 1` now results in an error (as it should): `Skolem variable s2 would escape its scope`.

PR #1164, opened 2023-03-17
Add Type Ascription Syntax to Resolve Issue #10

Alexander-Block:
This pull request implements type ascription syntax and should resolve Issue #10.

### DIFF
--- a/src/Swarm/Language/Syntax.hs
+++ b/src/Swarm/Language/Syntax.hs
@@ -788,6 +788,8 @@ data Term' ty
     --   be a special syntactic form so its argument can get special
     --   treatment during evaluation.
     SDelay DelayType (Syntax' ty)
+  | -- | Annotate a term with a type
+    SAnnotate (Syntax' ty) Polytype
   deriving (Eq, Show, Functor, Foldable, Traversable, Data, Generic, FromJSON, ToJSON)
 
 -- The Traversable instance for Term (and for Syntax') is used during
@@ -960,6 +962,7 @@ erase (SApp s1 s2) = TApp (eraseS s1) (eraseS s2)
 erase (SLet r x mty s1 s2) = TLet r (lvVar x) mty (eraseS s1) (eraseS s2)
 erase (SDef r x mty s) = TDef r (lvVar x) mty (eraseS s)
 erase (SBind mx s1 s2) = TBind (lvVar <$> mx) (eraseS s1) (eraseS s2)
+erase (SAnnotate s _) = eraseS s
 
 ------------------------------------------------------------
 -- Free variable traversals
@@ -1001,6 +1004,7 @@ freeVarsS f = go S.empty
     SDef r x xty s1 -> rewrap $ SDef r x xty <$> go (S.insert (lvVar x) bound) s1
     SBind mx s1 s2 -> rewrap $ SBind mx <$> go bound s1 <*> go (maybe id (S.insert . lvVar) mx bound) s2
     SDelay m s1 -> rewrap $ SDelay m <$> go bound s1
+    SAnnotate s1 _ -> go bound s1
    where
     rewrap s' = Syntax' l <$> s' <*> pure ty
 
--- a/src/Swarm/Language/Syntax.hs
+++ b/src/Swarm/Language/Syntax.hs
@@ -967,7 +967,7 @@ erase (SApp s1 s2) = TApp (eraseS s1) (eraseS s2)
 erase (SLet r x mty s1 s2) = TLet r (lvVar x) mty (eraseS s1) (eraseS s2)
 erase (SDef r x mty s) = TDef r (lvVar x) mty (eraseS s)
 erase (SBind mx s1 s2) = TBind (lvVar <$> mx) (eraseS s1) (eraseS s2)
-erase (SAnnotate s _) = eraseS s
+erase (SAnnotate s pty) = TAnnotate (eraseS s) pty
 
 ------------------------------------------------------------
 -- Free variable traversals
--- a/src/Swarm/Language/Syntax.hs
+++ b/src/Swarm/Language/Syntax.hs
@@ -56,6 +56,7 @@ module Swarm.Language.Syntax (
   pattern TDef,
   pattern TBind,
   pattern TDelay,
+  pattern TAnnotate,
 
   -- * Terms
   Var,
@@ -900,8 +901,12 @@ pattern TBind mv t1 t2 <- SBind (fmap lvVar -> mv) (STerm t1) (STerm t2)
 pattern TDelay :: DelayType -> Term -> Term
 pattern TDelay m t = SDelay m (STerm t)
 
+-- | Match a TAnnotate without syntax
+pattern TAnnotate :: Term -> Polytype -> Term
+pattern TAnnotate t pt = SAnnotate (STerm t) pt
+
 -- | COMPLETE pragma tells GHC using this set of pattern is complete for Term
-{-# COMPLETE TUnit, TConst, TDir, TInt, TAntiInt, TText, TAntiText, TBool, TRequireDevice, TRequire, TVar, TPair, TLam, TApp, TLet, TDef, TBind, TDelay #-}
+{-# COMPLETE TUnit, TConst, TDir, TInt, TAntiInt, TText, TAntiText, TBool, TRequireDevice, TRequire, TVar, TPair, TLam, TApp, TLet, TDef, TBind, TDelay, TAnnotate #-}
 
 -- | Make infix operation (e.g. @2 + 3@) a curried function
 --   application (@((+) 2) 3@).
--- a/src/Swarm/Language/Syntax.hs
+++ b/src/Swarm/Language/Syntax.hs
@@ -1009,7 +1009,7 @@ freeVarsS f = go S.empty
     SDef r x xty s1 -> rewrap $ SDef r x xty <$> go (S.insert (lvVar x) bound) s1
     SBind mx s1 s2 -> rewrap $ SBind mx <$> go bound s1 <*> go (maybe id (S.insert . lvVar) mx bound) s2
     SDelay m s1 -> rewrap $ SDelay m <$> go bound s1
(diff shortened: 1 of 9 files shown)


src/Swarm/Language/Syntax.hs
```diff
@@ -1001,6 +1009,7 @@ freeVarsS f = go S.empty
     SDef r x xty s1 -> rewrap $ SDef r x xty <$> go (S.insert (lvVar x) bound) s1
     SBind mx s1 s2 -> rewrap $ SBind mx <$> go bound s1 <*> go (maybe id (S.insert . lvVar) mx bound) s2
     SDelay m s1 -> rewrap $ SDelay m <$> go bound s1
+    SAnnotate s1 _ -> go bound s1
```
byorgey, line 1012: This should be something more like `rewrap $ SAnnotate <$> go bound s1 <*> pure ty` (untested).  There's no reason that traversing over free variables should remove `SAnnotate` nodes.

test/unit/TestLanguagePipeline.hs
```diff
@@ -282,6 +282,12 @@ testLanguagePipeline =
                       ]
                     )
             )
+        , testCase
+            "simple type ascription"
+            (valid "(3 : int) + 5")
+        , testCase
+            "invalid type ascription"
+            (process "1 : text" "1: Can't unify text and int")
```
byorgey, line 290: Can we add a few more tests?  e.g. let's at least have a test where we annotate something with a polytype.  Actually, trying it out just now, this doesn't seem to work:
```
((\x . x) : a -> a) 3
```
This ought to return `3`, but instead we get an error `Can't unify s0 and int`.  I think I see what the issue is, I will add a comment to the typechecker.

src/Swarm/Language/Typecheck.hs
```diff
@@ -452,6 +452,9 @@ infer s@(Syntax l t) = (`catchError` addLocToTypeErr s) $ case t of
     c2' <- maybe id ((`withBinding` Forall [] a) . lvVar) mx $ infer c2
     _ <- decomposeCmdTy (c2' ^. sType)
     return $ Syntax' l (SBind mx c1' c2') (c2' ^. sType)
+  SAnnotate c pty -> do
+    uty <- skolemize $ toU pty
+    check c uty
```
byorgey, line 457: This is the right idea but I think it has a couple issues:
1. The general idea is that the typechecker annotates the AST with types, but otherwise leaves the AST intact.  I think even in the annotation case we should leave the annotation there instead of stripping it off (for example, because we might later want to pretty-print the term in an error message and make sure we preserve the way the user wrote it; or even more relevant, for #7 we might want to later pretty-print the term to save it in a file, in which case it might not even typecheck any more if we got rid of the annotation).  So instead of just `check c uty` this should say something like `c' <- check c uty; return $ Syntax' l (SAnnotate c' pty) ...`
2. The second issue is that after skolemizing the type in order to check it, we need to re-generalize so that we don't end up later trying to match skolem variables against other types when the annotated thing is used.  In other words, the skolemized type should only be used internally to *check* that the annotated thing is valid with the given type annotation; it should not be used *externally* as the inferred type of the thing.  Actually it may be as simple as using `toU pty` as the external type.  So something like `return $ Syntax' l (SAnnotate c' pty) (toU pty)`.  I haven't tested this, however.

byorgey (changes requested):
Thanks!  I was just thinking about whether I should implement this myself since it might help with #1148 , so I'm really glad someone else did instead! :smile:   This is looking very nice, there are just a few things I think need to be addressed before we can merge.

src/Swarm/Language/Syntax.hs
Alexander-Block: I completely agree. Thank you for pointing out this mistake. I think that your proposed fix works and will change this line accordingly.

src/Swarm/Language/Typecheck.hs
Alexander-Block: Thank you! Again I immediately agree with your point 1 that it is not a good idea to erase type signatures at this point. I should have paid closer attention to the `SLam` and `SLet` cases to realize this. 

Regarding your point 2 I had to think and experiment a bit to get to a workable solution. Your proposal doesn't quite work, since `toU pty` is of type `UPolytype` where `UType` is needed. I think the actual problem is that I invoke `skolemize` to replace universally quantified type variables with instances that cannot unify with anything other than themselves; surely we have a problem if they escape to the external world. In think we should rather `instantiate` them with unification variables. 

However, if I implement it this way, I get strange errors about a Skolem variable escaping its scope under certain circumstances.  An example for this is the following expression:
```
((\x.x):(a->a)->(a->a))(\x:a->a.x)
```
I can evaluate it once and get the type `(a->a)->a->a`, while a second evaluation of this expression raises the error `Skolem variable a would escape its scope`. It seems to me that the reason for this problem lies not in the new type ascription but rather in how type annotations _inside_ a lambda expression are handled. As the corresponding constructor in the AST is `SLam LocVar (Maybe Type) (Syntax' ty)` and not something like `SLam LocVar (Maybe Polytype) (Syntax' ty)` a type annotation in this context is not polymorphic and free type variables seem to be interpreted as a kind of type constant. 

I would like to look into this problem a little longer to evaluate whether it makes sense to solve it in this merge request or enter a separate issue for it.

test/unit/TestLanguagePipeline.hs
Alexander-Block: I have added this test case, but I will add more test cases tomorrow.

Alexander-Block:
I have noticed that type inference in the presence of type acriptions still doesn't quite work as expected after my last change. Now it is too permissive in that for example `3:a` typechecks. I think I should go through the type inference again step by step for some examples and compare them to the Hendley-Milner rules to figure out the correct solution at this point. 

For the moment I convert this pull request to a draft, since in this state it clearly should not be merged.

src/Swarm/Language/Typecheck.hs
```diff
@@ -453,8 +453,10 @@ infer s@(Syntax l t) = (`catchError` addLocToTypeErr s) $ case t of
     _ <- decomposeCmdTy (c2' ^. sType)
     return $ Syntax' l (SBind mx c1' c2') (c2' ^. sType)
   SAnnotate c pty -> do
-    uty <- skolemize $ toU pty
-    check c uty
+    uty <- instantiate $ toU pty
```
byorgey, line 456: Right, this isn't going to work because it allows things like `3 : a`, as you noted.  If `a` turns into a unification variable then it will be free to unify with `int`.  We really do need `skolemize` here for checking the type of `c`, so that we can check that it really is polymorphic as advertised.  But then we need to throw away the type we got from checking `c` and return a version which has been `instantiate`d.  So maybe something like this (untested):

```
sty <- skolemize $ toU pty
uty <- instantiate $ toU pty
c' <- check c sty
return $ Syntax' l (SAnnotate c' pty) uty
```

An alternative would be to just say that polytypes are not allowed in type annotations.  That would make things a lot simpler.  I don't have a good sense of how big/annoying of a restriction that would be.

Alexander-Block: Thank you for this explanation. In fact I have just pushed a commit that is very similar to your proposed solution (which I have just seen afterwards). What really helped me understand what is going on here is that I read through your rather detailed and (at least in my opinion) well-written [blog post](https://byorgey.wordpress.com/2021/09/08/implementing-hindley-milner-with-the-unification-fd-library) on your Hendley-Milner implementation. 

I have added a few more test cases to validate the correctness of my last change. Essentially it only differes from your proposal by the presence of a check for escaping Skolem variables. Without it, `\f. (f:forall a. a->a) 3` would typecheck, which it really shouldn't without higher ranked types. If I understand it correctly, this would be the same problem as in #211.

test/unit/TestLanguagePipeline.hs
Alexander-Block: I have added a few additional test cases. Do you still see a gap in the test coverage?

src/Swarm/Language/Typecheck.hs
Alexander-Block: I see another difference now which puzzles me a little bit. You only check `c` against `sty` and use the result in the resulting AST node. I did something similar in an intermediate version of my implementation, but wondered whether this is a good idea, since `check` puts the unified type into the resulting AST node. So wouldn't `Syntax' l (SAnnotate c' pty) uty` lead to a possibly problematic missmatch between `uty` and `c' ^. sType` ?

src/Swarm/Language/Syntax.hs
```diff
@@ -960,6 +967,7 @@ erase (SApp s1 s2) = TApp (eraseS s1) (eraseS s2)
 erase (SLet r x mty s1 s2) = TLet r (lvVar x) mty (eraseS s1) (eraseS s2)
 erase (SDef r x mty s) = TDef r (lvVar x) mty (eraseS s)
 erase (SBind mx s1 s2) = TBind (lvVar <$> mx) (eraseS s1) (eraseS s2)
+erase (SAnnotate s _) = eraseS s
```
byorgey, line 970: Oh, I didn't notice this before, we should preserve the `SAnnotate` constructor here too, for the same reasons as elsewhere.  Erasing types just means erasing the inferred types the typechecker filled in, not erasing an explicit type annotation the user wrote.

src/Swarm/Language/Typecheck.hs
byorgey: Ah, glad you found the blog post useful!

Checking for escaping skolems makes sense, and of course you're right about the potential problems with my version having a mismatch between `uty` and `c' ^. sType`.  I think the way you have done it makes sense.  In an abstract sense it bothers me somewhat to call `check` twice on `c`, once with a skolemized type and once with an instantiated type---but I don't actually know of a better way to do it!

byorgey (approved):
Looks good!  There's just the one remaining issue of preserving the `SAnnotate` constructor in `erase`, but feel free to just fix that up and then merge.

PR #1165, opened 2023-03-18
Drill command returns first inventory addition

### LINKED ISSUE #1157: drill command should return the name of the entity obtained
When drilling, either an entity will be obtained in one's inventory, or it will be changed in place on the ground.

This proposal is to change the type of the `drill` command from

    dir -> cmd unit

to

     dir -> cmd (unit + text)

The command shall return either `unit` (if nothing was brought into the inventory) or the name of the entity obtained.
Since it is possible for recipes to have multiple `portable` outputs, the entity returned by the command shall be the **first** portable output listed in the recipe.

This new behavior aligns it with the `grab` command, which also returns the name of the thing that was added to the inventory.

kostmo:
Closes #1157

Demo:

    scripts/play.sh --scenario data/scenarios/Testing/1157-drill-return-value.yaml --autoplay

### DIFF
--- a/data/scenarios/Challenges/Ranching/_gated-paddock/enclosure-checking.sw
+++ b/data/scenarios/Challenges/Ranching/_gated-paddock/enclosure-checking.sw
@@ -59,6 +59,7 @@ def leaveBreadcrumbs =
                     // Nothing will remain on the ground.
                     // after making the "steam" via
                     // the drilling recipe.
+                    return ();
                 } {
                     grab;
                     return ();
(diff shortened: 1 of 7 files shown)


byorgey (approved):
Looks good.  Should we also update the description of the `drill` entity at all?

kostmo:
> update the description

Maybe not the "entity", but I have now also added to the description of the `drill` command.

byorgey:
> Maybe not the \"entity\", but I have now also added to the description of the drill command.

Ah, that makes sense.  Some entity descriptions include examples of using associated commands, but I see that is not the case with the `drill`.

PR #1168, opened 2023-03-19
simplify color parsing

kostmo:
I happened upon these utility functions from the `colour` package while working on #1155.  Also adding some JSON instances for the purpose of the same PR.

### DIFF
--- a/swarm.cabal
+++ b/swarm.cabal
@@ -167,6 +167,7 @@ library
                       brick                         >= 1.5 && < 1.7,
                       bytestring                    >= 0.10 && < 0.12,
                       clock                         >= 0.8.2 && < 0.9,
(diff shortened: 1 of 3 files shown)


swarm.cabal
```diff
@@ -167,6 +167,7 @@ library
                       brick                         >= 1.5 && < 1.7,
                       bytestring                    >= 0.10 && < 0.12,
                       clock                         >= 0.8.2 && < 0.9,
+                      colour                        >= 2.3.6 && < 2.3.7,
```
byorgey, line 170: These version bounds seem overly restrictive.  At least we should put `< 2.4`, right?

byorgey (approved):

swarm.cabal
kostmo: done

PR #1170, opened 2023-03-20
Implement 'detect' command

### LINKED ISSUE #1140: isPresent command
Let's define a command that can answer whether a certain entity exists within a rectangular region.

It takes arguments of a scalar integer "distance" (using [Chebyshev distance](https://en.wikipedia.org/wiki/Chebyshev_distance)) and an entity name:

    isPresent :: int -> text -> bool

It would require `god` capability, or perhaps some other system-robot-exclusive capability that we might call `"detect"` or `"senseremote"` (a counterpart to `sensefront` and `sensehere`).  Come to think of it, maybe the command itself should be named `detect`?

There are probably good non-system applications for such a command, provided that the radius is capped.

Perhaps an [olfactometer](https://en.wikipedia.org/wiki/Machine_olfaction) device could provide the `senseremote` capability.

## Motivation

I found myself wanting to create a Goal that involved completely clearing an irregularly-shaped forest (not just harvesting, but removing entirely).  However, to check this with a system robot would entail a raster scan over the area in question and is inefficient.

## Alternative approaches

Another, more precise alternative may be to supply exact coordinates:

    isPresent :: (int * int) -> (int * int) -> text -> bool

The coordinates would be relative to the robot issuing the command.

kostmo:
Closes #1140

### DIFF
--- a/src/Swarm/Language/Typecheck.hs
+++ b/src/Swarm/Language/Typecheck.hs
@@ -532,6 +532,7 @@ inferConst c = case c of
   Create -> [tyQ| text -> cmd unit |]
   Time -> [tyQ| cmd int |]
   Whereami -> [tyQ| cmd (int * int) |]
+  Detect -> [tyQ| text -> ((int * int) * (int * int)) -> cmd bool |]
   Heading -> [tyQ| cmd dir |]
   Blocked -> [tyQ| cmd bool |]
   Scan -> [tyQ| dir -> cmd (unit + text) |]
--- a/src/Swarm/Language/Typecheck.hs
+++ b/src/Swarm/Language/Typecheck.hs
@@ -532,7 +532,7 @@ inferConst c = case c of
   Create -> [tyQ| text -> cmd unit |]
   Time -> [tyQ| cmd int |]
   Whereami -> [tyQ| cmd (int * int) |]
-  Detect -> [tyQ| text -> ((int * int) * (int * int)) -> cmd bool |]
+  Detect -> [tyQ| text -> ((int * int) * (int * int)) -> cmd (unit + (int * int)) |]
   Heading -> [tyQ| cmd dir |]
   Blocked -> [tyQ| cmd bool |]
   Scan -> [tyQ| dir -> cmd (unit + text) |]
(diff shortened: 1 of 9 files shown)


test/integration/Main.hs
```diff
@@ -263,6 +263,8 @@ testScenarioSolution _ci _em =
         , testSolution Default "Testing/956-GPS"
         , testSolution Default "Testing/958-isempty"
         , testSolution Default "Testing/1024-sand"
+        , testSolution Default "Testing/1140-detect-command"
+        , testSolution Default "Testing/1157-drill-return-value"
```
kostmo, line 267: Forgot to add this drill test in #1165/#1167, so sneaking it in here.

src/Swarm/Language/Typecheck.hs
```diff
@@ -532,6 +532,7 @@ inferConst c = case c of
   Create -> [tyQ| text -> cmd unit |]
   Time -> [tyQ| cmd int |]
   Whereami -> [tyQ| cmd (int * int) |]
+  Detect -> [tyQ| text -> ((int * int) * (int * int)) -> cmd bool |]
```
byorgey, line 535: What do you think about making the type something like `text -> ((int * int) * (int * int)) -> cmd (unit + (int * int))` instead?  That is, instead of just returning a boolean, if it finds the thing you're looking for it actually tells you (the offset to) where it is.  That seems like it would be strictly more useful.

byorgey (approved):

src/Swarm/Language/Typecheck.hs
kostmo: Good idea.

Now that I think of it, eventually we could expose a spectrum of different sensory commands with various precision/constraints.  Some ideas:

```
sniff : text -> cmd int
```
Olfactory.  Physically interpreted as a "point sensor", thus cannot infer direction.  Return the **distance** to the nearest entity of a given name.  To use it, the player can move around in different directions to see whether they are getting "hot" or "cold".

```
chirp : text -> cmd dir
```
Sonar based on two "ears" will return the **direction** to the nearest entity of a given name.  To use it, the player can move in said direction until the reported direction reverses, and then progressively home in on it (or switch to another close-range sensing modality).

```
resonate : text -> ((int * int) * (int * int)) -> cmd bool
```
The original Boolean type signature.  Usage would entail a [quadtree](https://en.wikipedia.org/wiki/Quadtree) search strategy.  The physical justification is more of a stretch, but could entail something like MRI, where a strong magnetic field is applied over a given area, and a stimulating pulse is emitted, exciting the particles of the entity in question.  A non-directional signal is returned from the excited area, with a receiver being tuned to the frequency of the entity's particles.

PR #1172, opened 2023-03-21
Offload value conversions to a typeclass

kostmo:


src/Swarm/Game/Values.hs
```diff
@@ -0,0 +1,35 @@
+{-# OPTIONS_GHC -fno-warn-orphans #-}
+
+-- |
+-- SPDX-License-Identifier: BSD-3-Clause
+module Swarm.Game.Values where
```
byorgey, line 5: It's not a big deal, but for the most part we have been following the convention where nouns in module names are singular (*e.g.* `Swarm.Game.Robot` instead of `Robots`), and in particular we have `Swarm.Language.Value`, so it seems odd to me to have this module name be plural.

src/Swarm/Game/Values.hs
```diff
@@ -0,0 +1,35 @@
...
+
+import Control.Lens (view)
+import Data.Int (Int32)
+import Linear (V2 (..))
+import Swarm.Game.Entity
+import Swarm.Game.Robot
+import Swarm.Language.Value
+
+class Valuable a where
```
byorgey, line 14: Can we add a Haddock comment to the `Value` class explaining its purpose?  I can see what it's for but I'm thinking from the perspective of potential contributors who are trying to understand the codebase.

byorgey (approved):
What a great idea, I love this!

src/Swarm/Game/Values.hs
kostmo: done

src/Swarm/Game/Values.hs
kostmo: done

Issue #1105, opened 2023-02-13
"turn down" is unrecoverable without compass


kostmo:
**Describe the bug**
`turn down` sets the direction vector to zero, and no other command using relative directions can restore it.  You have now become literally "disoriented"!  Setting an absolute direction by way of the `compass` allows one to recover.

This issue has probably existed for a long time.

**To Reproduce**
The Ice Cream Shop scenario entails a mobile `base` without a `compass`.

    scripts/play.sh --scenario Challenges/ice-cream.yaml

then

    turn down

Now the base can no longer move in any direction.  Somewhat appropriately, the base is now displayed as `X`.

**Expected behavior**

May want to either exclude `turn down` as a command, either entirely or if a compass is not equipped.

On the other hand, if we can find a use for `turn down`, getting "stuck" this way may make for an amusing scenario.  Say, playing ostrich?

xsebek:
Not an issue, as you point out, you could use it to implement some algorithms where you let the robot unwind its recursive stack at the end.

Leaving it out makes less sense since you should not need a compass to look down. :slightly_smiling_face:

Since it's really easy to not use it and even Tab complete will not suggest it (N,E,S,W) I am going to close this.
We can certainly revisit `down` if/when we do multiple Z-levels. In the meantime, we can use it to play ostrich.

kostmo:
I suppose this phenomenon is analogous to [2-D gimbal lock](https://en.wikipedia.org/wiki/Gimbal_lock#In_two_dimensions).

kostmo:
Related (historical): #332

Issue #916, opened 2022-12-14
A toplevel notification system (for achievements, etc.)


kostmo:
Should have either a status bar or **non-modal** popup/dialog overlays (c.f. [Toast](https://developer.android.com/guide/topics/ui/notifiers/toasts) on Android) to alert the user of certain events (e.g. attaining achievements (#797)).

This should not be the same as the "robot messages" window.

xsebek:
As a note, one implementation that we discussed could be to add another layer to the entire game.

Currently, only the running game has two layers (dialogues and base layer) while the main menu only has one layer.

This way, achievements could popup even on the `About` page.

byorgey:
Hmm, yes, adding another layer with non-modal popups could be a good, consistent way to add notifications across the entire game.  A few more thoughts from #1506:

- This would also work for notifications when new recipes or entities are unlocked.
- We'd have to think about what should happen when multiple notifications are triggered at once (e.g. an achievement + new recipes unlocked).  The simplest idea would probably be to just have a queue of notifications and display each for a certain amount of time.
- Now that we have pretty-printing for Markdown we could make use of that for notification messages.
- If we use pop-ups, I suppose putting them at or near the top of the window would probably be the least disruptive place.

xsebek:
> If we use pop-ups, I suppose putting them at or near the top of the window would probably be the least disruptive place.

How about popup only for warning level greater or equal to user configured value? 🙂 (default error)

A UI at my work has it like that with buttons Clear and Hide.

> We'd have to think about what should happen when multiple notifications are triggered at once.

If the popup is a scrollable list, then the preview could just show the categories: `I 3  W 1  E 0` (or ℹ️ ⚠️ 🛑)

Alternatively I would be fine with a preview of the latest most serious notification, which would only be explicitly cleared by the player. Clearing notification could be as simple as right mouse click.

byorgey:
> How about popup only for warning level greater or equal to user configured value? 🙂 (default error)

I was not thinking about popups in terms of errors at all, but rather for notifications of __good__ things like unlocking achievements and recipes.  After merging #1487, most errors and warnings can now be shown inline in the REPL.  Were there additional sorts of "out-of-band" errors that you could imagine most effectively being shown via popups?

Also, for notifications of achievements, recipes, etc. I personally would rather have them just pop up for a certain (configurable) amount of time and then automatically disappear, rather than having to explicitly click to close them.

xsebek:
[...]

PR #1173, opened 2023-03-24
Achievement for disorientation

kostmo:
Inspired by #1105

image


byorgey (approved):
Looks nice.  (At first I thought it didn't work but then I remembered we don't have #916 yet so when you get the achievement nothing happens.)

PR #1175, opened 2023-03-24
Display scenario for obtained achievement

kostmo:
# Before
image
# After
image


byorgey (approved):
Looks good to me.  This made me think that perhaps we should also display the name of the scenario file in the "new game" menu, in case you want to play a scenario from the command line.  Currently there's no good way to figure out the name of the file corresponding to a given scenario other than looking through the scenario files manually.

Issue #1171, opened 2023-03-20
Implement more sensory commands


kostmo:
- [x] `sniff` command (#1181)
- [x] `sniff` capability (#1181)
- [ ] `sniff` recipe (device defined in #1263)
- [x] `chirp` command (#1181)
- [x] `chirp` capability (#1181)
- [ ] `chirp` recipe
- [x] `resonate` command (#1204)
- [x] `resonate` capability (#1204)
- [ ] `resonate` recipe

From [this comment](https://github.com/swarm-game/swarm/pull/1170#discussion_r1142555693):
> 
> ```
> sniff : text -> cmd int
> ```
> Olfactory.  Physically interpreted as a "point sensor", thus cannot infer direction.  Return the **distance** to the nearest entity of a given name.  To use it, the player can move around in different directions to see whether they are getting "hot" or "cold".
> 
> ```
> chirp : text -> cmd dir
> ```
> Sonar based on two "ears" will return the **direction** to the nearest entity of a given name.  To use it, the player can move in said direction until the reported direction reverses, and then progressively home in on it (or switch to another close-range sensing modality).
> 
> ```
> resonate : text -> ((int * int) * (int * int)) -> cmd bool
> ```
> The original Boolean type signature.  Usage would entail a [quadtree](https://en.wikipedia.org/wiki/Quadtree) search strategy.  The physical justification is more of a stretch, but could entail something like MRI, where a strong magnetic field is applied over a given area, and a stimulating pulse is emitted, exciting the particles of the entity in question.  A non-directional signal is returned from the excited area, with a receiver being tuned to the frequency of the entity's particles.

byorgey:
Collecting some ideas here for recipes + new devices (will fill in more detail later).

- `rolex`:
    - 1 `clock`
    - 1 `scanner`
    - This would be kind of difficult to make, but `watch` doesn't really add a whole lot of expressive capability, it's just more efficient (without `watch` you can just use a `scanner` to poll the adjacent location).
- `olfactometer`:
- `dozer blade`:
- Proposed device for `chirp`:
- Proposed device for `resonate`:

PR #1181, opened 2023-03-27
sniff/chirp commands

kostmo:
towards #1171

## New tests

Unit tests:

    scripts/run-tests.sh --test-arguments '--pattern "Relative direction"'

Integration tests for `sniff` and `chirp` demonstrate how to home in on an item using distance and orientation, respectively.

## Efficiency

For the sake of execution time, I have capped the max "diameter" (`N`) of both commands to `200` cells.  In the worst case (the entity is not present), `O(N^2)` cells are inspected, which manifests as a perceptible delay when the command is run.  I came across the `getElemsInArea` function that seems to suggest that an `O(N * log N)` search may be possible.  Is that the case?

Otherwise we may be able to add some new data structures to the game state for efficient entity location querying.


src/Swarm/Game/Location.hs
```diff
@@ -139,6 +141,28 @@ cardinalDirs =
...
 toDirection v = M.lookup v cardinalDirs
 
+-- | Example:
+--      DWest `relativeTo` DSouth == DRight
+relativeTo :: AbsoluteDir -> AbsoluteDir -> RelativeDir
+relativeTo targetDir referenceDir =
+  [DForward, DLeft, DBack, DRight] !! indexDiff
+ where
+  enumCount = length (Util.listEnums :: [AbsoluteDir])
+  indexDiff = (((-) `on` fromEnum) targetDir referenceDir + enumCount) `mod` enumCount
```
byorgey, line 151:
```suggestion
  indexDiff = ((-) `on` fromEnum) targetDir referenceDir `mod` enumCount
```
Adding `enumCount` before doing the `mod` is a workaround only needed for languages that do not have a sensible semantics for `mod` (*e.g.* Java, C).

src/Swarm/Game/Location.hs
```diff
@@ -139,6 +141,28 @@ cardinalDirs =
...
+
+-- | Logic adapted from:
+-- https://gamedev.stackexchange.com/questions/49290/#comment213403_49300
+nearestDirection :: Heading -> AbsoluteDir
+nearestDirection coord =
+  -- Using `cycle` obviates the `mod` operation
+  cycle orderedDirs !! index
```
byorgey, line 158: Personally I think just using `mod` on the rounded index would be clearer, especially since we don't have to worry about adding `enumCount` before performing the `mod`.

src/Swarm/Language/Syntax.hs
```diff
@@ -100,14 +102,27 @@ import Swarm.Language.Types
 import Swarm.Util qualified as Util
 import Witch.From (from)
 
+-- | Maximum perception distance for
+-- 'chirp' and 'sniff' commands
+maxSniffRange :: Int32
+maxSniffRange = 200
```
byorgey, line 108: Most other constants in the game are powers of 2. Would it make sense to use either 128 or 256?  I don't think it's a big deal though.

src/Swarm/Language/Capability.hs
```diff
@@ -72,6 +72,10 @@ data Capability
     CSensehere
   | -- | Execute the 'Detect' command
     CDetectloc
+  | -- | Execute the 'Sniff' command
+    CDetectdistance
+  | -- | Execute the 'Chirp' command
+    CDetectdirection
```
byorgey, line 78: These capabilities don't have any devices that provide them, right?  Is the plan to leave that until later?

src/Swarm/Game/Step.hs
```diff
@@ -1845,6 +1867,18 @@ execConst c vs s k = do
       , prettyText (Out (VCApp c (reverse vs)) s k)
       ]
 
+  findNearest name = do
```
byorgey, line 1870: Some type signatures on `findNearest` and possibly some of its local variables (`sortedLocs`, `genDiamondSides`, etc) might be really helpful.  I spent a while trying to understand it which was made more difficult by having to do type inference in my head.

src/Swarm/Game/Step.hs
```diff
@@ -1206,6 +1206,28 @@ execConst c vs s k = do
...
+        inst <- use equippedDevices
+        let processDirection entityDir =
+              if countByName "compass" inst >= 1
+                then Just $ DAbsolute entityDir
+                else case mh >>= toDirection of
+                  Just (DAbsolute robotDir) -> Just $ DRelative $ entityDir `relativeTo` robotDir
+                  _ -> Nothing -- This may happen if the robot is facing "down"
+            val = VDir $ fromMaybe (DRelative DDown) $ do
+              entLoc <- firstFound
+              guard $ sum entLoc /= 0
```
byorgey, line 1227: What is this `guard` doing?  I can't make heads or tails of it.

src/Swarm/Language/Syntax.hs
```diff
@@ -100,14 +102,27 @@ import Swarm.Language.Types
...
 --   use them.
-data AbsoluteDir = DNorth | DSouth | DEast | DWest
+--
+-- NOTE: These values are ordered by increasing angle according to
+-- the standard mathematical convention.
+-- That is, the right-pointing direction, East, is considered
+-- the "reference angle" and the order proceeds counter-clockwise.
+-- See https://en.wikipedia.org/wiki/Polar_coordinate_system#Conventions
+--
+-- Do not alter this ordering, as there exist functions that depend on it.
```
byorgey, line 124: This makes sense.  Just for completeness' sake, it might be nice to include a reference to an example of a function which depends on the ordering.

byorgey (approved):
I have some questions/suggestions/etc. but overall this looks good to me.

Re: efficiency, `getElemsInArea` is O(n lg n) only in the context of having a `Map` with `Location` keys, and trying to find all the values with locations in a certain area.  Since a `Map` is already stored using an ordering on `Location`s, we can do better than simply enumerating all the locations in the area and looking up each in the `Map`.  `findNearest` is doing something a bit different --- there's no way we can currently avoid having to look at each one of the O(n^2) locations.   I could imagine building some data structures to cache information about individual entities that would allow us to search faster, but intuitively it would be pretty complex and maybe not worth it.

If we eventually add devices and make these commands available in classic mode, it might make sense to lower the max range (e.g. to 64) in that case.

src/Swarm/Game/Location.hs
kostmo: Ideal for my use case (if not sensible) would be to return a positive value even if the left operand is negative.  But that's not what the Haskell `mod` operator does: https://replit.com/@KarlOstmo/LostWretchedDatamining

Is that what you were referring to?

byorgey: No, that *is* what the Haskell `mod` function does.  ``-5 `mod` 3`` parses as ``-(5 `mod` 3)``.  Try ``(-5) `mod` 3``. :smile:

byorgey: When `b` is positive, `mod` always satisfies ``0 <= a `mod` b < b`, and ``a = (a `div` b) * b + (a `mod` b)``.

kostmo: > No, that _is_ what the Haskell `mod` function does. `` -5 `mod` 3 `` parses as `` -(5 `mod` 3) ``. Try `` (-5) `mod` 3 ``. 😄

HA! Oh wow.

kostmo: I feel like there should be a lint rule that warns on misleading use of whitespace/lack of parenthesization with the `-` operator

byorgey: There 100% should be. Negation in Haskell is a bit strange... also cf how you have to write `subtract 2` instead of `(-2)` if you want an operator section.

src/Swarm/Game/Step.hs
byorgey: But wouldn't `sum entLoc == 0` be true also for vectors like `V2 (-2) 2`?

byorgey: Also, I think `entLoc` is actually a tuple of a distance and a vector, right?  That was making me even more confused about the use of `sum`.

kostmo: Ugh.  Yes.  I think I had intended to write

    sum . fmap abs

...which is even worse.

src/Swarm/Language/Capability.hs
kostmo: Yes, will come up with those later.

src/Swarm/Game/Location.hs
kostmo: https://github.com/ndmitchell/hlint/issues/1484

PR #1183, opened 2023-03-27
`requirements` command for viewing requirements of any expression

### LINKED ISSUE #922: Display definition requirements
> What I see as the actual problem is that the requirements are hidden. We should be able to show them...

_Originally posted by @xsebek in https://github.com/swarm-game/swarm/issues/921#issuecomment-1356190044_

Conceptually, the requirements of a command are part of its type (eventually, with #231, they will be literally so), and it should be something a player can see about a definition, just as it is currently possible to display the type of a definition.

Since the set of requirements can be quite large, however, displaying them should probably not be the default.  But perhaps there could be a key to toggle whether requirements are shown as part of types.  Or perhaps there could be some sort of built-in command to display the requirements of a name.

byorgey:
Add a special built-in command `requirements` which can be applied to any expression.  Typing `requirements e` at the REPL will cause a log message to be generated listing the requirements for the given expression.  For example, `requirements (move; grab)` generates
```
Requirements for `(move; grab)`:
  Equipment:
    - fast grabber OR grabber OR harvester
    - solar panel
    - tank treads OR treads
```

Closes #922 .


src/Swarm/Game/Step.hs
```diff
@@ -748,6 +749,46 @@ stepCESK cesk = case cesk of
...
+          | S.null caps && S.null devs = []
+          | otherwise =
+              "  Equipment:"
+                : (("    - " <>) . (T.intercalate " OR " . S.toList) <$> S.toList deviceSets)
+
+        inventoryLog
+          | M.null inv = []
+          | otherwise =
+              "  Inventory:"
+                : (("    - " <>) . (\(e, n) -> e <> " (" <> showT n <> ")") <$> M.assocs inv)
```
kostmo, line 782: Having a[`parens` function](https://github.com/swarm-game/swarm/pull/974/files#r1151430564) would make this cleaner

src/Swarm/Game/Step.hs
```diff
@@ -748,6 +749,46 @@ stepCESK cesk = case cesk of
...
+            -- Generate the set of possible devices for each capability, along
+            -- with the outright required devices.
+            S.map (S.fromList . map (^. entityName) . (`deviceForCap` em)) caps
+              `S.union` S.map S.singleton devs
+
+        equipmentLog
+          | S.null caps && S.null devs = []
+          | otherwise =
+              "  Equipment:"
+                : (("    - " <>) . (T.intercalate " OR " . S.toList) <$> S.toList deviceSets)
```
kostmo, line 776: Factor out common list-rendering logic for `Equipment:` and `Inventory:`?

src/Swarm/Game/Step.hs
```diff
@@ -748,6 +749,46 @@ stepCESK cesk = case cesk of
...
+    em <- use entityMap
+    let (R.Requirements caps devs inv, _) = R.requirements currentContext t
+        deviceSets :: Set (Set Text)
+        deviceSets =
+          -- Remove any device sets which are a superset of another
+          -- set.  For example, if (grabber OR fast grabber OR
+          -- harvester) is required but (grabber OR fast grabber) is
+          -- also required then we might as well remove the first set,
+          -- since satisfying the second device set will automatically
+          -- satisfy the first.
```
kostmo, line 765: Are both sets equally likely to be the one removed? Or is there bias toward removing the larger set?

src/Swarm/Game/Step.hs
```diff
@@ -748,6 +749,46 @@ stepCESK cesk = case cesk of
...
+          removeSupersets $
+            -- Generate the set of possible devices for each capability, along
+            -- with the outright required devices.
+            S.map (S.fromList . map (^. entityName) . (`deviceForCap` em)) caps
+              `S.union` S.map S.singleton devs
+
+        equipmentLog
+          | S.null caps && S.null devs = []
```
kostmo, line 773: Would this be overkill?
```suggestion
          | all S.null [caps, devs] = []
```

src/Swarm/Game/Step.hs
```diff
@@ -748,6 +749,46 @@ stepCESK cesk = case cesk of
...
+          -- set.  For example, if (grabber OR fast grabber OR
+          -- harvester) is required but (grabber OR fast grabber) is
+          -- also required then we might as well remove the first set,
+          -- since satisfying the second device set will automatically
+          -- satisfy the first.
+          removeSupersets $
+            -- Generate the set of possible devices for each capability, along
+            -- with the outright required devices.
+            S.map (S.fromList . map (^. entityName) . (`deviceForCap` em)) caps
+              `S.union` S.map S.singleton devs
```
kostmo, line 770: Separate intermediate variables for both arguments of the `S.union` that correspond to the above comment would be nice, particularly since the line-breaking/formatting is out of our hands.

src/Swarm/Util.hs
```diff
@@ -349,7 +351,13 @@ _NonEmpty :: Lens' (NonEmpty a) (a, [a])
...
 
 ------------------------------------------------------------
--- Some utilities for NP-hard approximation
+-- Some set utilities
+
+-- | Remove any sets which are supersets of other sets.  In other words,
+--   (1) no two sets in the output are in a subset relationship
+--   (2) every element in the input is a superset of some element in the output.
+removeSupersets :: Ord a => Set (Set a) -> Set (Set a)
+removeSupersets ss = S.filter (\s -> all (liftA2 (||) (== s) (not . (`S.isSubsetOf` s))) ss) ss
```
kostmo, line 360: Would some intermediate variables improve comprehensibility?

kostmo (approved):
Just some documentation and style nits.

src/Swarm/Game/Step.hs
byorgey: A set will be removed if and only if it is a superset of another set.  So the larger set will always be removed.

src/Swarm/Game/Step.hs
byorgey: :shrug: I think it's fine.

byorgey: Ah, spoke too soon.  It's a nice idea but it turns out that it doesn't work, since `caps` and `devs` have different types.

src/Swarm/Util.hs
byorgey: I've refactored it a lot and I think it's much more readable now:
```
removeSupersets ss = S.filter (not . isSuperset) ss
 where
  isSuperset s = any (`S.isSubsetOf` s) (S.delete s ss)
```

kostmo: Very nice!

Issue #1089, opened 2023-02-06
Define Tutorial Bug


AishmanTaylor:
Unable to place flower or pickerel weed in the inventory of the base despite the robot showing that it has been deposited. 

Using the following four definitions the navigate the maze: 
def move4 : cmd unit = move; move; move; move end
def move16 : cmd unit = move4; move4; move4; move4 end
def moveup : cmd unit = turn north; move; move end
def move4 : cmd unit = turn south; move; move end

When navigating the maze using these I have used both a combination of the 'harvest', the 'grab' the ' t <- harvest', and the 't <- grab' command to grab the flower. Then use a combination of 'place "flower"', 'place "pickerel weed"', 'place t' to try and get it into the base. This registers as grabbing the flower and then placing it somewhere when using the build command to run a robot through the maze and back. The robot has nothing in its inventory after running the place command but nothing is placed into the inventory of the base.

byorgey:
Thanks for the report!  The command to get it into the inventory of the base is `give base t` (if you stored its name in the variable `t`).  I thought the tutorial explained the `give` command previously but it seems I was wrong.  See #1086 .

xsebek:
It may also be that the unmoving "base" robot suggests that it is a location instead of a robot. :thinking:

I can imagine a simplified model where placing entities in the base adds them to the inventory, but that is not how it currently works. :smile: 

Side note for scenario authors: A system robot could await (#687) changes to the base location and grab and give entities to the base.

Issue #1086, opened 2023-02-05
`Define` tutorial requires using `give` but that command has not yet been explained by the tutorial


byorgey:
The `give` command does technically show up in the `Debug` tutorial, but you are simply told to execute `give base "Win"`; it is not really explained.  Since this command is needed to solve the `Define` tutorial, we should add a paragraph explaining it.

PR #1186, opened 2023-03-29
Dump a list of tutorials and commands they first introduce

kostmo:
May be able to reduce likelihood of #1086 by having a autogenerated table of the commands and the tutorials they are introduced in.

This work is related to #1106 in that robot commands should be given distinct markup (backticks).

## Demo

    scripts/play.sh generate pedagogy

**Updated:** See [rendered output](https://gist.github.com/kostmo/d4af3a6814e65fc6d455fd34b41552d3).

## New tests

    scripts/run-tests.sh --test-arguments '--pattern "Pedagogical soundness"'

Note that the test for `crash.yaml` is currently failing because the `Give` command (and some others) are utilized in this tutorial without having been mentioned in the description or in earlier tutorials.

Additionally, the "bug" described in [this comment](https://github.com/swarm-game/swarm/issues/1089#issuecomment-1418497686) **should** be caught by these tests (i.e. fail the tests) since the embedded `solution` code in the YAML file utilizes the `give` command.  That is, if it currently weren't for [this issue](https://github.com/swarm-game/swarm/pull/1186/files#r1152789153).


src/Swarm/Docs/Pedagogy.hs
```diff
@@ -0,0 +1,201 @@
...
+--
+-- Also, the code from `run` is not parsed transitively yet.
+getCommands :: ProcessedTerm -> [Const]
+getCommands (ProcessedTerm (Module stx _) _ _) =
+  mapMaybe isCommand nodelist
+ where
+  ignoredCommands = S.fromList [Run, Noop]
+
+  nodelist :: [Syntax' Polytype]
+  nodelist = universe stx
```
kostmo, line 117: I think this may not actually be sufficient to find all of the `Const` instances for the commands?  It might not be finding the commands hidden inside a `build` command.  Will I have to use an approach like the [`requirements'` function](https://github.com/swarm-game/swarm/blob/9e2f8b29613223b7eaded5690257abbc6553ac93/src/Swarm/Language/Requirement.hs#L177) instead of this simple `universe` function?

kostmo:
This isn't finished yet, but wanted to give a heads-up @byorgey on the work in progress, and I had one [question](https://github.com/swarm-game/swarm/pull/1186#discussion_r1152789153).

src/Swarm/Docs/Util.hs
```diff
@@ -0,0 +1,26 @@
+{-# LANGUAGE OverloadedStrings #-}
```
kostmo, line 1: Moved from `DocGen.hs`.

data/scenarios/Tutorials/bind2-solution.sw
kostmo, line 1: Moved inline in the `.yaml` file so its commands can be extracted.

src/Swarm/Docs/Pedagogy.hs
byorgey: There's no reason commands hidden inside `build` would not be found.  `universe` works in terms of the `Plated` instance which is derived entirely generically in terms of the auto-derived `Data` instance.  `build` is just an application of the `Build` constant to an argument; to `universe`, the `build` command is not special in any way.

src/Swarm/Docs/Pedagogy.hs
```diff
@@ -0,0 +1,201 @@
...
+  TutorialInfo siPair solnCommands $ getDescCommands s
+ where
+  solnCommands = S.fromList $ maybe [] getCommands maybeSoln
+  maybeSoln = view scenarioSolution s
+
+-- | Obtain the set of all commands mentioned by
+-- name in the tutorial's goal descriptions.
+--
+-- NOTE: It may be more robust to require that a command reference
+-- be surrounded by backticks and parse for that accordingly.
```
byorgey, line 92: Yes, I was going to suggest this.  Based on the sample output in the gist you linked, it seems like there are many false positives where we happened to use an English word which is also the name of a command.

src/Swarm/Docs/Pedagogy.hs
byorgey: Perhaps the reason #1086 is not caught is that `give` technically shows up in the `Debug` tutorial (although it is not really explained there, you are just told to do it).  Also, the English word "give" (but NOT the command) shows up in the description of the `Type errors` tutorial.

test/unit/TestPedagogy.hs
```diff
@@ -0,0 +1,44 @@
...
+ where
+  tutorialInfos = generateIntroductionsSequence $ view scenarios gs
+
+  testFromTut :: Int -> CoverageInfo -> TestTree
+  testFromTut idx (CoverageInfo (TutorialInfo (_s, si) _ descCommands) novelCommands) =
+    testCase
+      (unwords [show idx, scPath])
+      $ assertBool errMsg allCommandsCovered
+   where
+    missingCmds = novelCommands `S.difference` descCommands
```
byorgey, line 34: Yes, because of the many false positives, for this to be at all useful I think we will have to be more selective about recognizing only commands that show up inside backticks in descriptions.

src/Swarm/Game/ScenarioInfo.hs
```diff
@@ -237,11 +243,25 @@ loadScenarios em = do
  where
   p = "scenarios"
 
+loadScenariosWithWarnings :: EntityMap -> IO ([SystemFailure], ScenarioCollection)
+loadScenariosWithWarnings entities = do
+  eitherLoadedScenarios <- runExceptT $ loadScenarios entities
+  return $ case eitherLoadedScenarios of
+    Left xs -> (xs, SC mempty mempty)
+    Right (warnings, x) -> (warnings, x)
```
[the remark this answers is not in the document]
kostmo: #1194

src/Swarm/Docs/Util.hs
kostmo: fixed

byorgey:
@kostmo , apropos of nothing, I am wondering why you often force-push completely new commits to PRs instead of just pushing additional commits on top of the existing ones.  In my experience it makes PRs somewhat harder to review, because (a) I cannot easily see what has been updated, *e.g.* when there are additional changes after I reviewed the initial PR, and (b) it is slightly annoying if I pulled the branch locally to test it; when I go to update it to test a later version of the PR I cannot just pull, I have to delete the local branch and re-pull the entire thing.

One good reason might be to make a nicer, more logical commit history; but since we squash PRs into a single commit when merging anyway, this doesn't make much difference.  But there might be other benefits I am not thinking of.

src/Swarm/Docs/Pedagogy.hs
kostmo: I realize now that the "Define" tutorial (`def.yaml`) comes *after* the "Debug" tutorial (`crash.yaml`), which is why the new tests caught the problem in `crash.yaml` but not `def.yaml`.  So the `universe` function works fine.

kostmo:
> why you often force-push completely new commits to PRs instead of just pushing additional commits on top of the existing ones

Thanks for bringing this up; I hadn't sufficiently considered the review difficulties.  For what it's worth, have you tried the ["Compare"](https://github.com/swarm-game/swarm/compare/71444bc83c8e5d64f2ef14fa707ed89cb7ddfcb8..f9a7e9142ec4a827b9e764ef2e0656221db1f4bb) button next to the "force-pushed" GitHub log entries?

There are a couple of situations where I might want to force push:
1. A rebase, e.g. for #1194.  In addition to the rebase itself requiring a force push, sometimes I have squashed the commits to make the rebase process simpler.  However, then the output of the "Compare" button may conflate changes specific to the PR with the changes since made to the `main` branch and be less useful.  It might then be helpful to have left commits un-squashed.
2. While grouping related changes into a single commit, e.g. a specific bit of reviewer feedback, I may later notice an oversight that semantically belongs to that same commit.  It might then make sense to amend the commit for the sake of the reviewer.

But in general, I will try to avoid gratuitously squashing after the point a reviewer has taken the first look at a PR.

Even though merging a PR will squash the commits, I sometimes like to preserve individual commits for historical interest that one can go back to the PR to view.  I can wait to do that until reviewers are done.

src/Swarm/Docs/Pedagogy.hs
```diff
@@ -0,0 +1,201 @@
...
+getCommands :: ProcessedTerm -> [Const]
+getCommands (ProcessedTerm (Module stx _) _ _) =
+  mapMaybe isCommand nodelist
+ where
+  ignoredCommands = S.fromList [Run, Noop]
+
+  nodelist :: [Syntax' Polytype]
+  nodelist = universe stx
+  isCommand (Syntax' _ t _) = case t of
+    TConst c -> guard (isCmd c && c `S.notMember` ignoredCommands) >> Just c
```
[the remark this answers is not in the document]
kostmo: > Maybe it's not just commands we're worried about but, more generally, any built-in constants

`north` is neither a "command" nor a `TConst`, but in a subsequent PR perhaps we could widen the keyword list to include `TDir`.

Meanwhile I have loosened `isCmd` to `isUserFunc`.

byorgey:
> For what it's worth, have you tried the ["Compare"](https://github.com/swarm-game/swarm/compare/71444bc83c8e5d64f2ef14fa707ed89cb7ddfcb8..f9a7e9142ec4a827b9e764ef2e0656221db1f4bb) button next to the "force-pushed" GitHub log entries?

I... hadn't noticed that before!  Thanks, that definitely helps.

> There are a couple of situations where I might want to force push: ...

Yes, those all make sense.

> Even though merging a PR will squash the commits, I sometimes like to preserve individual commits for historical interest that one can go back to the PR to view.

Agreed, I have sometimes gone back to look at individual commits in a PR, e.g. when tracking down the specific change that introduced a bug via `git bisect`.

data/scenarios/Tutorials/crash.yaml
```diff
@@ -16,10 +16,12 @@ objectives:
         For example:
       - |
         build {log "Hi!"; turn east; move; move; move; log "3"; move; log "OK"}
-      - After the robot crashes, execute `view it0` (or whichever
+      - |
+        `wait` for the robot to crash, then execute `view it0` (or whichever
         `itN` variable corresponds to the result of the `build`
         command) to see how far it got. Further instructions should
-        appear in the crashed robot's log.
+        appear in the crashed robot's log and `give` you an opportunity to `salvage`
```
byorgey, line 23: :laughing:

src/Swarm/Doc/Pedagogy.hs
```diff
@@ -0,0 +1,239 @@
...
+  otherLines =
+    concat
+      [ pure . surround "`" . T.pack $ view scenarioPath si
+      , [""]
+      , pure . surround "*" . T.strip $ view scenarioDescription s
+      , [""]
+      , renderSection "Introduced in solution" . renderCmdList $ M.keysSet novelCmds
+      , [""]
+      , renderSection "Referenced in description" $ renderCmdList dCmds
+      ]
```
byorgey, line 176: Could some kind of `intersperse` be used here to avoid having to repeat `[""]`?

src/Swarm/Language/Syntax.hs
```diff
@@ -590,7 +590,10 @@ constInfo c = case c of
       , "Note that you can see the messages either in your logger device or the message panel."
       ]
   Log -> command 1 short "Log the string in the robot's logger."
-  View -> command 1 short "View the given actor."
+  View ->
+    command 1 short . doc "View the given actor." $
+      [ "This will recenter the map on the target robot and allow its inventory and logs to be inspected."
```
byorgey, line 595: :+1:

byorgey (commented):
This is looking good, I think.  What do you still have in mind before it is ready?

kostmo:
> This is looking good, I think. What do you still have in mind before it is ready?

I think it's ready for now, @byorgey.  In the future maybe we could augment the output to measure "tutorial coverage" of all of the extant commands.

byorgey:
> In the future maybe we could augment the output to measure "tutorial coverage" of all of the extant commands.

Perhaps, though (in my mind at least) the goal is not to cover all commands in the tutorial; some commands you should just learn about by playing the game --- crafting new devices, reading their descriptions, and so on.

byorgey (approved):

PR #1190, opened 2023-04-01
prefer dots over dollars

kostmo:
See [this comment](https://github.com/swarm-game/swarm/pull/974#discussion_r1112381178).

### DIFF
--- a/.hlint.yaml
+++ b/.hlint.yaml
@@ -33,6 +33,8 @@
 # Will suggest replacing "wibbleMany [myvar]" with "wibbleOne myvar"
 # - error: {lhs: "wibbleMany [x]", rhs: wibbleOne x}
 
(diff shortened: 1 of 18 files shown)


byorgey:
Hmm... that would be a LOT of changes, not all of which look nice to me.  Do you all think this would be worth it?  Personally I would be OK to just leave "prefer dots over dollars" as a socially-enforced rule rather than having it enforced in CI.

Another idea might be to make our own custom rule that suggests replacing `a $ b $ c $ d` with `a . b . c $ d`, so it would only apply if there are more than 2 consecutive dollar signs.  I wonder if that would give us a much smaller footprint while still catching the ones we really care about.

byorgey:
That seems more reasonable/manageable to me.  @kostmo , what do you think?

byorgey (approved):

PR #1201, opened 2023-04-06
Implement 'watch' and 'surveil'

### LINKED ISSUE #687: Location watch command

kostmo:
Closes #687

## Demo

    scripts/play.sh --scenario data/scenarios/Testing/687-watch-command.yaml --autoplay

or use `run` instead of `--autoplay`, and monitor the state:

    curl -s http://localhost:5357/robot/0 | jq ._watchedLocations

### DIFF
--- a/src/Swarm/Game/State.hs
+++ b/src/Swarm/Game/State.hs
@@ -39,6 +39,7 @@ module Swarm.Game.State (
   robotMap,
   robotsByLocation,
   robotsAtLocation,
+  robotsWatching,
   robotsInArea,
   baseRobot,
   activeRobots,
@@ -102,6 +103,7 @@ module Swarm.Game.State (
   addRobot,
   addTRobot,
   emitMessage,
+  wakeWatchingRobots,
   sleepUntil,
   sleepForever,
   wakeUpRobotsDoneSleeping,
@@ -151,7 +153,7 @@ import Servant.Docs (ToSample)
 import Servant.Docs qualified as SD
 import Swarm.Game.Achievement.Attainment
 import Swarm.Game.Achievement.Definitions
-import Swarm.Game.CESK (emptyStore, finalValue, initMachine)
+import Swarm.Game.CESK (CESK (Waiting), emptyStore, finalValue, initMachine)
 import Swarm.Game.Entity
 import Swarm.Game.Failure
 import Swarm.Game.Failure.Render
@@ -371,6 +373,13 @@ data GameState = GameState
     -- append to a list than to a Set.
     _waitingRobots :: Map Integer [RID]
   , _robotsByLocation :: Map Location IntSet
+  , -- This member exists as an optimization so
+    -- that we do not have to iterate over all of the robots,
+    -- since there may be many.
+    -- In contrast with "_waitingRobots", the values of "_robotsWatching"
+    -- can actually be represented as a more semantically-appropriate Set
+    -- because insertion is "rare".
+    _robotsWatching :: S.Set RID
   , _allDiscoveredEntities :: Inventory
   , _availableRecipes :: Notifications (Recipe Entity)
   , _availableCommands :: Notifications Const
@@ -473,6 +482,10 @@ robotsAtLocation loc gs =
     . M.lookup loc
     . view robotsByLocation
     $ gs
+
+-- | Get a list of all the robots that are "watching" any location.
+-- This is the counterpart to the per-robot "watchedLocations" member.
+robotsWatching :: Lens' GameState (S.Set RID)
 
 -- | Get all the robots within a given Manhattan distance from a
 --   location.
@@ -842,6 +855,7 @@ activateRobot rid = internalActiveRobots %= IS.insert rid
 wakeUpRobotsDoneSleeping :: Has (State GameState) sig m => m ()
 wakeUpRobotsDoneSleeping = do
   time <- use ticks
+
   mrids <- internalWaitingRobots . at time <<.= Nothing
   case mrids of
     Nothing -> return ()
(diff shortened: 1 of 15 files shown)


src/Swarm/Game/State.hs
```diff
@@ -846,10 +858,20 @@ wakeUpRobotsDoneSleeping = do
...
       let aliveRids = filter (`IM.member` robots) rids
       internalActiveRobots %= IS.union (IS.fromList aliveRids)
 
+      forM_ aliveRids $ \rid -> do
+        let maybeRobotObj = IM.lookup rid robots
+        case maybeRobotObj of
+          Nothing -> return ()
+          Just obj -> do
+            let updatedObj = obj . watchedLocations .= mempty
+            robotMap %= IM.insert rid updatedObj
```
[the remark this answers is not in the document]
byorgey: Something like `forM_ aliveRids $ \rid -> robotMap . at rid . _Just . watchedLocations %= mempty` should work, I think?

src/Swarm/Game/State.hs
```diff
@@ -842,6 +855,7 @@ activateRobot rid = internalActiveRobots %= IS.insert rid
 wakeUpRobotsDoneSleeping :: Has (State GameState) sig m => m ()
 wakeUpRobotsDoneSleeping = do
   time <- use ticks
+
```
kostmo, line 858: Is the `<<.= Nothing` on the next line equivalent to:
```
internalWaitingRobots %= delete time
```
?

Wondering how that works, since `internalWaitingRobots` doesn't have `Maybe` values.

src/Swarm/Game/State.hs
```diff
@@ -850,6 +864,61 @@ wakeUpRobotsDoneSleeping = do
...
+      newWakeTime = currentTick + 1
+      newInsertions = M.singleton newWakeTime $ S.toList watchingBots
+
+      filteredWaitingMap =
+        M.filter (not . null) $
+          M.map (filter (`S.notMember` watchingBots)) waitingMap
+
+  -- NOTE: There are two "sources of truth" for the waiting state of robots:
+  -- 1. In the GameState via "internalWaitingRobots"
+  -- 2. In each robot, via the CESK machine state
```
kostmo, line 908: This was a gotcha for me, after having plumbed just the waking logic for `internalWaitingRobots` but observing that the robots still did not wake up.

Would it be feasible to have a single source of truth?  The irony of now introducing both a `robotsWatching` in `GameState` and `watchedLocations` in each `Robot` is not lost on me.

src/Swarm/Game/Step.hs
```diff
@@ -604,6 +606,7 @@ updateWorld c (ReplaceEntity loc eThen down) = do
     then throwError $ cmdExn c ["The", eThen ^. entityName, "is not there."]
     else do
       world %= W.update (W.locToCoords loc) (const down)
+      wakeWatchingRobots loc
```
kostmo, line 609: Here's the other place I call `wakeWatchingRobots`.

src/Swarm/Game/Step.hs
```diff
@@ -419,7 +419,9 @@ entityAt loc = zoomWorld (W.lookupEntityM @Int (W.locToCoords loc))
 -- | Modify the entity (if any) at a given location.
 updateEntityAt ::
   (Has (State GameState) sig m) => Location -> (Maybe Entity -> Maybe Entity) -> m ()
-updateEntityAt loc upd = zoomWorld (W.updateM @Int (W.locToCoords loc) upd)
+updateEntityAt loc upd = do
+  zoomWorld (W.updateM @Int (W.locToCoords loc) upd)
+  wakeWatchingRobots loc
```
kostmo, line 424: It's not ideal that I have to call `wakeWatchingRobots` both here (in `updateEntityAt`) and inside the `updateWorld` function.  I wonder if there is consolidation that could be done.  My first question is whether specifically `W.updateM` is needed here and specifically `W.update` is needed in `updateWorld`, or if both places could use the same.

src/Swarm/Game/State.hs
```diff
@@ -474,6 +483,10 @@ robotsAtLocation loc gs =
     . view robotsByLocation
     $ gs
 
+-- | Get a list of all the robots that are "watching" any location.
+-- This is the counterpart to the per-robot "watchedLocations" member.
+robotsWatching :: Lens' GameState (S.Set RID)
```
kostmo, line 488: Should I even bother with having this as a performance optimization?  Or is it fine to just iterate over all extant robots to obtain this information?

src/Swarm/Game/Step.hs
byorgey: I think probably `updateWorld` should be refactored to call `updateEntityAt`.  I don't know if there's a specific reason why `updateWorld` calls `update` instead of `updateM`.  I can't think of any reason why it would ever be important to call `update` specifically.  It's always safe to call `updateM` but perhaps sometimes calling `update` could be an optimization if you know the tile must already be loaded.

src/Swarm/Game/World.hs
```diff
@@ -247,15 +247,27 @@ lookupEntityM c = do
...
 -- | Update the entity (or absence thereof) at a certain location,
---   returning an updated 'World'.  See also 'updateM'.
-update :: Coords -> (Maybe e -> Maybe e) -> World t e -> World t e
+--   returning an updated 'World' and a Boolean indicating whether.
+--   the update changed the entity here.
+--   See also 'updateM'.
+update :: Eq e => Coords -> (Maybe e -> Maybe e) -> World t e -> (World t e, Bool)
 update i g w@(World f t m) =
-  World f t (M.insert i (g (lookupEntity i w)) m)
+  (wNew, entityAfter /= entityBefore)
```
kostmo, line 255: I suppose this equality check may be expensive, since `Entity` is a wide record.  Maybe if so long as the result of the equality check is consumed lazily, it is ok for the comparison code to be here?

src/Swarm/Game/Step.hs
```diff
@@ -419,7 +419,10 @@ entityAt loc = zoomWorld (W.lookupEntityM @Int (W.locToCoords loc))
 -- | Modify the entity (if any) at a given location.
 updateEntityAt ::
   (Has (State GameState) sig m) => Location -> (Maybe Entity -> Maybe Entity) -> m ()
-updateEntityAt loc upd = zoomWorld (W.updateM @Int (W.locToCoords loc) upd)
+updateEntityAt loc upd = do
+  didChange <- zoomWorld $ W.updateM @Int (W.locToCoords loc) upd
+  when didChange $
```
kostmo, line 424: Instead of using `when` here, I could pass `didChange` as an argument to `wakeWatchingRobots`, and lazy consumption could avoid performing expensive entity equality checks when a cell is unwatched.

src/Swarm/Game/State.hs
kostmo: Changed to `Map Location (Set RID)`.

src/Swarm/Game/State.hs
kostmo: I've at least added some comments so I guess I can live with the the dual sources of truth for now.

kostmo:
All comments addressed, this is ready for another look, @byorgey.

src/Swarm/Game/World.hs
byorgey: You could also compare their hashes instead of checking them for equality directly, i.e.  ``(/=) `on` view entityHash``.

byorgey (approved):
Looks good!

PR #1204, opened 2023-04-08
resonate command

kostmo:
Towards #1171

The `resonate` command counts entities of a given type within a rectangle.

This PR also fixes a bug in `detect` when the rectangle coords are non-ascending.

### DIFF
--- a/src/Swarm/Game/Step.hs
+++ b/src/Swarm/Game/Step.hs
@@ -1243,18 +1243,18 @@ execConst c vs s k = do
       let Location x y = loc
       return $ Out (VPair (VInt (fromIntegral x)) (VInt (fromIntegral y))) s k
     Detect -> case vs of
-      [VText name, r] -> do
+      [VText name, VRect x1 y1 x2 y2] -> do
         loc <- use robotLocation
-        let locs = rectCells r
+        let locs = rectCells x1 y1 x2 y2
         -- sort offsets by (Manhattan) distance so that we return the closest occurrence
         let sortedLocs = sortOn (\(V2 x y) -> abs x + abs y) locs
         firstOne <- findM (fmap (maybe False $ isEntityNamed name) . entityAt . (loc .+^)) sortedLocs
         return $ Out (asValue firstOne) s k
       _ -> badConst
     Resonate -> case vs of
-      [VText name, r] -> do
+      [VText name, VRect x1 y1 x2 y2] -> do
         loc <- use robotLocation
-        let locs = rectCells r
+        let locs = rectCells x1 y1 x2 y2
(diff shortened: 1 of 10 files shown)


src/Swarm/Game/Value.hs
```diff
@@ -14,6 +15,16 @@ import Swarm.Game.Entity
 import Swarm.Game.Robot
 import Swarm.Language.Value
 
+-- * Patterns
+
+type VRect = Value
+pattern VRect :: Integer -> Integer -> Integer -> Integer -> VRect
+pattern VRect x1 y1 x2 y2 = VPair (VPair (VInt x1) (VInt y1)) (VPair (VInt x2) (VInt y2))
+
+{-# COMPLETE VRect #-}
```
byorgey, line 24: I don't think we want this `COMPLETE` pragma.  The meaning of such a pragma would be that pattern-matching on `VRect` is sufficient to cover all possible `Value`s, but that is certainly not true.  And indeed, the `rectCells` function is in fact a partial function, and we don't want to just paper over that with a bogus `COMPLETE` pragma.  We never want partial functions (even in cases where bad inputs "shouldn't ever happen"), since as a matter of policy/quality control we want to try to make sure that the game can not possibly crash.

src/Swarm/Game/Step.hs
```diff
@@ -1243,14 +1243,21 @@ execConst c vs s k = do
       let Location x y = loc
       return $ Out (VPair (VInt (fromIntegral x)) (VInt (fromIntegral y))) s k
     Detect -> case vs of
-      [VText name, VPair (VPair (VInt x1) (VInt y1)) (VPair (VInt x2) (VInt y2))] -> do
+      [VText name, r] -> do
```
byorgey, line 1246: There is a subtle issue here, which is that this pattern now matches *any* `Value` as the second argument to `Detect` instead of only pairs of pairs of ints.  In theory, if typechecking works, that is the only possible value we could ever get here. But in case there is ever any kind of bug that causes the wrong kind of `Value` to end up here, previously it would have fallen through to the `badConst` case, which throws an appropriate in-game exception; with the current code, it will instead actually *crash* the whole game with a non-exhaustive pattern match error in the `rectCells` function.

We should be using a `VRect` pattern *here*, in the pattern for `vs`, and then pass the resulting four `Int` values to `rectCells`.  That way `rectCells` will be a total function.

byorgey (changes requested):
I like the new `resonate` command itself, that all looks good to go.  We just need to fix up the way we do pattern-matching on values + the `rectCells` function so we can't possibly crash with a non-exhaustive pattern match.

byorgey (approved):
Looks good, thanks!

Issue #1205, opened 2023-04-09
Exception in goal condition code is treated as a win


kostmo:
It seems that any exceptions that are not caught by `try` will cause the `condition` of a goal to evaluate as `True`.

PR #1206, opened 2023-04-09
Unit test for exception in goal code

kostmo:
This test case passes, but should fail.

Towards #1205

    scripts/play.sh --scenario data/scenarios/Testing/1205-goal-condition-exception.yaml --autoplay

### DIFF
--- a/data/scenarios/Testing/1205-goal-condition-exception.yaml
+++ b/data/scenarios/Testing/1205-goal-condition-exception.yaml
@@ -1,0 +1,39 @@
+version: 1
+name: Test goal exceptions
+description: |
+  Goal exceptions should not be counted as a win.
+objectives:
+  - condition: |
+      as base {
+        let x = 1 / 0 in
+        has "boulder";
+      }
+    goal:
+      - |
+        Pick up a "boulder".
+solution: |
+  move; move; grab;
+robots:
+  - name: base
+    dir: [1, 0]
+    devices:
+      - scanner
+      - branch predictor
+      - grabber
+      - strange loop
+      - lambda
+      - dictionary
+      - logger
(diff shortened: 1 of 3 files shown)


data/scenarios/Testing/1205-goal-condition-exception.yaml
```diff
@@ -0,0 +1,39 @@
+version: 1
+name: Test goal exceptions
+description: |
+  Goal exceptions should not be counted as a win.
+objectives:
+  - condition: |
+      as base {
+        let x = 1 / 0 in
```
kostmo, line 8: This division by zero is counted as a win.

test/integration/Main.hs
```diff
@@ -269,6 +269,8 @@ testScenarioSolution _ci _em =
         , testSolution Default "Testing/1157-drill-return-value"
         , testSolution Default "Testing/1171-sniff-command"
         , testSolution Default "Testing/1171-chirp-command"
+        , testSolution Default "Testing/1171-resonate-command"
```
kostmo, line 272: Also sneaking in this test.

test/integration/Main.hs
```diff
@@ -269,6 +269,8 @@ testScenarioSolution _ci _em =
...
+        , testSolution Default "Testing/1205-goal-condition-exception"
```
kostmo, line 273: Not sure how we should set up a scenario-based test that, when the behavior is working correctly, the test should fail.

I think we would need to teach the test suite to differentiate between a scenario fail and a scenario win, and then add an extra "failure" goal condition that is "achieved" upon an exception.  When this behavior is fixed, then the normal win condition will be achieved instead.

PR #1209, opened 2023-04-11
scout command

### LINKED ISSUE #1207: Line-of-sight robot detection command
So far there are many commands for locating entities, but few for locating robots.

A `scout` command would answer whether any (visible) robot exists along a direct line of sight in a cardinal direction.

Introducing such a command with "limited power"---in this case, returning a Boolean rather than some richer type, and requiring that the user position themselves "inline" with the target robot---facilitates more interesting challenges.

Furthermore, a new property of entities, `Opaque`, should be introduced, which can block the line of sight for this command.

kostmo:
Closes #1207

This PR also involves some refactoring that simplifies more return values in `Step.hs` with `asValue`.

### DIFF
--- a/src/Swarm/Game/Location.hs
+++ b/src/Swarm/Game/Location.hs
@@ -28,6 +28,7 @@ module Swarm.Game.Location (
   manhattan,
   euclidean,
   getElemsInArea,
+  sigdir,
 
   -- ** reexports for convenience
   Affine (..),
@@ -149,6 +150,18 @@ relativeTo targetDir referenceDir =
  where
   enumCount = length (Util.listEnums :: [AbsoluteDir])
   indexDiff = ((-) `on` fromEnum) targetDir referenceDir `mod` enumCount
+
+-- | Substitutes all nonzero values with one, preserving sign.
+-- Compare to "signorm", which is constrained to class "Floating":
+-- https://hackage.haskell.org/package/linear-1.22/docs/Linear-Metric.html#v:signorm
+sigdir :: (Ord a, Num a) => V2 a -> V2 a
+sigdir = fmap $ signOrdering . compare 0
+
+signOrdering :: Num a => Ordering -> a
+signOrdering = \case
+  LT -> -1
+  EQ -> 0
+  GT -> 1
 
 -- | Logic adapted from:
 -- https://gamedev.stackexchange.com/questions/49290/#comment213403_49300
--- a/src/Swarm/Game/Location.hs
+++ b/src/Swarm/Game/Location.hs
@@ -154,14 +154,8 @@ relativeTo targetDir referenceDir =
 -- | Substitutes all nonzero values with one, preserving sign.
 -- Compare to "signorm", which is constrained to class "Floating":
 -- https://hackage.haskell.org/package/linear-1.22/docs/Linear-Metric.html#v:signorm
-sigdir :: (Ord a, Num a) => V2 a -> V2 a
-sigdir = fmap $ signOrdering . compare 0
-
-signOrdering :: Num a => Ordering -> a
-signOrdering = \case
-  LT -> -1
-  EQ -> 0
-  GT -> 1
+sigdir :: Num a => V2 a -> V2 a
+sigdir = fmap signum
 
 -- | Logic adapted from:
 -- https://gamedev.stackexchange.com/questions/49290/#comment213403_49300
--- a/src/Swarm/Game/Location.hs
+++ b/src/Swarm/Game/Location.hs
@@ -28,7 +28,6 @@ module Swarm.Game.Location (
   manhattan,
   euclidean,
   getElemsInArea,
-  sigdir,
 
   -- ** reexports for convenience
   Affine (..),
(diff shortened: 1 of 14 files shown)


src/Swarm/Game/Location.hs
```diff
@@ -150,6 +151,18 @@ relativeTo targetDir referenceDir =
   enumCount = length (Util.listEnums :: [AbsoluteDir])
   indexDiff = ((-) `on` fromEnum) targetDir referenceDir `mod` enumCount
 
+-- | Substitutes all nonzero values with one, preserving sign.
+-- Compare to "signorm", which is constrained to class "Floating":
+-- https://hackage.haskell.org/package/linear-1.22/docs/Linear-Metric.html#v:signorm
+sigdir :: (Ord a, Num a) => V2 a -> V2 a
+sigdir = fmap $ signOrdering . compare 0
```
byorgey, line 158: Isn't this just `fmap signum`?

src/Swarm/Game/Step.hs
```diff
@@ -1920,11 +1942,25 @@ execConst c vs s k = do
...
+      (fromIntegral x2)
+      (fromIntegral y2)
+
+  rectCellsInt32 :: Int32 -> Int32 -> Int32 -> Int32 -> [V2 Int32]
+  rectCellsInt32 x1 y1 x2 y2 = [V2 x y | x <- [xMin .. xMax], y <- [yMin .. yMax]]
    where
     (xMin, xMax) = sortPair (x1, x2)
     (yMin, yMax) = sortPair (y1, y2)
 
+  hasSightLineTo :: HasRobotStepState sig m => Location -> Location -> m Bool
```
byorgey, line 1958: Maybe add a comment that it only makes sense to use this with two locations that share the same x- or y-coordinate?

src/Swarm/Game/Step.hs
```diff
@@ -1238,10 +1238,35 @@ execConst c vs s k = do
         inv <- use robotInventory
         return $ Out (VInt (fromIntegral $ countByName name inv)) s k
       _ -> badConst
+    Scout -> case vs of
```
byorgey, line 1241: So does the `scout` command have an unlimited range?  That seems unrealistic.

src/Swarm/Game/Location.hs
kostmo: Indeed, thanks.

src/Swarm/Game/Step.hs
kostmo: Fixed

src/Swarm/Game/Step.hs
```diff
@@ -1238,10 +1238,35 @@ execConst c vs s k = do
...
+              isVisible = not . (^. robotDisplay . invisible)
+            isStraightAhead = (== heading) . sigdir . (myLoc .-.)
+            predicate x =
+              all
+                ($ x)
+                [ isStraightAhead . fst
+                , hasVisibleBot . snd
+                ]
+
+        botsByLocs <- use robotsByLocation
```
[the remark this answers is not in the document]
kostmo: Yes, that's much better. Fixed!

src/Swarm/Game/Step.hs
```diff
@@ -1240,10 +1240,47 @@ execConst c vs s k = do
...
+      [VDir d] -> do
+        rMap <- use robotMap
+        myLoc <- use robotLocation
+        heading <- deriveHeading d
+        botsByLocs <- use robotsByLocation
+        selfRid <- use robotID
+
+        -- Includes the base location, so we exclude the base robot later.
+        let locsInDirection :: [Location]
+            locsInDirection = take maxScoutRange $ iterate (.+^ heading) myLoc
```
byorgey, line 1253:
```suggestion
            locsInDirection = take maxScoutRange . drop 1 $ iterate (.+^ heading) myLoc
```
Would it be easier to do this, so we don't include the base location and don't have to exclude it later?

src/Swarm/Game/Step.hs
```diff
@@ -1240,10 +1240,47 @@ execConst c vs s k = do
...
+            evalLocation isOpaque loc
+              | isOpaque = Just False
+              | hasVisibleBot loc = Just True
+              | otherwise = Nothing
+
+        let evalLocationM loc = do
+              opaque <- hasOpaqueEntity loc
+              return $ evalLocation opaque loc
+
+        result <- asum <$> mapM evalLocationM locsInDirection
```
byorgey, line 1277: If I'm reading this right, I think this still has the property that it will always evaluate all 64 cells in the given direction no matter what?  It would be nice (though not critical) if it stops early as soon as it encounters either an opaque entity or a visible robot.

src/Swarm/Game/Step.hs
```diff
@@ -1240,10 +1240,47 @@ execConst c vs s k = do
...
+        let hasVisibleBot :: Location -> Bool
+            hasVisibleBot = any botIsVisible . IS.toList . excludeSelf . botsHere
+             where
+              excludeSelf = (`IS.difference` IS.singleton selfRid)
+              botsHere loc = M.findWithDefault mempty loc botsByLocs
+              botIsVisible = maybe False canSee . (`IM.lookup` rMap)
+              canSee = not . (^. robotDisplay . invisible)
+
+        -- A robot on the same cell as an opaque entity is considered hidden.
+        let evalLocation :: Bool -> Location -> Maybe Bool
```
byorgey, line 1267: What does the `Maybe Bool` represent?  The name `evalLocation` does not really give any hints.  I wonder if it would make the code more readable to use a custom data type like `data ScoutStatus = Opaque | VisibleBot | Clear`.

byorgey (approved):

src/Swarm/Game/Step.hs
kostmo: I actually started with that version, but realized that I want to be able to detect a robot on the same cell as the base.

src/Swarm/Game/Step.hs
kostmo: Fixed by using `firstJustM`.

src/Swarm/Game/Step.hs
kostmo: Added some code comments.

PR #1211, opened 2023-04-13
use brick-list-skip package

### LINKED ISSUE #1127: Consider using `brick-list-search` package
Following [this discussion on Reddit](https://www.reddit.com/r/haskell/comments/114i0pj/comment/j8z6sg0/), @amano-kenji published [`brick-list-search`](https://hackage.haskell.org/package/brick-list-search) to Hackage.  We should evaluate whether we can/should depend on it instead of using [our own custom list search code](https://github.com/swarm-game/swarm/blob/main/src/Swarm/TUI/List.hs), and/or whether there is any way we can contribute to the package.

kostmo:
Closes #1127


kostmo:
@amano-kenji, this works nicely.  I opened [this PR](https://codeberg.org/amano.kenji/brick-list-skip/pulls/1) to try to fix the `ghc-8.10.7` build error.  However, I am not sure how to fix the `ghc-9.0.2` build error.

byorgey:
I'm pretty sure the GHC 9.0.2 failure has to do with simplified subsumption; see e.g. https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0287-simplify-subsumption.rst and https://www.reddit.com/r/haskell/comments/ujpzx3/was_simplified_subsumption_worth_it_for_industry/ .  So it can probably be solved by some eta-expansion and/or explicit type signatures.  I may see if I can come up with a PR that fixes it.

amano-kenji:
The upper version boundary of `<0.1.1.3` is a bit too restrictive.

brick-list-skip doesn't export any orphan instances. According to PVP, the upper version boundary should be `<0.2`.

amano-kenji:
stack.yaml contains brick-list-skip-0.1.1.0 instead of brick-list-skip-0.1.1.2

byorgey:
Thanks for pointing out those issues, @amano-kenji .  See #1213 .

PR #1214, opened 2023-04-16
Key input handler

### LINKED ISSUE #102: Input handler command

### LINKED ISSUE #1210: It's possible to get stuck in pilot mode
**Describe the bug**
If the base is executing a long-running command (or infinite loop), one can switch into pilot mode but then not get out of it.

**To Reproduce**
1. Start a creative mode game.
2. `def forever = \c. c ; forever c end`
3. `forever (scan down)`
4. Now switch into pilot mode with `M-p`
5. Observe that you can't `Ctrl-c` to stop the infinite loop (since the keypress is handled by pilot mode) but neither can you get out of pilot mode: if you try to hit `M-p`, an error dialog pops up telling you that you should clear the REPL first.

**Expected behavior**
It should be possible to switch *out* of pilot mode no matter what the REPL is doing.

I'll probably just fix this as part of my work on #102 .

byorgey:
Ability to code your own input handler routines.  Closes #102 .  Fixes #1210 .

* Adds a new type `key` to represent keypresses
* Adds a primitive function `key : text -> key` which can handle usual letters, numbers, etc. as well as special keys like `"Down"` etc, as well as modifier key prefixes like `key "A-C-Del"`.  `swarm generate keys` generates a list of all recognized special key names.
* New command `installKeyHandler : text -> (key -> cmd unit) -> cmd unit` which sets the "current key handler".  The `text` value is a hint line to display in the secondary key hints menu while the handler is running.  The global shortcut `M-k` toggles the currently installed handler.
* Add a `keyboard` device to provide these commands, as well as a `key` entity (the recipe for a `keyboard` is 16 `key`s + 1 `board`).
* Add a few examples in the `examples` folder.
* Add an installed `keyboard` to the `building-bridges` challenge.


example/multi-key-handler.sw
```diff
@@ -0,0 +1,20 @@
+// Proof of concept illustrating the possibility of key
+// handlers that process multi-key sequences.
+
+def cons : a * b -> (a -> b) -> (a -> b) = \p. \k. \a.
+  if (a == fst p) {snd p} {k a}
+end
+
+def nil : a -> cmd unit = \a. return () end
```
kostmo, line 8: Neat trick for emulating lists.  I will have to borrow this for a challenge scenario.

example/pilotmode.sw
```diff
@@ -0,0 +1,24 @@
...
+  cons (key "Left",    turn left) $
+  cons (key "Right",   turn right) $
+  cons (key "S-Up",    turn north) $
+  cons (key "S-Down",  turn south) $
+  cons (key "S-Left",  turn west) $
+  cons (key "S-Right", turn east) $
+  cons (key "g",       grab; return ()) $
+  cons (key "h",       harvest; return ()) $
+  cons (key "d",       drill forward; return ()) $
+  cons (key "s",       scan forward; return ()) $
```
kostmo, line 21: While trying out this demo, I found that I missed being able to see the result of `scan` and `grab` in the REPL.  But I take it that it is necessary force all commands to have the same return type in the key handler, and therefore suppress the REPL output via `return ()`?

I guess this is a use case to continue to retain the original pilot mode.

src/Swarm/Game/State.hs
```diff
@@ -400,6 +402,8 @@ data GameState = GameState
   , _needsRedraw :: Bool
   , _replStatus :: REPLStatus
   , _replNextValueIndex :: Integer
+  , _inputHandler :: Maybe Value
+  , _inputHandlerHint :: Maybe Text
```
kostmo, line 406: Looks like these are always modified together.

Perhaps combine them into a single field `Maybe (Value, Text)`?

kostmo (approved):

example/multi-key-handler.sw
byorgey: Well, specifically lists-as-append-only-lookup-maps.  This same trick won't work for emulating more general lists.

example/pilotmode.sw
byorgey: Yes, that's right.  Though if we implement a simple version of #683, so *e.g.* we have `print : text -> cmd unit` which will output something on the console, then we could use that to print out the results of `grab` and `scan`.

src/Swarm/Game/State.hs
byorgey: Good idea, done: :heavy_check_mark:

example/pilotmode.sw
kostmo: Come to think of it, simply using `log` would achieve the same end.

kostmo: Maybe this example should do so?

data/scenarios/Challenges/bridge-building.yaml
```diff
@@ -157,6 +157,7 @@ robots:
       - counter
       - dictionary
       - grabber
+      - handle
```
byorgey, line 160: @kostmo I also added a `handle` device to the `bridge-building` challenge.  I think it makes it a little more convenient but for some parts it's still way quicker to write some code to automate it, so it doesn't detract from the challenge.  But if you feel strongly it *shouldn't* be there, we can take it out!

data/entities.yaml
```diff
@@ -185,6 +185,22 @@
   - It also possess some electro-magnetic properties.
   properties: [portable]
 
+- name: handle
```
kostmo, line 188: Actually `handle` already exists: #1028

kostmo (changes requested):

data/entities.yaml
kostmo: Created #1221 to catch this

byorgey: Whoops, thanks!  I'll fix this soon.

byorgey: Can confirm that once I merged #1223 into this branch it did catch the duplicate `handle`.  I changed `handle` to `keyboard`.

example/pilotmode.sw
byorgey: Oh, good idea!  I tried adding logging but it seems to have no effect.  Currently trying to figure out why.

kostmo (approved):

PR #1215, opened 2023-04-17
Fix objective management upon goal condition exception

### LINKED ISSUE #1205: Exception in goal condition code is treated as a win

kostmo:
Fixes #1205.

Bug has existed since #927.

### DIFF
--- a/src/Swarm/Game/Step.hs
+++ b/src/Swarm/Game/Step.hs
@@ -325,7 +325,7 @@ hypotheticalWinCheck em g ws oc = do
       Left exnText ->
         CompletionsWithExceptions
           (exnText : exnTexts)
-          currentCompletions
+          (OB.addIncomplete obj currentCompletions)
           announcements
       Right boolResult ->
         CompletionsWithExceptions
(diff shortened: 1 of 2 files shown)


byorgey (approved):
This looks fine, and I trust you that it fixes the issue, but can you say a bit more about what the issue was and why this is the right fix?  (Either here, and/or, if it may be helpful to others/ourselves in the future, in a source comment.) I'm curious.

kostmo:
> can you say a bit more about what the issue was

I actually had a [comment a bit above](https://github.com/swarm-game/swarm/blob/2e6207e4aac82059cde8715c8fbafd780e60aa94/src/Swarm/Game/Step.hs#L313-L317) that explained how the code is supposed to work.  But I have now added an inline comment right at the line where the bug was.

In short, prior to the `fold`, all "incomplete" goals are "popped" from their list, and then either restored to that list or promoted to "complete" inside the `fold`.  I had forgotten to restore it in the "exception" case.

kostmo:
> what should we do with #1206 ?

I've closed that PR for now.  The test scenario served its purpose to find this bug, but would be more work to design it into the continuous test suite.

PR #1216, opened 2023-04-17
More efficient gopher solution using 'watch'

kostmo:
Tested using:

    scripts/run-tests.sh --test-arguments '--pattern "gopher"'

| Before | After |
| --- | --- |
| 6.28s | 1.33s |

### DIFF
--- a/data/scenarios/Challenges/gopher.yaml
+++ b/data/scenarios/Challenges/gopher.yaml
@@ -73,6 +73,7 @@ robots:
       - [120, swivel]
       - [120, branch predictor]
       - [120, drill]
+      - [120, rolex]
       - [120, flower]
   - name: gopher
     system: true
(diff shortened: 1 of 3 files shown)


data/scenarios/Challenges/gopher.yaml
```diff
@@ -117,6 +118,14 @@ entities:
...
       - Unearthed by industrious gophers.
     properties: [known, portable]
+  - name: rolex
+    display:
+      char: 'R'
+      attr: silver
+    description:
+      - Enables robots to use the 'watch' command.
+    properties: [known, portable]
+    capabilities: [wakeself]
```
byorgey, line 128: Fun! Is the plan to eventually add this to the default `entities.yaml`?

byorgey (approved):

data/scenarios/Challenges/gopher.yaml
kostmo: I could do that.

byorgey: It doesn't have to be in this PR.  We'd have to come up with a recipe for it too.

PR #1219, opened 2023-04-20
implement stride command

### LINKED ISSUE #1218: Multi-cell movement command
A command should exist that allows moving multiple tiles in a single tick:

    stride : int -> cmd unit

A maximum distance (say, `64`) shall be hard-coded.
The robot proceeds up to the specified cell count or until an `unwalkable` entity is encountered.

Perhaps a new entity property `resistive` (or synonym) could be introduced that is able to halt a `stride`, but not prevent passage with the standard `move` command.

This command operates in contrast to the trick of using a non-normalized initial orientation in the `chess` scenario, in which the movement vector is fixed at the outset and unchangeable.

We may choose not to make this command available in the "Classic" game, but could provide custom devices in Challenge scenarios which would allow the player to implement algorithms that require `O(1)` array access.

kostmo:
Closes #1218

    scripts/play.sh --scenario data/scenarios/Testing/1218-stride-command.yaml --autoplay

I decided to implement this command with the same exception behaviors as the `move` command.
If any obstructions exist on the path, then an exception is thrown and no movement is made.

The implications of this are:
1. With `try`, one can learn in `O(1)` time whether any obstructions exist along a line of up to 64 units long
2. One can learn in `O(log N)` time the exact distance of the nearest obstruction along that line.

### DIFF
--- a/src/Swarm/Game/Step.hs
+++ b/src/Swarm/Game/Step.hs
@@ -1068,6 +1068,18 @@ execConst c vs s k = do
       return $ Out VUnit s k
     Stride -> case vs of
       [VInt d] -> do
+        when (d > fromIntegral maxStrideRange) $
+          throwError $
+            CmdFailed
+              Stride
+              ( T.unwords
+                  [ "Can only stride up to"
+                  , T.pack $ show maxStrideRange
+                  , "units."
+                  ]
+              )
+              Nothing
+
         -- Figure out where we're going
         loc <- use robotLocation
         orient <- use robotOrientation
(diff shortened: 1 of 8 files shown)


src/Swarm/Game/Step.hs
```diff
@@ -1064,6 +1064,37 @@ execConst c vs s k = do
...
+      [VInt d] -> do
+        -- Figure out where we're going
+        loc <- use robotLocation
+        orient <- use robotOrientation
+        let heading = orient ? zero
+
+        -- Excludes the base location.
+        let locsInDirection :: [Location]
+            locsInDirection =
+              take (min (fromIntegral d) maxStrideRange) $
```
byorgey, line 1077: Should trying to `stride` more than the `maxStrideRange` be an error?  Currently, if you say e.g. `stride 65` it will just silently stride 64 instead.

Issue #1221, opened 2023-04-22
Validate uniqueness of entity names in CI


kostmo:
E.g. we don't want to define two devices both named `handle`.

byorgey:
See https://github.com/haskell/aeson/issues/531 and https://hackage.haskell.org/package/aeson-2.1.2.1/docs/Data-Aeson-Parser-Internal.html#v:jsonWith , perhaps.  Also related is #587 .

PR #1223, opened 2023-04-25
enforce no duplicate entity names

kostmo:
New test added:

    scripts/run-tests.sh --test-arguments '--pattern 1221-duplicate-entities'


byorgey (approved):
Looks good!  This was easier than I thought.  I forgot that as far as the `.yaml` file goes, the entities are just encoded as a list, not as a map from entity names.  I hope my comments on #1221 didn't lead you too far astray. :smile:

PR #1231, opened 2023-04-29
unrestricted variant of atomic

### LINKED ISSUE #1227: Allow system robots to execute a large program in a single tick
There should be a privileged variant of `atomic` that allows an arbitrarily large program to be executed in a single tick.
This would be useful for setting up a play area with randomness, for example in the word search challenge (#999).  As is, the player robot needs to be confined and to wait patiently while the system robot does a bunch of work.  Unless the system robot is programmed carefully, the solution might be revealed to an observant player as the playfield setup is performed.

kostmo:
Closes #1227.

Demo:

    scripts/play.sh --scenario data/scenarios/Challenges/word-search.yaml --autoplay

### DIFF
--- a/data/scenarios/Challenges/_word-search/create-puzzle.sw
+++ b/data/scenarios/Challenges/_word-search/create-puzzle.sw
@@ -244,4 +244,4 @@ def createPuzzle = \width. \height.
     removeBoulder;
     end;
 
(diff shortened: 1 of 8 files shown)


byorgey (approved):
LGTM!

PR #1235, opened 2023-04-29
Implement push command

### LINKED ISSUE #1234: Sokoban support
Swarm already supports most of the elements required to implement [Sokoban](https://en.wikipedia.org/wiki/Sokoban).  All that is really needed now is a `push` command.

The `push` command would have the same type signature as `move`.  Both the robot and the entity directly in front of it will move forward by one space, so long as no entity already occupies the destination space.  An entity can be `push`ed if it is `portable`, even if that entity is `unwalkable`.

kostmo:
Closes #1234

A `"dozer blade"` is defined locally to provide `push` capability.  To decouple this PR from bikeshedding, will defer the definition of a global entity that offers `push` capability to another PR.

### DIFF
--- a/src/Swarm/Game/Step.hs
+++ b/src/Swarm/Game/Step.hs
@@ -1066,6 +1066,36 @@ execConst c vs s k = do
           }
       updateRobotLocation loc nextLoc
       return $ Out VUnit s k
+    Push -> do
+      -- Figure out where we're going
+      loc <- use robotLocation
+      orient <- use robotOrientation
+      let heading = orient ? zero
+          nextLoc = loc .+^ heading
+          placementLoc = nextLoc .+^ heading
+
+      -- If unobstructed, the robot will move even if
+      -- there is nothing to push.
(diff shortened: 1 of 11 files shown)


src/Swarm/Game/Step.hs
```diff
@@ -1066,6 +1066,36 @@ execConst c vs s k = do
...
+          placementLoc = nextLoc .+^ heading
+
+      -- If unobstructed, the robot will move even if
+      -- there is nothing to push.
+      maybeCurrentE <- entityAt nextLoc
+      case maybeCurrentE of
+        Just e -> do
+          -- Make sure there's nothing already occupying the destination
+          nothingHere <- isNothing <$> entityAt placementLoc
+          nothingHere `holdsOrFail` ["Blocked by an entity."]
```
byorgey, line 1084: It seems unnecessary to use the somewhat internal, technical word "entity" in this error message.  Maybe just say "Something is in the way!" or something like that.

byorgey (approved):
I like it!

PR #1237, opened 2023-05-01
sliding puzzle

kostmo:
Note: I put some effort toward a completely general, automated solution, but eventually just reduced the board size to 3x3 and solved it manually, and used the API to obtain the manual commands to paste into the solution:

    curl http://localhost:5357/repl/history/full | jq .[].in -r

The board generator is generalized to any square dimension, but some scaffolding must be built to re-use the code across multiple scenarios.  The new `Sliding Puzzles` directory is intended to contain a 3x3, 4x4, and 5x5 puzzle, but currently only contains 3x3.

### 3x3 board
image

### 4x4 demo
board

## Demo

    scripts/play.sh --scenario "data/scenarios/Challenges/Sliding Puzzles/3x3.yaml" --autoplay


byorgey:
You have a bunch of these WIP challenges in the pipeline.  I've mostly been ignoring them until you request a review, but would it be helpful to get more feedback along the way?

kostmo:
This one is ready!

data/scenarios/Challenges/Sliding Puzzles/3x3.yaml
```diff
@@ -0,0 +1,742 @@
...
+    display:
+      char: 'd'
+      attr: eventile
+    description:
+      - Four
+    properties: [known, portable]
+  - name: e-tile
+    display:
+      char: 'e'
+      attr: eventile
```
byorgey, line 227: How come `d-tile` and `e-tile` are both `eventile`?

byorgey (approved):

data/scenarios/Challenges/Sliding Puzzles/3x3.yaml
byorgey: Actually it's not just `d` and `e` that seem off, I don't understand the rationale for which tiles are "even" and "odd" at all.

kostmo: Good catch.  I originally wanted the solved 4x4 puzzle to have a checkerboard-like appearance, which meant that every other row would have to invert the parity-based color.  But then I realized that this would not hold for a 3x3 or 5x5 puzzle so I kind of threw up my hands.

Now the tiles are consistently colored based on parity.  A solved 3x3 puzzle or 5x5 puzzle will appear as a checkerboard, whereas solved 4x4 would have vertical stripes.

PR #1241, opened 2023-05-03
finer texture for static

### LINKED ISSUE #1145: Improve visual/spatial resolution of static
> I wonder what it would look like with a couple of shades of gray mixed in.  Typically static has a "grayish" appearance, due to the specks being so small and blurred together.  But these specks are large and chunky.  What we lack in spatial resolution in our TUI display, we could make up for in "color depth".

> For a more granular effect, in the long term we could explore using [half blocks](https://en.wikipedia.org/wiki/Block_Elements) (or even quadrant blocks).

_Originally posted by @kostmo in https://github.com/swarm-game/swarm/pull/1110#discussion_r1125884205_

byorgey:
Closes #1145.

Instead of using simple black or white cells to display static, use cells with any one of the 16 possible arrangements of on/off quarter-blocks.  This makes static appear at almost 2x resolution as before (though there is still a bit of blockiness since the quarter-blocks can't occur independently). See https://asciinema.org/a/582095 for a demonstration.

### DIFF
--- a/src/Swarm/TUI/View/CellDisplay.hs
+++ b/src/Swarm/TUI/View/CellDisplay.hs
@@ -1,10 +1,9 @@
 -- |
 -- SPDX-License-Identifier: BSD-3-Clause
-
 module Swarm.TUI.View.CellDisplay where
 
 import Brick
-import Control.Lens (to, view, (^.), (.~), (&))
+import Control.Lens (to, view, (&), (.~), (^.))
 import Data.ByteString (ByteString)
 import Data.Hash.Murmur
 import Data.List.NonEmpty qualified as NE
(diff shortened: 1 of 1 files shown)


src/Swarm/TUI/View/CellDisplay.hs
```diff
@@ -79,21 +78,48 @@ displayLocRaw showRobots g coords = sconcat $ terrain NE.:| entity <> robots
...
-displayStatic = bool mempty (terrainMap M.! IceT)
+staticDisplay :: GameState -> W.Coords -> Display
+staticDisplay g coords = maybe mempty displayStatic (getStatic g coords)
+
+-- | Draw static given a number from 0-15 representing the state of
+--   the four quarter-pixels in a cell
+displayStatic :: Word32 -> Display
+displayStatic s =
+  defaultEntityDisplay (staticChar s)
+    & displayPriority .~ 100 -- Static has higher priority than anything else
```
kostmo, line 91:
```suggestion
    & displayPriority .~ maxBound -- Static has higher priority than anything else
```

kostmo (approved):
Looks great!

PR #1243, opened 2023-05-03
Fix scenario menu update after quitting to menu

kostmo:
Ensures that the "played" status and the "best scores" get updated in the scenario menu after "Quit to menu" is selected from within a scenario.

This bug appeared after merging #974.

### DIFF
--- a/src/Swarm/TUI/Controller.hs
+++ b/src/Swarm/TUI/Controller.hs
@@ -454,52 +454,76 @@ getNormalizedCurrentScenarioPath =
       gs <- use $ gameState . scenarios
       Just <$> liftIO (normalizeScenarioPath gs p')
 
+saveScenarioInfoOnFinish :: (MonadIO m, MonadState AppState m) => FilePath -> m (Maybe ScenarioInfo)
+saveScenarioInfoOnFinish p = do
+  initialCode <- use $ gameState . initiallyRunCode
+  t <- liftIO getZonedTime
+  wc <- use $ gameState . winCondition
+  let won = case wc of
+        WinConditions (Won _) _ -> True
+        _ -> False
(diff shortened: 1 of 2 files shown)


src/Swarm/TUI/Controller.hs
```diff
@@ -454,56 +454,81 @@ getNormalizedCurrentScenarioPath =
       gs <- use $ gameState . scenarios
       Just <$> liftIO (normalizeScenarioPath gs p')
 
+saveScenarioInfoOnFinish :: (MonadIO m, MonadState AppState m) => FilePath -> m (Maybe ScenarioInfo)
+saveScenarioInfoOnFinish p = do
```
kostmo, line 458: This is just a cut-and-paste of the body of `saveScenarioInfoOnFinishNocheat` (formerly `saveScenarioInfoOnFinish`).

src/Swarm/TUI/Controller.hs
```diff
@@ -454,56 +454,81 @@ getNormalizedCurrentScenarioPath =
...
-                when (won && firstDir == tutorialsDirname) $
-                  attainAchievement' t (Just p) (GlobalAchievement CompletedSingleTutorial)
-              _ -> return ()
-            liftIO $ saveScenarioInfo p si
+      Just p -> void $ saveScenarioInfoOnFinish p
 
 -- | Write the @ScenarioInfo@ out to disk when exiting a game.
 saveScenarioInfoOnQuit :: (MonadIO m, MonadState AppState m) => m ()
 saveScenarioInfoOnQuit = do
-  saveScenarioInfoOnFinish
```
kostmo, line 496: This function call is moved 8 lines down to avoid a redundant "cheat" check.

PR #1244, opened 2023-05-05
reduce lag in monitor bot for Maypole scenario

kostmo:
As I was testing #1010, I observed that the `maypole.yaml` scenario (#1158) would reach the "won" condition if launched from the command line via `--autoplay`, for which the `solution` content is the following code:

    run "scenarios/Challenges/_maypole/solution.sw"

but would **not** achieve a win if executed via `--run` with the solution file:

    scripts/play.sh --scenario data/scenarios/Challenges/maypole.yaml --run data/scenarios/Challenges/_maypole/solution.sw

Placing a `wait 16` at the beginning of the solution script was enough to fix the `--run` invocation.  Presumably, the overhead of the `run` command for the player's `solution` code had previously given the system robot enough of a head start in counting the player's loops to reach the goal count.

Ultimately, instead of placing a delay in the solution, I sped up the system robot with an `instant` command (#1231).  However, inserting `instant` requires some care; it can cause the game to freeze if it gets tangled with a recursion.

Perhaps a better solution in the long term would be to have some way to make system robots (in particular, invisible "judge" robots) execute *all* commands in zero ticks.  This should not be the default for all system robots, especially those that are not `invisible`; sometimes we prefer that their commands have finite duration to interact plausibly with the player.

### DIFF
--- a/data/scenarios/Challenges/_maypole/monitor.sw
+++ b/data/scenarios/Challenges/_maypole/monitor.sw
@@ -58,6 +58,18 @@ def getCurrentQuadrant : (int * int) -> cmd int = \myLoc.
   return $ getQuadrant baseLoc myLoc;
   end;
 
+def checkNewQuadrant = \myLoc. \prevQuadrant. \quadrantTraversalCount.
+  currentQuadrant <- getCurrentQuadrant myLoc;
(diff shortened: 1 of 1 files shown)


byorgey (approved):

kostmo:
@Mergifyio refresh

PR #1248, opened 2023-05-07
Do not show static when no robots exist in creative mode

kostmo:
"Stub" maps generated by the World Editor (#873) do not define any robots.

It is convenient to view these via:

    scripts/play.sh --scenario stub.yaml --cheat

However, the static needs to be suppressed in creative mode when robots do not exist.


byorgey (approved):
Makes sense!

PR #1250, opened 2023-05-10
Inventory search/filter mode

### LINKED ISSUE #126: Inventory search

byorgey:
When the inventory panel is focused, hit `/` to enter search mode (maybe "search" is a bit of a misnomer, it's really more of a "filter" mode).  Then type stuff to modify the current search term, shown at the bottom of the inventory panel.  You can still navigate the filtered inventory list with arrow keys etc. Hit Esc to exit search mode or Enter to pop out the focused item (and also exit search mode).

Closes #126.

### DIFF
--- a/src/Swarm/TUI/Controller.hs
+++ b/src/Swarm/TUI/Controller.hs
@@ -1312,35 +1312,56 @@ adjustTPS (+/-) = uiState . lgTicksPerSecond %~ (+/- 1)
 
 -- | Handle user input events in the robot panel.
 handleRobotPanelEvent :: BrickEvent Name AppEvent -> EventM Name AppState ()
-handleRobotPanelEvent = \case
-  (Key V.KEnter) ->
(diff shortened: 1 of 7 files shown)


byorgey:
It was very gratifying to realize how much functionality could be reused here: we already have machinery for rebuilding the current view of the inventory and flagging when it needs to be rebuilt.  All I had to do was hook into that and add an extra condition to filter out things that don't match the current search.

kostmo (approved):

PR #1252, opened 2023-05-12
Better type variable names

### LINKED ISSUE #1050: Nicer type variable names in inferred types
Currently we get things like 
```
atWS : forall a258 a259 a260. cmd a260 -> cmd a259 -> cmd a258 -> cmd a258
```
:vomiting_face: 

After type inference finishes, we should try renaming the type variables that end up in the types.  Level 1 would be to just pick names `a`, `b`, `c` in order.  Level 2 would be to also try to remember type variable names that were given by users in type annotations, and use those when possible.

byorgey:
Uses type variable names like `a`, `b`, `c`, ... instead of using `aNNN` with numbers based on the internal numbers used for fresh type variables. Closes #1050.

### DIFF
--- a/src/Swarm/Language/Typecheck.hs
+++ b/src/Swarm/Language/Typecheck.hs
@@ -43,6 +43,7 @@ module Swarm.Language.Typecheck (
   isSimpleUType,
 ) where
 
+import Control.Arrow ((***))
 import Control.Category ((>>>))
 import Control.Lens ((^.))
 import Control.Lens.Indexed (itraverse)
@@ -218,10 +219,12 @@ generalize uty = do
   ctxfvs <- freeVars ctx
   let fvs = S.toList $ tmfvs \\ ctxfvs
       alphabet = ['a' .. 'z']
-      xs =
-        take (length fvs) $
-          map T.pack (map (: []) alphabet ++ [x : show n | n <- [0 :: Int ..], x <- alphabet])
-  return $ Forall xs (substU (M.fromList (zip (map Right fvs) (map UTyVar xs))) uty')
+      -- infinite supply of pretty names a, b, ..., z, a0, ... z0, a1, ... z1, ...
+      prettyNames = map T.pack (map (: []) alphabet ++ [x : show n | n <- [0 :: Int ..], x <- alphabet])
+  return $
+    Forall
+      (zipWith const prettyNames fvs)   -- one pretty name per free variable
+      (substU (M.fromList . map (Right *** UTyVar) $ zip fvs prettyNames) uty')
 
 ------------------------------------------------------------
 -- Type errors
--- a/src/Swarm/Language/Typecheck.hs
+++ b/src/Swarm/Language/Typecheck.hs
@@ -60,6 +60,7 @@ import Data.Map qualified as M
 import Data.Maybe
 import Data.Set (Set, (\\))
 import Data.Set qualified as S
+import Data.Text qualified as T
 import Swarm.Language.Context hiding (lookup)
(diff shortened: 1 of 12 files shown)


src/Swarm/Language/Typecheck.hs
```diff
@@ -206,14 +207,20 @@ skolemize (Forall xs uty) = do
...
 generalize uty = do
   uty' <- applyBindings uty
   ctx <- ask
   tmfvs <- freeVars uty'
   ctxfvs <- freeVars ctx
   let fvs = S.toList $ tmfvs \\ ctxfvs
-      xs = map (mkVarName "a") fvs
+      alphabet = ['a' .. 'z']
+      xs =
+        take (length fvs) $
```
kostmo, line 222: Is this `take (length fvs) $` necessary?  The `zip` on line 224 would to the proper list truncation regardless?

kostmo (approved):

kostmo (approved):

src/Swarm/Language/Typecheck.hs
byorgey: You know, I thought the same thing, and so I left out this call to `take` when I initially wrote the code.  Then my computer almost fell over when the test suite started to use all available memory. :smile:   The problem is that `xs` is *also* used as the first argument to `Forall`, so if `xs` is infinite...

I think I can refactor it to be more readable/obvious though...

src/Swarm/Language/Typecheck.hs
```diff
@@ -218,10 +219,12 @@ generalize uty = do
...
-      xs =
-        take (length fvs) $
-          map T.pack (map (: []) alphabet ++ [x : show n | n <- [0 :: Int ..], x <- alphabet])
-  return $ Forall xs (substU (M.fromList (zip (map Right fvs) (map UTyVar xs))) uty')
+      -- infinite supply of pretty names a, b, ..., z, a0, ... z0, a1, ... z1, ...
+      prettyNames = map T.pack (map (: []) alphabet ++ [x : show n | n <- [0 :: Int ..], x <- alphabet])
+  return $
+    Forall
+      (zipWith const prettyNames fvs)   -- one pretty name per free variable
+      (substU (M.fromList . map (Right *** UTyVar) $ zip fvs prettyNames) uty')
```
byorgey, line 227: @kostmo is this more readable?

kostmo: I like that `prettyNames` is an infinite list now.  But I wonder if there is a readable way to avoid referencing it+zipping with it twice on lines 226-227 :smile: ?

src/Swarm/Language/Typecheck.hs
```diff
@@ -219,12 +219,14 @@ generalize uty = do
...
+      -- Infinite supply of pretty names a, b, ..., z, a0, ... z0, a1, ... z1, ...
       prettyNames = map T.pack (map (: []) alphabet ++ [x : show n | n <- [0 :: Int ..], x <- alphabet])
+      -- Associate each free variable with a new pretty name
+      renaming = zip fvs prettyNames
   return $
     Forall
-      (zipWith const prettyNames fvs)   -- one pretty name per free variable
-      (substU (M.fromList . map (Right *** UTyVar) $ zip fvs prettyNames) uty')
+      (map snd renaming)
+      (substU (M.fromList . map (Right *** UTyVar) $ renaming) uty')
```
byorgey, line 229: @kostmo good point re: doing only one `zip`, how's this?

kostmo: Nice :clap:

PR #1256, opened 2023-05-13
`halt` command

### LINKED ISSUE #392: `halt` command to cancel a robot's running program

byorgey:
Closes #392 .  Adds a command `halt : actor -> cmd unit` which halts the given robot if it is within a distance of 1 (no distance limit for system robots or in creative mode).  `halt self` works too.

I'd love some help thinking up a good and/or reasonable and/or funny recipe for a `halting oracle`.  I think we actually want it to be available fairly early on, since making your robots stop when they get out of hand is very useful.

### DIFF
--- a/data/entities.yaml
+++ b/data/entities.yaml
@@ -1249,3 +1249,19 @@
     example `key "Down"` or `key "C-S-x"`.
   properties: [portable]
   capabilities: [handleinput]
+
+- name: halting oracle
+  display:
+    attr: device
+    char: '?'
+  description:
+  - A device to solve the halting problem. When asked if a
+    particular robot program will halt, it always answers YES.
+    And it is always correct... or else!
+  - |
+    Enables the command `halt : actor -> cmd unit` which takes
+    a robot as an argument and, if it is up to one cell away,
+    cancels its currently running program (if any).  In creative mode,
+    there is no distance limitation.
+  properties: [portable]
+  capabilities: [halt]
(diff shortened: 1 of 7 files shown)


data/entities.yaml
```diff
@@ -1249,3 +1249,19 @@
...
   capabilities: [handleinput]
+
+- name: halting oracle
+  display:
+    attr: device
+    char: '?'
+  description:
+  - A device to solve the halting problem. When asked if a
+    particular robot program will halt, it always answers YES.
+    And it is always correct... or else!
```
kostmo, line 1260: This is great.

kostmo (approved):

byorgey:
Oh, I forgot to add some tests!  I will do that before merging.

PR #1261, opened 2023-05-15
decouple relative and absolute time capabilities

kostmo:
`time` now requires `CTimeabs`, and `wait` now requires `CTimerel`.
Introduced a new `hourglass` device to provide `CTimerel` capability.

The `clock` device remains backward compatible, providing both relative and absolute time capabilities.


byorgey (approved):
Looks good to me!

PR #1272, opened 2023-05-20
Allow zero-tick recipes to apply immediately

kostmo:
# Summary

This PR allows a recipe's effects to be applied within the same tick if `time: 0` is specified for the recipe.

The previous behavior was that if the `time` of the recipe was zero, it would nonetheless behave the same as `time: 1`, and the entity updates would not be applied until the next tick.

This new behavior is useful when "abusing" the `drill` command to emulate various other games via help from a system robot.

# Motivation
For the "Lights Out" game (#1273) I have implemented a system robot that `instant`-ly toggles the four adjacent lights and the center light when a center light is `drill`-ed by the player.  The light entities are named `"off"` and `"on"`.

I did some experiments to see if the system robot would react "fast enough" if the player drilled quickly in succession.

# Experiments

The 5x5 light grid is initially `"off"`.  I moved the player robot to the center of the grid, facing `east`. From this position, I tried three command sequence variations both *before* and *after* this PR.  The recipes for toggling lights specify `time: 0`.

### 1. Drilling twice in succession

    drill down; drill left;

This resulted in an error:
no-instant-error-message

so only the first `drill` got applied:

no-instant

This is because I believe the sequence of events was:

| Tick / Robot | Action |
| --- | --- |
| **Tick 1:** Player | Execute the first `drill` command, queue entity modification |
| **Tick 1:** System | No changes observed |
| **Tick 2:** Player | Entity modification from first `drill` is applied.  Execute second `drill` command, queue second entity modifications |
| **Tick 2:** System | Observe modified entity, toggle the adjacent lights |
| **Tick 3:** Player | Try to apply entity modification from second `drill` command. Throw an error because the system robot has toggled the drill's target entity to `on` whereas the `drill` was initially executed when it was still `off`. |

### 2. Instant for both drills

    instant $ (drill down; drill left)

In this case, both of the player's `drill` commands were executed, but the system robot then failed to doubly-invert one of the cells:

instant-both-drills

This is due to some kind of race condition in the system robot's logic.  However, it is irrelevant because the `instant` command shall not be available to the player.

### 3. Instant for first drill only, then second drill

    instant (drill down); drill left;

This accomplished what approach number 1 was supposed to; both drill commands were applied and the two overlapping cells got properly inverted:

instant-first-drill

**Explanation:** In this case, the entire drilling operation (including applying all effects to entities on the ground) was completed within the player's first tick, so the system robot updated the lights appropriately on its turn.  Then, in the next tick the player robot applied another drill command, and again the system robot appropriately did another inversion on the two overlapping middle cells.

### 1. Drilling twice in succession

    drill down; drill left;

Now works as expected:

after-no-instant

### 2. and 3.

Behaved identically to before the PR.

### DIFF
--- a/src/Swarm/Game/Step.hs
+++ b/src/Swarm/Game/Step.hs
@@ -599,7 +599,7 @@ stepRobot r = do
   -- sendIO $ appendFile "out.txt" (prettyString cesk' ++ "\n")
   return $ r' & machine .~ cesk'
 
--- replace some entity in the world with another entity
+-- | replace some entity in the world with another entity
 updateWorld ::
   (Has (State GameState) sig m, Has (Throw Exn) sig m) =>
   Const ->
@@ -608,9 +608,17 @@ updateWorld ::
 updateWorld c (ReplaceEntity loc eThen down) = do
   w <- use world
   let eNow = W.lookupEntity (W.locToCoords loc) w
+  -- Can fail if a robot started a multi-tick "drill" operation on some entity
+  -- and meanwhile another entity swaps it out from under them.
   if Just eThen /= eNow
     then throwError $ cmdExn c ["The", eThen ^. entityName, "is not there."]
     else updateEntityAt loc $ const down
+
+applyRobotUpdates :: (Has (State GameState) sig m, Has (State Robot) sig m) => [RobotUpdate] -> m ()
+applyRobotUpdates =
+  mapM_ \case
+    AddEntity c e -> robotInventory %= E.insertCount c e
+    LearnEntity e -> robotInventory %= E.insertCount 0 e
 
 -- | The main CESK machine workhorse.  Given a robot, look at its CESK
 --   machine state and figure out a single next step.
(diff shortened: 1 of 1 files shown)


src/Swarm/Game/Step.hs
```diff
@@ -2138,11 +2144,19 @@ execConst c vs s k = do
...
     m CESK
-  finishCookingRecipe r v wf rf = do
-    time <- use ticks
-    let remTime = r ^. recipeTime
-    return . (if remTime <= 1 then id else Waiting (remTime + time)) $
-      Out v s (FImmediate c wf rf : k)
+  finishCookingRecipe r v wf rf =
+    if remTime <= 0
+      then do
+        mapM_ (\(ReplaceEntity loc _ newEntity) -> updateEntityAt loc $ const newEntity) wf
```
byorgey, line 2150: Why not call the `updateWorld` function here?  This lambda seems dangerous since it will become partial if we ever add other kinds of world updates.

src/Swarm/Game/Step.hs
```diff
@@ -631,10 +639,8 @@ stepCESK cesk = case cesk of
     case wc of
       Left exn -> return $ Up exn s k
       Right () -> do
-        forM_ rf $ \case
-          AddEntity c e -> robotInventory %= E.insertCount c e
-          LearnEntity e -> robotInventory %= E.insertCount 0 e
-        needsRedraw .= True
+        applyRobotUpdates rf
+        flagRedraw
```
byorgey, line 643: Can we abstract this even further into a single function called both here and in `finishCookingRecipe` when `remTime <= 0`?  As it stands, this seems ripe for a subtle bug where we decide to change the behavior of recipes somehow, and someone changes this code but forgets/doesn't realize that they also have to change the code in `finishCookingRecipe` to match, resulting in inconsistent behavior between recipes with and without `time: 0`.

kostmo:
How's it look now?

byorgey (approved):
Looks good!

PR #1276, opened 2023-05-20
Use SKPair in CESK constructors

kostmo:
The pair of `s` and `k` often get threaded through functions together unchanged.  Bundling them in the constuctors of `In`, `Out`, and `Up` allows some code to be simplified.

### DIFF
--- a/src/Swarm/Game/CESK.hs
+++ b/src/Swarm/Game/CESK.hs
@@ -59,6 +59,7 @@ module Swarm.Game.CESK (
   RobotUpdate (..),
 
   -- * Store
+  SKPair (..),
   Store,
   Addr,
   emptyStore,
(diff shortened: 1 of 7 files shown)


byorgey:
I'm not necessarily against this, but guess I'm not sure what it really buys us.  In a bunch of places we now get to write `sk` instead of `s k`, but there are also a bunch of places where we now need an extra `SK` constructor, extra parentheses, etc.  Were there any other big simplifications I missed?

kostmo:
OK, brace yourself, my idea is to introduce another sum type layer to factor out `SKPair` from the `In`/`Out`/`Up` constructors: see #1281.

That change itself does not simplify the code; in fact it adds another `Go` constructor and lots more parentheses!  However, my wager is that it opens up refactoring potential which I am still exploring.

PR #1277, opened 2023-05-22
Recreate `GameState` from scratch when starting a scenario

### LINKED ISSUE #516: Fix scenario loading to game state

### LINKED ISSUE #689: Custom entities persist when quitting scenario

byorgey:
All info that needs to be persistent (*e.g.* the set of standard/default entities and recipes) is now stored in the `RuntimeState`.  The `GameState` record is now recreated completely from scratch every time upon starting a new scenario, so custom entities and recipes from the previous scenario do not persist into the next.

Fixes #516.  Fixes #689.

Also a bunch of related refactoring/simplification relating to how we were initializing state for tests and benchmarks.

I also want to refactor the tests so we only load data from disk once, but I have split that out into a separate issue (#1279).

### DIFF
--- a/bench/Benchmark.hs
+++ b/bench/Benchmark.hs
@@ -16,14 +16,14 @@ import Swarm.Game.Display (defaultRobotDisplay)
 import Swarm.Game.Location
 import Swarm.Game.Robot (TRobot, mkRobot)
 import Swarm.Game.State (GameState, addTRobot, creativeMode, world)
-import Swarm.TUI.Model (gameState)
-import Swarm.TUI.Model.StateUpdate (classicGame0)
 import Swarm.Game.Step (gameTick)
 import Swarm.Game.Terrain (TerrainType (DirtT))
 import Swarm.Game.World (WorldFun (..), newWorld)
 import Swarm.Language.Context qualified as Context
 import Swarm.Language.Pipeline (ProcessedTerm)
 import Swarm.Language.Pipeline.QQ (tmQ)
+import Swarm.TUI.Model (gameState)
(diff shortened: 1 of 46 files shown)


byorgey:
@kostmo I always forget about the benchmarks. :facepalm:   Should be ready for you to take a look now!

kostmo (approved):
lgtm

PR #1283, opened 2023-05-27
Rework type checker using "propagate types inward" trick

### LINKED ISSUE #99: Redo type checker using "propagate types inward" trick

byorgey:
The basic idea is that we try as hard as possible to stay in "checking mode", where we have an "expected type" from the context and "push it down" through a term.  We also try hard to avoid generating fresh unification type variables unless absolutely necessary.  That way, when we come to something that doesn't match the expected type, we can immediately flag a specific error about that part of the code, rather than later having a generic unification error.

This is mostly just refactoring, but it already improves the type error messages somewhat:

* Instead of saying something like "can't unify int and text", it might say something like "expected type int, but got text" with a reference to a specific source code location.
*  It's probably possible to still get it to sometimes fail with a "can't unify" type error, but with only this technique I don't think we can ever 100% get rid of that.
* Type errors now report a specific line and column number, not just a line number like they used to. 

There is still a lot more I want to do to improve the type error messages but this is a start.

Closes #99.


src/Swarm/Language/Typecheck.hs
```diff
@@ -99,11 +100,18 @@ runInfer ctx =
...
-lookup :: SrcLoc -> Var -> Infer UType
+lookup :: SrcLoc -> Var -> TC UType
 lookup loc x = do
   ctx <- ask
   maybe (throwError $ UnboundVar loc x) instantiate (Ctx.lookup x ctx)
 
+-- | Add a source location to a type error and re-throw it.
+addLocToTypeErr :: Syntax' ty -> TypeErr -> TC a
+addLocToTypeErr s te = case te of
+  UnifyErr NoLoc a b -> throwError $ UnifyErr (s ^. sLoc) a b
```
kostmo, line 111: can `throwError` be moved outside of the `case` expression?

src/Swarm/Language/Typecheck.hs
```diff
@@ -146,14 +154,44 @@ substU m =
...
+  upty' <- applyBindings upty
+  let tyvs =
+        ucata
+          (const S.empty)
+          (\case TyVarF v -> S.singleton v; f -> fold f)
+          upty'
+      ftyvs = tyvs `S.difference` S.fromList xs
+  unless (S.null ftyvs) $
+    throwError $
+      EscapedSkolem l (head (S.toList ftyvs))
```
kostmo, line 169: How about:
```suggestion
  forM_ (S.lookupMin ftyvs) $ throwError . EscapedSkolem l
```

kostmo (approved):

src/Swarm/Language/Typecheck.hs
byorgey: Good call, but I actually want to completely rewrite `addLocToTypeErr` (it should actually have a type like `SrcLoc -> TC a -> TC a`, and maybe a different name to go with it; it should handle all the different kinds of type errors, to facilitate which we should factor the `SrcLoc` out so it's not duplicated in every constructor, etc.) so I'd prefer to just wait and do all the refactoring at once.

PR #1285, opened 2023-05-28
Render contiguous boundaries

### LINKED ISSUE #1271: Draw annotated contiguous boundaries with proper wall glyphs
Can create a new entity property called `"boundary"` for which, upon loading the scenario, contiguous boundaries of this entity will be automatically replaced with the appropriate corner or straight wall entities.

E.g.:

```
entities:
  - name: wall
    display:
      char: '#'
    description:
    - Solid boundary
    properties: [known, unwalkable, boundary]

........
..####..
..#..#..
..#..#..
..####..
........
```
would be replaced with:

```

........
..┌──┐..
..│..│..
..│..│..
..└──┘..
........
```

Should be a fun programming exercise.

kostmo:
Closes #1271

## Demo

    scripts/play.sh --scenario data/scenarios/Testing/1271-wall-boundaries.yaml --autoplay

Screenshot from 2024-09-18 09-27-14

## Implementation

This implements neighbor-aware boundary rendering strictly on the display layer.  Underlying entities remain unchanged (for the purpose of world cell contents and e.g. `scan`).  Eventually we should be able to retire all of the niche boundary glyph entities.

Only a single function `getBoundaryDisplay` is exposed for computing the glyph appropriate for adjacent boundaries.  It takes a function from `AbsoluteDir -> Bool` indicating whether the neighbor cell in a particular direction is (also) a "boundary" entity.

The internal `glyphForNeighbors` function is implemented with pattern-matching such that exhaustive coverage of cases is guaranteed.

### DIFF
--- a/data/scenarios/Testing/00-ORDER.txt
+++ b/data/scenarios/Testing/00-ORDER.txt
@@ -36,6 +36,7 @@ Achievements
 1218-stride-command.yaml
 1234-push-command.yaml
 1256-halt-command.yaml
+1271-wall-boundaries.yaml
 1262-display-device-commands.yaml
 1295-density-command.yaml
 1138-structures
--- a/data/scenarios/Testing/1271-wall-boundaries.yaml
+++ b/data/scenarios/Testing/1271-wall-boundaries.yaml
@@ -1,0 +1,96 @@
+version: 1
+name: Wall boundary display
+creative: false
+description: |
+  Demonstrate rendering of contiguous boundaries.
+
+  Only adjacent with the 'boundary' property
+  will be rendered with box border glyphs.
+attrs:
+  - name: purpleWall
+    fg: '#ff00ff'
+  - name: cyanWall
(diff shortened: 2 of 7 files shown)


xsebek (approved):
@kostmo Looks good! 👍 

I was initially put off by having the variants hardcoded, but on further thought this simple approach makes sense.

The PR description confused me though, is the discussion up to date?

byorgey (approved):
Looks good to me, though I can't help thinking that the complication with the `orientationMap` necessitates adding yet more complication in the form of a `boundaryOverride` field in a `Display`.

This is just speculation for a future PR... what if a `Display` contained just a single `Char`, and an "orientation map" was stored at a higher level as a function from direction to `Display`?  Likewise, then `Display` wouldn't have to know anything about boundary overrides --- if you wanted to show a boundary you would just take the `Display` you got an replace its `Char` with the appropriate box drawing character.

Issue #1238, opened 2023-05-01
"act" command as automatic variant of "use"


kostmo:
Building upon #1007...

Whereas the proposed type signature of `use` is:

    use :: entity -> direction -> cmd unit

I propose an additional command `act` that automatically uses the "best" device to perform an **action**, similar to the current semantics of `drill`:

    act :: direction -> cmd unit

This would preserve the convenience and interesting gameplay mechanics of `drill` while shedding the connotations of the `drill` name.

Implementing this could be as simple as renaming `drill` to `act`.  Though `drill` may be re-introduced as a specialization of `use`.

byorgey:
See #1286 .

PR #1287, opened 2023-05-29
'use' command

kostmo:
Towards #1007


xsebek:
Thanks for taking this up @kostmo! 👍 

Out of curiosity, is removing `drill` the only missing thing to close the Issue?

xsebek (approved):
LGTM! But I have not touched Haskell in weeks so any code would look beautiful. 🤩

xsebek (approved):
LGTM! But I have not touched Haskell in weeks so any code would look beautiful. 🤩

kostmo:
> is removing `drill` the only missing thing to close the Issue?

I think that removing `drill` can be a separate issue (if we decide to do so at all), so I've closed it.  Note there is also a separate issue #1238 to implement something even more drill-like.

data/scenarios/Testing/1007-use-command.yaml
```diff
@@ -0,0 +1,83 @@
...
+  description:
+  - Impassable barrier
+  properties: [known, unwalkable]
+- name: gate key
+  display:
+    attr: iron
+    char: 'k'
+  description:
+  - Can open a closed gate
+  capabilities: [drill]
```
byorgey, line 44: It's a bit strange the the capability is still named `drill`.  But maybe we'll generalize that in a future PR?

src/Swarm/Game/Step.hs
```diff
@@ -2048,30 +2055,33 @@ execConst c vs s k = do
...
         -- E.g. "metal drill" vs. "drill"
         preferredDrill = listToMaybe $ sortOn (Down . T.length . (^. entityName)) equippedDrills
 
-    drill <- preferredDrill `isJustOr` Fatal "Drill is required but not equipped?!"
+    tool <- preferredDrill `isJustOr` Fatal "Drill is required but not equipped?!"
+    applyDevice ins "drill" d tool
 
-    (nextLoc, nextE) <- getDrillTarget "drill" d
+  applyDevice ins verbPhrase d tool = do
+    (nextLoc, nextE) <- getDrillTarget verbPhrase d
```
byorgey, line 2069: It might be nice to clean up the remaining variable names with `drill` in them and make them more generic, like `getDeviceTarget` instead of `getDrillTarget`.

src/Swarm/Language/Capability.hs
```diff
@@ -224,6 +224,7 @@ constCaps = \case
   Meet -> Just CMeet
   MeetAll -> Just CMeet
   Drill -> Just CDrill
+  Use -> Nothing -- Recipes alone shall dictate whether things can be "used"
```
byorgey, line 227: I will note that one possibly strange thing about this is that if you write a program like `build {move; use "flerb" forward; move}` it will fail, because the requirements analysis will not figure out that we should automatically equip a `flerb` when `build`ing the child robot.  You would have to explicitly write e.g. `build {move; require "flerb"; use "flerb" forward; move}`.  We might consider adding something to the requirements analysis which will at least work when `use` is applied to a statically known device name, as in this example.

data/scenarios/Testing/1007-use-command.yaml
kostmo: In fact, no capability is required; I'll remove this line.

src/Swarm/Language/Capability.hs
kostmo: Created #1301

src/Swarm/Game/Step.hs
kostmo: Fixed in #1299

PR #1294, opened 2023-05-31
Refactor to factor out type error `SrcLoc`

byorgey:
Some refactoring related to type errors:

- Change the type of `addLocToTypeErr` to something more sensible, and get rid of some redundant calls
- Instead of storing a `SrcLoc` in every single `TypeErr` constructor, factor it out into a new type `ContextualTypeErr`.  This also paves the way for additional contextual information which will help generate better type error messages.

This is 100% refactoring; there should be no behavioral changes.


src/Swarm/Language/Typecheck.hs
```diff
@@ -164,7 +165,7 @@ noSkolems l (Forall xs upty) = do
           (\case TyVarF v -> S.singleton v; f -> fold f)
           upty'
       ftyvs = tyvs `S.difference` S.fromList xs
-  forM_ (S.lookupMin ftyvs) $ throwError . EscapedSkolem l
+  forM_ (S.lookupMin ftyvs) $ throwTypeErr l . EscapedSkolem
```
xsebek, line 168: Does this get to throw more than once?

I would expect this to boil down to “find at most one unbound variable”.

src/Swarm/Language/Typecheck.hs
```diff
@@ -267,42 +268,60 @@ generalize uty = do
 ------------------------------------------------------------
 -- Type errors
 
+-- | A type error along with various contextual information to help us
+--   generate better error messages.  For now there is only a SrcLoc
+--   but there will be additional context in the future.
```
xsebek, line 273: Intriguing! 🧐 It would be nice to give a hint here and maybe link to an Issue.

xsebek (approved):
I like it! 👍

src/Swarm/Language/Typecheck.hs
byorgey: Yes, `S.lookupMin :: Set a -> Maybe a` returns the smallest element of a set, or `Nothing` if the set is empty.  We don't actually care about getting the *smallest* type variable, but `lookupMin` is a convenient way to say "just get me one element if any exist".  The `forM_` is actually over the `Maybe` so it represents doing the `throwTypeErr` either zero or one time, depending on whether `lookupMin` returns `Nothing` or `Just`.

I will add a comment to this effect in the code.

PR #1298, opened 2023-06-02
Better type error messages

byorgey:
Improvements to type inference and type error messages.  Includes:
- More informative error when record literal fields don't match the expected set of fields
- More informative type mismatch error which includes the term where the mismatch occurred.
- Reinstates a case for lambdas with explicit type annotations in inference mode (which was removed in #1283), which helps preserve better error messages in some cases.


xsebek (approved):
One small code change, huge leap for error messages! 🚀 👍

ussgarci (approved):
:+1:

Issue #1087, opened 2023-02-05
`suspend`, i.e. the Secret Master Plan for Fixing All The Things


byorgey:
The big idea
------------

What happens, theoretically and practically, when you type stuff at the REPL?  The way I always think about it is that it's as if you're writing one big long block of commands chained by semicolons; it just so happens that you get to enter them one at a time.  This explains why names defined at the REPL should be in scope for later things entered at the REPL.  (Well, you can also enter expressions to be evaluated in addition to commands, but this can easily be explained by saying that anything without a command type gets implicitly wrapped in `return`.)

So what is the `base` doing in between REPL entries?  Currently, it is idle, just like any robot that has finished running its program. But **this is wrong**!  We should have a **new CESK machine state** to represent what the base is doing while waiting for another REPL input. 

- Fundamentally, the base is doing something different than an idle   robot. An idle robot will never do anything more unless another   robot comes to `salvage` or `reprogram` it.  The base will resume   executing as soon as something else is entered at the REPL.  If only   for conceptual reasons it makes sense to distinguish between these   states.

- If the base is idle, its CESK machine no longer remembers any   context or store.  So we have to save them into the `robotContext`   just so we can load them back into the new CESK machine state on the   next REPL input.  I now realize how silly this is!  Why not just   have the CESK machine remember it all along?

Details
-------

I propose adding a **new CESK machine state** `Suspended` and a **new primitive constant** `suspend`.

- The new machine state could be called `Suspended`, and will contain   a result `Value`, a current `Env`, and a current `Store`.  I do not   think it needs to have a continuation (since I cannot think when we   would ever want anything besides the empty continuation), but we   could easily add a continuation if it turns out to be useful later. 
- The new primitive is `suspend : a -> cmd a`.
    - When `suspend v` is executed, it has the effect of putting the machine immediately into the `Suspended` state, with `v` as the value, and the current environment and store.
    - `suspend` should not be available for players to use, but will only be used internally.
    - Whenever a command like `x; y; z` is entered at the REPL, it will be automatically rewritten to `x; y; v <- z; suspend v`.
      - Note that the `suspend` comes last, so there will in fact be an empty continuation at the point when it is executed, which is why the `Suspended` state does not need to store one.
      - On the other hand if we later decided we wanted to allow things like `x; suspend v; y` then we would need to save a continuation.
- Instead of checking to see when the base is idle, the REPL will now check to see when it becomes `Suspended`
[...]

PR #1303, opened 2023-06-03
combine all cesk simplifications

kostmo:
This is a flattening of the three stacked PRs:
1. #1276
2. #1281
3. #1302

As I was working on this, I realized that certain adjacent cases inside the `stepCESK` function read like a dialogue (e.g. evaluating the terms of a pair), so it wouldn't make sense to separate them by grouping all of the `In` and all of the `Out` cases together.

I also realized that the pattern matching in `stepCESK` often considers the *combination* of the `Go` and `SKPair` fields, so this limits the opportunities to group into sub-cases by field.

The obvious wins in this refactoring are in `prettyPrec` and `getStore`, but in terms of code volume they may be overshadowed by the extra constructors in the `stepCESK` function.

To mitigate the overhead of these extra constructors, I introduced helper functions where possible: `returnVal`, `returnResult`, `returnException`, and `pushFrame`.  Eventually, applying Pattern Synonyms may get us even more mileage out of this refactoring.

Overall, @byorgey I'll let you be the judge if this refactoring as a whole is worth it, or if perhaps just parts of it should be split off.

### DIFF
--- a/src/Swarm/Game/CESK.hs
+++ b/src/Swarm/Game/CESK.hs
@@ -60,6 +60,7 @@ module Swarm.Game.CESK (
 
   -- * Store
   SKPair (..),
+  Go (..),
   Store,
   Addr,
   emptyStore,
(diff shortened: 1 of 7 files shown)


byorgey (commented):
@kostmo, I appreciate all your work on this, and it was worth a try, but unfortunately this doesn't seem worth it to me.  The extra constructors introduce a lot of noise (definitely far more than what we save in functions like `prettyCESK`) and I think the functions you introduced to try to mitigate this (`returnVal` et al) actually make things worse, in the sense of making the code harder to read (especially when thinking about a new contributor trying to read the code).  Before, to understand what the `stepCESK` function was doing you just had to know about the different `CESK` constructors and you could see it literally sending certain states to other states.  Now you have to understand several nested data types and what constructors will be output by the various functions like `returnVal`.

The `SKPair` type doesn't really seem meaningful beyond "the part of the machine that we always have no matter which state we're in" (well, except for when we are `Waiting`... and except for when we are `Suspended` (#1087 )...).  Perhaps `Go` is meaningful, as in, "we are in a state where the machine is actively doing something", but overall splitting up the `CESK` into so many sub-types seems to just make it harder to understand, without any big benefits that I can see (though I'm very open to being corrected as far as benefits go... perhaps there are benefits to the refactoring that I've missed).

There are some typo fixes, extra comments etc. in here which should obviously be split out into a separate PR and merged.  Whether there is anything else worth splitting out I'm not sure; open to suggestions.

kostmo:
Yes, that's a fair assessment, and I appreciate you spending the time to review the changes.  If nothing else, this exercise did help me understand the CESK machine!  In trying to "improve" it, I got a better sense of why it is written the way it was.

byorgey:
> If nothing else, this exercise did help me understand the CESK machine! In trying to "improve" it, I got a better sense of why it is written the way it was.

If you can think of any extra comments that could be added to help preserve your insights for others reading the code later, that would be a great thing to include in a new PR.

PR #1305, opened 2023-06-04
ensure progress gets saved

kostmo:
I believe that since #1277 the progress has not been saved when exiting a scenario back to the menu.

It is because `getNormalizedCurrentScenarioPath`, which is utilized by `saveScenarioInfoOnQuit`, was returning `Nothing`, in turn because `currentScenarioPath` was `Nothing`.  This is because `scenarioToAppState` was clearing it immediately after being set inside `startGameWithSeed`.

Foreshadowed by this comment: https://github.com/swarm-game/swarm/pull/1243#pullrequestreview-1411539743 .

### DIFF
--- a/src/Swarm/TUI/Model/StateUpdate.hs
+++ b/src/Swarm/TUI/Model/StateUpdate.hs
@@ -110,7 +110,6 @@ startGameWithSeed userSeed siPair@(_scene, si) toRun = do
   t <- liftIO getZonedTime
   ss <- use $ runtimeState . scenarios
   p <- liftIO $ normalizeScenarioPath ss (si ^. scenarioPath)
-  gameState . currentScenarioPath .= Just p
   runtimeState
     . scenarios
     . scenarioItemByPath p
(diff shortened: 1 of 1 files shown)


src/Swarm/TUI/Model/StateUpdate.hs
```diff
@@ -119,6 +118,7 @@ startGameWithSeed userSeed siPair@(_scene, si) toRun = do
     . scenarioStatus
     .= Played (Metric Attempted $ ProgressStats t emptyAttemptMetric) (prevBest t)
   scenarioToAppState siPair userSeed toRun
+  gameState . currentScenarioPath .= Just p
```
xsebek, line 121: @kostmo could you add a short comment here please? I would hate to break this again in refactoring. 😅

xsebek (approved):
Nice find @kostmo! 😮 

I was just about to create an Issue that achievements don't work in clean install from `main` branch, but this fixes it for me. 👍

Issue #1297, opened 2023-06-01
Add more context to type errors


byorgey:
Type errors now give a precise source location, but it would be nice to add extra information about the context at the time the type error occurred, for example
- The expression we were checking
- Why we were expecting the type that we did
- Relevant things that are in scope

See:
- [x] #1308
- [x] #1314
- [x] #1336

PR #1308, opened 2023-06-08
More typechecking error message improvements

byorgey:
Towards #1297 .  In this PR:

- Moves some functions out of `Pipeline` and into more appropriate modules
- Adds a new type `Join` to represent two types or other things, one "expected" and one "actual", at the point where they join + we check whether they are equal; a bunch of refactoring to use this new type.  This way we don't have to remember which argument is which when calling a function that takes one "expected" thing and one "actual" thing.
- Refactors all the `decomposeXXX` functions to take a `Syntax` node to use in error messages, and to take a `Sourced UType` (*i.e.* a `UType` along with whether it is "expected" or "actual") instead of just a `UType` so we can generate accurate error messages
- Introduce a basic typechecking stack that keeps track of "what we are currently in the middle of checking", so that when we encounter an error we can say things like "while checking the definition of `foo`".  Currently this is just a basic proof of concept.
- A few other miscellaneous error message improvements.

### DIFF
--- a/src/Swarm/Language/Typecheck.hs
+++ b/src/Swarm/Language/Typecheck.hs
@@ -19,11 +19,16 @@ module Swarm.Language.Typecheck (
   InvalidAtomicReason (..),
 
   -- * Type provenance
-  Source(..), Join, getJoin,
+  Source (..),
+  Join,
+  getJoin,
 
   -- * Typechecking stack
-
-  TCFrame(..), TCStack, withFrame, getTCStack,
+  TCFrame (..),
+  LocatedTCFrame (..),
+  TCStack,
+  withFrame,
+  getTCStack,
 
   -- * Typechecking monad
   TC,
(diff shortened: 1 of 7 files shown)


kostmo:
FYI rebasing on `main` will fix the `restyled` error.

src/Swarm/Language/Typecheck.hs
```diff
@@ -527,19 +615,38 @@ infer s@(Syntax l t) = addLocToTypeErr l $ case t of
...
-    return $ Syntax' l (SApp f' x') resTy
+
+    -- Call applyBindings explicitly, so that anything we learned
+    -- about unification variables while checking the type of the
+    -- argument can flow to later steps.  This is especially helpful
+    -- while checking applications of polymorphic multi-argument
+    -- functions such as 'if'.  Without this call to 'applyBindings',
+    -- type mismatches between the branches of an 'if' tend to get
+    -- caught in the unifier, resulting in vague "can't unify"
+    -- messages (for example, "if true then {3} {move}" yields "can't
```
kostmo, line 630: typo?
```suggestion
    -- messages (for example, "if true {3} {move}" yields "can't
```

kostmo (approved):

PR #1318, opened 2023-06-09
Better type error messages when there are unification variables involved

### LINKED ISSUE #1313: Use more descriptive English terms in type error messages
This is a follow-up to #1308.

We should try harder to use more descriptive English terms in error messages, especially where unification variables or skolem variables occur in types. For example, instead of saying "expecting xyz to have type u3 -> u4 but it actually..." we could say something like "expecting xyz to be a function but it actually..."  In other words, when printing type error messages, we should do some work to analyze the types involved and see if we can describe them in some way that doesn't involve a bunch of type variables, rather than simply pretty-printing them.  At the very least we could replace unification variables with underscores or ellipses, as in "expecting xyz to have a type like `(_ -> _) * _`, but ..."

So, concretely, I propose the following.  In the context of `expecting xyz to ...`
- If a type has no variables, pretty-print it directly: ``have type `int -> text` ``
- If a type consists of a top-level type constructor with variables in it, print it as a phrase: `be a function`
- If a type is more complex but has variables in it, print it with underscores in place of the variables: ``have a type like `(_ -> _) * _` ``

byorgey:
Say things like "expecting `xyz` to be a function" instead of "expecting `xyz` to have type `u3 -> u4`".  Closes #1313.

### DIFF
--- a/test/unit/TestLanguagePipeline.hs
+++ b/test/unit/TestLanguagePipeline.hs
@@ -362,6 +362,18 @@ testLanguagePipeline =
                 "def m : int -> int -> int = \\x. \\y. {3} end"
                 "1:37: Type mismatch:\n  From context, expected `{3}` to have type `int`,\n  but it is actually a delayed expression\n\n  - While checking the definition of m"
             )
+        , testCase
+            "comparing two incompatible functions"
+            ( process
+                "(\\f:int -> text. f 3) (\\x:int. 3)"
+                "1:32: Type mismatch:\n  From context, expected `3` to have type `text`,\n  but it actually has type `int`\n"
+            )
+        , testCase
+            "comparing two incompatible functions 2"
+            ( process
+                "(\\f:int -> text. f 3) (\\x:int. \\y:int. \"hi\")"
+                "1:32: Type mismatch:\n  From context, expected `\\y:int. \"hi\"` to have type `text`,\n  but it is actually a function\n"
+            )
         ]
     ]
  where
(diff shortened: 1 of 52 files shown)


test/unit/TestLanguagePipeline.hs
```diff
@@ -333,7 +333,34 @@ testLanguagePipeline =
...
+            "mismatched if branches"
+            ( process
+                "if true {grab} {}"
+                "1:16: Type mismatch:\n  From context, expected `noop` to have type `cmd text`,\n  but it actually has type `cmd unit`"
+            )
+        , testCase
+            "definition with wrong result"
+            ( process
+                "def m : int -> int -> int = \\x. \\y. {3} end"
+                "1:37: Type mismatch:\n  From context, expected `{3}` to have type `int`,\n  but it is actually a delayed expression"
```
xsebek, line 363: Could you please  add a test case for two functions?

It seems to be handled based on the `cmd` test case above, but it would be nice to be sure we don't say:
> expected `f` to be a _function_, but it is actually a _function!_

xsebek (approved):
Those new error messages sure sound more human! 👍 I also like more context in type checking, even if that comes at the cost of some heavy `lift`ing. 😄

test/unit/TestLanguagePipeline.hs
byorgey: I added a couple test cases involving incompatible functions to show what happens.  We will never say "expected f to be a function, but it is actually a function", for two reasons: first, if we are trying to unify two types and they are both function types, then we will proceed to unifying their input and output types, so we would never fail at the point where we have two function types.  Second, we only say "blah is a function" when it specifically has a type of the form `u1 -> u2` where `u1` and `u2` are unification variables.  Any two such types will be guaranteed to be unifiable.

PR #1322, opened 2023-06-09
Don't step robots that are waiting, even during `instant`

### LINKED ISSUE #1267: `instant (wait 1)` causes the game to freeze
In creative mode, or with a system robot, it is possible to completely freeze the game by executing something like `instant (wait 1)`.  See the original comment here: https://github.com/swarm-game/swarm/issues/1245#issuecomment-1546731939

What *should* `instant (wait 1)` do?  I think the fact that it causes a freeze should be considered a bug, but it is unclear to me what it ought to do instead.  Options include:
1. `wait` commands executed inside `instant` are simply ignored, so *e.g.* `instant (move; wait 1; move)` would be equivalent to `instant (move; move)`
2. `wait` commands inside `instant` terminate:
    - No error: If a `wait` command is executed inside `instant` then it goes into effect and immediately terminates the `instant` command; any subsequent commands are ignored.  *e.g.* `instant (move; wait 1; move)` would only end up moving once.
    - With error: If a `wait` command is executed inside `instant` then an exception is thrown.  *e.g.* `instance (move; wait 1; move)` would only move once, and would fail with an exception.
4. If a `wait` command is executed inside `instant`, then it goes into effect immediately, any any subsequent commands will be executed `instant`ly once the robot wakes up.  *e.g.* `instant (move; move; wait 1; move; move)` would move twice in one tick and then twice in the next tick.

I think these are ordered roughly from easiest to hardest in terms of implementation difficulty.

byorgey:
This fixes #1267 in the best possible way (option 3): if a `wait` command is executed inside `instant`, then it goes into effect immediately, and any subsequent commands will be executed `instant`ly once the robot wakes up. For example, `instant (move; move; wait 1; move; move)` moves twice in one tick and then twice in the next tick.  Even calls to `wait` nested inside recursive function calls inside a call to `instant` work fine.

### DIFF
--- a/src/Swarm/Game/Step.hs
+++ b/src/Swarm/Game/Step.hs
@@ -1,7 +1,6 @@
 {-# LANGUAGE BlockArguments #-}
 {-# LANGUAGE ConstraintKinds #-}
 {-# LANGUAGE DataKinds #-}
-{-# LANGUAGE MultiWayIf #-}
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE QuasiQuotes #-}
 {-# LANGUAGE RecordWildCards #-}
@@ -587,10 +586,9 @@ tickRobot r = do
 tickRobotRec :: (Has (State GameState) sig m, Has (Lift IO) sig m) => Robot -> m Robot
 tickRobotRec r = do
   time <- use ticks
-  if
-    | wantsToStep time r && (r ^. runningAtomic || r ^. tickSteps > 0) ->
-        stepRobot r >>= tickRobotRec
-    | otherwise -> return r
+  case wantsToStep time r && (r ^. runningAtomic || r ^. tickSteps > 0) of
+    True -> stepRobot r >>= tickRobotRec
+    False -> return r
 
 -- | Single-step a robot by decrementing its 'tickSteps' counter and
 --   running its CESK machine for one step.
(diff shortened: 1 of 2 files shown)


byorgey:
This was way easier than I thought it was going to be!  Essentially, we have some logic in the `tickRobotRec` function which decides when to keep stepping a robot and when to stop and let another robot run.  The `instant` command sets a flag which overrides the usual logic about a maximum number of steps; fixing this was as simple as making sure that robots which are in a waiting state will never be run, even if the `instant` flag is set.  This actually makes normal `wait` commands more efficient too, since in the past if a robot executed `wait` we would keep stepping it (which turned into a no-op) until it ran out of steps for the current tick.  Now we will stop stepping a robot immediately when it executes `wait`.

src/Swarm/Game/Step.hs
```diff
@@ -584,10 +585,12 @@ tickRobot r = do
...
-tickRobotRec r
-  | isActive r && (r ^. runningAtomic || r ^. tickSteps > 0) =
-      stepRobot r >>= tickRobotRec
-  | otherwise = return r
+tickRobotRec r = do
+  time <- use ticks
+  if
+      | wantsToStep time r && (r ^. runningAtomic || r ^. tickSteps > 0) ->
+          stepRobot r >>= tickRobotRec
+      | otherwise -> return r
```
kostmo, line 593: Seems that `hlint` doesn't like this.

kostmo (approved):

src/Swarm/Game/Step.hs
byorgey: Ugh, I hate `if ... then ... else ...` syntax in Haskell.  But I think I just remembered we already have a rule in `.hlint.yaml` to allow `case` expressions on Booleans, so I will switch to using that.

PR #1330, opened 2023-06-10
wrap TickNumber in newtype

kostmo:
newtypes > aliases

### DIFF
--- a/src/Swarm/Game/CESK.hs
+++ b/src/Swarm/Game/CESK.hs
@@ -79,7 +79,7 @@ module Swarm.Game.CESK (
   -- ** Extracting information
   finalValue,
   TickNumber (..),
-  addTo,
+  addTicks,
 ) where
 
 import Control.Lens ((^.))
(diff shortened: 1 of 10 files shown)


byorgey (approved):
Looks great, and I 100% agree that a newtype is better than a type alias here.  I just want to bikeshed about the names a little bit, since `asInteger` and `addTo` seem too generic.
* Rather than `asInteger`, the traditional name for a `newtype` field projection would be something like `getTickNumber` or `unTickNumber`, or we could just use `getTicks` for short.  But in any case something that reminds us it has to do with `TickNumber`.  (If we name it `asInteger`, what are we going to name the field projection of the next `newtype` wrapper around `Integer` we decide to make?  **edited to add**: I could imagine naming it `asInteger` if it was intended to be imported qualified; but then `TickNumber` would have to go in its own separate module.)
* `addTo` also seems a bit too generic to me.  Maybe `addTicks`?  (In my ideal world we would note that the group of integers under addition form a torsor, i.e. free group action, over `TickNumber`, and we could make some appropriate instances... but in the actual world we live in that would create too much extra noise and boilerplate.)

Issue #1320, opened 2023-06-09
World function DSL


xsebek:
**Is your feature request related to a problem? Please describe.**

The world function is currently static except for _seed_. Being able to parametrise it would be helpful in several places:
1. in custom scenarios
2. in scenario launch options (#1290)
3. in save files (#50)
4. in world editor? (#873)

Another related problem is that we currently have competing ways to specify the world.
Namely specifying the seed, default entity and a part of map.
When the default entity is specified, it will disregard the seed and default world function.

**Describe the solution you'd like**
We should create a simple configuration scheme, that we can play with, modify and extend.

In YAML the current world function could look like this:
```yaml
world:
  parameters:
    noise: # WorldGen.hs: Perlin noises p0, p1, p2, cl0
      smallBig: [0, 6, 0.05, 0.6]  # seed, number of octaves, scale, and persistance.
      softHard: [1, 6, 0.05, 0.6]  # we could specify those as object parameters or tuple
      naturalArtificial: [2, 6, 0.05, 0.6]  # here I am using tuple to be brief, but object is likely better
      choice: [0, 4, 0.08, 0.5]
    biomes:
      big:
        hard:
          natural:  # WorldGen.hs: genBiome Big Hard Natural
            # use object for conditions
            - [{sample: 0.5}, "stone", "mountain"]
            # interpret top-down
            - [{murmur: {mod: 30, equal: 0}}, "stone", "boulder"]
            # combine conditions
            - [{sample: 0.5; murmur: {mod: 30, equal: 0}}, "stone", "LaTeX"]
            - [{sample: 0.5}, "stone", "tree"] 
            # use otherwise string like in Haskell and null for no entity
            - [otherwise, "grass", null]
          artificial:
             # ...
      small:
        # ...
```

The Haskell datatype `WorldParameters` would follow this structure and get interpreted like `testWorld2`.

**Describe alternatives you've considered**

In general (ignoring the specific syntax proposed here) the only alternative is passing parameters ad-hoc in code.
That is fine for current seed and default entity, but will need more structured approach later.

**Additional context**

- We should consider how changes to the scheme will impact saves.
  Adding a new field should not brick all player save files... at least once we have a stable release. 😄 
- We should measure the performance impact.
  A new benchmark would be needed.

xsebek:
@kostmo @byorgey what do you think about this proposal?

The proposed format could subsume current world function, but might be more generic then what was imagined in #1290.

[...]

PR #1332, opened 2023-06-10
structure templates

### LINKED ISSUE #1138: Structure templates in YAML
Building upon #112, there should be a mechanism to define multi-cell structures.  It should follow the model of Entities, which can be defined both globally in a dedicated YAML file and can have custom definitions in scenario files.

Structures should support:
* orientation
* transparency (aka "masking")
* location

Structures would be defined similar to maps are today, and will include their own "palette".  The palette shall reference pre-existing entities (either built-in or defined elsewhere in the scenario).  See also [this comment](https://github.com/swarm-game/swarm/issues/242#issuecomment-1133773741) on composing structures from plain entities.

One of the entries in the palette shall represent "transparent" cells, allowing the underlying map content to show through.  This could be represented by the "Blank" terrain element, by default.

## Locating structures
The location of the structure shall be specified externally to the scenario map, as with the [`loc` field of robots](https://github.com/swarm-game/swarm/tree/main/data/scenarios#robots).
However, in contrast with how the `robots` section operates today, to facilitate multiple copies of structures, there should be a separate toplevel "placement" section of the scenario file, aside from the structure definitions section.

### Recursive structures

Potentially, we could allow a structure definition to have its own "placement" section.  This would facilitate recursive structures, allowing one to statically compose buildings and towns from smaller units.

## Handling cell conflicts

Overlaying structures on the map might be done *destructively* or *non-destructively*.  For "static" structures, the entities of the overlaid structure shall supersede whatever entity placement had been defined in the map.  Since a cell can only contain one entity, the previous entity is simply "gone".

Looking forward, we may want to use "dynamic" structures to represent "vehicles".  In that case, a parallel "plane" of the world would have to be defined to keep track of the dynamic structures, so that the underlying entities in the world are preserved as the "vehicle" passes over.

At the outset, perhaps we should focus only on static structures.

kostmo:
Closes #1138.

Supports all of:
* Nesting
* Transparency
* Flip
* Rotate

image

image

image

## Demo

    scripts/play.sh --scenario scenarios/Testing/1138-structures/nested-structure.yaml
    scripts/play.sh --scenario scenarios/Testing/1138-structures/flip-and-rotate.yaml
    scripts/play.sh --scenario data/scenarios/Testing/1138-structures/sibling-precedence.yaml

### DIFF
--- a/data/scenarios/Testing/1138-structures/basic-structure.yaml
+++ b/data/scenarios/Testing/1138-structures/basic-structure.yaml
@@ -1,0 +1,84 @@
+version: 1
+name: Basic structure
+description: |
+  Define a structure and place it in the map.
+robots:
+  - name: base
+    loc: [11, 0]
+    dir: [1, 0]
+known: [tree, flower, bit (0), bit (1)]
+world:
+  default: [blank]
+  palette:
+    '.': [grass]
+    'T': [stone, tree]
+    '*': [stone, flower]
+    '┌': [stone, upper left corner]
+    '┐': [stone, upper right corner]
+    '└': [stone, lower left corner]
+    '┘': [stone, lower right corner]
+    '─': [stone, horizontal wall]
+    '│': [stone, vertical wall]
+    '~': [stone, water]
+  upperleft: [-1, 1]
+  structures:
+    - name: bitpair
+      structure:
+        palette:
+          '0': [stone, bit (0)]
+          '1': [stone, bit (1)]
+        map: |
+          1
+          0
+    - name: bigbox
+      structure:
+        palette:
+          '.': [stone]
+          'T': [stone, tree]
+        structures:
+          - name: minibox
+            structure:
+              palette:
+                '.': [stone]
+                '┌': [stone, upper left corner]
+                '┐': [stone, upper right corner]
+                '└': [stone, lower left corner]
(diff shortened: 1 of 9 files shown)


data/scenarios/Testing/1138-structures/basic-structure.yaml
```diff
@@ -0,0 +1,84 @@
...
+            structure:
+              palette:
+                '.': [stone]
+                '┌': [stone, upper left corner]
+                '┐': [stone, upper right corner]
+                '└': [stone, lower left corner]
+                '┘': [stone, lower right corner]
+              placements:
+                - src: bitpair
+                  upperleft: [-1, 0]
```
kostmo, line 49: Maybe should rename this to `offset` and invert the coords

byorgey (approved):
Nice work! This gives us a nice bang for our buck in terms of being not much code for a big increase in expressive power when building worlds.

In a larger sense, I don't think this is general enough, and I am currently fleshing out a design for a more general + compositional DSL that will subsume both this and #1320 .  But I don't want to let the [perfect become the enemy of the good](https://en.wikipedia.org/wiki/Perfect_is_the_enemy_of_good).  I think we should merge this and proceed to use it, and if and when we implement a more general version we can easily convert any existing scenarios.

PR #1334, opened 2023-06-12
traffic vignette

kostmo:
A simulation of traffic on intersecting roads.  This is a "dynamic vignette"---an animated, looping scene that a player might come upon in their exploration of the `swarm` world.  Such scenes would presumably be "paused" until within range of the player.

Makes use of structure templates (#1332) and records (#1148).

    scripts/play.sh --scenario data/scenarios/Vignettes/roadway.yaml

image


byorgey (approved):
Nice! :car: :traffic_light: :vertical_traffic_light:

PR #1337, opened 2023-06-20
canonical device for 'scout' command

kostmo:
Supplements #1207

### DIFF
--- a/data/entities.yaml
+++ b/data/entities.yaml
@@ -765,6 +765,15 @@
   capabilities: [grab, swap, give, place, atomic]
   properties: [portable]
 
+- name: binoculars
+  display:
+    attr: device
+    char: 'B'
+  description:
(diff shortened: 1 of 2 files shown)


data/entities.yaml
```diff
@@ -765,6 +765,15 @@
   capabilities: [grab, swap, give, place, atomic]
   properties: [portable]
 
+- name: binoculars
+  display:
+    attr: device
+    char: 'B'
+  description:
+  - Allows one to `scout` for other robots
```
byorgey, line 773: Personally, I would tend to go for a somewhat more whimsical/detailed description, but I don't think it matters all that much.

byorgey (approved):

data/entities.yaml
kostmo: Sure, there's room to revisit when inspiration strikes.

PR #1339, opened 2023-06-24
Separate entities for each text operation

### LINKED ISSUE #1239: Separate string splitting capability from concatenation
See comment [here](https://github.com/swarm-game/swarm/pull/861#discussion_r1020964550).

Having only "one-way" string transformations in the form of concatenation avails a certain gameplay mechanic.

Propose a separate "wedge" device that provides string-splitting capability.

kostmo:
Closes #1239

The `string` device maintains all four of the split capabilities for backwards compatibility.

### DIFF
--- a/data/entities.yaml
+++ b/data/entities.yaml
@@ -336,6 +336,62 @@
   properties: [portable, growable]
   growth: [100, 800]
 
+- name: linotype
+  display:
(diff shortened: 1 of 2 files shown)


byorgey (approved):
I like it!

PR #1342, opened 2023-06-25
powerset scenario

kostmo:
image

    scripts/play.sh --scenario data/scenarios/Challenges/Ranching/powerset.yaml --autoplay

### DIFF
--- a/data/scenarios/Challenges/Ranching/00-ORDER.txt
+++ b/data/scenarios/Challenges/Ranching/00-ORDER.txt
@@ -1,2 +1,3 @@
 capture.yaml
+powerset.yaml
 gated-paddock.yaml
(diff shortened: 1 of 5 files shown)


byorgey (approved):
Nice!  My solution seems like it's a bit different than yours: https://github.com/byorgey/swarm-defs/blob/04b613d2d7be5a0e736c8bd40a3c9db12f89a415/powerset.sw

I just read each fruit combo as a binary number, sum them all, then subtract from the magic value of 8001 to find out the binary number of the missing combo.

I did not end up using any of the string-related devices (Elmer's glue or linotype).

kostmo:
> I just read each fruit combo as a binary number, sum them all, then subtract from the magic value of 8001 to find out the binary number of the missing combo.

Works well!  Would receive extra points if the derivation of the `8001` constant was in code :)

> I did not end up using any of the string-related devices (Elmer's glue or linotype).

Yes, they are not essential.  The fruits were previously numbered (`fruit0`, `fruit1`, ...), which would avail an alternative solution with these devices.  But I still used the devices for debugging (see #1349).

Issue #144, opened 2021-10-02
Sub-worlds


byorgey:
A crazy idea that just came up in IRC is to have certain cells function as entrances and exits to other sub-worlds.  e.g. there could be caves, buildings, ...  This seems like it would be fun and could be an easy way to add new stuff to the world without taking up a ton of space.

Concretely I don't think this would be all that hard to add (famous last words, right?).  A meta-world will now consist of a collection of worlds, each with an integer ID.  A location now consists of a pair of (world id, coordinates).  An entrance to another area just stores the location that it links to.

xsebek:
So to store this we would have sections world<sub>1</sub>,world<sub>2</sub>,... and entrances `Bimap Location Location`?

I'd really like to `drill` through mountains and `view` the robot. The border tiles of mountain terrain could be the entrances. :thinking: 

Possibly related:
- could improve the user experience of #143 
- could slightly complicate #50 (but we can design the save format with sub-worlds in mind)

byorgey:
> So to store this we would have sections world1,world2,... and entrances `Bimap Location Location`?

Well, I think it is up for debate whether the entrances should be a `Bimap`.  That would mean you can always immediately get back to where you came from and the worlds would form a proper undirected graph.  Letting it be a simple `Map` instead of a `Bimap` opens the possibility for bugs ("oops, I forgot to connect this thing back to that thing...") but also for more interesting topologies, e.g. several worlds connected by a loop, or one-way dungeons (once you go in, the only way out is to fight your way through to the exit...).

byorgey:
Just thinking about the mechanics of moving between sub-worlds... a simple idea might be to use the `down` and `up` directions (well, we already have `down` but we could add `up`).  So `turn down; move` would move into a subworld.  To be clear, I'm not proposing to make the world 3D, but rather, that every sub-world portal would be specified as being in either the "up" or "down" direction.

Amusingly, when testing this I realized that executing `turn down` causes a robot to look like the base.

byorgey:
An important question is how you would be able to see the sub-worlds.  Can you `view` a robot that is in a subworld?  Maybe you would have to develop mapping technology first, so you could send in robots to explore?  Or maybe you just have to put tank treads on the base and go explore yourself?

PR #1356, opened 2023-07-02
waypoints and portals

kostmo:
**Portals** are a prerequisite to subworlds (#144).

However, portals as a first-class feature have utility that is independent of subworlds.  The Hackman scenario (#1135), for example, currently has to provide redundant implementations of portal logic for the `base` robot and for system robots.  Furthermore, the locations specified in these bespoke portal implementations are hard-coded in the `.sw` file, so are brittle with changes to the map.

# Demos
    scripts/play.sh --scenario data/scenarios/Testing/1356-portals/automatic-waypoint-patrol.yaml
    
image

    scripts/play.sh --scenario scenarios/Testing/1356-portals/portals-flip-and-rotate.yaml --autoplay

image

# Portal and waypoint semantics

## Waypoints

**Waypoints** are introduced as an intermediate abstraction to better accommodate substructures.

A `waypoint` command, independent of portal mechanics, is also introduced to allow system robots to query named locations.

* **Waypoints** may be defined either in a standalone list (as a sibling field to the `map`), or within the `map` via the `palette`.
    * An **object notation** is now supported as an alternative to, and backwards compatible with, the **list notation** for palette values.
    * Using the palette to define waypoints is typically more convenient, as it allows specification of both a waypoint **and** an entity which can be used to depict the waypoint location.
* Waypoints can be defined at the root level or within "substructures".
* In contrast with "destructive" entity overlay from overlapping substructures, waypoints are aggregated from all substructures, even when overlapping.
* Multiple waypoints under different names can exist at the same location.  I.e. waypoints can have *plural multiplicity*.
    * Uniqueness of locations for a given waypoint name is not enforced unless the `unique` flag is set to `true` for one of the instances of that waypoint name.

## Portals
* **Portals** are defined as `(entrance, exit)` pairs of waypoints.
    * The `entrance` is a bare waypoint name, but the `exit` is a waypoint name paired with an optional **subworld name**.  Omitting the subworld name implies a portal that exits to the same subworld as its entrance.
* Portals can have multiple entrances but only a single exit.
* A waypoint with multiplicity > 1 will result in multiple portals (all to the same exit) when specified as an `entrance`.

### Obstructions
* Portal **entrances** can be blocked, which will prevent movement into the portal.  Portal **exits** are not checked for obstructions when traversing the portal.

# Other musings
* Should some robots not be able to go through portals? e.g. based on capability (c.f. [The Terminator](https://en.wikipedia.org/wiki/The_Terminator) which needs an organic skin  covering)
* We could keep track of how many times a robot has passed through a portal in its internal state. Is there a use for such a statistic?
    * Come to think of it, there are other figures that may also be interesting to track, such as total distance moved.

### DIFF
--- a/data/scenarios/Testing/1356-portals/automatic-waypoint-patrol.yaml
+++ b/data/scenarios/Testing/1356-portals/automatic-waypoint-patrol.yaml
@@ -1,0 +1,73 @@
+version: 1
+name: Querying waypoints
+description: |
+  Demonstrate patrolling between waypoints
+creative: true
+robots:
+  - name: base
+    loc: [0, 0]
+    dir: [1, 0]
+  - name: patroller
+    loc: [5, -4]
+    dir: [1, 0]
+    display:
+      invisible: false
+      attr: robot    
+    program: |
+      run "scenarios/Testing/1356-portals/_automatic-waypoint-patrol/program.sw"
+known: [flower, boulder]
+world:
+  upperleft: [-1, 1]
+  default: [blank]
+  palette:
+    '.': [grass]
+    '*': [stone, flower]
+    '┌': [stone, upper left corner]
+    '┐': [stone, upper right corner]
+    '└': [stone, lower left corner]
+    '┘': [stone, lower right corner]
+    '─': [stone, horizontal wall]
+    '│': [stone, vertical wall]
+  structures:
+    - name: bigbox
+      structure:
+        palette:
+          '.': [dirt]
+          '@': [dirt, boulder]
+          'w':
+            cell: [dirt]
+            waypoint:
+              name: wp
+        map: |
+          @@@
+          @w.
+          @.@
+  placements:
+    - src: bigbox
+      offset: [2, -2]
+      orient:
+        up: "DNorth"
+    - src: bigbox
+      offset: [8, -2]
+      orient:
+        up: "DEast"
+    - src: bigbox
+      offset: [8, -6]
+      orient:
+        up: "DSouth"
+    - src: bigbox
+      offset: [2, -6]
+      orient:
+        up: "DWest"
+  map: |
+    ┌───────────┐
+    │*..*..*..*.│
+    │.*..*..*..*│
+    │..*..*..*..│
+    │*..*..*..*.│
+    │.*..*..*..*│
+    │..*..*..*..│
+    │*..*..*..*.│
+    │.*..*..*..*│
+    │..*..*..*..│
+    └───────────┘
--- a/src/Swarm/Game/Scenario/Topography/Navigation/Portal.hs
+++ b/src/Swarm/Game/Scenario/Topography/Navigation/Portal.hs
@@ -68,13 +68,13 @@ failUponDuplication message binnedMap =
 -- * portals can have multiple entrances but only a single exit
 -- * no two portals share the same entrance location
 -- * global waypoint uniqueness when the "unique" flag is specified
-validateLandmarks ::
+validateNavigation ::
   (MonadFail m, Traversable t) =>
   V2 Int32 ->
   [Originated Waypoint] ->
   t Portal ->
   m Navigation
-validateLandmarks upperLeft unmergedWaypoints portalDefs = do
+validateNavigation upperLeft unmergedWaypoints portalDefs = do
   failUponDuplication "is required to be unique, but is duplicated in:" waypointsWithUniqueFlag
 
   -- TODO(#144) Currently ignores subworld references
--- a/src/Swarm/Game/Scenario/Topography/Navigation/Portal.hs
+++ b/src/Swarm/Game/Scenario/Topography/Navigation/Portal.hs
@@ -1,0 +1,120 @@
+{-# LANGUAGE OverloadedStrings #-}
+
+-- |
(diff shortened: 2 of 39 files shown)


byorgey:
Looks fun! I'm travelling but hope to be able to look at this within the next few days sometime.

src/Swarm/Game/Step.hs
```diff
@@ -2677,12 +2688,17 @@ updateRobotLocation ::
...
       robotsByLocation . at oldLoc %= deleteOne rid
-      robotsByLocation . at newLoc . non Empty %= IS.insert rid
-      modify (unsafeSetRobotLocation newLoc)
+      robotsByLocation . at newlocWithPortal . non Empty %= IS.insert rid
+      modify (unsafeSetRobotLocation newlocWithPortal)
       flagRedraw
  where
+  applyPortal loc = do
+    lms <- use worldNavigation
+    return $ M.findWithDefault loc loc $ portals lms
```
byorgey, line 2700: So am I understanding correctly that if you simply walk on top of a portal, you are instantly teleported?  What's the argument for doing it that way as opposed to, say, having a command like `travel` (or whatever) that activates a portal on which you are standing?  I'm honestly not sure which I would prefer, I can think of pros and cons of both.

data/scenarios/Testing/1356-portals/automatic-waypoint-patrol.yaml
```diff
@@ -0,0 +1,73 @@
...
+          '@': [dirt, boulder]
+          'w':
+            cell: [dirt]
+            waypoint:
+              name: wp
+        map: |
+          @@@
+          @w.
+          @.@
+  placements:
```
byorgey, line 45: How do you control the order of all the waypoints in a specific cycle?  Is it determined by the order of the `placements` list?  If so, that seems fragile, especially if the order of placements may also affect which structures get drawn on top of others.

src/Swarm/Game/Scenario/Topography/Navigation/Portal.hs
```diff
@@ -0,0 +1,120 @@
...
+ where
+  duplicated = M.filter ((> 1) . NE.length) binnedMap
+
+-- | Enforces the following constraints:
+-- * portals can have multiple entrances but only a single exit
+-- * no two portals share the same entrance location
+-- * global waypoint uniqueness when the "unique" flag is specified
+validateNavigation ::
+  (MonadFail m, Traversable t) =>
+  V2 Int32 ->
```
byorgey, line 73: Why is this a `V2 Int32` as opposed to a `Location`?

byorgey (approved):
Looks cool!  Just a few questions/comments.

test/integration/Main.hs
```diff
@@ -296,6 +296,7 @@ testScenarioSolution _ci _em =
         , testSolution Default "Testing/1234-push-command"
         , testSolution Default "Testing/1256-halt-command"
         , testSolution Default "Testing/1295-density-command"
+        , testSolution Default "Testing/1356-portals/portals-flip-and-rotate.yaml"
```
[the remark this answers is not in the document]
kostmo: This is the only one I defined a goal and solution for; the other two are just "manual" tests.

src/Swarm/Game/Step.hs
kostmo: Ultimately I think we should support both :)

Teleporter-style, which is currently implemented, should be instantaneous.  But "ladders" and "stairs" would also be implemented as portals, and we could give the player the option to proceed down or up explicitly, after walking on to the cell.

To specify this, there will be an additional YAML attribute "direction" which can be, `up`, `down`, or omitted.

data/scenarios/Testing/1356-portals/automatic-waypoint-patrol.yaml
kostmo: Good point.  Perhaps an explicit sort should be performed upon construction.  Opened #1366.

src/Swarm/Game/Scenario/Topography/Navigation/Portal.hs
```diff
@@ -0,0 +1,120 @@
...
+            [ "No waypoint named"
+            , quote rawName
+            ]
+    Just xs -> return xs
+
+  extractLoc (Originated _ (Waypoint _ loc)) = loc
+  correctedWaypoints =
+    binTuples $
+      map
+        (\x -> (wpName $ wpConfig $ value x, fmap (offsetWaypoint upperLeft) x))
```
kostmo, line 116: Using the `upperLeft` here as a vector to apply an offset

src/Swarm/Game/Scenario/Topography/Navigation/Portal.hs
kostmo: I see how that is confusing; I [use it as a vector to offset a location](https://github.com/swarm-game/swarm/pull/1356/files#r1260462962), but really I should pass it into this function as a `Location` and defer the conversion.  Will follow up with fix.

kostmo:
Oops, `mergebot` got stuck so I clicked the merge button.  Should have manually rebased first to pre-empt the breakage.

PR #1375, opened 2023-07-17
Change binding on Hide REPL to M-,

### LINKED ISSUE #1325: Remap dialog collapse key to ALT-comma
Remap dialog collapse key to CTRL-comma.

Currently there is a conflict with CTRL-s used for "saving" the map in the world editor (#873).

_Originally posted by @kostmo in https://github.com/swarm-game/swarm/discussions/1312#discussioncomment-6135624_

noahyor:
Fixes #1325.

### DIFF
--- a/src/Swarm/TUI/Controller.hs
+++ b/src/Swarm/TUI/Controller.hs
@@ -387,7 +387,7 @@ handleMainEvent ev = do
       -- Paint with the World Editor
       EC.handleCtrlLeftClick mouseLoc
     -- toggle collapse/expand REPL
-    ControlChar 's' -> do
+    ControlChar ',' -> do
       invalidateCacheEntry WorldCache
       uiState . uiShowREPL %= not
     MouseDown n _ _ mouseLoc ->
--- a/src/Swarm/TUI/Controller.hs
+++ b/src/Swarm/TUI/Controller.hs
@@ -387,7 +387,7 @@ handleMainEvent ev = do
       -- Paint with the World Editor
       EC.handleCtrlLeftClick mouseLoc
     -- toggle collapse/expand REPL
-    ControlChar ',' -> do
+    MetaChar ',' -> do
       invalidateCacheEntry WorldCache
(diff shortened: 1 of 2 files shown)


src/Swarm/TUI/Controller.hs
```diff
@@ -387,7 +387,7 @@ handleMainEvent ev = do
       -- Paint with the World Editor
       EC.handleCtrlLeftClick mouseLoc
     -- toggle collapse/expand REPL
-    ControlChar 's' -> do
+    ControlChar ',' -> do
```
kostmo, line 390: I just realized that, unfortunately, `CTRL + comma` is [not handled properly by the terminal](https://stackoverflow.com/a/7544277/105137).

However `ALT + comma` does work.  So we could change this to 
```suggestion
    MetaChar ',' -> do
```

kostmo (changes requested):

src/Swarm/TUI/View.hs
```diff
@@ -929,7 +929,7 @@ drawKeyMenu s =
       , may isPaused (NoHighlight, "^o", "step")
       , may (isPaused && hasDebug) (if s ^. uiState . uiShowDebug then Alert else NoHighlight, "M-d", "debug")
       , Just (NoHighlight, "^zx", "speed")
-      , Just (NoHighlight, "^s", if s ^. uiState . uiShowREPL then "hide REPL" else "show REPL")
+      , Just (NoHighlight, "^,", if s ^. uiState . uiShowREPL then "hide REPL" else "show REPL")
```
kostmo, line 932: The `Alt` key is the same as `Meta`, so this should be:
```suggestion
      , Just (NoHighlight, "M-,", if s ^. uiState . uiShowREPL then "hide REPL" else "show REPL")
```

kostmo:
Thank you for pushing this one along!

kostmo (approved):

byorgey (approved):
I tested the new `M-,` , seems to work.

Issue #29, opened 2021-09-21
Generate world by tiles


byorgey:
Right now a `World` is specified by a `WorldFun`, defined as
```
type WorldFun t e = Coords -> (t, Maybe e)
```
This is just a function that specifies a terrain value and possible entity for every cell.  However, this has several limitations.  For one thing, it is difficult to specify things that should be coherent across multiple cells (*i.e.* structures).  For another, if we want the world to contain creatures that move around (implemented as system robots), there is no good way to specify when they should spawn.

The implementation of `World`s already uses a notion of tiles for efficiency, and we can build on this to solve the limitations above.  Instead of specifying a world cell by cell, we specify it tile by tile.  That is, a `WorldFun` will now be something like
```
type WorldFun t e = TileCoords -> (TerrainTile t, EntityTile e, [Robot])
```
which says that the world function takes as input the coordinates of a *tile* and generates an entire tile's worth of terrain and entities all at once, in addition to some robots which should spawn when this tile is loaded.  When calling the function, it will build tiles first by filling in initial values based on a `Coords -> (t, Maybe e)` function as before, but then has the opportunity to further modify them, e.g. by deciding where to draw in certain structures.

byorgey:
As a first step, we should just refactor the existing game to work this way, but generate exactly the same world (with no robots or structures) that we currently do.

byorgey:
Maybe we should keep the old `type WorldFun t e = Coords -> (t, Maybe e)` (since we will certainly still want that type) and define `type TiledWorldFun t e = TileCoords -> (TerrainTile t, EntityTile e, [Robot])`.  Then as a starting point we can define
```
simpleTiledWorldFun :: WorldFun t e -> TiledWorldFun t e
```
which just fills in everything according to the `WorldFun`.  Maybe we can also make `TiledWorldFun` a `Monoid` such that the lists of robots compose and entities and tiles on the right overwrite ones on the left.  We would have to make a special distinguished "empty terrain" value to serve as the identity.  `EntityTile`s will already contain `Maybe e`s so `Nothing` can serve as the identity.  Given this monoid, we can start with a base generated by a `WorldFun` and then layer on things like structures etc.

byorgey:
Here is another idea which is hopefully both simpler and more flexible: let's just keep the `WorldFun` type as it is, but to the output we can add `[Robot]`, describing robots which should spawn when the given cell is first loaded.

[...]

PR #1376, opened 2023-07-17
World description DSL

byorgey:
DSL for programming worlds, towards #1320 and #29 (and, indirectly, toward #50, since the world DSL should make a nice target for world saves) .  Eventually this should be able to recreate all the world description/building features we have, though there is still a long way to go.  But currently we can at least recreate the "classic" procedurally-generated world.  I think this is a solid foundation we can merge as a first step, and then work on adding more features in subsequent PRs.  Below are some notes that should help in reviewing.  Note that the large number of files changed is due in large part to the elimination of the `default` field in scenario descriptions; see the "changed files" section below for an overview of the important/interesting changes.

Issues split off from this one: #1394 #1395 #1396 #1397 

Major changes
============

- New `data/worlds` subdirectory
    - All `.world` files are parsed at load time and saved in a `WorldMap` which gets threaded through, similar to `EntityMap` (perhaps we should think about passing around a single record instead)
- Standard "classic" world
    - Used to be `testWorld2`, defined in Haskell code; now it is defined via the DSL in `worlds/classic.world`.  This should make it much easier to experiment with variations.
    - We can now automatically extract entities mentioned in a world DSL term with `extractEntities`.  There used to be an explicit list in `testWorld2Entities`, used to check pedagogy, generate documentation, etc., but it turns out it had (predictably) gotten out of date!  This can't happen anymore.
    - It is now referenced in several tutorials (backstory, farming, world101, speedruns, etc.)
- The `default` field of world descriptions is no more: one can use `dsl` to just specify a constant
    - Note in `Swarm.Game.State`, `dslWF` and `arrayWF` are combined using the `Monoid` instance to create `wf`.
- `Erasable`
    - It used to be the case that if some kind of default terrain + entity was specified (e.g. stone + water), any `map` would completely override the default.  However, we want to move towards combining everything with a `Monoid` instance.  But by default this means the default entity would show through anywhere the `map` did not specify an entity.  So we need a way to explicitly "erase" an entity from a lower layer.
    - If `e` is a `Semigroup`, then `Maybe e` is a `Monoid` where `Nothing` acts as an identity element.  Likewise, `Erasable e` is a `Monoid` but adds two new elements: `ENothing` to be an identity, and `EErase` to be an *annihilator*.  i.e. combining with `EErase` is like multiplying by zero.
    - We can now specify `erase` as an entity to override entity underneath.
    - There are several Haskell files with only changes related to `Erasable`, relating to e.g. the world editor, `PCells`, etc.; I'm not 100% sure I've always done the right thing here.

DSL overview
===========

- Integer, float, and Boolean literals.  Note that `3` is *always* an `int`, and `3.0` is a `float`.  It makes things much easier to not have to deal with `3` possibly being either `int` or `float`, though it does make things slightly more annoying for programmers.
- Standard boolean, arithmetic, and comparison operators
- `if ... then ... else ...`
- `<>` operator for combining via `Semigroup` instance
- Cell literals are enclosed in curly braces.  Unlike the previous awkward world description syntax with one, two, or three-element lists denoting terrain, terrain + entity, or terrain + entity + robot, there can now be any number of elements in any order.
    - `{foo}` will be resolved as either terrain, an entity, or a robot, whichever is successful.  So if the names are unambiguous one can just write `{tree}` or `{stone}`.
    - It is possible to explicitly indicate the type of cell value with syntax like `{entity: tree}` or `{terrain: stone}`.
    - Multiple items separated by commas is syntax sugar for combining with `<>`.  e.g. `{tree, entity: boulder, stone} = {tree} <> {entity: boulder} <> {stone}`.
- Ability to refer to the `seed`
- Refer to the current `x` or `y` coordinates or the `hash` of the current coordinates
- `let`-expressions for multiple variables: `let x1 = e1, x2 = e2, ... in ...`
- `overlay [e1, e2, ...]` layers `e1` on the bottom, `e2` on top of that, etc., using the `Semigroup` instance for world functions
- `"foo"` imports the DSL term in `worlds/foo.world`
- `perlin` function to generate perlin noise
- `mask` function to mask with a condition

Changed files
===========

- `Swarm.Util`: moved the `acquire` function here and gave it a more descriptive name.
- `Swarm.Doc.Gen`: can now extract mentioned entities directly.
- `Swarm.Game.Failure`: added new failure modes
- `Swarm.Game.Scenario.Topography.WorldDescription`: get rid of `defaultTerrain` field, add `worldProg` for DSL.
- `Swarm.Game.State`: see comment.
- `Swarm.Game.World`: a bit of reorganization.  Added a bunch of modules under this.
    - `Swarm.Game.World.Coords`: moved some code here from `Swarm.Game.World`.
    - `Swarm.Game.World.Gen`: moved some things here from `Swarm.Game.WorldGen` (also deleted a bunch of irrelevant code), and also added the `extractEntities` function to get all entities mentioned by a DSL term.
    - `Swarm.Game.World.Syntax`: raw, untyped syntax for world DSL terms.
    - `Swarm.Game.World.Parse`: parser for world DSL terms. Fairly standard.
    - `Swarm.Game.World.Typecheck`: takes raw, untyped terms produced by the parser and both typechecks and elaborates them into a simpler core language.  An interesting feature is that the core language is *type-indexed*, so that the Haskell type system is actually ensuring that our typechecker is correct; every typechecked world DSL term value has a type which is indexed by a Haskell type corresponding to the type of the underlying DSL term.  For example, `{entity: tree}` would have a type like `TTerm [] (World CellVall)` etc.  Once terms make it through the typechecker, there cannot possibly be any bugs in the rest of the pipeline which would result in a crash, because the Haskell type system.  (There could of course be *semantic* bugs.)  Understanding exactly how the typechecker works is not too important.  Of interest may be the `resolveCell` function, which determines how we decide what `Cell` is represented by a cell expression in curly braces.
    - `Swarm.Game.World.Abstract`: compile elaborated, typechecked world DSL terms down into an extremely simple core language with only constants and function application.  This gives us very fast evaluation of world DSL terms.  Understanding this module is not really necessary but there is a link to a blog post for those who are interested in how it works.
    - `Swarm.Game.World.Compile`: a further processing/compilation step after `Swarm.Game.World.Abstract`.  Currently we don't actually use this, since it doesn't seem like it makes a big efficiency difference.
    - `Swarm.Game.World.Interpret`: interpreter for abstracted world DSL terms.
    - `Swarm.Game.World.Eval`: just puts together the pieces of the pipeline to evaluate a typechecked world DSL term.
    - `Swarm.Game.World.Load`: just loading world DSL terms from disk.


byorgey:
@kostmo This is far from being complete, but of course you're welcome to take a look at your leisure.  I'll probably try to break this whole world DSL thing up into multiple PRs.

src/Swarm/Game/Scenario/Topography/WorldDescription.hs
```diff
@@ -72,14 +77,26 @@ instance FromJSONE (InheritedStructureDefs, (EntityMap, RobotMap)) WorldDescript
         unmergedWaypoints
         portalDefs
 
-    WorldDescription terr
+    mwexp <- liftE (v .:? "dsl")
+    dslTerm <- case mwexp of
+      Nothing -> return Nothing
```
kostmo, line 82: Does `forM` work here?

src/Swarm/Game/State.hs
```diff
@@ -1233,23 +1241,24 @@ scenarioToGameState scenario (LaunchParams (Identity userSeed) (Identity toRun))
...
-  worldArray :: Array (Int32, Int32) (TerrainType, Maybe Entity)
+  worldArray :: Array (Int32, Int32) (TerrainType, Erasable Entity)
   worldArray = listArray ((ulr, ulc), (ulr + rs - 1, ulc + cs - 1)) (concat worldGrid)
 
-  wf = case defaultTerrain of
-    Nothing ->
-      (if offsetOrigin then findGoodOrigin else id) . testWorld2FromArray em worldArray
-    Just (Cell t e _) -> const (worldFunFromArray worldArray (t, e))
+  dslWF, arrayWF :: Seed -> WorldFun TerrainType Entity
+  dslWF = maybe mempty (((if offsetOrigin then findGoodOrigin else id) .) . runTTerm) worldProg
```
kostmo, line 1258: [`applyWhen`](https://hackage.haskell.org/package/base-4.18.0.0/docs/Data-Function.html#v:applyWhen)?

src/Swarm/Game/World/Parse.hs
```diff
@@ -0,0 +1,244 @@
...
+-- Utility
+
+sepByNE :: (MonadPlus m) => m a -> m sep -> m (NonEmpty a)
+sepByNE p sep = NE.fromList <$> p `sepBy1` sep
+
+------------------------------------------------------------
+-- Lexing
+
+reservedWords :: [Text]
+reservedWords = ["not", "true", "false", "seed", "x", "y", "hash", "let", "in", "overlay", "hcat", "vcat", "if", "then", "else", "perlin", "mask", "empty", "abs"]
```
kostmo, line 41: One word per line for diff friendliness

src/Swarm/Game/Scenario/Topography/WorldDescription.hs
byorgey: Yes, good suggestion, thanks!

src/Swarm/Game/State.hs
byorgey: Cool, I didn't know about `applyWhen`!  Though it looks like it has only been in `base` since version 4.18, so if we want Swarm to build with older versions of GHC/`base` we can't use it yet.

...oh, wait!  We can depend on the version from `base-compat`, that's exactly what `base-compat` is for.

byorgey: Well, actually, the LTS version we are currently pegged to has `base-compat-0.12.2`, which does not contain `applyWhen`.  I guess we'll have to wait a little while yet.

byorgey: For now I stuck a version of `applyWhen` in `Swarm.Util` with a note to switch to the version from `base-compat` once it becomes available.

src/Swarm/Game/World/Parse.hs
```diff
@@ -0,0 +1,264 @@
...
+  , "abs"
+  ]
+
+-- | Skip spaces and comments.
+sc :: Parser ()
+sc =
+  L.space
+    space1
+    (L.skipLineComment "//")
+    (L.skipBlockComment "/*" "*/")
```
xsebek, line 69: I don't see any comments in the classic world or other world description while scrolling through.

Maybe we could have a simpler training world with comments that note the basic features? 🙂

xsebek (commented):
@byorgey I will try to take a look at the code later. I like how the DSL looks nice and clean from your description though. 👍

Could you maybe copy the DLS description from the PR notes to some README?

data/schema/world.json
```diff
@@ -5,10 +5,10 @@
     "description": "Description of the world in the Swarm game",
     "type": "object",
     "properties": {
-        "default": {
+        "dsl": {
             "default": null,
-            "type": "array",
-            "description": "A tuple representing the contents of a default cell (see Cells, except that the default cell may not contain a robot). If this key is present, it means that the whole world besides the part specified with the map will be filled with this default cell. If omitted, the world besides the part specified with the map will be procedurally generated."
+            "type": "string",
```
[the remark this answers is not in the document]
byorgey: Yeah, I don't think we have been keeping it up to date, I know @kostmo has added a bunch of features that I think we missed adding to the schema.  If there is an automated way to make sure all existing scenario descriptions conform to the schema, we should add it to our CI!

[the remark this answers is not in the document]
byorgey: There is supposed to be a way to get Emacs to use the schemas when editing `.yaml` files, but I have never been able to get it to work.

src/Swarm/Game/World/Parse.hs
byorgey: Good idea, I'll add both.

src/Swarm/Util/Erasable.hs
```diff
@@ -0,0 +1,46 @@
...
+-- | Extend a semigroup to a monoid by adding an identity ('ENothing') /and/ an
+--   annihilator ('EErase').  That is,
+--
+--   * `ENothing <> e = e <> ENothing = e`
+--   * `EErase <> e = e <> EErase = EErase`
+--
+--   This allows us to "erase" previous values by combining with
+--   'EErase'.  The 'erasableToMaybe' function turns an @Erasable@
+--   into a @Maybe@ by collapsing 'ENothing' and 'EErase' both back
+--   into 'Nothing'.
```
kostmo, line 17:
```suggestion
--   * @ENothing <> e = e <> ENothing = e@
--   * @EErase <> e = e <> EErase = EErase@
--
--   This allows us to "erase" previous values by combining with
--   'EErase'.  The 'erasableToMaybe' function turns an 'Erasable'
--   into a 'Maybe' by collapsing 'ENothing' and 'EErase' both back
--   into 'Nothing'.
```

kostmo (approved):
Fantastic!  This opens up a lot of possibilities for scenario design.

Agree with @xsebek that the PR description is good and as much of it as possible should be copied into Haddocks.

src/Swarm/Game/World/Parse.hs
byorgey: Added some comments to the classic world and created a `README` in `data/worlds` with a full description of the DSL.

data/worlds/README.md
```diff
@@ -0,0 +1,189 @@
...
+Identifiers consist of any non-empty sequence of letters, digits,
+underscores, or single quotes, but must begin with a letter or
+underscore.
+
+The extended BNF syntax `S*,` denotes the syntax `S` repeated zero
+or more times, separated by `,`.  Likewise, `S+,` denotes one or more
+repetitions of `S` separated by `,`.
+
+```
+<int> ::= integer literal
```
xsebek, line 64: I like the formal specification here, could we check that it applies to all World DSL?

I expect we use megaparsec.

xsebek (approved):
Impressive, I look forward to playing with this. 👍 🚀

data/worlds/README.md
xsebek: Or other way - create this description from code?

data/scenarios/Testing/00-ORDER.txt
```diff
@@ -37,6 +37,7 @@
 1256-halt-command.yaml
 1295-density-command.yaml
 1138-structures
+1320-world-DSL
```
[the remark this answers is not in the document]
byorgey: There's not really any logic.  I think they have generally ended up in the order they were added.  I don't think the order matters all that much though.

data/worlds/README.md
byorgey: Not easily at the moment, since it's not fully formal (notice it contains English phrases like "any single character besides...".  In theory we could have a fully formal specification from which we extract both the parser code and some documentation, but it sounds like a lot of work.  I'll just add a note in the code as a reminder to update the README to match when modifying the parser.

data/scenarios/Testing/00-ORDER.txt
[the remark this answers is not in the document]
xsebek: Sure, looking at integration tests there are groups/directories of similar tests and otherwise it's mostly natural sort.

xsebek:
🗺️ 🥳

PR #1378, opened 2023-07-18
Use robotMap directly instead of addRobot

### LINKED ISSUE #1372: Use a lens to modify robot state for "Say" command
See [this code](https://github.com/swarm-game/swarm/blob/485e6ace1f9d4860416667568f24929d357a0680/src/Swarm/Game/Step.hs#L1585-L1592) (originally added [here](https://github.com/swarm-game/swarm/pull/565/files#r1264589864)).

It makes use of the `addRobot` function which seems like a non-idiomatic way to modify robot state.

kostmo:
Closes #1372

### DIFF
--- a/src/Swarm/Game/Step.hs
+++ b/src/Swarm/Game/Step.hs
@@ -1584,12 +1584,16 @@ execConst c vs s k = do
                 | otherwise -> es |> e
         let addToRobotLog :: Has (State GameState) sgn m => Robot -> m ()
             addToRobotLog r = do
-              r' <- execState r $ do
+              maybeRidLoc <- evalState r $ do
                 hasLog <- hasCapability CLog
                 hasListen <- hasCapability CListen
                 loc' <- use robotLocation
-                when (hasLog && hasListen) (robotLog %= addLatestClosest loc')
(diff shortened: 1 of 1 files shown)


src/Swarm/Game/Step.hs
```diff
@@ -1584,12 +1584,16 @@ execConst c vs s k = do
...
+              maybeRidLoc <- evalState r $ do
                 hasLog <- hasCapability CLog
                 hasListen <- hasCapability CListen
                 loc' <- use robotLocation
-                when (hasLog && hasListen) (robotLog %= addLatestClosest loc')
-              addRobot r'
+                rid <- use robotID
+                return $ do
+                  guard $ hasLog && hasListen
+                  Just (rid, loc')
```
byorgey, line 1594:
```suggestion
                return $ (rid, loc') <$ (guard $ hasLog && hasListen)
```
I find this simpler/easier to read; YMMV.

byorgey: Hmm, actually, never mind, that might lead to an ambiguous type error...?

byorgey (approved):

PR #1383, opened 2023-07-19
Load persistent state from disk only once and reuse for all integration tests

### LINKED ISSUE #1279: Only initialize `AppState` once when running scenario tests
Right now, every single scenario test loads all the entity/recipe data etc. from disk, from scratch.  This slows down the tests a lot.

We should load everything once at the beginning and share among all the scenario tests.

This is a follow-up from #1277.

byorgey:
This seems to make a big difference --- the integration test suite now takes only about 35% as long as it used to. Fixes #1279.


kostmo (approved):
Benchmark on my machine:
| Before | After |
| --- | --- |
| 0m53.054s | 0m26.622s |

PR #1385, opened 2023-07-20
Show framerate slowdown

### LINKED ISSUE #1348: Draw framerate slowdown indicator
One suggestion in #42:
> When it hits the cap, the game will slow down: it will essentially be running at a slower ticks/second rate than requested. Perhaps we could signal this by e.g. drawing the ticks per second rate in red, or something like that.

Possibly, we could display a percentage of the frame interval that the computation took.

byorgey:
Highlight ticks per frame in red when it reaches the cap.  Closes #1348 .


byorgey:
Oops, created the branch from the wrong commit.

PR #1386, opened 2023-07-20
Highlight ticks per frame in red when it reaches the cap

### LINKED ISSUE #1348: Draw framerate slowdown indicator

byorgey:
Closes #1348.


kostmo:
Can you share a "test plan" to reproduce hitting the cap and demo the new behavior?

noahyor:
Run swarm with any scenario and increasing the speed is a good way to reproduce this.

kostmo (approved):

PR #1390, opened 2023-07-21
Automatic re-orientation via portal

### LINKED ISSUE #1379: Automatically re-orient robot when passing through portals
This should be an optional attribute on the portal in the scenario specification.

kostmo:
Closes #1379

Related: #950

Also moves "directions" types/logic into its own module.

## Demo

    scripts/play.sh --scenario data/scenarios/Testing/1379-single-world-portal-reorientation.yaml --autoplay

### DIFF
--- a/data/scenarios/Testing/1379-single-world-portal-reorientation.yaml
+++ b/data/scenarios/Testing/1379-single-world-portal-reorientation.yaml
@@ -1,0 +1,101 @@
+version: 1
+name: Portal reorientation within a single subworld
+description: |
+  Turning without turning
+attrs:
+  - name: portal_in
+    fg: "#ff9a00"
+    bg: "#ff5d00"
+objectives:
+  - goal:
+    - |
+      `place` the "flower" on the white cell.
+    condition: |
+      j <- robotnamed "judge";
+      as j {ishere "flower"}
+solution: |
+    def doN = \n. \f. if (n > 0) {f; doN (n - 1) f} {}; end;
+    doN 23 move;
+    f <- grab;
+    doN 23 move;
+    place f;
+entities:
+  - name: telepad entrance
+    display:
+      attr: portal_in
+      char: "o"
+    description:
+    - Portal entrance
+    properties: [known]
+robots:
+  - name: base
+    dir: [0, 1]
+    devices:
+      - branch predictor
+      - calculator
+      - comparator
+      - dictionary
+      - grabber
+      - lambda
+      - logger
+      - strange loop
+      - treads
+  - name: judge
+    dir: [1, 0]
+    system: true
+    display:
+      char: 'J'
+      invisible: true
+known: [flower]
+world:
+  name: root
+  default: [blank]
+  palette:
+    '.': [grass]
+    'f': [grass, flower]
+    'g': [ice, null, judge]
+    'B': [grass, null, base]
+    '0':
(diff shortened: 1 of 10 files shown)


src/Swarm/Language/Directions.hs
```diff
@@ -0,0 +1,116 @@
+{-# LANGUAGE DerivingStrategies #-}
+{-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE UndecidableInstances #-}
+
+-- |
+-- SPDX-License-Identifier: BSD-3-Clause
+--
+-- Abstract syntax for terms of the Swarm programming language.
```
byorgey, line 9: This comment needs to be updated.

src/Swarm/Language/Directions.hs
```diff
@@ -0,0 +1,116 @@
...
+instance ToJSON PlanarRelativeDir where
+  toJSON = genericToJSON directionJsonOptions
+
+-- | The type of directions. Used /e.g./ to indicate which way a robot
+--   will turn.
+data Direction = DAbsolute AbsoluteDir | DRelative RelativeDir
+  deriving (Eq, Ord, Show, Read, Generic, Data, Hashable, ToJSON, FromJSON)
+
+-- | Direction name is generated from Direction data constructor
+-- e.g. DLeft becomes "left"
```
byorgey, line 101: This comment is not really accurate any more.

src/Swarm/Language/Directions.hs
```diff
@@ -0,0 +1,116 @@
+{-# LANGUAGE DerivingStrategies #-}
+{-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE UndecidableInstances #-}
+
+-- |
+-- SPDX-License-Identifier: BSD-3-Clause
+--
+-- Abstract syntax for terms of the Swarm programming language.
+module Swarm.Language.Directions (
```
byorgey, line 10: It's not a big deal, but I think we have generally settled on the convention of module names being singular.

data/scenarios/Testing/1379-single-world-portal-reorientation.yaml
```diff
@@ -0,0 +1,101 @@
+version: 1
```
byorgey, line 1: This file needs to be added to both `00-ORDER.txt` and also to the integration tests.

src/Swarm/Game/Scenario/Topography/Navigation/Portal.hs
```diff
@@ -72,6 +74,7 @@ data Portal = Portal
   { entrance :: WaypointName
   , exitInfo :: PortalExit
   , consistent :: Bool
+  , reorient :: PlanarRelativeDir
```
byorgey, line 77: Do you think there is any need/desire for portals that can do *absolute* (instead of relative) reorientation?  *e.g.* imagine descending into a cave where you start at the end of a long, narrow passage; it might be nice to always start facing down the passage no matter which direction you were facing when you entered the cave.  Just to be clear, I don't think this is something that needs to be addressed in this PR, only some idle thoughts.

byorgey (approved):
Looks great!  Just a few simple things to address before merging, I think.

PR #1392, opened 2023-07-25
Refactor to more consistently use "capability style" in loading + initializing code

### LINKED ISSUE #1122: Consistently use "capability style" for all error handling
> In general it would be nice to use e.g. a `Has (Throws SystemFailure) m` constraint rather than using an explicit `ExceptT`. I know we already have `ExceptT` in a few places, but a lot of our code is already in the "capability style" where we have a generic monad m and just say what constraints it must satisfy, and this kind of style is more flexible since you can use it at multiple concrete types, as opposed to using explicit monad transformers which limits you to use exactly that type.

From [this comment](https://github.com/swarm-game/swarm/pull/1116#discussion_r1111139757).

byorgey:
Closes #1122 .  General principles:

- Use `SystemFailure` as error rather than `Text` as much as possible, and use `prettyFailure` only at the very top level.
- Replace `ExceptT` with `Has (Throw SystemFailure)` constraint.
- Use `Accum (Seq SystemFailure)` constraints to accumulate warnings that should not abort computation, rather than returning a pair of a list of warnings + result.
- Use `Has (Lift IO)` constraint instead of `MonadIO`, which means using `sendIO` instead of `liftIO`.
- In general, use `runThrow` to dispatch a `Throw` constraint (results in returning an `Either`, just like `runExceptT`), and `runM` to dispatch a final `Lift IO` constraint to result in an `IO` computation.
- Use `withThrow` to adapt from one type of error to another.


byorgey:
@kostmo, eager to hear what you think of this.  I think overall it simplifies quite a bit of code, but I want to make sure it makes sense going forward.  Happy to answer any questions about how to work with code written in this style.

src/Swarm/Game/Failure.hs
```diff
@@ -26,4 +28,5 @@ data LoadingFailure
   = DoesNotExist Entry
   | EntryNot Entry
   | CanNotParse ParseException
+  | Duplicate Text
```
kostmo, line 31: Should this `Duplicate` constructor have another field that indicates what kind of thing was duplicated?  E.g. an `Entity`, a `Subworld`, etc.

src/Swarm/Game/Entity.hs
```diff
@@ -313,11 +316,11 @@ deviceForCap cap = fromMaybe [] . M.lookup cap . entitiesByCap
 -- | Build an 'EntityMap' from a list of entities.  The idea is that
 --   this will be called once at startup, when loading the entities
 --   from a file; see 'loadEntities'.
-buildEntityMap :: [Entity] -> Either Text EntityMap
+buildEntityMap :: Has (Throw LoadingFailure) sig m => [Entity] -> m EntityMap
 buildEntityMap es = do
   case findDup (map fst namedEntities) of
-    Nothing -> Right ()
-    Just duped -> Left $ T.unwords ["Duplicate entity named", quote duped]
```
kostmo, line 320: In reference to [this comment](https://github.com/swarm-game/swarm/pull/1392/files#r1274184205), here is where the information is dropped that an "Entity" is being duplicated

byorgey:
Hah, I think I found a bug in the implementation of the `Accum` effect in the `fused-effects` library, which is causing warnings to be repeatedly duplicated.  I will prepare an upstream patch but this definitely shouldn't be merged until it is fixed upstream.

xsebek:
Sorry about the conflicts @byorgey. 😅

byorgey:
> Sorry about the conflicts @byorgey. sweat_smile

Hah, no worries at all.  I would rather have to fix some conflicts for this PR than hold everything else up, especially since this one is kind of blocked at the moment on the `fused-effects` bug.  Since I haven't gotten a response on that yet I'm thinking of just sticking a fixed copy of that module in our repo for now.

src/Swarm/Doc/Pedagogy.hs
```diff
@@ -159,11 +157,13 @@ generateIntroductionsSequence =
...
-  when (notNull failures) . liftIO $
-    hPutStrLn stderr "Loading failures: " >> mapM_ (T.hPutStrLn stderr . prettyFailure) failures
-  return loadedScenarios
+  entities <- loadEntities
+  evalAccum (mempty :: Seq SystemFailure) $ loadScenarios entities
+
+-- Note we ignore any warnings generated by 'loadScenarios' above.
+-- Any warnings will be caught when loading all the scenarios via
+-- the usual code path; we do not need to do anything with them
+-- here while simply rendering pedagogy info.
```
kostmo, line 166: I think that this comment should be inline with the body of `loadScenarioCollection` (i.e. just above line 161).  As is, it's kind of floating in no-man's land.

src/Swarm/Game/Achievement/Persistence.hs
```diff
@@ -30,23 +28,25 @@ getSwarmAchievementsPath createDirs = getSwarmXdgDataSubdir createDirs "achievem
...
+        isFile <- sendIO $ doesFileExist fullPath
         if isFile
           then do
             eitherDecodedFile <- sendIO (Y.decodeFileEither fullPath)
             return $ left (AssetNotLoaded Achievement p . CanNotParse) eitherDecodedFile
           else return . Left $ AssetNotLoaded Achievement p (EntryNot File)
-      return $ partitionEithers eithersList
-    else return ([AssetNotLoaded Achievement "." $ DoesNotExist Directory], [])
+    else do
+      warn $ AssetNotLoaded Achievement "." $ DoesNotExist Directory
```
kostmo, line 48: I like having some kind of writer monad to propagate warnings.  Passing them all explicitly as function return values was pretty clunky.

src/Swarm/Game/Failure.hs
```diff
@@ -38,26 +47,54 @@ data LoadingFailure
...
+prettyAsset a = showLowT a
 
 prettyLoadingFailure :: LoadingFailure -> Text
 prettyLoadingFailure = \case
-  DoesNotExist e -> "The " <> tShowLow e <> " is missing!"
-  EntryNot e -> "The entry is not a " <> tShowLow e <> "!"
-  CanNotParse p -> "Parse failure:\n" <> T.pack (indent 8 $ prettyPrintParseException p)
+  DoesNotExist e -> "The " <> showLowT e <> " is missing!"
+  EntryNot e -> "The entry is not a " <> showLowT e <> "!"
+  CanNotParse p -> "Parse failure:\n" <> into @Text (indent 8 $ prettyPrintParseException p)
```
kostmo, line 76: `unlines`?

src/Swarm/Game/Failure.hs
```diff
@@ -38,26 +47,54 @@ data LoadingFailure
...
+prettyAsset :: Asset -> Text
+prettyAsset (Data ad) = prettyAssetData ad
+prettyAsset a = showLowT a
 
 prettyLoadingFailure :: LoadingFailure -> Text
 prettyLoadingFailure = \case
-  DoesNotExist e -> "The " <> tShowLow e <> " is missing!"
-  EntryNot e -> "The entry is not a " <> tShowLow e <> "!"
-  CanNotParse p -> "Parse failure:\n" <> T.pack (indent 8 $ prettyPrintParseException p)
+  DoesNotExist e -> "The " <> showLowT e <> " is missing!"
```
kostmo, line 74: Opportunity to switch to `unwords` (since we're touching this line anyway)

src/Swarm/Game/Failure.hs
```diff
@@ -38,26 +47,54 @@ data LoadingFailure
...
+  OrderFileWarning orderFile w ->
+    "Warning: while processing " <> into @Text orderFile <> ":\n" <> T.unlines (prettyOrderFileWarning w)
+  CustomFailure m -> m
+
+prettyOrderFileWarning :: OrderFileWarning -> [Text]
+prettyOrderFileWarning = \case
+  NoOrderFile -> ["  File not found; using alphabetical order"]
+  MissingFiles missing ->
+    "  Files not listed will be ignored:"
+      : map (("  - " <>) . into @Text) missing
```
kostmo, line 97: Opportunity to use [`BulletList`](https://github.com/swarm-game/swarm/blob/c69d76edf3c4547c363bdb2bc387a7fa773e3b12/src/Swarm/Language/Pretty.hs#L75-L78)?

src/Swarm/Game/Failure.hs
```diff
@@ -38,26 +47,54 @@ data LoadingFailure
...
+  | Duplicate AssetData Text
   | CustomMessage Text
   deriving (Show)
 
-tShowLow :: Show a => a -> Text
-tShowLow = T.toLower . T.pack . show
+data OrderFileWarning
+  = NoOrderFile
+  | MissingFiles [FilePath]
+  | DanglingFiles [FilePath]
```
kostmo, line 57: Can strengthen the type (see reference below)
```suggestion
  | MissingFiles (NonEmpty FilePath)
  | DanglingFiles (NonEmpty FilePath)
```

src/Swarm/Game/ScenarioInfo.hs
```diff
@@ -126,85 +134,71 @@ flatten (SISingle p) = [p]
...
-        liftIO . putStr . unlines $
-          ( "Warning: while processing "
-              <> (dirName </> orderFileName)
-              <> ": files not listed in "
-              <> orderFileName
-              <> " will be ignored"
-          )
-            : map ("  - " <>) missing
+      unless (null missing) . warn $
+        OrderFileWarning (dirName </> orderFileName) (MissingFiles missing)
```
kostmo, line 183: Here, we can exploit the stronger `MissingFiles (NonEmpty FilePath)` type:
```suggestion
      forM_ (NE.nonEmpty missing) $ warn .
        OrderFileWarning (dirName </> orderFileName) . MissingFiles
```

src/Swarm/Game/ScenarioInfo.hs
```diff
@@ -126,85 +134,71 @@ flatten (SISingle p) = [p]
...
 
-      unless (null dangling) $
-        liftIO . putStr . unlines $
-          ( "Warning: while processing "
-              <> (dirName </> orderFileName)
-              <> ": nonexistent files will be ignored"
-          )
-            : map ("  - " <>) dangling
+      unless (null dangling) . warn $
```
kostmo, line 185: Can use same `forM_ (NE.nonEmpty dangling) $ ...` idiom

src/Swarm/Language/Pipeline.hs
```diff
@@ -48,14 +48,14 @@ import Witch
 data ProcessedTerm = ProcessedTerm TModule Requirements ReqCtx
   deriving (Data, Show, Eq, Generic)
 
-processTermEither :: Text -> Either String ProcessedTerm
+processTermEither :: Text -> Either Text ProcessedTerm
 processTermEither t = case processTerm t of
-  Left err -> Left $ "Could not parse term: " ++ from err
+  Left err -> Left $ "Could not parse term: " <> err
```
kostmo, line 53: `T.unwords`?

kostmo (approved):

src/Swarm/Game/Achievement/Persistence.hs
byorgey: Indeed, that was one of my major motivations for doing this refactoring.

src/Swarm/Game/ScenarioInfo.hs
byorgey: Nice idea!

byorgey:
@kostmo all feedback addressed.  While looking at your suggestions for refactoring the `SystemFailure` pretty-printing code I realized I should just go whole hog and turn the code into `PrettyPrec` instances instead.

kostmo:
Looks good!

PR #1400, opened 2023-08-01
Implement 'backup' command

### LINKED ISSUE #1399: Reverse direction without generalized turning
I want to create a scenario in which the only two movements available are **forward** and **backwards**.  The two ways to accomplish this I can think of are:
1. Further partition the capabilities required to reference specific `Direction`s (see `COrient`), such that the `turn` command can execute `turn back` but not `turn left` (or any other direction, for that matter).
2. Provide a new command that simply moves the robot backwards (call it `Backup`).  `Backup` shall be identical to `Move` except in the opposite direction.

Option 2 seems simpler for an implementation standpoint.

kostmo:
Closes #1399.

## Demo

    scripts/play.sh --scenario data/scenarios/Testing/1399-backup-command.yaml --autoplay

### DIFF
--- a/data/entities.yaml
+++ b/data/entities.yaml
@@ -771,6 +771,15 @@
   capabilities: [move, turn, moveheavy]
   properties: [portable]
 
+- name: tape drive
+  display:
+    attr: device
+    char: '%'
+  description:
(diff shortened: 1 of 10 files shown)


byorgey (approved):
Looks good.  I like the `tape drive`. :smile:

PR #1404, opened 2023-08-03
mention `drill` command in drill entity description

byorgey:
Based on feedback from `__monty__` in IRC.


xsebek (approved):
Originally, entity descriptions were quite secretive. 😄

PR #1413, opened 2023-08-06
Render markdown in entity descriptions

### LINKED ISSUE #1408: Markdown formatting is not displayed properly in entity descriptions
> Oh, I see, the goal dialog is formatted properly, but the formatting doesn't show up when you look at the description of an entity.

_Originally posted by @byorgey in https://github.com/swarm-game/swarm/issues/1106#issuecomment-1666568218_

markdown

### LINKED ISSUE #1409: Words reflowed across a line break in objective text are missing a space between them
markdown2

Notice "nextto" and "samedirection" in the screenshot above.  Notice that "to build" has a space even though those words are also on separate lines in the `.yaml` file.  I think the difference may be that "next to" and "same direction" are contained in quoted (i.e. starting with a `|`) text in the `.yaml` file whereas "to build" is in unquoted text.  For reference, here is the relevant `.yaml` file:

https://github.com/swarm-game/swarm/blob/1eb2f9c567b51a5242323bdf59277aff5879618a/data/scenarios/Tutorials/build.yaml

xsebek:
* use `Markdown.Document` as `entityDescription`
* add missing spaces in `chunksOf`
* fix code in `entities.yaml` (mostly types and few outdated snippets)
* add code markdown in craft tutorial
* use colours for types and entities

- closes #1408
- closes #1409

### DIFF
--- a/data/entities.yaml
+++ b/data/entities.yaml
@@ -363,10 +363,7 @@
   - |
     Facilitates the concatenation of text values.
   - |
-    The infix operator
-    ```
-    ++ : text -> text -> text
-    ```
+    The infix operator `++ : text -> text -> text`{=snippet}
     can be used to concatenate two text values.  For example,
   - |
     "Number of widgets: " ++ format numWidgets
@@ -412,22 +409,23 @@
     also be woven into larger configurations such as cloth or nets.
   - |
     An equipped `string`{=entity} device enables several commands for working with
-    `text` values:
+    `text`{=type} values:
   - |
     `format : a -> text` can turn any value into a suitable text
     representation.
   - |
-    The infix operator `++ : text -> text -> text`
+    The infix operator `++ : text -> text -> text`{=snippet}
     can be used to concatenate two text values.  For example,
   - |
     ```
-    "Number of widgets: " ++ format numWidgets
+    let numWidgets = 42
+    in "Number of widgets: " ++ format numWidgets
     ```
   - |
     `chars : text -> int` computes the number of characters in a
-    `text` value.
+    `text`{=type} value.
   - |
-    `split : int -> text -> text * text` splits a `text` value into
+    `split : int -> text -> text * text` splits a `text`{=type} value into
     two pieces, one before the given index and one after.
   properties: [portable]
   capabilities: [format, concat, charcount, split]
@@ -443,9 +441,9 @@
     enables two functions:
   - |
     `charAt : int -> text -> int` returns the numeric code of the
-    character at a specific index in a (0-indexed) `text` value.
+    character at a specific index in a (0-indexed) `text`{=type} value.
   - |
-    `toChar : int -> text` creates a singleton (length-1) `text`
+    `toChar : int -> text` creates a singleton (length-1) `text`{=type}
     value containing a character with the given numeric code.
   properties: [portable]
   capabilities: [code]
@@ -462,7 +460,7 @@
     ```
     def thrice : cmd unit -> cmd unit = \c. c;c;c end
     ```
-  - defines the function `thrice` which repeats a command three times.
+  - defines the function `thrice`{=snippet} which repeats a command three times.
   properties: [portable, growable]
   growth: [100, 200]
   capabilities: [lambda]
@@ -797,7 +795,7 @@
     attr: device
     char: '%'
   description:
-  - A "tape drive" allows you to `backup`; that is, to `drive` in reverse.
+  - A "tape drive" allows you to `backup`; that is, to drive in reverse.
   capabilities: [backup]
   properties: [portable]
 
@@ -1014,7 +1012,8 @@
   - 'Example:'
   - |
     ```
-    if (x > 3) {move} {turn right; move}'
+    let x = 2 in
+    if (x > 3) {move} {turn right; move}
     ```
   properties: [portable]
   capabilities: [cond]
@@ -1129,7 +1128,7 @@
   - "To wait for a message and get the string value, use:"
   - |
     ```
-    l <- listen; log $ \"I have waited for someone to say \" ++ l
+    l <- listen; log $ "I have waited for someone to say " ++ l
     ```
   properties: [portable]
   capabilities: [listen]
@@ -1140,7 +1139,7 @@
     char: 'C'
   description:
   - |
-    A counter enables the command `count : string -> cmd int`,
+    A counter enables the command `count : text -> cmd int`,
     which counts how many occurrences of an entity are currently
     in the inventory.  This is an upgraded version of the `has`
     command, which returns a bool instead of an int and does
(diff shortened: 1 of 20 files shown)


data/entities.yaml
```diff
@@ -1323,16 +1322,17 @@
...
   - |
     `meetAll : (b -> actor -> cmd b) -> b -> cmd b` runs a command on
     every nearby actor (other than oneself), folding over the results
-    to compute a final result of type `b`.  For example, if `x`, `y`,
-    and `z` are nearby actors, then `meetAll f b0` is equivalent to
-    `b1 <- f b0 x; b2 <- f b1 y; f b2 z`.
+    to compute a final result of type `b`{=type}.  For example, if
+    `x`{=snippet}, `y`{=snippet}, and `z`{=snippet}
+    are nearby actors, then `meetAll f b0`{=snippet} is equivalent to
+    `b1 <- f b0 x; b2 <- f b1 y; f b2 z`{=snippet}.
```
xsebek, line 1335: It would be kind of neat if we could introduce variables with markup and then apply them to the rest of the document. Something like:
```markdown
If `x`{=variable} is a nearby actor, then `give "apple" x` will end all hunger.
```

xsebek:
This is how I debugged it:
```Haskell
ghci> -- parse entities.yaml
ghci> em :: [Entity] <- either (fail . prettyPrintParseException) pure =<< decodeFileEither "data/entities.yaml"
ghci> -- format 3D printer descriptions like in TUI
ghci> let Just d3 = L.find (\e -> view entityName e == T.pack "3D printer") em in chunksOf 99 . toStream $ view entityDescription d3
```
And it finally seems to look OK:
Screenshot 2023-08-06 at 2 19 12

data/entities.yaml
```diff
@@ -1374,9 +1374,9 @@
   - |
     Also allows manipulating composite values consisting of a
     collection of named fields.  For example, `[x = 2, y = "hi"]`
-    is a value of type `[x : int, y : text]`.  Individual fields
+    is a value of type `[x : int, y : text]`{=type}.  Individual fields
     can be projected using dot notation.  For example,
-    `let r = [y="hi", x=2] in r.x` has the value 2.  The order
+    `let r = [y="hi", x=2] in r.x` has the value `2`.  The order
```
kostmo, line 1379: Annotate snippet?
```suggestion
    `let r = [y="hi", x=2] in r.x`{=snippet} has the value `2`.  The order
```

src/Swarm/Language/Text/Markdown.hs
```diff
@@ -181,7 +216,8 @@ fromTextPure :: Text -> Either String (Document Text)
 fromTextPure t = do
   let spec = Mark.rawAttributeSpec <> Mark.defaultSyntaxSpec <> Mark.rawAttributeSpec
   let runSimple = left show . runIdentity
-  runSimple $ Mark.commonmarkWith spec "markdown" t
+  Document tokenizedDoc <- runSimple $ Mark.commonmarkWith spec "markdown" t
+  return . Document $ normalise <$> tokenizedDoc
```
kostmo, line 220: Random thought: would there be any value in having a `NormalizedDocument` newtype whose "smart constructor" performs the `normalize` function?

data/entities.yaml
xsebek: Not if we want it to be parsed, maybe I should have named it `{=invalid}` or `{=non-swarm}`.

src/Swarm/Language/Text/Markdown.hs
xsebek: Honestly, if `Document` used `Seq` I would consider doing it immediately in `(<>)`, the naive approach here merges the text one-by-one anyway.

data/entities.yaml
byorgey: Yes, that would be neat, and not too hard in theory, though it might require a bit of work to plumb the right information around.  The basic idea would be that when we typecheck code in backticks we don't simply start it in an empty context, but rather start it in a context that already contains any declared variables (with types like `forall a. a`).

byorgey:
> And it finally seems to look OK:

The formatting looks great, but now it looks cramped.  Can we add blank lines in between every paragraph, like we used to?

data/entities.yaml
kostmo: [Literate Swarm](https://wiki.haskell.org/Literate_programming)?

data/entities.yaml
```diff
@@ -797,7 +795,7 @@
     attr: device
     char: '%'
   description:
-  - A "tape drive" allows you to `backup`; that is, to `drive` in reverse.
+  - A "tape drive" allows you to `backup`; that is, to drive in reverse.
```
byorgey, line 798:
```suggestion
  - A `tape drive`{=entity} allows you to `backup`; that is, to drive in reverse.
```

byorgey:
CI is failing since there is some code in a docspec test in `Swarm.Game.Exception` that needs to be fixed, I think.

xsebek:
> The formatting looks great, but now it looks cramped. Can we add blank lines in between every paragraph, like we used to?

@byorgey should be fixed now, I had to rethink the code a little and keep paragraphs for printing.

xsebek:
Btw. the docspec test would not fail if we had `OverloadedLists` extension enabled, but as it is I used `mempty`.

But for markdown in Haskell code (e.g. generated entity/robot description) we can use that instance.

src/Swarm/Language/Text/Markdown.hs
```diff
@@ -208,11 +251,8 @@ unStream = \case
   TextNode a t -> (TextNode a, t)
   CodeNode t -> (CodeNode, t)
   RawNode a t -> (RawNode a, t)
-  ParagraphBreak -> error "Logic error: Paragraph break can not be unstreamed!"
```
kostmo, line 211: Always nice to get rid of partial functions...

src/Swarm/Language/Text/Markdown.hs
```diff
@@ -230,13 +269,16 @@ chunksOf n = chop (splitter True n)
...
+        startSpace = T.takeWhile isSpace t
+        twords = T.words t & _head %~ (startSpace <>) & _last %~ (<> endSpace)
+     in case splitWordsAt i twords of
           ([], []) -> (con "", con "")
           ([], ws@(ww : wws)) ->
             both (con . T.unwords) $
               -- In case single word (e.g. web link) does not fit on line we must put
               -- it there and guarantee progress (otherwise chop will cycle)
-              if start then ([ww], wws) else ([], ws)
+              if start then ([T.take i ww], T.drop i ww : wws) else ([], ws)
```
kostmo, line 281: Can we utilize `T.splitAt i ww` here?

src/Swarm/Language/Text/Markdown.hs
```diff
@@ -265,15 +306,36 @@ streamToText = T.concat . map nodeToText
...
+--------------------------------------------------------------
+
+nodeToMark :: PrettyPrec a => Node a -> Text
+nodeToMark = \case
+  LeafText a t -> foldl attr t a
+  LeafRaw _ c -> wrap "`" c
+  LeafCode c -> wrap "`" (prettyText c)
+  LeafCodeBlock f c -> codeBlock f $ prettyText c
+ where
+  codeBlock f t = "```" <> T.pack f <> "\n" <> t <> "\n```"
```
kostmo, line 330:
```suggestion
  codeBlock f t = wrap "```" $ T.pack f <> "\n" <> t <> "\n"
```

kostmo (approved):
LGTM with minor suggestions.  Also don't forget [this one](https://github.com/swarm-game/swarm/pull/1413/files#r1285142879) from @byorgey.

src/Swarm/Language/Text/Markdown.hs
xsebek: Indeed, I don't think that paragraph break worked anyway! 😄 It shows that I wrote it before knowing where and how we use the documents.

src/Swarm/Language/Text/Markdown.hs
xsebek: Maybe, but I doubt it would make the code simpler or more performant.

kostmo: Sure, it's up to your preference.  I suppose it could look like this.
```suggestion
              if start then bimap pure (: wws) $ T.splitAt i ww else ([], ws)
```

xsebek: Thanks, I was about to try write it myself, but this is definitely harder to read for me. 😆

xsebek:
@kostmo I went ahead and added some colours anyway:
Screenshot 2023-08-12 at 12 04 07

I also added a script to run scenarios _in order_ with `--autoplay` and show goal descriptions. 😁 
```
❯ scripts/autoplay-tutorials.sh
backstory.yaml	CONTINUE [Y/n]:
move.yaml	CONTINUE [Y/n]:
craft.yaml	CONTINUE [Y/n]:
grab.yaml	CONTINUE [Y/n]:
place.yaml	CONTINUE [Y/n]: n
```

Issue #351, opened 2022-06-03
Generalize properly when inferring the type of a bind


xsebek:
Returning a generic type seems to crash the typing:
```
> r <- return (\i.undefined)
\i. undefined : ∀ a1 a2. a1 -> a2
> 1
```
```
swarm: Maybe.fromJust: Nothing
CallStack (from HasCallStack):
  error, called at libraries/base/Data/Maybe.hs:148:21 in base:Data.Maybe
  fromJust, called at src/Swarm/Language/Types.hs:285:11 in swarm-0.1.0.0-inplace:Swarm.Language.Types
```
The game crashes immediately once it starts to type the next expression. :bomb:  :boom:

byorgey:
Weird, just typing `\i. undefined` or `return (\i. undefined)` by themselves don't trigger the bug, the bind `r <-` seems to be necessary...

byorgey:
So I found the place in `inferModule` where we typecheck a binder.  At first I thought it was as simple as adding a call to `generalize`, just like we have in the branch of `inferModule` that infers the type of a `def`.  However, it didn't work, and it took me a long time to figure out why.  The problem is that when trying to generalize over the type of `r : u0 -> u0` (where `u0` represents a unification variable, i.e. `IntVar (-9223372036854775807)` in the above), it doesn't consider `u0` free, since `u0` actually shows up in the context: the type of `r` itself is in the context---as it should be, since `r` may occur later in the term in addition to having its type exported to the top level.  (Indeed, even in this example, `r <- return (\i.i)` is really syntax sugar for `r <- return (\i.i); return r`.)

Perhaps the right answer is that it simply doesn't make sense to bind polymorphic things via `<-` at the top level.

xsebek:
Thanks, @byorgey for investigating this, I found it quite surprising and the GHCi example is even more baffling. :exploding_head:

byorgey:
https://stackoverflow.com/questions/61805899/why-ghc-types-any-here is relevant, and further strengthens my conviction that we shouldn't allow polymorphic types as the result of a top-level `<-`.  

[...]

PR #1418, opened 2023-08-06
Throw an error instead of crashing on impredicative types

byorgey:
Towards #351.  This is not a solution to the underlying problem, but at least it prevents the game from crashing and returns a placeholder error message instead.


src/Swarm/Language/Pretty.hs
```diff
@@ -294,6 +294,8 @@ instance PrettyPrec TypeErr where
     "Record does not have a field with name" <+> pretty x <> ":" <+> ppr t
   prettyPrec _ (InvalidAtomic reason t) =
     "Invalid atomic block:" <+> ppr reason <> ":" <+> ppr t
+  prettyPrec _ Impredicative =
+    "Impredicative type encountered. See https://github.com/swarm-game/swarm/issues/351"
```
xsebek, line 298: Could we fit some more actionable info that would be useful even to people who don't know what impredicative types are?

xsebek (approved):

src/Swarm/Language/Pretty.hs
byorgey: Well, not really, the problem is this is really just a band-aid on a bug.  We should have rejected this type before getting to this point, at some point when we would have been in a position to give a more reasonable error message.  But at the point where we find unification variables in a type it is kind of too late to give any useful info.  I guess I can at least make clear in the message that this is a bug in Swarm.

PR #1425, opened 2023-08-13
Refactor mystery baton type

xsebek:
* add more descriptive name to the mysterious `MVar` in Web API startup function


byorgey (approved):
:+1:

PR #1431, opened 2023-08-18
system robots should build system robots

### LINKED ISSUE #1430: Robots built by system robots should also be system robots
As title.

kostmo:
Closes #1430.

# Demo

    scripts/play.sh -i data/scenarios/Testing/1430-built-robot-ownership.yaml

Before, pressing `F2` would show two robots: the `base` and the system-built robot.
Now, only the `base` is shown, as the built robot is properly classified as another system robot.


byorgey (approved):
Makes sense!

Issue #455, opened 2022-06-21
Entities: identity, class, and type


byorgey:
The problem
-----------------

There are several open issues struggling with how to deal with entities when we need to be able to distinguish among entities that would otherwise be considered identical.

- #114 wonders about how to manage things like `box`es with things put inside them: once we can start putting things in boxes, it matters a lot *which* box we are referring to.  For example, we want to be able to say things like "take one of my 50 `box` entities, put a bunch of things in it, and now give that specific box to another robot".
- #116 wonders about how to distinguish among different `paper` entities that have had values printed on them.

I'm sure other, similar situations will arise as we continue designing more interesting entities and commands.

In general, we have considered entities to be uniquely identified by their *name*.  However, this causes a lot of problems.
- Entities aren't *actually* uniquely identified or stored by name, but rather by hash value.  I think there are multiple places in the code where we look up an entity by name and just *assume* that we will get only one, even though in theory there could be multiple.
- Changing the name of an entity to make it unique is problematic.  Consider the example of a `box` again.  #114 suggests taking a `box` entity and changing its name to something unique (*e.g.* `box1425`) so that we can now refer to it uniquely.  The problem is that we now no longer know that it is a `box`!  As in, we might want to say something like "give every box in my inventory to `base`", but we can't: the `box1425` is no longer a `box`, it is now a `box1425`.
- Moreover, representing entities as `string` values is error prone and not type-safe: e.g. `give base "treee"` only fails at runtime.

Proposed solution concept
-------------------------------------

The germ of this idea came from @noahyor .  #393 noticed that principles of object-oriented programming were helpful in thinking about the way robots interact; the same thing is true here.  The basic idea is to formally distinguish between the *class* of an entity (represented by a `string` such as `box`) and its *identity* (represented by some kind of internal `int` ID, probably).  For efficiency, we can still have multiple identical entities represented simply by a count (for example, if we have harvested 1257 `tree`s, we do not need to give each one a separate, unique identity! --- we just have 1257 copies of the generic `tree` entity).  But we can also have multiple entities with distinct identities but the same `class`.  This makes the `box` and `paper` examples above work cleanly.  For example, we can refer to a specific `box` by its unique identity, but it is still a `box`, so we can do things like "give all `box`es" even if some of them have distinct identities.

Proposed solution details
----------------------------------

- Make a new opaque `entity` type, similar to what we did with robots in #303 .
    - Internally, an `entity` value can be stored as an `Int`, the entity hash value we already compute.
[...]

PR #1432, opened 2023-08-18
combustion

### LINKED ISSUE #1355: Fire that spreads
Perhaps this could be generalized to "spreadable growth", but I will discuss with the use case of fire in mind.
A new entity property, `combustible`, may be introduced.

Setting an entity on fire would work something like the inverse of "growth" and the `harvest` command; the entity is replaced by a system robot that depicts (and animates) "burning", as opposed a to a bot that represents "growing".

While burning, there should be a random (customizable) chance of spreading to an adjacent tile each tick.  For now perhaps all entities will have a default duration that they may "burn" before self-extinguishing.  As some future extension, perhaps the duration could be customized analogously to the `"growth"` property.

By default, after extinguishing, all burned entities will yield an `"ash"` entity.  Again, as a future extension, this might be customized to yield other kinds of entities based on what entity it started as.

Other potential spreadable phenomena:
* radiation
* weeds
* water (coastline erosion)
* sand (desert erosion)

Scenario ideas:
* plow a fire break
* Build a radiation [containment dome](https://en.wikipedia.org/wiki/Chernobyl_New_Safe_Confinement)

This may make use of #1343 to customize the display sequence.

kostmo:
Closes #1355

# Demo

    scripts/play.sh -i data/scenarios/Testing/1355-combustion.yaml --autoplay

image

### DIFF
--- a/src/Swarm/Game/Entity.hs
+++ b/src/Swarm/Game/Entity.hs
@@ -13,10 +13,14 @@
 -- are mutually recursive (an inventory contains entities, which can
 -- have inventories).
 module Swarm.Game.Entity (
+  EntityName,
+
   -- * Properties
   EntityProperty (..),
   GrowthTime (..),
   defaultGrowthTime,
+  Combustibility (..),
+  defaultCombustibility,
 
   -- * Entities
   Entity,
@@ -31,6 +35,7 @@ module Swarm.Game.Entity (
   entityDescription,
   entityOrientation,
   entityGrowth,
+  entityCombustion,
   entityYields,
   entityProperties,
   hasProperty,
@@ -118,6 +123,8 @@ import Text.Read (readMaybe)
 import Witch
 import Prelude hiding (lookup)
 
+type EntityName = Text
+
 ------------------------------------------------------------
 -- Properties
 ------------------------------------------------------------
@@ -133,6 +140,8 @@ data EntityProperty
     Opaque
   | -- | Regrows from a seed after it is harvested.
     Growable
+  | -- | Can use the Ignite command on it
+    Combustible
   | -- | Regenerates infinitely when grabbed or harvested.
     Infinite
   | -- | Robots drown if they walk on this without a boat.
@@ -161,6 +170,24 @@ newtype GrowthTime = GrowthTime (Integer, Integer)
 
 defaultGrowthTime :: GrowthTime
 defaultGrowthTime = GrowthTime (100, 200)
+
+-- | Properties of combustion
+data Combustibility = Combustibility
+  { ignition :: Double
+  -- ^ Rate of ignition by a neighbor, per tick.
+  -- When denoted as "lambda",
+  -- probability of ignition over a period "t" is:
+  -- 1 - e^(-(lambda * t))
+  -- See: https://math.stackexchange.com/a/1243629
+  , duration :: (Integer, Integer)
+  -- ^ min and max tick counts for combustion to persist
+  , product :: Maybe EntityName
+  -- ^ what entity, if any, is left over after combustion
+  }
+  deriving (Eq, Ord, Show, Read, Generic, Hashable, FromJSON, ToJSON)
+
+defaultCombustibility :: Combustibility
+defaultCombustibility = Combustibility 0.5 (100, 200) (Just "ash")
 
 ------------------------------------------------------------
 -- Entity
@@ -224,6 +251,8 @@ data Entity = Entity
   --   a robot moves, it moves in the direction of its orientation.
   , _entityGrowth :: Maybe GrowthTime
   -- ^ If this entity grows, how long does it take?
+  , _entityCombustion :: Maybe Combustibility
+  -- ^ If this entity is combustible, how spreadable is it?
   , _entityYields :: Maybe Text
   -- ^ The name of a different entity obtained when this entity is
   -- grabbed.
@@ -243,7 +272,7 @@ data Entity = Entity
 -- | The @Hashable@ instance for @Entity@ ignores the cached hash
 --   value and simply combines the other fields.
 instance Hashable Entity where
-  hashWithSalt s (Entity _ disp nm pl descr orient grow yld props caps inv) =
+  hashWithSalt s (Entity _ disp nm pl descr orient grow combust yld props caps inv) =
     s
       `hashWithSalt` disp
       `hashWithSalt` nm
@@ -251,6 +280,7 @@ instance Hashable Entity where
       `hashWithSalt` docToText descr
       `hashWithSalt` orient
       `hashWithSalt` grow
+      `hashWithSalt` combust
       `hashWithSalt` yld
       `hashWithSalt` props
       `hashWithSalt` caps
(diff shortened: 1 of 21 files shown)


src/Swarm/Game/Step.hs
```diff
@@ -1230,6 +1316,34 @@ execConst c vs s k = do
       _ -> badConst
     Grab -> doGrab Grab'
     Harvest -> doGrab Harvest'
+    Ignite -> case vs of
```
byorgey, line 1319: Might there be a benefit to doing something like `use "torch"` instead of adding yet another command?  I am not sure how it would work exactly, just wondering.

src/Swarm/Game/Entity.hs
```diff
@@ -162,6 +169,22 @@ newtype GrowthTime = GrowthTime (Integer, Integer)
 defaultGrowthTime :: GrowthTime
 defaultGrowthTime = GrowthTime (100, 200)
 
+-- | Properties of combustion
+data Combustibility = Combustibility {
+    ignition :: Double
+    -- ^ Likelihood of ignition by a neighbor, per tick.
+    -- This gets multiplied by the number of (4-adjacent) neighbors
```
byorgey, line 176: This doesn't sound right, and it's not what the code actually does.  If you have 4 neighbors burning your probability of igniting is not 4 times the probability with only one neighbor burning.  It's actually something like 1 - (1 - p)^4, but we don't need to put that in a comment.

src/Swarm/Game/Step.hs
```diff
@@ -1065,6 +1065,92 @@ seedProgram minTime randTime thing =
...
+        []
+        []
+        True
+        False
+        ts
+  where
+    combustionProg = combustionProgram minT (maxT - minT) combustionOutput spreadLikelihood
+    -- FIXME should support Maybe
+    Combustibility spreadLikelihood (minT, maxT) maybeCombustionProduct = combustibility
+    combustionOutput = fromMaybe "ash" maybeCombustionProduct
```
byorgey, line 1152: The special default entity name `"ash"` is repeated here and in `defaultCombustibility`.  Let's DRY it up and just have it in one place.

src/Swarm/Game/Entity.hs
kostmo: You caught me with a "draft" PR! :satisfied: 😇 

But the attention is appreciated anyway.

byorgey: Oh, I knew it was a draft, just thought I'd take a look and leave a few comments.  If I suggest doing something you were already going to do, or make a comment about part of the code that's going to change anyway, feel free to ignore me. :smile:

src/Swarm/Game/Step.hs
kostmo: So far, the `use` command exclusively applies a **recipe** to an **entity** to produce another **entity**.  `ignite` is closer to the `harvest` command, in that the application of the command produces a **robot**.  I think it would be weird to special-case `use` for this purpose.

kostmo: But perhaps a `"torch"` could provide the `Capability` that enables the `ignite` command.

xsebek: If we end up with a lot of commands like ignite/harvest/… that have effects, then we could generalise `use` and drop the commands.

It might be nice to have #455 for that.

src/Swarm/Game/Entity.hs
```diff
@@ -345,6 +388,7 @@ instance FromJSON Entity where
               <*> (v .: "description")
               <*> v .:? "orientation"
               <*> v .:? "growth"
+              <*> v .:? "combustion"
```
byorgey, line 391: Can we also update `data/schema/entities.json` and `scenarios/README.md` to match?  I know we haven't been keeping it completely up-to-date but it would be nice to get it back in shape.

src/Swarm/Language/Syntax.hs
```diff
@@ -539,6 +541,10 @@ constInfo c = case c of
       [ "Leaves behind a growing seed if the harvested item is growable."
       , "Otherwise it works exactly like `grab`."
       ]
+  Ignite ->
+    command 1 short . doc "Ignite a combustible item in the specified direction." $
+      [ "Combustion persists for a random duration and may spread."
```
byorgey, line 546: Can we say something about what the return value represents?

src/Swarm/TUI/View.hs
```diff
@@ -1115,6 +1115,7 @@ displayProperties :: [EntityProperty] -> Widget Name
 displayProperties = displayList . mapMaybe showProperty
  where
   showProperty Growable = Just "growing"
+  showProperty Combustible = Just "burning"
```
byorgey, line 1118: Why not `combustible`?  `burning` sounds like it is currently on fire.

src/Swarm/Game/Step.hs
```diff
@@ -1230,6 +1385,34 @@ execConst c vs s k = do
...
     Harvest -> doGrab Harvest'
+    Ignite -> case vs of
+      [VDir d] -> do
+        let verb = "ignite"
+            verbed = "ignited"
+
+        (loc, me) <- lookInDirection d
+        -- Ensure there is an entity here.
+        e <-
+          me `isJustOrFail` ["There is nothing here to", verb <> "."]
```
byorgey, line 1396: Don't we have code very similar to this in other places?  Is there an opportunity to factor out some repeated code?

src/Swarm/Game/Step.hs
```diff
@@ -1065,6 +1065,161 @@ seedProgram minTime randTime thing =
     selfdestruct
   |]
 
+-- Triggers the ignition of the entity underfoot with some delay.
```
byorgey, line 1068: All the code for dealing with ignition & combustion is somewhat complex.  I think I understand it and it makes sense.  But it might be nice to have a comment somewhere giving an overall explanation of the flow of execution and how all the pieces fit together.

We could also move all the combustion-related code into a submodule, `Swarm.Game.Step.Combustion`?  (The `Step` module is long enough as it is!)  That would also provide a natural place for a comment explaining the big picture.

byorgey (approved):
:fire:

src/Swarm/Game/Entity.hs
kostmo: Updated.  Though it was tedious to reproduce the info in `scenarios/README.md`.  Do we have a way to auto-generate that from `data/schema/entities.json`?

src/Swarm/Language/Syntax.hs
kostmo: Ah, it is not supposed to have a return value.  Good catch.

src/Swarm/Game/Entity.hs
byorgey: Agreed.  No, we don't, but that would be nice.

src/Swarm/Game/Step.hs
kostmo: I moved just enough helper functions out of `Swarm.Game.Step` and into `Swarm.Game.Step.Util` to allow for extraction of the combustion code into `Swarm.Game.Step.Combustion`.  We can follow up to offload more stuff to `Swarm.Game.Step.Util` and re-organize.

src/Swarm/Game/Entity.hs
kostmo: > it was tedious to reproduce the info in scenarios/README.md. Do we have a way to auto-generate that from data/schema/entities.json?

For posterity: fixed in #1441

PR #1448, opened 2023-08-23
Normalise versions for comparison

xsebek:
* make sure that `0.4` and `0.4.0.0` are the same

This fixes the issue worked around by 5c509d2 for future releases.


byorgey (approved):
:+1:

PR #1454, opened 2023-08-23
Draft: Add autoplay option

xsebek:
* add Autoplay to scenario launch options

### DIFF
--- a/src/Swarm/TUI/Launch/Controller.hs
+++ b/src/Swarm/TUI/Launch/Controller.hs
@@ -16,6 +16,7 @@ import Data.Maybe (listToMaybe)
 import Graphics.Vty qualified as V
 import Swarm.Game.Scenario.Status (ParameterizableLaunchParams (LaunchParams))
 import Swarm.Game.ScenarioInfo
+import Swarm.Language.Pipeline (ProcessedTerm (..))
 import Swarm.TUI.Controller.Util
 import Swarm.TUI.Launch.Model
 import Swarm.TUI.Launch.Prep (initFileBrowserWidget, makeFocusRingWith, parseSeedInput, parseWidgetParams, toValidatedParams)
@@ -24,7 +25,6 @@ import Swarm.TUI.Model.Name
 import Swarm.TUI.Model.StateUpdate
 import Swarm.TUI.Model.UI
 import Swarm.Util (listEnums)
-import Swarm.Language.Pipeline (ProcessedTerm (..))
 
 updateFocusRing :: EditingLaunchParams -> EventM Name LaunchOptions ()
 updateFocusRing parsedParams = do
--- a/src/Swarm/TUI/Launch/Controller.hs
+++ b/src/Swarm/TUI/Launch/Controller.hs
@@ -24,6 +24,7 @@ import Swarm.TUI.Model.Name
 import Swarm.TUI.Model.StateUpdate
 import Swarm.TUI.Model.UI
 import Swarm.Util (listEnums)
+import Swarm.Language.Pipeline (ProcessedTerm (..))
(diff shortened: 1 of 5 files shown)


src/Swarm/TUI/Launch/Prep.hs
```diff
@@ -51,11 +52,14 @@ parseSeedInput seedEditor =
...
+parseWidgetParams :: Maybe ProcessedTerm -> LaunchControls -> IO EditingLaunchParams
+parseWidgetParams solution (LaunchControls autoPlay (FileBrowserControl _fb maybeSelectedScript _) seedEditor _ _) = do
   eitherParsedCode <-
-    runThrow . withThrow (prettyText @SystemFailure) $
-      traverse parseCodeFile maybeSelectedScript
+    if autoPlay
+      then pure . Right $ CodeToRun ScenarioSuggested <$> solution
+      else
+        runThrow . withThrow (prettyText @SystemFailure) $
+          traverse parseCodeFile maybeSelectedScript
```
xsebek, line 62: I tried to get current scenario solution here, but I am not sure this path is even run. 😕

I attempted to debug it using `hPutStrLn stderr`, but did not see either branch print anything.

Maybe it's just something crazy like in #1456 and I need to test it in a clean docker container.

xsebek: OK, so it's only run if script is selected.

I will need to look around and see why.

src/Swarm/TUI/Launch/Controller.hs
```diff
@@ -136,7 +137,24 @@ handleLaunchOptionsEvent siPair = \case
         activateFocusedControl x
       _ -> return ()
   CharKey ' ' -> activateControl
-  Key V.KEnter -> activateControl
+  Key V.KEnter -> do
+    activateControl
+    fr <- use $ uiState . uiLaunchConfig . controls . scenarioConfigFocusRing
+    case focusGetCurrent fr of
+      Just (ScenarioConfigControl (ScenarioConfigPanelControl AutoPlaySelector)) -> do
+        mScenario <- use $ uiState . scenarioRef
```
xsebek, line 145: It seems this is not populated yet - where might I find the current scenario when I am in Menu?

src/Swarm/TUI/Launch/Controller.hs
```diff
@@ -136,7 +137,24 @@ handleLaunchOptionsEvent siPair = \case
...
+        let solution = view scenarioSolution . fst =<< mScenario
+        e <- use $ uiState . uiLaunchConfig . editingParams
+        uiState . uiLaunchConfig . editingParams
+          .= ( e
+                 { initialCode =
```
xsebek, line 150: * [ ] TODO: Create lenses for editing params

src/Swarm/TUI/Launch/Controller.hs
```diff
@@ -136,7 +137,24 @@ handleLaunchOptionsEvent siPair = \case
         activateFocusedControl x
       _ -> return ()
   CharKey ' ' -> activateControl
-  Key V.KEnter -> activateControl
+  Key V.KEnter -> do
+    activateControl
+    fr <- use $ uiState . uiLaunchConfig . controls . scenarioConfigFocusRing
```
xsebek, line 142: @kostmo should I just factor out a "cacheAutoplay" function?

I am not sure how the `cacheValidated<X>Input` functions are supposed to work, so any tips would be appreciated.

xsebek (commented):
I hacked on this a little and it now correctly shows that the solution is not where I hoped it would be:

Screenshot 2023-09-11 at 22 29 30

PR #1457, opened 2023-08-23
Test runtime log does not contain errors

### LINKED ISSUE #1449: Check scenario ordering errors in CI
Oddly it seems some warnings sneaked past our integration tests:

Screenshot 2023-08-23 at 10 14 15

While benign, there is no reason players should see those.

xsebek:
- closes #1449

### DIFF
--- a/src/Swarm/Game/Achievement/Persistence.hs
+++ b/src/Swarm/Game/Achievement/Persistence.hs
@@ -47,7 +47,6 @@ loadAchievementsInfo = do
             return $ left (AssetNotLoaded Achievement p . CanNotParseYaml) eitherDecodedFile
           else return . Left $ AssetNotLoaded Achievement p (EntryNot File)
     else do
-      warn $ AssetNotLoaded Achievement "." $ DoesNotExist Directory
       return []
 
 -- | Save info about achievements to XDG data directory.
--- a/src/Swarm/Game/Achievement/Persistence.hs
+++ b/src/Swarm/Game/Achievement/Persistence.hs
@@ -18,7 +18,7 @@ import Swarm.Game.Achievement.Attainment
 import Swarm.Game.Achievement.Definitions
 import Swarm.Game.Failure
 import Swarm.Game.ResourceLoading (getSwarmXdgDataSubdir)
-import Swarm.Util.Effect (forMW, warn)
+import Swarm.Util.Effect (forMW)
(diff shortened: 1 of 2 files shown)


byorgey (approved):
Looks good to me!

xsebek:
It seems there is an error about achievements in CI:
```
Tests
  Test runtime log does not contain errors:                                                                    FAIL
    test/integration/Main.hs:112:
    Error was produced during loading: Failed to acquire achievement from path '.':
      The 0 is missing!
    Use -p '/Test runtime log does not contain errors/' to rerun this test only.
```

@kostmo any ideas?

byorgey:
I don't understand how that message is even possible.  It is coming from here:

https://github.com/swarm-game/swarm/blob/8ed5b92edc0d7a3727625ae65edcff0a243d5a61/src/Swarm/Game/Failure.hs#L93

The `e` on that line of code is an `Entry`:

https://github.com/swarm-game/swarm/blob/8ed5b92edc0d7a3727625ae65edcff0a243d5a61/src/Swarm/Game/Failure.hs#L40-L44

I don't understand how calling `ppr` on an `Entry` could produce `0`...

...oh, lol, the `PrettyPrec Entry` instance is wrong:

https://github.com/swarm-game/swarm/blob/8ed5b92edc0d7a3727625ae65edcff0a243d5a61/src/Swarm/Game/Failure.hs#L88-L89

It looks so plausible!

xsebek:
I restarted ghc-9.6.2 job, it should at least show a better error message, but will still need more investigation.

xsebek:
@byorgey I thought it was running tests on merged state, but maybe I remembered that wrong. 😕

byorgey:
Ah, no, it looks like you're right.  The failed test now has a correct error message.

xsebek:
(That's because I force-pushed the rebased branch 😉)

Seems like we are missing the achievements directory in CI. 🤔

https://github.com/swarm-game/swarm/blob/49903de26d101d1284adc1460aaeaea6fdb61e48/src/Swarm/Game/Achievement/Persistence.hs#L49-L51

If I understand it correctly, having no achievements before playing should _not_ be an error. 🤔

byorgey:
Ah, yes, that makes sense.  Not having the achievements directory exist is just the normal state of affairs when playing the game for the first time.  It should not give a warning.  I think we should just get rid of that `else` block.

PR #1459, opened 2023-08-23
Make CLI format format

### LINKED ISSUE #1185: format subcommand is a misnomer
The `format` subcommand claims to "[format a file](https://github.com/swarm-game/swarm/blob/9e2f8b29613223b7eaded5690257abbc6553ac93/app/Main.hs#L46)", but what it really does is validate parseability and just [dump the input verbatim](https://github.com/swarm-game/swarm/blob/9e2f8b29613223b7eaded5690257abbc6553ac93/app/Main.hs#L155-L159).

xsebek:
- closes #1185

### DIFF
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -16,7 +16,9 @@ import Options.Applicative
 import Swarm.App (appMain)
 import Swarm.Doc.Gen (EditorType (..), GenerateDocs (..), PageAddress (..), SheetType (..), generateDocs)
 import Swarm.Language.LSP (lspMain)
-import Swarm.Language.Pipeline (processTerm)
+import Swarm.Language.Module (Module (..))
+import Swarm.Language.Pipeline (ProcessedTerm (..), processTerm)
+import Swarm.Language.Pretty (prettyText)
 import Swarm.TUI.Model (AppOpts (..), ColorMode (..))
 import Swarm.TUI.Model.UI (defaultInitLgTicksPerSecond)
 import Swarm.Version
@@ -161,8 +163,11 @@ formatFile :: Input -> IO ()
 formatFile input = do
   content <- getInput input
   case processTerm content of
-    Right _ -> do
-      Text.putStrLn content
+    Right t -> do
+      case t of
+        Nothing -> Text.putStrLn ""
+        Just (ProcessedTerm (Module ast _mctx) _req _ctx) ->
+          Text.putStrLn $ prettyText ast
       exitSuccess
     Left e -> do
       Text.hPutStrLn stderr $ showInput input <> ":" <> e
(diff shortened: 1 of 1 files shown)


app/Main.hs
```diff
@@ -161,8 +163,11 @@ formatFile :: Input -> IO ()
 formatFile input = do
   content <- getInput input
   case processTerm content of
-    Right _ -> do
-      Text.putStrLn content
+    Right t -> do
+      case t of
+        Nothing -> Text.putStrLn ""
+        Just (ProcessedTerm (Module ast _mctx) _req _ctx) ->
+          Text.putStrLn $ prettyText ast
```
xsebek, line 170: It seems there is no layout, maybe it needs some options? 😕 

```sh
cabal run swarm -- format scenarios/Challenges/_arbitrage/solution.sw
# def doN = \n. \f. if (n > 0) {f; force doN (n - 1) f} {noop} end; def intersperse = \n. \f2. \f1. if (n > 0) {f1; if (n > 1) {f2} {noop}; force intersperse (n - 1) f2 f1} {noop} end; def boolAsLeftRight = \b. if b {right} {left} end; def squareSide = drill down; force doN 5 move end; def goToFirstShop = force doN 3 move; force doN 5 (force doN 2 move; grab); turn left; force doN 2 move end; def doPartialFigureEight = force intersperse 2 (turn left) (force squareSide); force intersperse 4 (turn right) (force squareSide) end; def doFullFigureEight = \startLeft. force intersperse 4 (turn (force boolAsLeftRight (not startLeft))) (force squareSide); force intersperse 4 (turn (force boolAsLeftRight startLeft)) (force squareSide) end; force goToFirstShop; force doN 1 (force doPartialFigureEight); force doN 9 (force doFullFigureEight true); force doN 8 (force doFullFigureEight false); force squareSide; drill down
``

Comment (byorgey) [reply app/Main.hs:170]: It needs #11 !  We haven't really put any/much effort in to pretty-printing yet.

Comment (byorgey) [reply app/Main.hs:170]: We should not call `processTerm` in order to do formatting, since in particular `processTerm` does not only parse and typecheck but also elaborate, which does things like add extra calls to `force` and so on.  Also I think we should still be able to format code even if it does not typecheck, as long as it parses.  So instead of calling `processTerm` we need to just call the parser directly.  Something like this (untested):
```
content <- getInput input
case readTerm content of
  Right t -> case t of
    Nothing -> Text.putStrLn ""
    Just ast -> Text.putStrLn $ prettyText ast
```

**Comment by xsebek** [comment]
@byorgey once #1464 lands, this could be good to go:

```
> cabal run swarm -O0 -- format scenarios/Challenges/_blender/patrol-clockwise.sw
def forever = \c. c; forever c end;
def encircle = \lDir. \rDir.
  turn lDir;
  b <- blocked;
  if b {turn rDir} {wait 1};
  fwBlocked <- blocked;
  if fwBlocked {turn rDir} {move}
end;
def patrolCW = forever (encircle right left) end;
patrolCW
```

I noticed it drops comments, but we can fix that later.

xsebek:
Rebased and ready to go! 🙂

byorgey (approved):

PR #1460, opened 2023-08-23
Erase the erase function

xsebek:
* simplify the `erase` function using the `Functor` instance


byorgey (approved):
:exploding_head:

byorgey:
Wait, actually, now I am having second thoughts, is this new implementation really equivalent to the old one?

xsebek:
@byorgey it passes tests, so I think it's pretty equivalent (up to our test coverage). 😄 

I have been using `void` occasionally when I needed to get rid of the types and did not know about `erase`.

I there is a reason for dropping `SrcLoc`?

byorgey:
> Is there is a reason for dropping `SrcLoc`?

No, there's not, which is why I think `void` is better!

PR #1464, opened 2023-08-24
Make pretty Terms format prettier

### LINKED ISSUE #11: Better indentation & layout for pretty-printing
I kind of threw the pretty-printer together without much regard for how the layout and indentation works.  The `prettyprinter` library has lots of nice tools for specifying this kind of thing; someone just needs to sit down and figure out how they work and how to apply them to the Swarm language to get good results.

This is very related to #7 , which will depend on this to get nice-looking save files that users can actually edit without tearing their hair out.  Another semi-related issue is #644 .

At the moment, we use pretty-printing quite a bit for types, but we only pretty-print terms in certain error messages (*e.g.* when a command throws an exception).

xsebek:
- improve layout of terms:
  - break lines on binds, unless it fits on one line
  - lambdas go on same line, but the body _can_ go to next line
  - def and let can have long body on next line (indented for def)
  - parens and braces have body indented if it does not fit on line
- closes #11

Example using `--format` from #1459:

```
> cabal run swarm -O0 -- format scenarios/Challenges/_blender/patrol-clockwise.sw                          
def forever = \c. c; force forever c end;
def encircle = \lDir. \rDir.
  turn lDir;
  b <- blocked;
  if b {turn rDir} {wait 1};
  fwBlocked <- blocked;
  if fwBlocked {turn rDir} {move}
end;
def patrolCW = force forever (force encircle right left) end;
force patrolCW
```

### DIFF
--- a/src/Swarm/Language/Pretty.hs
+++ b/src/Swarm/Language/Pretty.hs
@@ -63,7 +63,7 @@ prettyString = docToString . ppr
 -- | Optionally surround a document with parentheses depending on the
 --   @Bool@ argument.
 pparens :: Bool -> Doc ann -> Doc ann
-pparens True = parens
+pparens True = group . parens
 pparens False = id
 
 -- | Surround a document with backticks.
@@ -180,7 +180,7 @@ instance PrettyPrec Term where
   prettyPrec p (TRequire n e) = pparens (p > 10) $ "require" <+> pretty n <+> ppr @Term (TText e)
   prettyPrec p (TRequirements _ e) = pparens (p > 10) $ "requirements" <+> ppr e
   prettyPrec _ (TVar s) = pretty s
-  prettyPrec _ (TDelay _ t) = braces $ ppr t
+  prettyPrec _ (TDelay _ t) = group . braces $ line' <> nest 1 (ppr t) <> line'
   prettyPrec _ t@TPair {} = prettyTuple t
   prettyPrec _ (TLam x mty body) =
     "\\" <> pretty x <> maybe "" ((":" <>) . ppr) mty <> "." <+> ppr body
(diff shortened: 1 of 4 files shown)


xsebek:
@byorgey it would be great to have this rebased on #1459 so the improvements can be tested interactively.

Should I rebase and close #1459, or should we merge `--format` first?

byorgey (approved):
This looks good to me, it's exactly the kind of thing I had in mind.  Thanks for figuring out the right combinators to use etc.

The CI is failing because the formatting of terms has changed in a few error messages.  It looks like the messages are long and then have a term on the end, so the pretty-printer is inserting newlines, but this looks bad, e.g.
```
1:9: Invalid atomic block: def, let, and lambda are not allowed: \c. c;
c
```
I think we should change it to put a newline before the (indented) term, e.g.
```
1:9: Invalid atomic block: def, let, and lambda are not allowed: 
  \c. c; c
```

As for #1459 , that one will require more work/thought, since as you point out, it is not valid to *e.g.* insert `force` when formatting.  So I think we should just go ahead and merge this.

xsebek:
@byorgey meanwhile, I came up with a different solution - add backticks to code in error messages (as we already did for some) and group it, so it tends to stay on one line.

Thinking about it, we could experiment with `flatAlt` and use triple backticks in case it goes multiline. But I am still a Prettyprinter newbie. 😅

src/Swarm/Language/Pretty.hs
```diff
@@ -299,37 +299,42 @@ prettyTypeErr code (CTE l tcStack te) =
...
-  prettyPrec _ (CantInferProj t) =
-    "Can't infer the type of a record projection:" <+> ppr t
-  prettyPrec _ (UnknownProj x t) =
-    "Record does not have a field with name" <+> pretty x <> ":" <+> ppr t
-  prettyPrec _ (InvalidAtomic reason t) =
-    "Invalid atomic block:" <+> ppr reason <> ":" <+> ppr t
-  prettyPrec _ Impredicative =
-    "Unconstrained unification type variables encountered, likely due to an impredicative type. This is a known bug; for more information see https://github.com/swarm-game/swarm/issues/351 ."
+  prettyPrec _ = \case
+    (UnifyErr ty1 ty2) ->
```
byorgey, line 303: Aren't all these parentheses around the different cases redundant?

byorgey:
Instead of enclosing in single or triple backticks etc., it would also be fun to play with using the "annotations" feature to do syntax highlighting etc. for error messages, just like we do for entity descriptions etc.  But we can leave that for another PR.

xsebek:
@byorgey yeah, I thought about that too.

I think that would lead to a standalone function:
```Haskell
prettyTypeErr :: TypeErr -> Doc MarkdownAnn
```
The `alterAnnotations` function could then be used to convert our annotations to [`AnsiStyle`](https://hackage.haskell.org/package/prettyprinter-ansi-terminal-1.1.3/docs/Prettyprinter-Render-Terminal.html#t:AnsiStyle) or `Attr`.

src/Swarm/Language/Pretty.hs
xsebek: They are, I just did not get them with my initial regex and then I was lazy to do it. 😅

xsebek:
OK, let's not be hasty this time...

Screenshot 2023-08-24 at 21 20 44

It seems some code can get messed up inside descriptions. 🤦

xsebek:
The descriptions should be OK again. Hopefully it was the only visible bug. 😅

byorgey:
@xsebek looks good to me!  I like how you handled the issue with code formatting in descriptions by distinguishing between single and triple backticks.

PR #1470, opened 2023-08-26
Fix lambda precedence

### LINKED ISSUE #1468: Parameters to case are not formatted with parens
**Describe the bug**

I noticed in #1459, that the case command parameters are formatted wrong.

They are both of type `a -> cmd b`, but are formatted without parentheses.

**To Reproduce**

Try to format the `hanoi-invariant.sw`:
```
❯ cabal run swarm -O0 -- format scenarios/Challenges/_hanoi/hanoi-invariant.sw
```

In the original, there are cases like this:
```
case mn (\_. //...
```
but in the formatted version, the parens are dropped:
```
case mn \_. wait 16; /* ... */ \_. return ()
```

**Expected behavior**

The precedence should force the bind chains in parentheses.

xsebek:
- give lambdas lower precedence (9) then application (10)
- closes #1468

Before (with #1459):
```
❯ cabal run swarm -O0 -- format <(echo '\\m. case m (\\x. x + 1) (\\y. y * 2)')
\m. case m \x. x + 1 \y. y * 2
```

After:
```
❯ cabal run swarm -O0 -- format <(echo '\\m. case m (\\x. x + 1) (\\y. y * 2)')
\m. case m (\x. x + 1) (\y. y * 2)
```

### DIFF
--- a/src/Swarm/Language/Pretty.hs
+++ b/src/Swarm/Language/Pretty.hs
@@ -188,7 +188,7 @@ instance PrettyPrec Term where
   prettyPrec _ (TVar s) = pretty s
   prettyPrec _ (TDelay _ t) = group . encloseWithIndent 2 lbrace rbrace $ ppr t
   prettyPrec _ t@TPair {} = prettyTuple t
-  prettyPrec p t@(TLam {}) = 
+  prettyPrec p t@(TLam {}) =
     pparens (p > 9) $
       prettyLambdas t
   -- Special handling of infix operators - ((+) 2) 3 --> 2 + 3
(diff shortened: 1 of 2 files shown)


src/Swarm/Language/Pretty.hs
```diff
@@ -188,7 +188,9 @@ instance PrettyPrec Term where
   prettyPrec _ (TVar s) = pretty s
   prettyPrec _ (TDelay _ t) = group . encloseWithIndent 2 lbrace rbrace $ ppr t
   prettyPrec _ t@TPair {} = prettyTuple t
-  prettyPrec _ t@(TLam {}) = prettyLambdas t
+  prettyPrec p t@(TLam {}) =
+    pparens (p > 9) $
+      prettyLambdas t
```
xsebek, line 193: @byorgey I assume this is overly pessimistic - in the example case, only the inner lambda requires parentheses.

However, the least amount of parentheses might be harder to read. 😄

byorgey: Hah, yes, in theory, a lambda at the very tail end of an expression does not need parentheses, but that would be annoying to read.  I think this is fine.

byorgey (approved):

PR #1472, opened 2023-08-26
Move traversable documentation to derived instance

xsebek:
This should render the documentation in haddock according to this blog:
https://kowainik.github.io/posts/haddock-tips#what-can-be-documented

It does not for me (maybe it will in release?), but at least it's where it belongs.


byorgey (approved):

byorgey:
Our CI would fail if there's a haddock syntax error, right?

xsebek:
@byorgey AFAIK that was the behaviour before GHC9.0, so yes.

Issue #1436, opened 2023-08-21
Autogenerate parts of `scenarios/README.md`


kostmo:
See https://github.com/swarm-game/swarm/pull/1432#discussion_r1299447625:

Much of the content of `scenarios/README.md` is identical to that in `data/schema/entities.json`.  We should autogenerate the former from the latter, in whole or in part.

Builds upon #344.

Related: #1428, as the json schema must be parsed.

PR #1475, opened 2023-08-27
Validate scenarios against json schema

### LINKED ISSUE #1428: Validate scenarios against schema as part of CI
Yeah, I don't think we have been keeping it up to date, I know @kostmo has added a bunch of features that I think we missed adding to the schema.  If there is an automated way to make sure all existing scenario descriptions conform to the schema, we should add it to our CI!

_Originally posted by @byorgey in https://github.com/swarm-game/swarm/pull/1376#discussion_r1289339157_

kostmo:
Closes #1428

Since the authoritative validation of scenario files is actually performed by virtue of `swarm` parsing them, this CI job actually exists to ensure the JSON Schema descriptions are accurate.  This is important for two purposes:
* Documentation is generated from the JSON Schema files (#1436)
* JSON Schema has integration with VS Code and other IDEs

# Testing

Verified that the schema checker action does indeed work by intentionally pushing an invalid scenario file in f789f81.

### DIFF
--- a/data/schema/planarloc.json
+++ b/data/schema/planarloc.json
@@ -1,7 +1,7 @@
 {
     "$schema": "http://json-schema.org/draft-07/schema#",
     "$id": "https://raw.githubusercontent.com/swarm-game/swarm/main/data/schema/planarloc.json",
-    "title": "Cosmic location",
+    "title": "Planar location",
     "description": "Planar location",
     "type": "array",
     "items": [
--- a/data/schema/planarloc.json
+++ b/data/schema/planarloc.json
@@ -1,0 +1,17 @@
+{
+    "$schema": "http://json-schema.org/draft-07/schema#",
+    "$id": "https://raw.githubusercontent.com/swarm-game/swarm/main/data/schema/planarloc.json",
+    "title": "Cosmic location",
+    "description": "Planar location",
+    "type": "array",
+    "items": [
+        {
+            "name": "X coordinate",
+            "type": "number"
+        },
+        {
+            "name": "Y coordinate",
+            "type": "number"
+        }
+    ]
+}
--- a/data/schema/recipes.json
+++ b/data/schema/recipes.json
@@ -26,16 +26,16 @@
         ],
         "properties": {
             "in": {
-                "$ref": "https://raw.githubusercontent.com/swarm-game/swarm/main/data/schema/inventory.json",
+                "$ref": "./inventory.json",
                 "description": "A list of ingredients consumed by the recipe. Each ingredient is a tuple consisting of an integer and an entity name, indicating the number of copies of the given entity that are needed."
             },
             "out": {
-                "$ref": "https://raw.githubusercontent.com/swarm-game/swarm/main/data/schema/inventory.json",
+                "$ref": "./inventory.json",
                 "description": "A list of outputs produced by the recipe. It is a list of [count, entity name] tuples just like in."
             },
             "required": {
                 "default": [],
-                "$ref": "https://raw.githubusercontent.com/swarm-game/swarm/main/data/schema/inventory.json",
+                "$ref": "./inventory.json",
                 "description": "A list of catalysts required by the recipe. They are neither consumed nor produced, but must be present in order for the recipe to be carried out. It is a list of [count, entity name] tuples just like in and out."
             },
             "time": {
--- a/data/schema/recipes.json
+++ b/data/schema/recipes.json
@@ -25,6 +25,7 @@
             }
         ],
         "properties": {
+            "additionalProperties": false,
             "in": {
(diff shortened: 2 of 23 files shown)


data/schema/recipes.json
```diff
@@ -26,16 +26,16 @@
...
             },
             "out": {
-                "$ref": "https://raw.githubusercontent.com/swarm-game/swarm/main/data/schema/inventory.json",
+                "$ref": "./inventory.json",
                 "description": "A list of outputs produced by the recipe. It is a list of [count, entity name] tuples just like in."
             },
             "required": {
                 "default": [],
-                "$ref": "https://raw.githubusercontent.com/swarm-game/swarm/main/data/schema/inventory.json",
+                "$ref": "./inventory.json",
```
xsebek, line 38: Huh, I thought this would break editors, but it seems to work fine with VSCode! 👍

xsebek (approved):
It seems I left several things unfinished in #620, thanks for picking this up @kostmo! 👍 

Did you manage to integrate the schema to your editor? I use VSCode with this workspace config:
```json
{
    "yaml.schemas": {
        "https://raw.githubusercontent.com/swarm-game/swarm/feature/json-schema-fixes/data/schema/scenario.json": [
            "data/scenarios/**/*.yaml"
        ],
        "https://raw.githubusercontent.com/swarm-game/swarm/feature/json-schema-fixes/data/schema/entities.json": [
            "data/entities.yaml"
        ],
        "https://raw.githubusercontent.com/swarm-game/swarm/feature/json-schema-fixes/data/schema/recipes.json": [
            "data/recipes.yaml"
        ],
    }
}
```

We should probably note that somewhere (`CONTRIBUTING.md` or `editors`?). I am kind of embarrassed that I only wrote that to the previous PR and can't see it in any repo docs. 😅

data/schema/planarloc.json
```diff
@@ -0,0 +1,17 @@
+{
+    "$schema": "http://json-schema.org/draft-07/schema#",
+    "$id": "https://raw.githubusercontent.com/swarm-game/swarm/main/data/schema/planarloc.json",
+    "title": "Cosmic location",
```
byorgey, line 4:
```suggestion
    "title": "Planar location",
```

data/schema/planarloc.json
```diff
@@ -0,0 +1,17 @@
...
+    "description": "Planar location",
```
byorgey, line 5: Maybe expand on this description a bit, *e.g.* "x and y coordinates of a location in a particular world" or something like that.

data/schema/robot.json
```diff
@@ -15,19 +15,11 @@
...
-                    "name": "X coordinate",
-                    "type": "number"
-                },
-                {
-                    "name": "Y coordinate",
-                    "type": "number"
-                }
-            ],
-            "description": "An optional (x,y) starting location for the robot. If the loc field is specified, then a concrete robot will be created at the given location. If this field is omitted, then this robot record exists only as a template which can be referenced from a cell in the world palette. Concrete robots will then be created wherever the corresponding palette character is used in the world map."
+            "description": "An optional (x,y) starting location for the robot. If the loc field is specified, then a concrete robot will be created at the given location. If this field is omitted, then this robot record exists only as a template which can be referenced from a cell in the world palette. Concrete robots will then be created wherever the corresponding palette character is used in the world map.",
```
byorgey, line 18:
```suggestion
            "description": "An optional starting location for the robot. If the loc field is specified, then a concrete robot will be created at the given location. If this field is omitted, then this robot record exists only as a template which can be referenced from a cell in the world palette. Concrete robots will then be created wherever the corresponding palette character is used in the world map.",
```

I suggest taking out the `(x,y)` since it is no longer entirely accurate; the `loc` can now either be a normal (x,y) location or a cosmic location.

byorgey (commented):
Where is all the stuff about `subworlds`?  I don't see that described anywhere in the schemas, but we have testing scenarios which use them.

byorgey:
I don't understand how this succeeds when the schemas have nothing about `subworlds` but there are scenarios which make use of subworlds.  Do `.yaml` files still validate when they have extra stuff which is not described in the schema?  That would be a problem because that is probably going to be the number one failure mode (someone adds a new feature but forgets to update the schema files).

kostmo:
> Do `.yaml` files still validate when they have extra stuff which is not described in the schema?

Yes, there are apparently two ways to handle this.
1. Add `"additionalProperties": false` for every `object` in the schema.  This is what I have done.
2. Use a schema validatation tool that allows specifying this globally.  The `check-jsonschema` tool does not support this.  Interestingly, the YAML VS Code extension does have a setting for this:
    Screenshot from 2023-08-27 18-57-09

Issue #1473, opened 2023-08-26
Optionally break up long types on `->` when pretty-printing


byorgey:
**Describe the bug**
This example code:
```
def harvestbox : dir -> (cmd unit -> cmd unit) -> (cmd unit -> cmd unit) -> cmd unit = \d. \rep1. \rep2.
  doboxP harvest d rep1 rep2 notempty
end
```
is currently pretty-printed as
```
def harvestbox : dir -> (cmd unit -> cmd unit) -> (
    cmd unit -> cmd unit
  ) -> cmd unit = \d. \rep1. \rep2.
  doboxP harvest d rep1 rep2 notempty
end;
```
which looks weird.  I would expect/desire something more like the following, perhaps:
```
def harvestbox : 
  dir ->
  (cmd unit -> cmd unit) ->
  (cmd unit -> cmd unit) ->
  cmd unit
  = \d. \rep1. \rep2.
  doboxP harvest d rep1 rep2 notempty
end;
```
I think the issue is that right now, the only option the pretty-printer has for spreading long function types across multiple lines is to split out parenthesized parts.  But I think we instead need to pretty-print function types in such a way that the pretty-printer knows the right way to optionally break lines at `->`, which I think would mean that parenthesized types are rarely split across multiple lines like that.  But I could imagine it still being the right thing to do in a case like
```
def foo : 
  lots ->
  of ->
  parameters ->
  (
    a ->
    very ->
    long ->
    function ->
    parameter
  ) ->
  more ->
  result
```
So I don't think there's anything wrong with the way we handle parenthesized types per se.

xsebek:
@byorgey I looked into this and the problem is that we want to basically get alternative layout for _lambda chain_.

How do I get that chain though? This current instance does not even know that it is a fixed type:
```Haskell
instance (PrettyPrec t) => PrettyPrec (TypeF t) where
instance (PrettyPrec (t (Fix t))) => PrettyPrec (Fix t) where
```

If I used `Type` instead, it would run into overlapping instances. 😖

I think moving the fix instance to a plain function and using it in the fixed type instances would be the most clean solution.

xsebek:
OK, I see getting rid of Fixed instance would lead to duplicating code... so maybe we could add some class instance for both `Fix` and `UTerm` to please give us the lambda chain...

This recursive business is breaking my mind a bit and I got stuck, so any pointers would be appreciated. 😅

byorgey:
[...]

PR #1476, opened 2023-08-27
Allow specific format width

xsebek:
* add option `--width` to `format` CLI
* if width is not specified use terminal width
* if terminal width is unknown, use default (100)

This should help with testing (not so) long layouts like in #1473.

### DIFF
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -22,6 +22,7 @@ import Swarm.Language.Parse (readTerm)
 import Swarm.Language.Pretty (ppr)
 import Swarm.TUI.Model (AppOpts (..), ColorMode (..))
 import Swarm.TUI.Model.UI (defaultInitLgTicksPerSecond)
+import Swarm.Util ((?))
 import Swarm.Version
 import Swarm.Web (defaultPort)
 import System.Console.Terminal.Size qualified as Term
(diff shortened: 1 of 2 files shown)


app/Main.hs
```diff
@@ -157,16 +164,24 @@ showInput :: Input -> Text
 showInput Stdin = "(input)"
 showInput (File fp) = pack fp
 
+infixl 3 <??>
+(<??>) :: Maybe a -> a -> a
+(<??>) = flip fromMaybe
```
byorgey, line 169: Note, we already have this defined in `Swarm.Util` as `(?)`. (But `infixr 1` instead of `infixl 3`.)

app/Main.hs
```diff
@@ -157,16 +164,24 @@ showInput :: Input -> Text
...
-        Just ast -> Text.putStrLn $ prettyText ast
-      exitSuccess
+    Right Nothing -> Text.putStrLn ""
+    Right (Just ast) -> do
+      mWindow <- Term.size
+      let mkOpt w = LayoutOptions (AvailablePerLine w 1.0)
+      let opt =
+            fmap mkOpt mWidth
+              <|> fmap (\(Term.Window _h w) -> mkOpt w) mWindow
+                <??> defaultLayoutOptions
```
byorgey, line 183: If you use an `infixr` version of `<??>` / `?` you can just write this as a chain (untested):
```
fmap mkOpt mWidth
  ? fmap (\(Term.Window _h w) -> mkOpt w) mWindow
  ? defaultLayoutOptions
```

byorgey (approved):

app/Main.hs
xsebek: Works great thanks!

app/Main.hs
xsebek: Ah, I saw Relude has [`(?:)`](https://hackage.haskell.org/package/relude-1.2.0.0/docs/Relude-Monad-Maybe.html#v:-63-:) as `infixr 0` but I did not know we have some operator too, thanks! 👍

PR #1478, opened 2023-08-28
Include VS Code workspace settings in repo

kostmo:
This saves a step in the development setup process.

Note that the settings file is included in the toplevel `.gitignore`, so any changes on top of this one will not result in a "dirty" repo.


xsebek (approved):
LGTM! 👍 I can imagine this will save me time once I delete it with `git clean -fxd`. 😄

Issue #1469, opened 2023-08-26
Parse, format and parse again swarm scripts in CI


xsebek:
The pretty-printer may have bugs in more complex scripts.

We should add formatting and parsing the formatted output of all our swarm scripts in integration tests.

This would make the pretty printer significantly more robust as it would work on our most complex scripts.

This idea was sparked by this bug:
- #1468

xsebek:
Some other ideas, that may or may not be worth to do in this issue:
* we have lots of scripts in scenario descriptions (YAML) - apply the same testing to them too
* we have a pretty printer of types, but it only gets used in formatting explicit annotations - consider inferring the types of def, let and lambdas and printing those too
* with #1467 fixed, we could use the reformat in editors - this would force the parser to work even on partially valid inputs

byorgey:
Yes, now that we actually have some rudimentary pretty-printing in place we absolutely need to test that we can round-trip with pretty-printing and parsing.

Another idea would be to test the round-trip in the other direction: have several scripts (it does not need to be all of them) which are formatted according to the output of `--format`, and we could then test that parsing + pretty-printing them does not modify them at all.

xsebek:
Another idea would be to use QuickCheck Arbitrary instance and see what kind of code it cooks up.

byorgey:
`Arbitrary` instances for large AST types can be problematic, but yes, in general, testing on some kind of randomly generated ASTs is a great idea.

xsebek:
Turns out we already have a function to round trip parse and format in unit tests:

https://github.com/swarm-game/swarm/blob/4b5ecb05e55103ef2afdddb98e650d09fccb7d7c/test/unit/TestLanguagePipeline.hs#L388-L393

That specific function needs to be updated though, namely to not elaborate and to ignore source location.

xsebek:
Actually, I would like to have this tested even before the comments (#1467) are fixed. Implementing the other issues from #1571 will be a lot of trial and error and having tests for round tripping would be helpful.

We can workaround the comments by doing the round trip twice, which will remove them the first time and we can still check terms match.

Any tests that do not pass we can mark as expected failure and link an Issue.

byorgey:
There are actually two directions to think about here.

1. Generating random ASTs, and testing that we can pretty-print and parse them and get back the exact same AST.  This is actually the best test of the pretty-printer and parser, but it requires being able to generate random ASTs, which is tricky.  However, there are some known techniques and libraries for doing so which we might look into.
[...]

PR #1479, opened 2023-08-28
Make function chains pretty

### LINKED ISSUE #1473: Optionally break up long types on `->` when pretty-printing

xsebek:
- closes #1473

### DIFF
--- a/scripts/compare-format.sh
+++ b/scripts/compare-format.sh
@@ -9,7 +9,7 @@ function help() {
     echo
     echo "This script helps to compare the format layout for a range"
     echo "of output widths. Afterwards it prints those that differ"
-    echo "as markdown with tripple backticks."
+    echo "as markdown with triple backticks."
 }
 
 # Simple argumet parsing from https://stackoverflow.com/a/14203146
--- a/scripts/compare-format.sh
+++ b/scripts/compare-format.sh
@@ -12,7 +12,7 @@ function help() {
     echo "as markdown with triple backticks."
 }
 
-# Simple argumet parsing from https://stackoverflow.com/a/14203146
+# Simple argument parsing from https://stackoverflow.com/a/14203146
 POSITIONAL_ARGS=()
 while [[ $# -gt 0 ]]; do
   case $1 in
--- a/scripts/compare-format.sh
+++ b/scripts/compare-format.sh
@@ -1,0 +1,80 @@
+#!/bin/bash
+
+MIN=10
+MAX=200
+STEP=5
+
+function help() {
+    echo "$0 [--min $MIN --max $MAX --step $STEP] EXAMPLE_FILE.sw"
+    echo
+    echo "This script helps to compare the format layout for a range"
+    echo "of output widths. Afterwards it prints those that differ"
+    echo "as markdown with tripple backticks."
+}
+
+# Simple argumet parsing from https://stackoverflow.com/a/14203146
+POSITIONAL_ARGS=()
+while [[ $# -gt 0 ]]; do
+  case $1 in
+    --min)
+      MIN="$2"
+      shift # past argument
+      shift # past value
+      ;;
+    --max)
+      MAX="$2"
+      shift # past argument
+      shift # past value
+      ;;
+    -s|--step)
+      STEP="$2"
+      shift # past argument
+      shift # past value
+      ;;
+    -h|--help)
+      help
+      exit 0
+      ;;
+    -*)
+      echo "Unknown option $1"
+      help
+      exit 1
+      ;;
+    *)
+      POSITIONAL_ARGS+=("$1") # save positional arg
+      shift # past argument
+      ;;
+  esac
+done
+
+# Build first, otherwise the ouput would go to temporary files
+cabal build -O0
+
+function compare_format() {
+    echo "# $1"
+    if ! test -f "$1"; then
+      echo "Could not find file '$1'!"
+      return
+    fi
+    # save each version in a temporary file
+    t=$(mktemp -d)
+
+    for i in $(seq "$MIN" "$STEP" "$MAX"); do
+        echo -en "${i}\r"  # show progress
+        cabal run swarm -O0 -- format --width "$i" "$1" > "$t/$i.sw";
+    done
+    echo -en "   \r"
+
+    for i in $(seq "$MAX" "-$STEP" "$MIN"); do
+    if ! cmp -s "$t/$i.sw" "$t"/$((i - STEP)).sw; then
+        echo "$i";
+        echo '```';
+        cat "$t/$i.sw";
+        echo '```';
+    fi;
+    done
+}
+
+for file in "${POSITIONAL_ARGS[@]}"; do
+  compare_format "$file";
+done
--- a/src/Swarm/Language/Pretty.hs
+++ b/src/Swarm/Language/Pretty.hs
@@ -238,28 +238,15 @@ instance PrettyPrec Term where
             _ -> prettyPrecApp p t1 t2
     _ -> prettyPrecApp p t1 t2
   prettyPrec _ (TLet _ x mty t1 t2) =
-    group . vsep $
-      [ hsep $
-          ["let", pretty x]
-            ++ maybe [] (\ty -> [":", ppr ty]) mty
-            ++ ["=", ppr t1, "in"]
+    sep
+      [ prettyDefinition "let" x mty t1 <+> "in"
       , ppr t2
       ]
   prettyPrec _ (TDef _ x mty t1) =
-    let (defBody, defLambdaList) = unchainLambdas t1
-        defHead = "def" <+> pretty x
-        defType = maybe "" (\ty -> ":" <+> flatAlt (line <> indent 2 (ppr ty)) (ppr ty)) mty
-        defType' = maybe "" (\ty -> ":" <+> ppr ty) mty
-        defLambdas = hsep (map prettyLambda defLambdaList)
-     in sep
-          [ nest 2 . sep $
-              [ flatAlt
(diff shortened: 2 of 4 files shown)


src/Swarm/Language/Pretty.hs
```diff
@@ -121,14 +121,27 @@ data Wildcard = Wildcard
...
   prettyPrec _ _ = "_"
 
+-- | Split a function type chain, so that we can pretty print
+--   the type parameters aligned on each line when they don't fit.
+class UnchainableFun t where
+  unchainFun :: t -> [t]
+
+instance UnchainableFun Type where
+  unchainFun (a :->: ty) = a : unchainFun ty
+  unchainFun ty = [ty]
```
xsebek, line 131: @byorgey almost there, just have to figure out why it got too unchained 😅 
```
❯ cabal run swarm -O0 -- format --width 20 example.sw
def harvestbox:
  dir ->
  cmd unit ->
  cmd unit ->
  cmd unit ->
  cmd unit ->
  cmd unit 
  = \d. \rep1. \rep2.
  doboxP harvest d rep1 rep2 notempty
end
```

byorgey: Haha, oops!  That's odd.  The `UnchainableFun` instance above looks fine to me...

Hmm, maybe we should break up chains of function applications too...
```
def harvestbox
  = \d. \rep1. \rep2.
  doboxP 
    harvest 
    d 
    rep1
    rep2
    notempty
end
```

xsebek: @byorgey found it! I need to pass precedence to all but the last parameter.

xsebek: @byorgey I thought about function parameters too, but those might just be a bit complicated because of pattern matching operators. Let's leave that for a separate issue. 😅

xsebek:
Technically we could also apply the same logic to sum and product types...

I will get back to this later and maybe refactor the pretty printing construct so that it works for more separators.

src/Swarm/Language/Pretty.hs
byorgey: Oh, definitely leave it for a separate issue, I was just wondering out loud.

xsebek: I added it to the check list in #1571. 🙂

src/Swarm/Util.hs
```diff
@@ -239,6 +240,23 @@ applyWhen False _ x = x
...
+--
+-- Example:
+--
+-- >>> unsnocNE (1 :| [2, 3])
+-- ([1,2],3)
+--
+-- >>> unsnocNE (1 :| [])
+-- ([],1)
+unsnocNE :: NonEmpty a -> ([a], a)
+unsnocNE (x :| xs) = go x xs
```
xsebek, line 255: @kostmo I noticed there is a new HLint warning, but I could not find how to do this using our libraries.

I found this using Hoogle and borrowed the code.

test/unit/TestLanguagePipeline.hs
```diff
@@ -417,3 +411,14 @@ testLanguagePipeline =
...
+-- | Check round tripping of term from and to text, then test ToJSON/FromJSON.
+roundTripTerm :: Text -> Assertion
+roundTripTerm txt = do
+  assertEqual "roundtrip (readTerm -> prettyText)" txt (prettyText term)
+  assertEqual "roundtrip (ToJSON -> FromJSON)" term (decodeThrow $ encode term)
+ where
+  decodeThrow v = case eitherDecode v of
+    Left e -> error $ "Decoding of " <> from (T.decodeUtf8 (from v)) <> " failed with: " <> from e
+    Right x -> x
+  term = fromMaybe (error "empty document") $ either (error . T.unpack) id $ readTerm txt
```
xsebek, line 424: @byorgey I changed the function to produce better errors on round tripping and not elaborate.
```diff
- roundtrip
- expected: ProcessedTerm (Module {moduleAST = Syntax' {_sLoc = SrcLoc 0 30, _sTerm = SBind Nothing (Syntax' {_sLoc = SrcLoc 0 14, _sTerm = SDef False (LV {lvSrcLoc = SrcLoc 4 6, lvVar = "x"}) Nothing (Syntax' {_sLoc = SrcLoc 8 11, _sTerm = TInt 41, _sType = Forall [] (TyBaseF BInt)}), _sType = Forall [] (TyCmdF (TyBaseF BUnit))}) (Syntax' {_sLoc = SrcLoc 16 30, _sTerm = SDef False (LV {lvSrcLoc = SrcLoc 20 22, lvVar = "y"}) Nothing (Syntax' {_sLoc = SrcLoc 24 27, _sTerm = TInt 42, _sType = Forall [] (TyBaseF BInt)}), _sType = Forall [] (TyCmdF (TyBaseF BUnit))}), _sType = Forall [] (TyCmdF (TyBaseF BUnit))}, moduleCtx = Ctx {unCtx = fromList [("x",Forall [] (TyBaseF BInt)),("y",Forall [] (TyBaseF BInt))]}}) (Requirements {capReqs = fromList [CPower,CEnv], devReqs = fromList [], invReqs = fromList []}) (Ctx {unCtx = fromList [("x",Requirements {capReqs = fromList [], devReqs = fromList [], invReqs = fromList []}),("y",Requirements {capReqs = fromList [], devReqs = fromList [], invReqs = fromList []})]})
- but got: ProcessedTerm (Module {moduleAST = Syntax' {_sLoc = SrcLoc 0 32, _sTerm = SBind Nothing (Syntax' {_sLoc = SrcLoc 0 15, _sTerm = SDef False (LV {lvSrcLoc = SrcLoc 4 6, lvVar = "x"}) Nothing (Syntax' {_sLoc = SrcLoc 9 12, _sTerm = TInt 41, _sType = Forall [] (TyBaseF BInt)}), _sType = Forall [] (TyCmdF (TyBaseF BUnit))}) (Syntax' {_sLoc = SrcLoc 17 32, _sTerm = SDef False (LV {lvSrcLoc = SrcLoc 21 23, lvVar = "y"}) Nothing (Syntax' {_sLoc = SrcLoc 26 29, _sTerm = TInt 42, _sType = Forall [] (TyBaseF BInt)}), _sType = Forall [] (TyCmdF (TyBaseF BUnit))}), _sType = Forall [] (TyCmdF (TyBaseF BUnit))}, moduleCtx = Ctx {unCtx = fromList [("x",Forall [] (TyBaseF BInt)),("y",Forall [] (TyBaseF BInt))]}}) (Requirements {capReqs = fromList [CPower,CEnv], devReqs = fromList [], invReqs = fromList []}) (Ctx {unCtx = fromList [("x",Requirements {capReqs = fromList [], devReqs = fromList [], invReqs = fromList []}),("y",Requirements {capReqs = fromList [], devReqs = fromList [], invReqs = fromList []})]})
+ roundtrip (readTerm -> prettyText)
+ expected: "def x = 41 end; def y = 42 end"
+ but got: "def x = 41 end;\ndef y = 42 end"
```

AFAIK `readTerm` will not typecheck, is that OK or should I use another function?

xsebek: I would like to eventually use this function for #1469, thought that might mean either a copy in integration tests or a common test utility library.

scripts/compare-format.sh
```diff
@@ -0,0 +1,80 @@
...
+MIN=10
+MAX=200
+STEP=5
+
+function help() {
+    echo "$0 [--min $MIN --max $MAX --step $STEP] EXAMPLE_FILE.sw"
+    echo
+    echo "This script helps to compare the format layout for a range"
+    echo "of output widths. Afterwards it prints those that differ"
+    echo "as markdown with tripple backticks."
```
byorgey, line 12:
```suggestion
    echo "as markdown with triple backticks."
```

scripts/compare-format.sh
```diff
@@ -0,0 +1,80 @@
...
+}
+
+# Simple argumet parsing from https://stackoverflow.com/a/14203146
```
byorgey, line 15:
```suggestion
# Simple argument parsing from https://stackoverflow.com/a/14203146
```

test/unit/TestLanguagePipeline.hs
byorgey: Yes, it makes sense to use `readTerm` here instead of `processTerm`.

byorgey (approved):
Looks good to me!

Issue #1341, opened 2023-06-24
Stats for activity level of each robot


kostmo:
It could be useful for debugging (or just interesting to observe) to have a live measurement of the level of activity of each robot.  There are a few different ways this information could be presented in the F2 "Robots" popup:

1. Absolute number of "steps" of the CESK machine executed over the robot's lifetime.  This would include both tangible and intangible commands.  It would lend observability especially to system robots that may be performing more work than intended.
2. **Steps per tick:** Keep track of a window over a certain number of ticks and compute the running average.
3. **Duty cycle:** Ratio of "clock ticks" that the robot has been Waiting/Idle vs Busy.  To be useful, the "window" of the duty cycle should probably depend on what "speed" the UI is currently set to.

byorgey:
Yes, all of these sound nice!

byorgey:
Just reading over this again, and I think pretty much everything described here has been implemented, e.g. in #1484 and #2140 , so I will close this.  If there's some aspect of this that we still haven't done, let's open a new issue for it.

Issue #217, opened 2021-10-12
Add flag to test for invariants


polux:
With all the subtle invariants introduced by #200 and 2b146586d2c756e9c1e86d0975957d1ad96f5097 we really want to write a function that checks that they hold, and either add a test that they aren't violated over the run of a game or add an assert in `gameTick`, or both.

byorgey:
Let's collect the relevant invariants in a nice list here.
- Robots which are neither idle nor waiting are always in the `activeRobots` set

This is the only real invariant that I found from reading through #200.  I am not sure what invariants you are referring to from 2b14658.

byorgey:
Concretely, I propose:
- Add a new command-line flag `--check-invariants`, and store a boolean in the `UIState`
- Write one or more functions to check invariants (the one listed above is a start, but we can keep adding more as we think of them --- for example, we should never have `uiPlaying = False` and `uiMenu = NoMenu` at the same time), and call them conditionally on the `--check-invariants` flag being set.  The checking functions should crash the app loudly if an invariant is violated.
- Make sure we always set this flag in our CI tests (but not in the benchmark suite).

xsebek:
Related Issue:
- #1112

xsebek:
> Add a new command-line flag `--check-invariants`, and store a boolean in the `UIState`

@byorgey do we test UI state at all? I thought we only create the game state.

I think it would be better to have a boolean in game state that causes a verification of the invariants on each tick and if that fails a `Critical` message is written to the global log, which we check in integration tests:
https://github.com/swarm-game/swarm/blob/3cf62ab2947be91dad033760438252fc9bcaf169/test/integration/Main.hs#L130-L131

byorgey:
> do we test UI state at all? I thought we only create the game state.

I'm not sure I understand what you mean.  There are definitely UI-level invariants as well.

xsebek:
I meant the integration/unit tests. Of course we test the game manually as well.

byorgey:
Oh, I see.  Right, I don't think we currently have any integration/unit tests for the UI state.

xsebek:
I have met those subtle invariants from #200 today in #2522 and a few assertions would have helped. 😆 

One unspoken invariant was that _`currentTickWakeableBots` must be empty_, otherwise it gets overwritten.
I wonder if there was a way to break the it with `instant`, but we will never know since I am going to fix it.

PR #1484, opened 2023-08-31
Robot activity counts in F2 menu

kostmo:
Towards #1341.

    scripts/play.sh -i data/scenarios/Testing/1341-command-count.yaml --autoplay --speed 1

image


src/Swarm/Util/WindowedCounter.hs
```diff
@@ -0,0 +1,135 @@
...
+  Integral a =>
+  -- | reference tick
+  a ->
+  WindowedCounter a ->
+  Ratio a
+getOccupancy referenceTick (WindowedCounter s lastLargest nominalSpan) =
+  if Set.null s || maybe False (< referenceTick - nominalSpan) lastLargest
+    then 0
+    else case Set.lookupMin s of
+      Nothing -> 0
```
kostmo, line 97: TODO: I think I forgot to implement the filtering step described in the long Haddock. Should probably use `Set.split` here.

src/Swarm/Util/WindowedCounter.hs
```diff
@@ -0,0 +1,135 @@
...
+
+mkWindow ::
+  -- | Window span
+  a ->
+  WindowedCounter a
+mkWindow = WindowedCounter Set.empty Nothing
+
+-- | Return the ratio of {members in the window} to the {integral span
+-- represented by the window} (actually, the beginning of the window up to
+-- some reference tick).
```
kostmo, line 81: "... which is typically the current time"

src/Swarm/Game/Step.hs
```diff
@@ -999,6 +1005,12 @@ execConst c vs s k = do
   -- First, ensure the robot is capable of executing/evaluating this constant.
   ensureCanExecute c
 
+  -- Increment command count regardless of success
+  when (isTangible c) $
+    activityCounts . tangibleCommandCount %= (+ 1)
```
byorgey, line 1010:
```suggestion
    activityCounts . tangibleCommandCount += 1
```

src/Swarm/Util/WindowedCounter.hs
```diff
@@ -0,0 +1,135 @@
...
+data WindowedCounter a = WindowedCounter
+  { _members :: Set a
+  , _lastLargest :: Maybe a
+  -- ^ NOTE: It is possible that '_lastLargest' may not exist in the 'Set'.
+  , _nominalSpan :: a
+  -- ^ Data retention window
+  }
+  -- NOTE: deriving 'FromJSON' circumvents the protection offered by "smart constructors",
+  -- and the 'ToJSON' instance may expose internal details.
+  deriving (Eq, Show, Generic, FromJSON, ToJSON)
```
xsebek, line 71: A manually written ToJSON instance might be more understandable and FromJSON safer.

We have a lot of similar derived instances I imagine, but a TODO might fit better then a NOTE if merged.

xsebek (commented):
This looks cool! 🚀 Also, thanks @kostmo for the detailed documentation 👍 

I have not yet read through all of it, but I will try to find time to do so this week and also test this out on the scenarios.

src/Swarm/Util/WindowedCounter.hs
```diff
@@ -0,0 +1,135 @@
...
+-- In the worst case, the entire dataset may "age out" without being garbage collected,
+-- so that an @O(log n)@ filtering operation might be performed upon every "frame refresh"
+-- of the UI view.
+-- However, we also store the largest element of the window separately from the 'Set' so that
+-- we can compare against it for a @O(1)@ short-circuited path once every member ages out.
+data WindowedCounter a = WindowedCounter
+  { _members :: Set a
+  , _lastLargest :: Maybe a
+  -- ^ NOTE: It is possible that '_lastLargest' may not exist in the 'Set'.
+  , _nominalSpan :: a
```
byorgey, line 66: It seems strange to me that the type parameter `a` is used both for members and for the nominal span.  That seems like it will end up being restricted to only being able to store sets of numbers.  And if that's so, I don't know what we gain by being polymorphic in `a`.  It might be more general to have truly arbitrary values of some type `a`, indexed by tick number, something like
```
  { _members :: Map TickNumber a
  , _lastLargest :: Maybe TickNumber
  , _nominalSpan :: Integer
  }
```

src/Swarm/Util/WindowedCounter.hs
```diff
@@ -0,0 +1,135 @@
...
+-- The value inserted must always be at least as large as the
+-- current largest element of the set!
+-- If it is equal, it is ignored.
+--
+-- The 'discardGarbage' function is called from inside this function
+-- so that maintenance of the data structure is simplified.
+insert :: (Show a, Integral a) => a -> WindowedCounter a -> WindowedCounter a
+insert x statusQuo@(WindowedCounter s lastLargest nominalSpan)
+  | maybe False (x <) lastLargest =
+      error $
```
byorgey, line 112: I really, really don't want to have calls to `error` in the codebase, because it means that if there is a bug it could end up crashing the entire game, so someone could lose all their progress.  In this particular case, some alternatives might include (1) just ignore invalid inserts; (2) return a `Maybe (WindowedCounter a)` and check for the error condition elsewhere; (3) create some kind of "strict checking mode", enabled by a flag, where we do crash when encountering such a bug, and we can turn on that flag in our CI tests, but have it off by default (in which case we simply ignore invalid inserts as in option (1)).

src/Swarm/TUI/View.hs
```diff
@@ -627,6 +634,10 @@ robotsListWidget s = hCenter table
     , "Position"
     , "Inventory"
     , "Status"
+    , "Actions"
+    , "Commands"
+    , "Cycles"
+    , "Activity"
```
byorgey, line 640: This robots modal is getting very wide.  When I created a robot which happened to have a sort-of-long auto-generated name, it became so wide that the `Log` column was cut off.  Perhaps we can use abbreviations for some of these headers?  The contents of the columns is not very wide, it's just the headers.

Maybe something like  `Pos Inv Status Acts Cmds Cycs Activity Log` ?

src/Swarm/Util/WindowedCounter.hs
```diff
@@ -0,0 +1,135 @@
...
+--
+-- The data retention of the window shall be maintained externally by
+-- invoking the 'discardGarbage' function. However, we should not
+-- unconditionally invoke this function upon each game tick.
+--
+-- For efficiency, we do not want to iterate over every robot
+-- upon every tick; we only want to "visit" a robot if it is actually
+-- doing work that tick.
+-- Because of this, there may be some ticks in which the oldest element
+-- that is still stored falls outside of the nominal retention window.
```
kostmo, line 45: "... (because that robot is inactive)"

src/Swarm/Util/WindowedCounter.hs
```diff
@@ -0,0 +1,135 @@
...
+          ]
+  | Just x == lastLargest = statusQuo
+  | otherwise =
+      discardGarbage x $
+        WindowedCounter (Set.insert x s) (Just x) nominalSpan
+
+-- | Drop the leading elements that are not larger than the cutoff.
+--
+-- This function is already called by the 'insert' function, so clients
+-- no not necessarily every have to call this directly.
```
kostmo, line 127: "do not"

src/Swarm/Util/WindowedCounter.hs
kostmo: What do you think of the latest type parameterization, @byorgey?

src/Swarm/Util/WindowedCounter.hs
```diff
@@ -0,0 +1,179 @@
...
+  referenceTick = offsetBy (negate nominalSpan) currentTime
+  -- Cull the window according to the current time
+  WindowedCounter culledSet _ _ = discardGarbage currentTime wc
+
+-- | Invocations of this function shall be guarded externally
+-- by the conditions meant to be tracked in the window.
+--
+-- Proper usage dictates that the value inserted should always
+-- be at least as large as the current largest element of the set.
+-- If not, the insertion has no effect.
```
byorgey, line 137: This comment doesn't seem to correspond to what the code actually does.  It seems to me like inserting a value less than the current largest element does actually have an effect, since it gets added to the set (though the `lastLargest` field does not change).  I don't have a strong opinion about which behavior would be better.  I guess I lean towards keeping the current behavior (and changing this comment) since I see no particular reason to disallow "out of order" insertion.

src/Swarm/Util/WindowedCounter.hs
byorgey: It's better, though I guess I am still not sure what the polymorphism is buying us.  Do you have any concrete plans to instantiate `a` with anything other than a `TickNumber`?  As it stands, `a` still has to be `Offsettable` and `Ord` which pretty much limits it to be some kind of number.

byorgey (commented):
This is looking good!  I think this will give a lot of useful information when debugging robot programs.  I think there are just a few more things to clean up / resolve before merging.

- Do you have any idea why `idler1` only ever gets to 98.4% Activity?  Seems to me like it ought to be 100%.
- We should probably add a wiki page somewhere that explains in detail what each of the columns means, and link to it from somewhere (e.g. the bottom of the Robots panel).  In particular I imagine many players will be confused about the difference actions, commands, and cycles.  I don't care whether we do that in this PR or split it out into another issue and do it later.

src/Swarm/Util/WindowedCounter.hs
```diff
@@ -0,0 +1,179 @@
...
+  }
+  deriving (Eq, Show)
+
+-- Automatically deriving 'FromJSON' circumvents the protection offered by "smart constructors",
+-- and the 'ToJSON' instance may expose internal details.
+-- Therefore, we write our own custom implementations.
+instance ToJSON (WindowedCounter a) where
+  toJSON (WindowedCounter membersSet _lastLargest nominalSpan) =
+    object
+      [ "members" .= Set.size membersSet
```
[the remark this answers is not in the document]
kostmo: > Why do you write a `members` field but then not read it in the `FromJSON` instance?

Added a comment.  IMO, any creation of internal state should have to go through the smart constructor, as arbitrary JSON does not necessarily uphold the same invariants.

> Also, what's the justification for throwing away information about the members of the set and only writing its size?

At first, I was thinking that we shouldn't "reveal" information that is supposed to be "encapsulated".  But, since the `ToJSON` instance is supposed to be for diagnostics, I now expose the entire `Set`.

src/Swarm/TUI/View.hs
```diff
@@ -623,14 +634,41 @@ robotsListWidget s = hCenter table
...
+        -- a per-robot pop-up
+        str . show . sum . M.elems $ robot ^. activityCounts . commandsHistogram
+      , str $ show $ robot ^. activityCounts . lifetimeStepCount
+      , dutyCycleDisplay
       , txt rLog
       ]
+
+    dutyCycleAttrIdx = floor $ dutyCycleRatio * fromIntegral (length meterAttributeNames - 1)
+    -- Since (!!) is partial, here is "proof" that it is safe:
+    -- \* 'dutyCycleRatio' lies within the unit interval
```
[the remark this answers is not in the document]
kostmo: Introduced a `UnitInterval` type.

src/Swarm/Util/WindowedCounter.hs
kostmo: Corrected the comment.

src/Swarm/Util/WindowedCounter.hs
kostmo: > Do you have any concrete plans to instantiate `a` with anything other than a `TickNumber`?

No concrete plans, but suppose it is also useable on Enums that aren't numbers.

In our codebase, I want to ensure that the API is used directly with `TickNumber` types to preserve the "safety" of the `TickNumber` newtype "end to end".  However, I also don't think that this `Swarm.Util.WindowedCounter` module should depend on the definition of `TickNumber` (which currently lives in the `Swarm.Game.CESK` module).

Therefore, I think it makes sense to keep `WindowedCounter` polymorphic in `a`.

src/Swarm/Util/WindowedCounter.hs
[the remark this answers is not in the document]
byorgey: Ah, OK, that makes sense.  And it won't really matter if the `WindowedCounter`s are not preserved when saving + loading a game---it just means it will take a little bit for the activity% values for the robots to "warm up" again.

src/Swarm/Util/UnitInterval.hs
```diff
@@ -0,0 +1,24 @@
...
+  { getValue :: a
+  }
+
+-- | Guarantees that the stored interval falls on the range
+--   @[0, 1]@. It is up to clients to ensure that the promotion
+--   to this type is lossless.
+mkInterval :: (Ord a, Num a) => a -> UnitInterval a
+mkInterval = UnitInterval . max 0 . min 1
+
+safeIndex :: RealFrac a => UnitInterval a -> [b] -> b
```
byorgey, line 22: I love this!  Can you also copy the original comment arguing why this indexing operation was safe?

byorgey: Also, this should take a `NonEmpty b` instead of a `[b]`, otherwise it could crash when given an empty list.

byorgey (approved):
Looks great!

src/Swarm/Util/WindowedCounter.hs
kostmo: Option (3) sounds like #217?

PR #1487, opened 2023-09-04
Print REPL errors inline and get rid of error popup

### LINKED ISSUE #1461: Get rid of error popup and show errors inline in REPL
> In case you have invalid (can not parse/typecheck) REPL prompt and hit Enter, we show an error popup.
>
> I propose we instead save it in history and show the error like output. I.e. like GHCi does it.

_Originally posted by @xsebek in https://github.com/swarm-game/swarm/issues/713#issuecomment-1690685702_

Prerequisites: #60 , #11

byorgey:
Closes #1461.  Errors which used to be displayed in a pop-up window (parse errors, type errors) are now displayed inline in the REPL window instead.

- Get rid of the pop-up error dialog
- Also fix a bug introduced in #1481 where the REPL history would not be properly cleared when first starting a new scenario, because the old cache was still being used


xsebek (approved):
LGTM! 👍 It might be worth including in CHANGELOG as it affects the player experience.

Issue #281, opened 2021-11-06
Add benchmark artifacts


xsebek:
It would be nice to mark the benchmark logs as artifacts.

We could then write a script that would use GitHub CLI tool to download the artifacts of main branch and plot the benchmarks.

byorgey:
Since I didn't know what this meant, here's a link for the benefit of others who don't know either: https://docs.github.com/en/actions/advanced-guides/storing-workflow-data-as-artifacts

byorgey:
Marking the benchmark logs as artifacts sounds fine to me.  But as for plotting them, couldn't we just add some commands to the github action to plot the benchmarks right after generating them, and save the plot(s) as artifacts too?

xsebek:
If we install `gh` (may also need secrets) and some plotter in the container, then probably yes. 

If the actions do not measure performance well, then at least we would have a script to do it locally 🙂 

The only other useful artefact I can think of is the generated documentation, which could be cool to show off. 🤓

xsebek:
As a note, we do not run benchmarks in the CI. We could add it if we wanted to.
For local comparison, criterion-compare is quite nice:
```
cabal run benchmark -- --csv "main.csv" --output "main.html"
git switch feature
cabal run benchmark -- --csv "feature.csv" --output "feature.html"
criterion-compare main.csv feature.csv
```
<sup>For example see #415.</sup>

xsebek:
We do run haddock in the CI so it is only a matter of getting it from there to GitHub pages.

For example, Cabal seems to have been using GitHub pages and automatically pushing the Haddock documentation there:
- [Cabal Issue](https://github.com/haskell/cabal/issues/3543), [Commit](https://github.com/haskell/cabal/commit/0c254ba50440d3c6ce71cb3339b28f3d6de3a1de), [Repo](https://github.com/haskell/cabal-website), [Website with Haddock](http://haskell.github.io/cabal-website/doc/html/Cabal/)

But that is outdated now, so we should check if there is an easy way for Github Actions.

Worst case scenario we add a secret to allow pushing to a new swarm-website repository and manually do the clone/push after Haddock is built. This at least we could copy-paste from the linked commit.

byorgey:
https://github.com/agda/agda/pull/5115 does this for Agda using GitHub actions.

xsebek:
[...]

PR #1488, opened 2023-09-05
Some documentation improvements + refactoring

byorgey:
Some idle documentation improvements done on a plane.


xsebek (approved):
Thanks for writing those missing docs @byorgey, I see a couple that I left out. 👍  

I wish we had the docs for nightly build on a website now, these docs would be neat.

byorgey:
> I wish we had the docs for nightly build on a website now, these docs would be neat.

Yes, you have wished that before: #281 . :smile:   It would be a nice resource for pointing new contributors to.  I think I would find it helpful as well.

PR #1501, opened 2023-09-09
generalize variables bound by `<-`

### LINKED ISSUE #351: Generalize properly when inferring the type of a bind

byorgey:
Closes #351.

I sat down with @ccasin today to look at #351. He convinced me that it is actually fine to return `forall a. a -> a` as the type of `r` in `r <- return (\i.i)`, even though `return (\i.i)` has type `forall a. cmd (a -> a)` and we can never have a type `cmd (forall a. a -> a)`. Basically, in a Hindley-Milner-style system it is always safe to generalize at any time.

The real problem --- which I was very close to discovering in https://github.com/swarm-game/swarm/issues/351#issuecomment-1146274801 --- is simply that we need to generalize the type of variables bound by `<-`.  However, we need to make sure to do so before putting that variable in the context, which I apparently failed to do when I tried fixing this before.

### DIFF
--- a/src/Swarm/Language/Typecheck.hs
+++ b/src/Swarm/Language/Typecheck.hs
@@ -680,14 +680,9 @@ infer s@(Syntax l t) = addLocToTypeErr l $ case t of
   SBind mx c1 c2 -> do
     c1' <- withFrame l TCBindL $ infer c1
     a <- decomposeCmdTy c1 (Actual, c1' ^. sType)
-    -- We do not need to generalize the type 'a' here (unlike
-    -- top-level binds) because a variable bound here (mx) can only
-    -- have local scope.  In other words, when we generalize at the
-    -- very end, the variable will no longer be in the context, so we
-    -- will be able to safely generalize over any unification
-    -- variables that occurred in its type.
+    genA <- generalize a
     c2' <-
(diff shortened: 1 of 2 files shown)


xsebek (approved):
This fixes the bug for me! 👍 

While the code is already heavily annotated, this probably warrants a note. 🙂 
It would also be nice to have a unit test for this.

byorgey:
As a note for the future, we may possibly want to reconsider this at some point; see "Let Should Not Be Generalized" by Vytiniotis, Peyton Jones, and Schrijvers ( https://www.microsoft.com/en-us/research/wp-content/uploads/2016/02/tldi10-vytiniotis.pdf ).  But I think this will be fine for now.

Issue #1494, opened 2023-09-07
Ban (or restrict) partial functions with hlint


kostmo:
Ideally we don't want any instances of `error` in the codebase (https://github.com/swarm-game/swarm/pull/1484#discussion_r1315211065).  For similar reasons, we should restrict use of partial functions like `(!!)`, `fromJust`, and `head` to a vetted subset of modules.  `hlint` can [whitelist where certain functions may be used](https://github.com/ndmitchell/hlint/blob/ed1259a7da88420e8d05d6241d6bdd4493a9997f/README.md?plain=1#L383).

- [X] `NonEmpty.fromList` (#1503)
- [X] `Prelude.head` (#1503)
- [x] `Prelude.tail`
- [x] `(!!)`
- [ ] `(Data.Map.!)`
- [x] `(Data.IntMap.!)` (#2294)
- [x] `error` (#2431)
- [x] `fromJust`
- [x] `undefined`

byorgey:
It looks like we don't have any uses of `fromJust`, so we could just add that to the `hlint` config.

xsebek:
There may be other more obscure functions - for example in tests, I use the sneaky `(^?!)`. 🙂

byorgey:
I'm not as concerned about tests, since those don't directly affect users if they crash.

byorgey:
Note there are currently 3 uses of `Data.Map.!` in the codebase:
```
src/swarm-engine/Swarm/Game/ScenarioInfo.hs:139:51-53: Warning: Avoid restricted function
Found:
  M.!
Note: may break the code

src/swarm-lang/Swarm/Language/Context.hs:346:59-61: Warning: Avoid restricted function
Found:
  M.!
Note: may break the code

src/swarm-lang/Swarm/Language/Typecheck.hs:1301:80: Warning: Avoid restricted function
Found:
  !
Note: may break the code
```

PR #1503, opened 2023-09-10
Replace/quarantine uses of `head` and `NE.fromList` functions

kostmo:
Towards #1494

* Totally eliminates partial `head`
* introduces an hlint rule to ban unsafe use of `head`
* restricts use of the partial `fromList` from `NonEmpty`

### DIFF
--- a/src/Swarm/Doc/Gen.hs
+++ b/src/Swarm/Doc/Gen.hs
@@ -24,6 +24,7 @@ module Swarm.Doc.Gen (
 ) where
 
 import Control.Effect.Lift
+import Control.Effect.Throw (throwError)
 import Control.Lens (view, (^.))
 import Control.Lens.Combinators (to)
 import Control.Monad (zipWithM, zipWithM_)
@@ -32,7 +33,7 @@ import Data.Foldable (find, toList)
 import Data.List (transpose)
 import Data.Map.Lazy (Map, (!))
 import Data.Map.Lazy qualified as Map
-import Data.Maybe (fromMaybe, isJust)
+import Data.Maybe (fromMaybe, isJust, listToMaybe)
 import Data.Sequence (Seq)
 import Data.Set (Set)
 import Data.Set qualified as Set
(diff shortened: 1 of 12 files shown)


src/Swarm/Doc/Gen.hs
```diff
@@ -419,6 +420,11 @@ recipeTable a rs = T.unlines $ header <> map (listToRow mw) recipeRows
 recipePage :: PageAddress -> [Recipe Entity] -> Text
 recipePage = recipeTable
 
+getBaseRobot :: Scenario -> Either Text Robot
```
byorgey, line 423:
```suggestion
getBaseRobot :: Has (Throw SystemFailure) sig m => Scenario -> m Robot
```
Should be able to use this signature, then in the `Nothing` case you can `throwError $ CustomFailure "Scenario...`.  Then you don't need the `either (throwError . CustomFailure) return` below in `generateRecipe` at all.

src/Swarm/Game/Scenario/Topography/Cell.hs
```diff
@@ -69,10 +69,12 @@ instance ToJSON Cell where
 
 instance FromJSONE (EntityMap, RobotMap) Cell where
   parseJSONE = withArrayE "tuple" $ \v -> do
-    let tup = V.toList v
-    when (null tup) $ fail "palette entry must nonzero length (terrain, optional entity and then robots if any)"
+    let tupRaw = V.toList v
+    tup <- case NE.nonEmpty tupRaw of
+      Nothing -> fail "palette entry must nonzero length (terrain, optional entity and then robots if any)"
```
byorgey, line 74:
```suggestion
      Nothing -> fail "palette entry must have nonzero length (terrain, optional entity and then robots if any)"
```

src/Swarm/Util/Parse.hs
```diff
@@ -17,3 +21,6 @@ fully sc p = sc *> p <* eof
 --   ensuring that the parser extends all the way to eof.
 fullyMaybe :: (MonadParsec e s f) => f () -> f a -> f (Maybe a)
 fullyMaybe sc = fully sc . optional
+
+sepByNE :: (MonadPlus m) => m a -> m sep -> m (NonEmpty a)
```
byorgey, line 25: We should just use `sepBy1` from https://hackage.haskell.org/package/parser-combinators-1.3.0/docs/Control-Monad-Combinators-NonEmpty.html .

byorgey (approved):
This is great!

Issue #1435, opened 2023-08-19
Not all achievements are achievable


byorgey:
Unless I am mistaken, I think there are some achievements listed in the Achievements menu which are not actually achievable:

- [x] `CompletedAllTutorials`
- [x] #2351 
- [x] `CraftedBitcoin` (#1688)
- [x] `RobotIntoWater` (#1504)
- [x] `DestroyedBase` (#2349)

xsebek:
We could make a integration test checking that given achievement is achievable. 🤔 (Is there a way to depend on other tests so we don’t have to retest all tutorials?)

Imagine if we ended up with achievement descriptions with solution like we have for scenarios. 😄

byorgey:
Note, it will not be easy to test all achievements since some may require doing certain things in the menu and so on.  But others should certainly be possible to test with a solution.

byorgey:
Does anyone know if there is actually any way to destroy the `base`?  It's actually rather difficult/impossible to do, even in creative mode.  We should either (1) get rid of the achievement for destroying the base, or (2) make it possible to do (via some appropriately obscure method).  Or perhaps (3) it is already possible and I just haven't figured out how.

xsebek:
@byorgey I don't of any way to destroy base now - if you wanted, we could make `unequip` do the check that move does, so that unequiping a boat while on water would sink the base. 

Placing water should probably be check too in that case, though it might be better to not drown this way as it is more common.

byorgey:
Perhaps placing water should simply do nothing.  Like, you can `grab` water from a lake, but if you pour that water on the ground somewhere else it doesn't create a lake.

byorgey:
I think they are all achievable now!

PR #1504, opened 2023-09-10
Grant `RobotIntoWater` achievement

byorgey:
Towards #1435.  Some refactoring, + grant `RobotIntoWater` achievement when a robot dies in the water.

### DIFF
--- a/src/Swarm/Game/Step.hs
+++ b/src/Swarm/Game/Step.hs
@@ -1024,7 +1024,7 @@ execConst c vs s k = do
         return $ Waiting (addTicks d time) (Out VUnit s k)
       _ -> badConst
     Selfdestruct -> do
-      destroyIfNotBase $ Just AttemptSelfDestructBase
+      destroyIfNotBase $ \case False -> Just AttemptSelfDestructBase; _ -> Nothing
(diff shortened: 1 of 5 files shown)


kostmo:
Heads up on the build error: #1505

src/Swarm/Game/Step.hs
kostmo, line 1: It would be great to have an integration test for this (see #1507), but in lieu of such a framework, can you add a "test plan" to the toplevel PR description?

byorgey: @kostmo good idea.  How's this?

xsebek:
If the achievements are part of GameState, then you can test it in `testSolution'` - the winning conditions can be something like:
1. build a robot (robotNumbered 1 exists)
2. drown it (robotNumbered 1 does not exist)

byorgey:
Good idea, done!

src/Swarm/Game/Step.hs
byorgey: Following a suggestion from @xsebek I added an actual integration test.

xsebek (approved):
🏊‍♂️

Issue #872, opened 2022-11-20
Group related fields of GameState into child records


kostmo:
`GameState` is a very "wide" record.  Currently it has 34 fields.  Several of those fields are closely related to each other, but unrelated to the remaining fields.

E.g.:
```
  , _recipesOut :: IntMap [Recipe Entity]
  , _recipesIn :: IntMap [Recipe Entity]
  , _recipesReq :: IntMap [Recipe Entity]
```
and
```
  , _seed :: Seed
  , _randGen :: StdGen
```
and
```
  , _viewCenterRule :: ViewCenterRule
  , _viewCenter :: V2 Int64
```
and
```
  , _messageQueue :: Seq LogEntry
  , _lastSeenMessageTime :: Integer
```

I propose that we create new records for each of the groupings above (and perhaps more).  Each of new records will become consolidated fields of `GameState`.

xsebek:
I don’t think we _need_ seed in game state.

AFAIK its used to initialise the generator and to show to the player in UI. We could realistically save it in one of our other records, namely it is already present in the scenario.

xsebek:
Message queue needs a small rewrite to create a more efficient Message panel. I would suggest leaving that as is for now and maybe adding a TODO note.

I completely agree with refactoring recipes and in general making the game state smaller. 👍

xsebek:
Related Issue:
- #707

byorgey:
@kostmo Is there still more decomposition of `GameState` you want/intend to do?  Or can we close this now?

PR #1510, opened 2023-09-10
Decompose GameState into sub-records

kostmo:
Towards #872

Previously, the `GameState` record had `41` toplevel members.  It now has `22`.   Logical grouping of the fields makes it easier to peruse the code and paves the way to split `State.hs` into smaller modules.  Some functions now may even be able to operate exclusively on subsets of the game state, rather than having to pass in `GameState` as a whole.

There is potential to go even farther, by extracting view-related and robot-related members to their own records, but I figured I'd pursue this refactoring incrementally.

### DIFF
--- a/bench/Benchmark.hs
+++ b/bench/Benchmark.hs
@@ -16,7 +16,7 @@ import Swarm.Game.CESK (emptyStore, initMachine)
 import Swarm.Game.Display (defaultRobotDisplay)
 import Swarm.Game.Location
 import Swarm.Game.Robot (TRobot, mkRobot)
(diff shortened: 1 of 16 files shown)


byorgey (approved):
This is great!  It's a good idea in general, and I like the subcategories you have chosen to organize things.  There are just two improvements I'd like to see:
* New subrecord types like `Temporal`, `Landscape`, etc. should be exported (just the types, not their constructors or record fields)
* We should put some effort into organizing the generated Haddock documentation.  At the very least we should put each of the new subrecords + their lenses under a separate subheading in the export list.

PR #1512, opened 2023-09-10
update CONTRIBUTING guide, and remove info from README

byorgey:
All README info is now on the website.


xsebek (approved):
I don't think we should gut README so much - it's useful offline and Github websites can also have outages.

I would suggest keeping at least the installation section.

Since it wont't affect me mostly installing from source, I will leave that to you. 😄

kostmo:
> I'd rather have only a single "source of truth"

I also tend to weight this more heavily.  An offline-available doc is less valuable if it is not enforced to be in sync.

How does the website deployment work?  Can a copy of it be downloaded and viewed locally?

README.md
```diff
@@ -22,152 +22,7 @@ Swarm
...
-[#1000](https://github.com/swarm-game/swarm/issues/1000)).  We
-recommend either building Swarm with a different version of GHC
-(*e.g.* 9.4.x), or building with optimizations turned off (which does
-not seem to affect the game performance very much).
-
-**NOTE**: Swarm requires a POSIX-style terminal environment that
-supports `terminfo`.  Linux and MacOS should work out of the box.  On
-Windows, you will need to use [Windows Subsystem for
-Linux](https://learn.microsoft.com/en-us/windows/wsl/); you should
-then be able to follow instructions for installing on Linux.
```
xsebek, line 67: Specifically these two notes make more sense to me in swarm repo rather than the website repo.

I can’t imagine going to a website to check if there are any notes for compiling with specific GHC versions.

xsebek:
@byorgey we could setup this action in main branch that would keep Markdown files in sync:

https://github.com/marketplace/actions/push-a-file-to-another-repository

byorgey:
> we could setup this action in main branch that would keep Markdown files in sync:

That's nifty, but it seems like more trouble than it's worth.  We probably won't want the files to have exactly the same format, etc. etc.

README.md
byorgey: Yes, I agree those notes feel like maybe they don't belong on the website.  How about I put all the instructions for building from source in the repo, and leave the other stuff on the website (with appropriate links)?  Let me try that and see how it looks.

byorgey: @xsebek What do you think now?

Note, I simply removed the note about GHC 9.2.  By now it's pretty old, and since we're recommending building with `stack` (which will now get GHC 9.4) no one would run into this issue unless they (1) want to use their own custom workflow for building via `cabal` or whatever, AND (2) are still using GHC 9.2, which seems like a very unlikely combination.

Issue #1483, opened 2023-08-31
generate warnings from within the typechecker


kostmo:
This is a pre-requisite for LSP support to warn on name shadowing.  See https://github.com/swarm-game/swarm/issues/1338#issuecomment-1602660192.

Issue #1039, opened 2023-01-24
Logging rework


xsebek:
From `src/Swarm/Game/Log.hs`:

> Currently, we abuse this system for system logs, which is fun, but we should make a dedicated `SystemLogEntry` type for the `RuntimeState` message queue.

Our logging needs are slowly growing beyond pretending that each message was logged by some hypothetical or system robot.

It should also allow us to use structured error types and export logs through Web API.

Related issues:
- #629
- #1021

xsebek:
Runtime state has proper logs now, so closing.

PR #1513, opened 2023-09-10
Refactor `LogEntry` type

byorgey:
In preparation for #1483.   `LogEntry` started life as something specific to robot logs.  It then evolved to be used in the system log as well (see #1039 and #652), but in a sort of hacky way.  This PR refactors `LogEntry` to be more generic.

- Move `Swarm.Game.Log` -> `Swarm.Log` since it's not specific to gameplay.
- Rename `ErrorLevel` to `Severity`, add a new `Info` level, and add a top-level `leSeverity` field
- Rename `leRobotName` to just `leName`, since it was already being used to name both robots and system components anyway
- Move robot-specific fields (*e.g.* robot ID) into the new `RobotLogSource` type, and add `LogSource` to differentiate between robot and system logs
- Various other minor improvements and tweaks

### DIFF
--- a/src/Swarm/Log.hs
+++ b/src/Swarm/Log.hs
@@ -4,18 +4,8 @@
 -- |
 -- SPDX-License-Identifier: BSD-3-Clause
 --
--- XXX Also make 'Notification' documentation more generic
---
--- A data type to represent in-game logs by robots.
---
--- Because of the use of system robots, we sometimes
--- want to use special kinds of logs that will be
--- shown to the player.
---
--- TODO: #1039 Currently we abuse this system for system
--- logs, which is fun, but we should eventually make
--- a dedicated `SystemLogEntry` type for 'RuntimeState'
--- message queue.
+-- A data type to represent log messages, both for robot logs and
+-- the system log.
 module Swarm.Log (
   Severity (..),
   RobotLogSource (..),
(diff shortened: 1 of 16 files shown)


src/Swarm/Log.hs
```diff
@@ -39,35 +30,44 @@ import Swarm.Game.Universe (Cosmic)
...
-  , _leRobotID :: Int
-  -- ^ The ID of the robot that generated the entry.
+  -- ^ Where this log message came from.
+  , _leSeverity :: Severity
+  -- ^ Severity level of this log message.
+  , _leName :: Text
+  -- ^ Name of the robot or subsystem that generated this log entry.
   , _leLocation :: LogLocation (Cosmic Location)
-  -- ^ Location of the robot at log entry creation.
+  -- ^ Location associated with this log message.
```
xsebek, line 70: Maybe we should move location to robot log source. 🤔

xsebek: @byorgey I would say those are different things - system message can talk _about_ a location that is not its own.

If we need to show message only to certain location, using a robot seems like the natural (and in scenarios easier) way to do it.

Either way is fine, but I suspect it will feel weird having to set default location each time a system logs something.

byorgey: OK, I buy that.  I've updated it to put the location exclusively in robot log messages.

PR #1518, opened 2023-09-12
Add syntax highlighting and LSP configuration for Vim/Neovim

gaganchandan:
This adds two files. The first is the syntax file responsible for basic highlighting. The highlight categories are based roughly on those seen in the Emacs file ([swarm-mode.el](https://github.com/swarm-game/swarm/blob/main/editors/emacs/swarm-mode.el)). The second is the file for configuring the language server. Since it is based on Neovim's native LSP client, it only works with Neovim and not Vim.

`README.md` in the `editor` folder has also been updated to include instructions for setting these up.

### DIFF
--- a/editors/README.md
+++ b/editors/README.md
@@ -34,5 +34,20 @@ To configure YAML editor tabs for schema validation, install the [YAML plugin](h
 
 ## Vim and Neovim
 
-Currently there is neither highlighting nor LSP support for Vim,
-but we would be happy to [accept a contribution](../CONTRIBUTING.md).
+Add the following lines to your Vim/Neovim configuration file for files with the `.sw` extension to be recognized as `swarm` programs:
+
+
+`init.vim`:
+
+`au BufRead,BufNewFile *.sw setfiletype swarm`
+
+
+`init.lua`:
+
+`vim.cmd[[au BufRead,BufNewFile *.sw setfiletype swarm]]`
+
+
+Basic syntax highlighting is available for both Vim and Neovim. To make use of this capability, copy [swarm.vim](vim/swarm.vim) to the `syntax` directory in your Vim or Neovim configuration directory.
+
+
+An LSP configuration leveraging Neovim's native LSP client is also available. It only works with Neovim. To enable it, copy [swarm.lua](vim/swarm.lua) to `after/ftplugin` in your Neovim configuration directory.
--- a/editors/vim/swarm.lua
+++ b/editors/vim/swarm.lua
@@ -1,0 +1,8 @@
+if vim.fn.executable('swarm') == 1 then
+    vim.lsp.start({
+        name = 'Swarm Language Server',
+        cmd = { 'swarm', 'lsp' },
+    })
+end
+
+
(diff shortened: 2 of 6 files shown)


byorgey:
@gaganchandan thanks so much!  I can't really verify whether this works but it looks good to me overall.

There are two other related things we'll want to update:
- A couple places in our documentation (*e.g.* in the README) we mention which editors are supported, so we'll want to update that---however, we are currently (#1512) reshuffling some of that information so don't worry about that, we'll update it once things settle a bit.
- We have some CI tests to ensure that our editor configurations don't get out of date when we add new keywords, commands, etc.  See here for the tests:
    https://github.com/swarm-game/swarm/blob/bb31126c7541871f3c497a992cfa8918f0e7f8cd/test/integration/Main.hs#L439
    and see here for the code to generate the configurations used in the tests:
    https://github.com/swarm-game/swarm/blob/bb31126c7541871f3c497a992cfa8918f0e7f8cd/src/Swarm/Doc/Gen.hs#L147
    Updating these is *not* a prerequisite for merging this PR.  If you think you would have fun updating the tests then by all means go for it.  But if you would be more comfortable having someone else do it, that's totally fine too.  Just let us know!

gaganchandan:
> A couple places in our documentation (e.g. in the README) we mention which editors are supported, so we'll want to update that

Should I remove the changes made to the README from the PR?

About the tests, I took a look and it seems doable for me. Can I ask here if I run into any issues, or would you prefer a different medium, say, the IRC?

byorgey:
> Should I remove the changes made to the README from the PR?

No, you only changed `editors/README`, I was talking about the top-level `README`.

> About the tests, I took a look and it seems doable for me. Can I ask here if I run into any issues, or would you prefer a different medium, say, the IRC?

Great!  Either one is fine.  Asking here makes sense since the questions and answers will be archived and more easily discoverable for, say, someone adding support for additional editors in the future.  But IRC is fine too, e.g. if you have quick questions that seem like they don't belong here.

byorgey:
> they are failing here because `editors/vim/swarm.vim` does not yet exist on the `main` branch.

Hmm, that shouldn't make a difference.  I suspect the tests are actually failing because `editors/vim/swarm.vim` needs to be mentioned in `swarm.cabal`.  Currently, that file is not included in the distribution tarball, hence it does not exist when the tarball is unpacked into the test environment.  Try adding `editors/vim/swarm.vim` (or `editors/vim/*.vim`, or whatever pattern(s) you think makes sense) to the `extra-source-files:` field in `swarm.cabal`.

byorgey:
@gaganchandan Should we add `swarm.lua` to the `extra-source-files` as well?  If someone does `cabal install swarm` they are only going to get the files mentioned in `swarm.cabal`.  I am not sure what `swarm.lua` is for but if we want users to be able to easily copy or refer to it, we might want to include it.

gaganchandan:
Yes, it does make sense to include it. I'll just add it real quick. Are there any more changes you would like?

byorgey (approved):

byorgey:
No, this looks great, thanks!  The CI failure looks spurious, I'm re-running that job now and hopefully this should merge once the CI passes.

byorgey:
@gaganchandan , congrats on having your first PR accepted to Swarm!   We appreciate the contribution and we're really glad to welcome you as part of the [community](https://github.com/swarm-game/swarm/blob/main/COMMUNITY.md).  In a minute we will send you an invite to the repo (see https://github.com/swarm-game/swarm/blob/main/CONTRIBUTING.md#i-have-push-access-to-the-swarm-repository-now-what ).  If you use IRC, feel free to also join the `#swarm` channel on Libera.Chat.

gaganchandan:
Thank you so much!

xsebek:
@gaganchandan thanks a lot, I just set it up it and it works great! 🥳

Screenshot 2023-09-16 at 18 22 48

PR #1527, opened 2023-09-16
Update Web module haddock

xsebek:
* update Haddock documentation in Web module
* add explicit export lists
* move handlers to top level so the code gets more spaced out


byorgey (approved):
LGTM.

PR #1529, opened 2023-09-16
Web API docs improvements

xsebek:
* generate Web API endpoint docs in CLI - `swarm generate endpoints`
* use default port in API docs using `renderCurlBasePath`
* fix code rendering in Web API so that the tree nodes try to fit one line
* try adding some API samples
* hand craft `ToJSON Robot` instance to match `FromJSONE` more closely
  * default values are skipped
  * inventory and devices are shortened to names and counts

### DIFF
--- a/src/Swarm/Web.hs
+++ b/src/Swarm/Web.hs
@@ -24,10 +24,12 @@
 module Swarm.Web (
   startWebThread,
   defaultPort,
+
   -- ** Docs
   SwarmAPI,
   swarmApiHtml,
   swarmApiMarkdown,
+
   -- ** Development
   webMain,
 ) where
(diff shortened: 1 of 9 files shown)


src/Swarm/Web.hs
```diff
@@ -51,14 +58,15 @@ import Network.Wai.Handler.Warp qualified as Warp
 import Servant
 import Servant.Docs (ToCapture)
 import Servant.Docs qualified as SD
+import Servant.Docs.Internal qualified as SD (renderCurlBasePath)
```
xsebek, line 61: This should not be a problem as the record field is reexported normally. I made a PR to eventually fix that:
* haskell-servant/servant#1706

byorgey (approved):
Looks good to me, modulo the failing CI (looks like a couple forgotten todos?).

PR #1530, opened 2023-09-16
'Be' command for unprivileged 'As'

kostmo:
Not sure if this is needed yet.


xsebek:
@kostmo I think I like the alternative of unprivileged `as` - could you please create an issue? I am not sure how the alternatives would work exactly. 🤔

byorgey:
What would the hypothetical `privileged`/`god` commands be good for besides wrapping `as`?  If it already requires God capability to use it (as it should) then it doesn't make any difference unless running commands as someone else.

Issue #1547, opened 2023-09-21
`path` type mentions `entity` which is not a real type


byorgey:
**Describe the bug**
The `Typecheck` module contains `Path -> [tyQ| (unit + int) -> ((int * int) + entity) -> cmd (unit + dir) |]`, but we do not have any built-in type named `entity` (not yet at least; see #455), so it is interpreted as a type variable.  Writing `path` at the prompt displays the type
```
∀ a. unit + int -> int * int + a -> cmd (unit + dir)
```
where you can see that the `entity` has been replaced with a generic quantified type variable `a`.  However, the `path` command is expecting a `text` value, so we can use this to cause a fatal crash by typing e.g. `path (inl ()) (inr 3)` (see the below screenshot).

**Expected behavior**
Presumably the type of `path` should be `(unit + int) -> ((int * int) + text) -> cmd (unit + dir)`.

**Screenshots**
![path-type-bug](https://github.com/swarm-game/swarm/assets/533859/7f123795-a93e-41a3-9c09-679892d72172)

xsebek:
@byorgey will these type variables be already renamed or could we test that their names are reasonable (a, b, c,…)?

byorgey:
@xsebek you mean as a test to make sure this kind of mistake doesn't happen again?    We could test them right after parsing.  *e.g.* perhaps we could add a flag to the parser for types whether it should allow type variable names other than single lowercase letters (since we would not want to restrict what type variables users can write).

We might also want to consider making built-in type names start with an uppercase letter (`Int`, `Text`, etc.).

xsebek:
@byorgey sounds good - could we apply that flag in all our Template Haskell splices? There might be some other place where we write types that would have the same issue. 🤔 

Since swarm is case insensitive I am not sure using uppercase would help or cause more issues. It might be easier to read and would look more like Haskell though.

byorgey:
> sounds good - could we apply that flag in all our Template Haskell splices?

Sure, I'll put together a PR fixing this issue and also adding that check.

Good point re: swarm being case insensitive, I had forgotten that.  So it seems we have 3 choices that I can think of:
1. Leave things as they are.
2. Update all our documentation and examples to use titlecase for types, to recommend it as good style, but keep the parser case insensitive.
3. Change the parser so type names must start with uppercase and type variables must start with lowercase, as in Haskell.

[...]

PR #1550, opened 2023-09-26
Require explicit `forall` on polymorphic types

byorgey:
Through the discussion on #1547 I realized that we have two features which don't play well together:

1. Case-insensitive parsing
2. Implicit quantification of type variables

The problem is that if there's no syntactic way to distinguish type variables from type constructors, then if you write down a type constructor that doesn't exist, or isn't in scope, or if you just misspell the name of a type constructor, it will silently be implicitly quantified as a type variable.  Fortunately it was a developer that first tripped over this in #1547 but players could very easily trip over this as well when writing their own code with explicit type signatures.

Haskell doesn't have this problem since it has (2) but not (1): type constructors must start with an uppercase letter, and type variables with lowercase.

This PR implements the opposite solution: keep (1) but do away with (2), *i.e.* require all type variables to be explicitly introduced with a `forall` (or `∀`).  But I'm open to discussion of other ideas/solutions.

Independently, we might also decide, as an issue of style, to encourage writing types with initial capitals, and to update all our tutorials etc. to use this style.  But that would be a bigger PR and I decided to leave that for future discussion.


data/scenarios/Tutorials/conditionals.yaml
```diff
@@ -25,28 +25,28 @@ objectives:
         it is simply a built-in function of type
       - |
         ```
-        if : bool -> {a} -> {a} -> a
+        if : ∀ a. bool -> {a} -> {a} -> a
```
xsebek, line 28: Is this syntax introduced before? If players did not know the `forall`/`∀` then this would be confusing.

xsebek (approved):
Long term I think the more common implicit variables and upper case types would be better.

But short term this prevents errors and we don't actually need explicit types all that often. 👍

byorgey:
> Long term I think the more common implicit variables and upper case types would be better.

Well, if we would all prefer that option I'd rather just bite the bullet and do it now (I don't mind doing the work to do the conversion). I put together this PR since I wasn't sure how others felt about the various options, but I think I have a mild preference for implicit quantification and uppercase types as well.  @kostmo, what do you think?

Issue #1558, opened 2023-09-30
Optimize robot insertion in tick


xsebek:
The performance analysis in #1557 showed that 5% of time will be spent inserting running robots in the map in each tick.

| Inspecting profile with [Profiteur](https://jaspervdj.be/posts/2014-02-25-profiteur-ghc-prof-visualiser.html) |
| - |
| <img width="1508" alt="Screenshot 2023-09-30 at 14 47 50" src="https://github.com/swarm-game/swarm/assets/44544735/47bbaba0-0e4f-4211-aacf-32db24751781"> |

## Proposed solution

Lets consider a more performant alternatives, like:

_No inserting._

The idea would be to:
* store robots in a _immutable_ vector
* make a mutable copy
* run the tick using `ST`/[`MVector`](https://hackage.haskell.org/package/vector-0.13.0.0/docs/Data-Vector.html#t:MVector)/... updating robots in place
* end with a nice immutable state again
* profit
```Haskell
-- thaw the vector on each tick, updates to index map will be less frequent, use unsafeFreeze when done
data RobotMap = RobotMap { robotMap :: IntMap, robotStore:: Vector Robot } 
```

xsebek:
Actually, I just noticed this is about a different new insert for `WindowedCounter`! So robot are not the issue. 😄

@kostmo any idea why insert is taking so much time? (It goes down to 1% after optimisations though.)

byorgey:
The `WindowedCounter` is used to keep track of ticks on which a robot actually did something, so we can show statistics about the robot's "load" (i.e. percentage of ticks on which it is active) in the F2 dialog.

What if we *only* update `WindowedCounter` when the F2 dialog is actually open?  The downside is that it would take the activity values a little bit of time to "warm up" and stabilize after opening the dialog instead of just having the values immediately available, but it should save a lot of time in the common case that we are not computing activity percentages.

I also wonder whether using an `IntSet` instead of a generic `Set` in `WindowedCounter` would help with performance.

kostmo:
> I also wonder whether using an `IntSet` instead of a generic `Set` in `WindowedCounter` would help with performance.

I think then that the underlying type of `TickNumber` would have to become `Int` rather than `Integer`.  I actually experimented with that randomly in #1491.  Any reason it would need to remain as `Integer`?

xsebek:
@kostmo as long as it is 64 bit it should be enough to store time, I am not sure what the situation is on Windows.

[...]

PR #1562, opened 2023-10-01
Limit system robot activity tracking

kostmo:
Extension of #1484 that partially mitigates #1558.

Let's not bother to track the activity levels of system robots when we're not in creative mode, because they won't even show up in the `F2` dialog.  This mitigation can be substantial, as we generally expect there to be many more system robots than player robots.

It doesn't have quite the same "warm up" drawback as the potential mitigation described in https://github.com/swarm-game/swarm/issues/1558#issuecomment-1741794123, because player robots will always be warmed up, and system robots will have been warmed up if we've been playing in creative mode.  We wouldn't necessarily expect frequent toggling in-and-out of creative mode while trying to observe performance in the `F2` dialog.

### DIFF
--- a/src/Swarm/Game/CESK.hs
+++ b/src/Swarm/Game/CESK.hs
@@ -98,7 +98,7 @@ import Swarm.Language.Requirement (ReqCtx)
 import Swarm.Language.Syntax
 import Swarm.Language.Types
 import Swarm.Language.Value as V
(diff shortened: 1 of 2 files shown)


xsebek (approved):
LGTM! If we come up with a better condition we can just plug it to `shouldTrackActivity`. 👍

byorgey (approved):

PR #1564, opened 2023-10-02
more partial function bans

kostmo:
Towards #1494.

Replaced/restricted uses of `Prelude.tail` and `Prelude.!!`.  Quarantined `undefined`.

Introduced a new function `listEnumsNonempty` that is guaranteed safe.


byorgey (approved):
Looks great!

PR #1565, opened 2023-10-02
Use IntSet for WindowedCounter

kostmo:
Towards #1558.

### DIFF
--- a/src/Swarm/Game/CESK.hs
+++ b/src/Swarm/Game/CESK.hs
@@ -99,7 +99,7 @@ import Swarm.Language.Requirement (ReqCtx)
 import Swarm.Language.Syntax
 import Swarm.Language.Types
 import Swarm.Language.Value as V
-import Swarm.Util.WindowedCounter (Offsettable (..))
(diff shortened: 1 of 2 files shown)


byorgey (commented):
So what would happen on a system with a 32-bit `Int` type which is left running long enough to cause overflow?

kostmo:
> So what would happen on a system with a 32-bit `Int` type which is left running long enough to cause overflow?

Hmm... I suppose the `IntSet` might start growing without bound.  May be worth catching that edge case and writing a new unit test.

xsebek:
@kostmo could you close the window at maxBound of Int? Afterwards it can start clean from minBound and comparison should always make sense.

If maxBound is evenly divisibly by window size, then it will even be correct.

byorgey:
We could also just leave it as `Integer`, since we know that will be correct, and we're already doing other things to mitigate the performance impact of the `WindowedCounter`.  I don't know how much of a difference using an `IntSet` will make in practice.

kostmo:
Closing this for now, since it's not certain that it is needed.

PR #1566, opened 2023-10-03
Restore the "classic world" background for `friend` challenge

byorgey:
This was broken by #1376.  Any challenge which did *not* specify a `default` field would have had the classic world as an implicit background, but in this case I failed to add the proper `dsl` field to specify it.  I don't *think* there are any other such challenge scenarios, although to be really sure we would have to identify all the scenarios did not specify a `default` field before #1376 was merged and ensure that each of them now has an appropriate `dsl` field.

Before:

before

After:

after

### DIFF
--- a/data/scenarios/Challenges/friend.yaml
+++ b/data/scenarios/Challenges/friend.yaml
@@ -88,6 +88,8 @@ world:
     '*': [grass, flower]
     '@': [stone, boulder]
   upperleft: [-20, 2]
+  dsl: |
+    "classic"
   map: |-
     c,..,,,,,,..,,,,...,.
     ,..,,,,,,,...........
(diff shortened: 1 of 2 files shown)


xsebek (approved):
Before #1376, these scenarios did not have default specified:
```bash
find -name '*.yaml' -exec echo -n {} "  " \; -exec yq '.world.default' {} \; | grep '  null'
# ./classic.yaml   null
# ./Challenges/gopher.yaml   null
# ./Challenges/friend.yaml   null
# ./creative.yaml   null
# ./Tutorials/world101.yaml   null
# ./Tutorials/backstory.yaml   null
# ./Tutorials/farming.yaml   null
# ./Speedruns/forester.yaml   null
# ./Speedruns/curry.yaml   null
# ./Speedruns/mithril.yaml   null
```
It looks like only `gopher` and `friend` are missing a DSL now.

PR #1567, opened 2023-10-03
set allow-lists for remaining partial functions

kostmo:


xsebek:
@kostmo could we do this inline? For example if the function changes will HLint say that it no longer requires this or for example that it is missing?

PR #999, opened 2023-01-10
word search

kostmo:
Inspired by [this Reddit post](https://www.reddit.com/r/mildlyinfuriating/comments/108ba5i/dog).
image

The mechanics are:
* You are "caged" while a system robot arranges the board.  It lets you out when finished.
* You have a finite amount of `"ink"` to mark (`drill`) the three solution letters.
* The letters on the board are not placed completely randomly.  One can peek at the header of `create-puzzle.sw` for a spoiler.

I include a general solution in `solution.sw`, but for the purpose of speeding up the integration tests, a solution hard-coded to the given seed is used.

PR #1579, opened 2023-10-09
Structure browser and recognizer

### LINKED ISSUE #1575: Structure recognizer
There already exists a mechanism to define structures at the scenario level (#1332).

It would make an interesting game mechanic to have predefined structures that the player robot can assemble out of raw materials (i.e. `place` on the ground according to the structure template), and then be automatically recognized when complete, either to achieve a scenario goal or confer some ability/privilege.

In a sense, these would be like "recipes" but with a spatial component.  A subset of the structure template definitions in a scenario file could be marked as "participating" in recognition, perhaps by virtue of associating a "reward" or "output".

# Philosophical considerations

As a programming game, a particularly appealing aspect of Swarm is the spatial dynamism of seeing one's programs affect the world.  

Besides programming, Swarm gameplay involves "spatial activity" and "stationary activity".  Spatial activity entails movement and is visually interesting.  As examples:

* exploration, farming, and hauling resources place-to-place (i.e. in a resource pipeline) are spatial activities
* crafting (the `make` command) is a stationary activity
* subsequent to exploration, mining itself (i.e. the `drill` command) is a stationary activity.

Construction of structures would be another general game mechanic involving "spatial activity".  It can integrate with the classic "open world" mode without requiring the hermetic contrived environment of standalone scenarios.  This game mechanic could apply to both large scale structures (e.g. buildings like power plants) and small structures (perhaps described as "runes", encompassing tetrominoes or other pixellated glyphs).

## As programming challenge

Repeating patterns or symmetry within single structures provide an opportunity for players to craft parsimonious construction programs.  Across multiple structures, one may attempt a compact tiling of structures on the plane, or (if there were mutual exclusivity constraints on inventory occupancy, see #944) optimizing a construction assembly line.

# Technical considerations

Programming the word search scenario (#999) provided a taste of the complexity of recognizing "structures" in swarm-lang.  For performance, let alone the complexity, structure recognition would require built-in support.

<table>
<tr><th>Reference structure</th><th>Completing one row</th><th>Partial assembly</th><th>Nearly complete</th></tr>
<tr><td>

```
aBaBaBa
BxxoxxB
axoooxa
BxxoxxB
aBaBaBa
```

</td><td>

```
.......
Bxxoxx.
.......
.......
.......
```
</td><td>

```
aBaBaBa
BxxoxxB
.......
Bxxo.xB
aBaBaBa
```

</td><td>

```
aBaBaBa
BxxoxxB
a.oooxa
BxxoxxB
aBaBaBa
```

</td></tr>
<caption>Construction progress</caption>
</table>

## Search triggering

The ideal gameplay experience would be that a structure is instantly recognized once the player `place`s the last piece, no matter where within the structure that is.

A callback would be set up [here](https://github.com/swarm-game/swarm/blob/346f960085bc382a95daa78dba61eb3ab82abe3a/src/Swarm/Game/Step/Util.hs#L70) to scan for structure matches when an item is placed.

## Matching algorithm

Two matching steps, possibly utilizing different strategies, may be used to identify the constructed 2D structure.

### Row matching

In the first stage, perhaps [`AhoCorasick`](https://hackage.haskell.org/package/AhoCorasick) could be used.  That library searches for multiple "needles" simultaneously in one "haystack".  Each needle (or "key"), and the haystack itself, is a list of "key symbols".

Upon scenario initialization, each "row" constituting each participating structure template would be [added to the searcher state machine](https://hackage.haskell.org/package/AhoCorasick-0.0.3/docs/Text-AhoCorasick.html#v:makeStateMachine).  Rows of all participating structure definitions are the "needles", while the "haystack" is the horizontal world span extending `N` cells both to the left and right of the placed entity, where `N` is `max(all structure row lengths)`.  The "key symbols" don't necessarily have to be textual characters; they could be actual `entities`.

As an optimization, an entity-keyed `Map` of different [`StateMachine`s](https://hackage.haskell.org/package/AhoCorasick-0.0.3/docs/Text-AhoCorasick.html#t:StateMachine) could be precomputed, each including only the rows which contain a given entity.The right and left offset `N` could also be precomputed and keyed by entity, so that we don't have to examine as many neighboring cells.

Each "match" returned by [`findAll`](https://hackage.haskell.org/package/AhoCorasick-0.0.3/docs/Text-AhoCorasick.html#v:findAll) would be added to a cache of "partial structure candidates" stored in the world's state (or the `GameState`, keyed by subworld), keyed by coordinates.  Perhaps [`IntervalMap`](https://hackage.haskell.org/package/fingertree-0.1.5.0/docs/Data-IntervalMap-FingerTree.html#g:2) would be used for this data structure.

### Structures from rows

Matches against a known structure's row would trigger a second stage of matching in the vertical direction.  Perhaps this stage could also use AhoCorasick, but this time the "key symbols" are entire rows of a structure.

Instead of querying world cells directly, the search would be performed entirely within the "partial structure candidates" cache, among rows that are aligned vertically with the just-placed row candidate.  Again, the search would extend `M` rows both above and below the current row, where `M` is `max(all structure heights)`.

A complete structure may then be recognized, and possession granted to its creator.

A cell can participate in only one structure at a time, so the new structure's constituent rows are evicted from the "partial structure candidates" cache.

## Orientation

The method described above requires that the constructed orientation matches the structure definition.  It could be generalized to orientation-independence by adding the three other possible orientations to the searcher.

## Structure morphology

The structure shapes that we allow will depend on the matching algorithm.

* Should the structure recognition templates allow "transparent" cells?  If so, should the structure be required to be "convex"?
* Should the structure templates have a minimum length (in at least one dimension)?

<table>
<tr><th>Convex template with transparent cells</th><th>Non-convex template</th></tr>
<tr><td>

```
   x     
  xxx    
 xxxxx
xxxxxxx
 xxxxx
  xxx   
   x
```

</td><td>

```
xx  xx
 xxxx
 xxxx
xx  xx
```

</table>

It might be possible to implement matching in AhoCorasick that accommodates "transparent" cells with an `Ord` instance that is always [`EQ`](https://hackage.haskell.org/package/base-4.18.1.0/docs/Prelude.html#v:EQ) (i.e. it is a "don't care" cell that is always a match).

### Ownership

Upon placing the last piece, the placing robot should be granted "ownership" of and is conferred the privileges of that structure.

This could facilitate an adversarial gameplay mechanic, perhaps where players take turns to place pieces (like a reverse [Chomp](https://en.wikipedia.org/wiki/Chomp)) and try be the last placement, or even one player (perhaps a system robot) does most of the construction and another (the player) swoops in to place the last piece.

# Other enhancements

Incidentally,  #1271 would make structures with walls look nice.

kostmo:
Closes #1575

Implements structure recognition.

## Features

* Structure browsing dialog (`F6`) that becomes available if a scenario declares any recognizable structures
* Automatically recognizes statically-placed structures upon scenario initialization, accounting for occlusion by other entity/structure placement
* New `structure` command for querying location of recognized structures (primarily intended for system robots and goal checking)
* Efficiently recognizes structures immediately upon completion
* Accounts for removal of structures
* Several new integration tests
* Structured web-interface log to help understand/debug the recognition process
* Re-uses much of the functionality built previously for defining structures (#1332)

Other changes:
* Improved validation for static structure placement (ensure valid structure names instead of failing silently)
* Moved a few functions (`getNeighborLocs`, `zoomWorld`, `entityAt`, `robotWithID`, `robotWithName`) out of `Step.Util` and into `State` so that recognizer initialization, which becomes a field in `GameState`, could use them
* split `scenarioToGameState` into pure and non-pure functions

## Optimizations

Scenarios that do not make use of structure recognition are entirely unaffected, performance-wise.

Some optimizations include:

* Structure definitions must "opt-in" to participate in automatic recognition
* Aho-Corasick automatons optimized by:
    * only initiate structure search if a placed entity exists on a participating structure template
    * initializing different automatons for each type of "placed entity"
    * pruning inapplicable row candidates for 2-D search

The row-level structure recognition cache described in #1575 was not implemented; it's probably not worth the complexity cost.

# UI Demo

    scripts/play.sh -i scenarios/Testing/1575-structure-recognizer/1575-browse-structures.yaml --autoplay

1. Press `F6` for Structure Browser dialog
2. View http://localhost:5357/recognize/log and http://localhost:5357/recognize/found

image

# Future improvements

* Refactor `State.hs` so that the new helper functions can live elsewhere
* Support non-rectangular recognizable structures
* Allow flip/rotate of recognizable structures
* Structure ownership by robots
* Consolidate code between the Goals and Structures modal dialogs, and the Achievements browser
* Enforce minimum/maximum dimensions for structure definitions

### DIFF
--- a/data/scenarios/Testing/1575-structure-recognizer/1575-browse-structures.yaml
+++ b/data/scenarios/Testing/1575-structure-recognizer/1575-browse-structures.yaml
@@ -1,0 +1,131 @@
+version: 1
+name: Structure browser
+description: |
+  Hit F6 to view the recognizable structures.
+
+  Only the subset of the structures marked with
+  "recognize: true" are browseable.
+  In particular, the "donut" structure is placed
+  in the map but not displayed in the F6 dialog.
+creative: false
+objectives:
+  - teaser: Build structure
+    goal:
+      - |
+        Build a "precious" structure
+    condition: |
+      foundStructure <- structure "precious" 0;
+      return $ case foundStructure (\_. false) (\_. true);
+robots:
+  - name: base
+    dir: [1, 0]
+    devices:
+      - grabber
+      - treads
+    inventory:
+      - [50, flower]
+      - [50, log]
+      - [50, rock]
+      - [50, copper pipe]
+      - [50, iron gear]
+      - [50, quartz]
+      - [50, gold]
+      - [50, silver]
+      - [50, mithril]
+      - [50, cotton]
+solution: |
+   move;
+   place "quartz";
+   move;
+   place "quartz";
+   move;
+   place "mithril";
+structures:
+  - name: donut
+    structure:
+      palette:
+        '@': [dirt, rock]
+      mask: '.'
+      map: |
+        .@@@.
+        @@@@@
+        @@.@@
+        @@@@@
+        .@@@.
+  - name: diamond
+    recognize: true
+    description: "A diamond pattern of flowers"
+    structure:
+      mask: '.'
+      palette:
+        'x': [stone, flower]
+      map: |
+        ...x...
+        ..xxx..
+        .xxxxx.
+        xxxxxxx
+        .xxxxx.
+        ..xxx..
+        ...x...
+  - name: contraption
+    recognize: true
+    description: "A device for assembling useful widgets"
+    structure:
+      mask: '.'
+      palette:
+        'r': [stone, log]
+        'I': [stone, rock]
+        'l': [stone, copper pipe]
+        'g': [stone, iron gear]
+      map: |
+       rllllr
+       lIIIIl
+       lIIIgg
+       rlllgg
+  - name: precious
+    recognize: true
+    structure:
+      mask: '.'
+      palette:
+        'q': [stone, quartz]
+        'g': [stone, gold]
+        's': [stone, silver]
+        'm': [stone, mithril]
+      map: |
+       qgs
+       gsq
+       qqm
+  - name: smallish
+    recognize: true
+    structure:
+      mask: '.'
+      palette:
+        'q': [stone, quartz]
+        'm': [stone, mithril]
+        'c': [stone, cotton]
+      map: |
+       qqm
+       cqq
+known: [flower, log, rock, copper pipe, iron plate]
+world:
+  name: root
+  dsl: |
+    {blank}
+  palette:
+    '.': [grass]
+    'q': [grass, quartz]
+    'g': [grass, gold]
+    's': [grass, silver]
+    'm': [grass, mithril]
+    'c': [grass, cotton]
+    'B': [grass, null, base]
+  upperleft: [0, 0]
+  placements:
+    - src: donut
+      offset: [6, 0]
+  map: |
+    .qgs.........
+    .gsq.........
+    B............
+    .cqq.........
+    .............
--- a/data/scenarios/Testing/1575-structure-recognizer/1575-handle-overlapping.yaml
+++ b/data/scenarios/Testing/1575-structure-recognizer/1575-handle-overlapping.yaml
@@ -1,0 +1,76 @@
+version: 1
+name: Structure recognizer - Overlaps
+description: |
+  Completing a row that lies between two partially-complete structures
+  may complete both of them, but only one will be registered.
+creative: false
+objectives:
+  - teaser: Build structure
+    goal:
+      - |
+        Build a "precious" structure
+    condition: |
+      foundStructure <- structure "precious" 0;
+      return $ case foundStructure (\_. false) (\_. true);
+robots:
+  - name: base
+    dir: [1, 0]
+    devices:
+      - grabber
+      - treads
+    inventory:
+      - [50, quartz]
+      - [50, gold]
+      - [50, silver]
+      - [50, mithril]
+      - [50, cotton]
+solution: |
+   move;
+   place "quartz";
+   move;
+   place "quartz";
+   move;
+   place "mithril";
+structures:
+  - name: precious
+    recognize: true
+    structure:
+      mask: '.'
+      palette:
+        'q': [stone, quartz]
+        'g': [stone, gold]
+        's': [stone, silver]
+        'm': [stone, mithril]
+      map: |
+       qgs
+       gsq
+       qqm
+  - name: smallish
+    recognize: true
+    structure:
+      mask: '.'
+      palette:
+        'q': [stone, quartz]
+        'm': [stone, mithril]
+        'c': [stone, cotton]
+      map: |
+       qqm
+       cqq
+world:
+  name: root
+  dsl: |
+    {blank}
+  palette:
+    '.': [grass]
+    'q': [grass, quartz]
+    'g': [grass, gold]
+    's': [grass, silver]
+    'm': [grass, mithril]
+    'c': [grass, cotton]
+    'B': [grass, null, base]
+  upperleft: [0, 0]
+  map: |
+    .qgs.
+    .gsq.
+    B....
+    .cqq.
--- a/data/scenarios/Testing/1575-structure-recognizer/1575-overlapping-tiebreaker-by-location.yaml
+++ b/data/scenarios/Testing/1575-structure-recognizer/1575-overlapping-tiebreaker-by-location.yaml
@@ -1,0 +1,76 @@
+version: 1
+name: Structure recognizer - Tiebreaking overlaps by position
+description: |
+  A more lower-left overlapping structure of
+  identical size should win the tiebreaker
+creative: false
+objectives:
+  - teaser: Build structure
+    prerequisite:
+      not: wrong_structure
+    goal:
+      - |
+        Build a structure
+    condition: |
+      foundStructure <- structure "topleft" 0;
+      return $ case foundStructure (\_. false) (\_. true);
+  - id: wrong_structure
+    teaser: Don't recognize small structure
+    optional: true
+    goal:
+      - |
+        The "bottomright" structure shouldn't be recognized.
+    condition: |
+      foundStructure <- structure "bottomright" 0;
+      return $ case foundStructure (\_. false) (\_. true);
+robots:
+  - name: base
+    dir: [1, 0]
+    devices:
+      - grabber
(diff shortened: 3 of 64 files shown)


data/scenarios/Testing/1575-structure-recognizer/1575-overlapping-tiebreaker-by-location.yaml
```diff
@@ -0,0 +1,76 @@
+version: 1
+name: Structure recognizer - Tiebreaking overlaps by position
+description: |
+  A more northwesterly (i.e. earlier in row-major order) overlapping
```
kostmo, line 4: Fix description

src/Swarm/Language/Typecheck.hs
```diff
@@ -774,6 +774,7 @@ inferConst c = case c of
   Scout -> [tyQ| dir -> cmd bool |]
   Whereami -> [tyQ| cmd (int * int) |]
   Waypoint -> [tyQ| text -> int -> cmd (int * (int * int)) |]
+  Structure -> [tyQ| text -> int -> cmd (unit + (int * (int * int))) |]
```
kostmo, line 777: Perhaps could remove the sum type in the return value and just indicate length of `0` instead of `inL ()`.

kostmo: Nevermind; if the return type was `cmd (int * (int * int))` instead, when length is `0`, there is nothing to populate in the right side of the tuple.  So it needs to be wrapped in the sum type.

byorgey:
@kostmo I'm working my way through this surely but slowly =)

data/schema/structure.json
```diff
@@ -2,7 +2,7 @@
     "$schema": "http://json-schema.org/draft-07/schema#",
     "$id": "https://raw.githubusercontent.com/swarm-game/swarm/main/data/schema/structure.json",
     "title": "Structure",
-    "description": "Structure properties",
+    "description": "Structure properties.  Structures may opt into \"automatic recognition\" for when they are constructed by a robot. There are certain limitations on the shape and placement of such \"recognizable\" structures.",
```
kostmo, line 5:
```suggestion
    "description": "Structure properties.  Structures may opt-in to \"automatic recognition\" for when they are constructed by a robot. There are certain limitations on the shape and placement of such \"recognizable\" structures.",
```

src/Swarm/Game/Scenario/Topography/Structure/Recognition/Precompute.hs
```diff
@@ -0,0 +1,166 @@
...
+  getRows g = zipWith (StructureRow g) [0 ..] $ entityGrid g
+
+mkOffsets :: Foldable f => Int32 -> f a -> InspectionOffsets
+mkOffsets pos xs =
+  InspectionOffsets (pure (negate pos)) $
+    pure $
+      fromIntegral (length xs) - 1 - pos
+
+-- | Given a row of entities observed in the world,
+-- yield a searcher that can determine of adjacent
```
byorgey, line 76:
```suggestion
-- yield a searcher that can determine if adjacent
```

byorgey (approved):
This seems really cool and obviously was a ton of work to put together!  I am a bit worried about blowing our complexity budget, and that this may make it harder/more work to refactor various things in the future.  I don't have anything specific in mind, it just seems to have a high ratio of complexity of implementation to importance of the feature.  But in the end I don't think there's really any good reason not to merge it.  I appreciated the comments you included with assurances that this will not affect scenarios that don't opt-in to structure recognition, performance-wise.

data/scenarios/Testing/1575-structure-recognizer/1575-browse-structures.yaml
```diff
@@ -0,0 +1,131 @@
...
+      palette:
+        'q': [stone, quartz]
+        'g': [stone, gold]
+        's': [stone, silver]
+        'm': [stone, mithril]
+      map: |
+       qgs
+       gsq
+       qqm
+  - name: smallish
```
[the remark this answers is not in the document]
kostmo: Yes, cell membership in a structure is mutually-exclusive.

data/scenarios/Testing/1575-structure-recognizer/1575-handle-overlapping.yaml
```diff
@@ -0,0 +1,76 @@
+version: 1
+name: Structure recognizer - Overlaps
+description: |
+  Completing a row that lies between two partially-complete structures
+  may complete both of them, but only one will be registered.
```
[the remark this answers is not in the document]
kostmo: I settled on mutually-exclusive structure membership for each cell by the rationale that 70 structure instances is "too many", in the [example you've given](https://github.com/swarm-game/swarm/pull/1579#discussion_r1376365696).

> there is no way to control which of the two possible structures is the one that will actually be registered.

It will in part be up to the scenario author not to define structure templates that are conducive to overlap, or to minimize the significance of exact positioning when overlaps happen.

src/Swarm/Game/Scenario/Topography/Structure/Recognition/Type.hs
```diff
@@ -0,0 +1,163 @@
...
+-- 1. Primarily, larger area.
+-- 2. Secondarily, lower X-Y coords (X is compared first)
+--
+-- Since the natural order of coordinates increases as described,
+-- we need to invert it with 'Down' so that this ordering is by
+-- increasing preference.
+instance Ord FoundStructure where
+  compare = compare `on` (f1 &&& f2)
+   where
+    f1 = computeArea . getAreaDimensions . entityGrid . structureWithGrid
```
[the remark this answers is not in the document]
kostmo: I suppose that the cost may come from the `O(N)` time to compute `length` of rows and columns?  Maybe `Seq` is a better representation than `[]` for the members of the structure grid, so as to provide constant-time `length` access?

But anyway the `Ord` instance is only exercised when a tiebreaker is needed for recognizing simultaneously-completed structures, which is relatively rare.

kostmo:
> it just seems to have a high ratio of complexity of implementation to importance of the feature

I plan to write a few scenarios that use structure recognition.  With some examples I hope to demonstrate that this feature carries its weight :)

This could probably also simplify the goal condition logic for #999.

PR #1583, opened 2023-10-17
Require types to start with an uppercase letter

### LINKED ISSUE #1547: `path` type mentions `entity` which is not a real type

byorgey:
Closes #1547 , which was caused by misspelled/nonexistent types being interpreted as type variables.  In #1550 I proposed one solution to the problem, namely, to stop implicitly quantifying types and require explicit `forall` on any polymorphic type.  This PR represents an alternative solution: to keep implicit quantification but require all types to start with an uppercase letter, as in Haskell.  I think I like this one better but I'm open to feedback.

Specifically, with this PR:
- All built-in type constructors (`Int`, `Cmd`, `Unit`, etc.) must start with a capital letter
- Type variables:
    - Must start with a lowercase letter or underscore
    - May not be named a lowercase version of a type constructor, *e.g.* `int`
        - This is important so that old code with lowercase types is not silently accepted by parsing the former types as implicitly quantified type variables; even in cases where old code no longer typechecks, this will give better error messages.
- Term variables:
    - May start with upper- or lowercase
    - May be named lowercase versions of type names, *e.g.* `let f : Int -> Int = \int. int + 1` is fine

This PR obviously represents a bigger breaking change.  Once we merge this we might consider adding an option to `swarm format` to be able to parse old code with lowercase types and then reformat it with uppercase, to aid in porting code.

Once this is merged I will also regenerate the wiki pages that mention types.

Closes #1550.


xsebek (approved):
The types look more readable to me 👍

Just to check, uppercase is parsed as a concrete type or fails if it does not exist, right? Apologies if I overlooked a test for it.

byorgey:
> Just to check, uppercase is parsed as a concrete type or fails if it does not exist, right? Apologies if I overlooked a test for it.

That is correct.  I will add some more tests.

xsebek:
Btw. I think that if you `show`ed data from Swarm 0.4 and `read` and `print`ed them by 0.5, that would work as that tool.

It could look like:
```sh
swarm-0.4 format --out=Haskell old_code.sw | swarm-0.5 format --in=Haskell --stdin
```

byorgey:
> Btw. I think that if you `show`ed data from Swarm 0.4 and `read` and `print`ed them by 0.5, that would work as that tool.

That's true, but (1) that would require adding `--out` and `--in` flags, (2) it would be kind of annoying to tell users that in order to upgrade they needed to keep around an old binary instead of just replacing that old binary with the newer one, as usual.

byorgey:
@kostmo I've resurrected this PR.  When you get a chance, take a look and let me know what you think.

I was going to give `swarm format` the ability to convert code from the old format to the new.  I may still do that but I don't think it has to be in this PR.

src/swarm-lang/Swarm/Language/Parser/Lex.hs
```diff
@@ -142,54 +149,83 @@ operatorChar = T.singleton <$> oneOf opChars
...
+-- | Names of types built into the language.
+primitiveTypeNames :: [Text]
+primitiveTypeNames = "Cmd" : baseTypeNames
+
+-- | List of keywords built into the language.
+keywords :: [Text]
+keywords = T.words "let in def end true false forall require requirements"
+
 -- | List of reserved words that cannot be used as variable names.
 reservedWords :: [Text]
```
kostmo, line 165: Make this a `Set` since `elem` is used in a number of places?

kostmo (approved):

src/swarm-tui/Swarm/TUI/Controller.hs
```diff
@@ -1264,8 +1265,8 @@ tabComplete CompletionContext {..} names em theRepl = case theRepl ^. replPrompt
 
   possibleWords =
     names <> case ctxCreativeMode of
-      True -> reservedWords
-      False -> filter (`notElem` creativeWords) reservedWords
+      True -> S.toList reservedWords
+      False -> filter (`notElem` creativeWords) (S.toList reservedWords)
```
kostmo, line 1269: This looks quadratic?  Convert `creativeWords` to a set and use `difference`

byorgey: Oops, good point.  I'll include this change with #1851.

PR #1587, opened 2023-10-18
workspace spelling dictionary

kostmo:
The spellchecker in VS code can be quite noisy, given all of the esoteric terminology and shorthand variable names in code.
However, it is also super helpful for catching typos, if you can wade through the noise.

Adding a project-specific dictionary will significantly cut down on that noise.


kostmo:
This is not complete, but quite a few modules are now "spelling error free", so it has value to merge even now.  We can make more passes to fill out the dictionary in subsequent PRs.

.vscode/settings.json
```diff
@@ -10,5 +10,121 @@
...
+        "quasiquoters",
+        "sconcat",
+        "scrollability",
+        "selfdestruct",
+        "Semigroup",
+        "snoc",
+        "Splittable",
+        "squote",
+        "SRGB",
+        "struc",
```
byorgey, line 94: Is this entry supposed to be here?  What is `struc`?

byorgey (approved):

.vscode/settings.json
kostmo: It's debatable.  It's my truncated version of "structure".  The fact that you asked suggests that I should rename it in the source code :)

PR #1588, opened 2023-10-19
Achievement for pointless swapping

kostmo:
# Demo

    scripts/play.sh -i creative --seed 0

Then:

    turn right; move; turn right; move; t <- grab; move; swap t;

image


byorgey (approved):
Looks good.  I like the refactoring with `mkReturn` etc. and having `doGrab` return an `Entity` instead of a `CESK`.

PR #1591, opened 2023-10-19
combination locks scenario

kostmo:
image

# Demo

    scripts/play.sh -i data/scenarios/Challenges/combo-lock.yaml --autoplay


kostmo:
I found it difficult to get the "judge" robot to run its detection in a single tick.  At best, it seems to take 2 ticks.  To demonstrate, an example program executes two `drill` commands in immediate succession, and will **not** unlock the first gate.

```
// fast-drill.sw

move; move; move; move; move;
drill down;
turn back;
move; move;
drill down;
// wait 0;
drill down;
```

However, when uncommenting the `wait 0;` line, it works.

Execute the example program via:

    scripts/play.sh -i data/scenarios/Challenges/combo-lock.yaml --run fast-drill.sw

data/scenarios/Challenges/_combo-lock/setup.sw
```diff
@@ -0,0 +1,85 @@
...
+    };
+    end;
+
+def go =
+    comboString <- instant (      
+        move;
+        createCombo "";
+    );
+    // say comboString;
+    instant $ doUntilCorrect comboString;
```
kostmo, line 82: Here's where `instant` would ideally evaluate the lock combination in a single tick.

xsebek: What is the idea behind instant wait inside doUntilCorrect?

kostmo: That may have been in an earlier revision?  In the current revision of the PR, `instant` is only called outside of the `doUntilCorrect` function.

byorgey:
My solution, written without looking at the "official" solution (not pretty, but it works):
```
def x3 : cmd unit -> cmd unit = \c. c; c; c end
def m4 = move; move; move; move end
def m3 = move; move; move end

def blocked_left : cmd bool =
  r <- scan left;
  case r (\_. return false) (\_. return true)
end

def try_all : int -> cmd unit = \n.
  if (n == 0) {}
  { x3 (
      b <- blocked_left;
      if b {
        move;
        try_all (n-1);
        turn back; move; turn back;
        drill down;
        return ()
      }
      {}
    )
  }
end

def next = turn left; m4; turn right end

def solve =
  m3;
  try_all 3;
  next;
  try_all 4;
  next;
  try_all 5;
  next;
  try_all 6;
  turn left; m3; turn right; move; move; grab
end
```

byorgey (approved):
I like it!

kostmo:
Not that this scenario demonstrates an algorithmic puzzle of the like described in https://github.com/swarm-game/swarm/issues/112#issuecomment-931165389 which could potentially be distributed throughout an open world.

PR #1595, opened 2023-10-21
path caching

### LINKED ISSUE #1569: Caching for pathfinding
To improve the performance of #836, a caching mechanism could be designed that optimizes for the common case: a robot repeatedly invokes `path` with identical arguments and follows the yielded directions precisely.

1. If the robot takes a step off of the optimal path, the computed path shall be invalidated.
2. If the robot invokes the `path` command again with different arguments, the previous computed path shall be discarded.
3. If the list of entities that are unwalkable for the invoking robot changes, invalidate the cache.
4. Map modifications
    1. If an entity is added or removed at a Manhattan distance farther from the robot than the computed path length, the cache is unaffected.
    1. "unwalkable" entities (an entity is placed or removed that is "unwalkable" with respect to the invoking robot)
        1. If an "unwalkable" entity is **added** to the map, the computed path shall only be invalidated **if the new entity lies on the path**.
        2. If an "unwalkable" entity is **removed** from the map, the computed path shall be invalidated.
    2. "target" entities (if the `path` command had been invoked with the modified entity as a target)
        1. Adding
            1. If a target entity is added that lies on the computed path, the computed path is truncated to that entity's location
            2. If a target entity is added that does not lie on the computed path, invalidate the cache
        2. Removing
            1. If a target entity is removed that is the destination of the computed path, invalidate the cache
            2. If a target entity is removed that is *not* the destination of the computed path, the cache is unaffected

kostmo:
Closes #1569

## Performance

Path cache invalidation upon world modifications (i.e. entities inserted or removed) entails iterating over all of the previously-cached paths [here](https://github.com/swarm-game/swarm/pull/1595/files#r1390158411).  For efficiency's sake, we avoid iterating over "all existing robots".

Any scenario that does not use the `path` command is entirely unaffected by this change.

## Demo

Previously, this demo was virtually unplayable, since when moving between widely-spaced "clusters" of flowers, an expensive A-star search was invoked at almost every tick.  Now, the vast majority of moves utilize the cache, and the demo exhibits minimal stuttering (e.g. when a single A-star search is performed when moving between distant clusters).

    scripts/play.sh -i scenarios/Fun/horton.yaml --autoplay --speed 7

### Event log

An event log specific to the path cache is maintained with its own ring buffer:

    scripts/play.sh -i scenarios/Testing/1569-pathfinding-cache/1569-harvest-batch.yaml --autoplay

 and view http://localhost:5357/paths/log

### DIFF
--- a/src/Swarm/Game/Step/Util.hs
+++ b/src/Swarm/Game/Step/Util.hs
@@ -5,6 +5,8 @@
 
 -- |
 -- SPDX-License-Identifier: BSD-3-Clause
+--
+-- Utilities for implementing robot commands.
 module Swarm.Game.Step.Util where
 
 import Control.Applicative (Applicative (..))
@@ -16,6 +18,7 @@ import Control.Monad (forM_, guard, when)
 import Control.Monad.Trans.Class (lift)
 import Control.Monad.Trans.Maybe (MaybeT (..), runMaybeT)
 import Data.Array (bounds, (!))
+import Data.IntMap qualified as IM
 import Data.Set qualified as S
 import Data.Text (Text)
 import Data.Text qualified as T
@@ -27,7 +30,10 @@ import Swarm.Game.ResourceLoading (NameGenerator (..))
 import Swarm.Game.Robot
 import Swarm.Game.Scenario.Topography.Structure.Recognition.Tracking qualified as SRT
 import Swarm.Game.State
+import Swarm.Game.Step.Path.Cache
+import Swarm.Game.Step.Path.Type
 import Swarm.Game.Step.Path.Walkability
+import Swarm.Game.Step.RobotStepState
 import Swarm.Game.Universe
 import Swarm.Game.World qualified as W
 import Swarm.Game.World.Modify qualified as WM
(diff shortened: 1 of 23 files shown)


src/Swarm/Game/Step/Util.hs
```diff
@@ -74,6 +72,9 @@ updateEntityAt cLoc@(Cosmic subworldName loc) upd = do
     wakeWatchingRobots cLoc
     SRT.entityModified modType cLoc
 
+    pcr <- use $ pathCaching . pathCachingRobots
+    mapM_ (revalidatePathCache cLoc modType) $ IM.toList pcr
```
kostmo, line 76: Cache invalidation callback is here.

byorgey (approved):
@kostmo , thanks so much for the detailed comments and the amazing attention to detail with the whole cache logging system + tests!

I must admit I was skeptical of the idea of shortest path caching initially --- I was worried about performance, etc. --- but (1) the Horton's flower scenario is fun, and is a convincing demonstration of the utility of path caching, and (2) I am convinced that this will have minimal/zero impact whenever the `path` command is not being used.

I would love to eventually see a series of challenges introducing the `path` command (not saying you have to develop them).

PR #1604, opened 2023-10-29
Use rich entity modification info

kostmo:
This is a refactoring that is a prerequisite for both #1579 and #1595.


byorgey (approved):
Replacing a Boolean value with a data type representing concrete evidence is always OK in my book. :smile:

PR #1609, opened 2023-11-07
update to `megaparsec-9.6.1`

byorgey:
This allows us to remove any annoying workaround. Also bump stack resolver to LTS-21.19.

### DIFF
--- a/.github/workflows/haskell-ci.yml
+++ b/.github/workflows/haskell-ci.yml
@@ -77,11 +77,6 @@ jobs:
             compilerVersion: 9.0.2
             setup-method: ghcup
             allow-failure: false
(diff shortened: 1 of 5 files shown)


kostmo (approved):

byorgey:
Interesting, it looks like CI is failing because `megaparsec-9.6` requires `base >= 4.15`, so it won't build on GHC 8.10.  @kostmo , @xsebek , how do you feel about Swarm dropping support for GHC < 9.0?  Seems like it's not a big deal to me --- supporting 4 versions of GHC should be plenty.

kostmo:
> dropping support for GHC < 9.0?

Out with the old, in with the new!

PR #1610, opened 2023-11-07
Fix REPL type display

### LINKED ISSUE #1597: Type displayed in UR corner of REPL is incomplete
After the recent changes to pretty-printing, some types are no longer printed completely in the upper-right corner of the REPL.  For example:

repl-type

We should see the complete type of `place`, that is, `text -> cmd unit`.

byorgey:
Make sure it again prints the type on one line, by using `prettyTextLine` instead of `prettyText` to format the type.

Fixes #1597.

### DIFF
--- a/src/Swarm/TUI/Panel.hs
+++ b/src/Swarm/TUI/Panel.hs
@@ -1,5 +1,3 @@
------------------------------------------------------------------------------
------------------------------------------------------------------------------
 {-# LANGUAGE TemplateHaskell #-}
 
 -- |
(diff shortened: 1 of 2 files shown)


xsebek (approved):
I wonder what happens if the type is too long, should there be some `take maxTypeWidth` followed by `...`?

byorgey:
Hmm, good question.  The type just gets cut off:

repl-type-long

I will see whether there's a quick fix for putting a `...` at the end.  One problem is we'd like it to be dynamic based on the size of the window.

byorgey:
@xsebek How about this?  It's not ideal the way it cuts off the type *textually* (note below how a `->` gets cut off in the middle so we see `-...`), but figuring out how to cut it off semantically (so e.g. we always put the `...` at some kind of natural break) feels like a project for another time.

repl-type-long2

xsebek:
Looks good! 👍 

I think this will be more than enough context for REPL, anything beyond that should really be done in editor with LSP. 🙂

PR #1612, opened 2023-11-09
Show how many robots are out of range in F2 dialog

kostmo:
Screenshot from 2023-11-09 12-49-51

### DIFF
--- a/src/Swarm/TUI/View.hs
+++ b/src/Swarm/TUI/View.hs
@@ -47,6 +47,7 @@ import Brick.Widgets.Dialog
 import Brick.Widgets.Edit (getEditContents, renderEditor)
 import Brick.Widgets.List qualified as BL
 import Brick.Widgets.Table qualified as BT
+import Control.Arrow ((&&&))
 import Control.Lens as Lens hiding (Const, from)
 import Control.Monad (guard)
 import Data.Array (range)
@@ -55,7 +56,7 @@ import Data.Foldable (toList)
 import Data.Foldable qualified as F
 import Data.Functor (($>))
 import Data.IntMap qualified as IM
-import Data.List (intersperse)
+import Data.List (intersperse, partition)
 import Data.List qualified as L
 import Data.List.NonEmpty (NonEmpty (..))
 import Data.List.NonEmpty qualified as NE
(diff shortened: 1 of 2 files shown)


byorgey:
I think we should only show this information in creative mode, on the principle that it allows instantaneous transmission of information over arbitrary distances (a very far away robot can send you one bit of information by choosing whether to self-destruct, which could change the displayed count of robots that are too far).  More generally, if a robot is too far away to sense, you should not be able to find out anything about it, including whether it is alive.

byorgey:
Hmm, but in creative mode we already show all robots regardless of distance, right?  So maybe I am simply opposed to this change.

kostmo:
The original impetus was my state of confusion when I popped up the F2 dialog and the robots **that I had built** did not show up; I thought it was a bug.

The intent of this change is only to display player-built robots.  Effectively the filtering logic makes this so, but to guarantee this, perhaps we could also do a check that ensures their direct parent is the `base`.

I think we can rationalize this in terms of the "game world logic" by the fact that the `base` must have known about each of these robots at the time they were built.  So, if they subsequently go out of range, the fact that we "reveal" their existence is not communicating any new information.

Besides, this is only available in the UI, and not accessible programmatically.  Technically speaking, when robots *are* within range, I think that the F2 dialog communicates other information to the chair-sitting human as well that isn't programmatically available.

kostmo:
Perhaps I could salvage this feature by implementing some filtering modes in the F2 dialog.  E.g. in creative mode, I might want to exclude the (dozens) of plant-growing robots and just show my `base`-built robots.  Or I might want to filter by distance **as if** I were not playing in creative mode (as a kind of debugging aid).

kostmo:
One more option may to implement some "persistence" of robot existence in the UI layer; the widget may retain knowledge of "last seen" robots, and if they go out of range their status may become "unknown".  If they had self-destructed while out of range, their status will remain unknown permanently.

byorgey:
> Perhaps I could salvage this feature by implementing some filtering modes in the F2 dialog.

That sounds like a nice feature, but obviously represents a lot more work.

> One more option may to implement some "persistence" of robot existence in the UI layer; the widget may retain knowledge of "last seen" robots, and if they go out of range their status may become "unknown". If they had self-destructed while out of range, their status will remain unknown permanently.

Yes, that option sounds good to me.

PR #1613, opened 2023-11-09
indicate whether a goal is optional

kostmo:
Screenshot from 2023-11-09 14-20-49

Screenshot from 2023-11-09 14-20-56


byorgey (approved):
Looks nice!

Issue #1607, opened 2023-10-31
Migrate to brick 2.0


xsebek:
vty 6.0 and brick 2.0 support windows! 🚀 So we could too and close #53

jtdaugherty:
We might have a wrinkle or two left to work out, but heavier users of `vty` like `swarm` will help us surface remaining issues. Keep me and @chhackett in the loop!

chhackett:
Hello,
I updated the cabal and App module to allow for running on Windows. I did some playtesting, and I didn't find any rendering or gameplay issues. All the keyboard commands seem to behave well. I've addressed comments on the PR.

Keep in mind that right now only command line and Powershell terminals are supported.

Btw, what a neat game!

chhackett:
Actually, I do see 2 issues so far:
- mouse mode isn't working for some reason
- program expects the data files to be present in the C:\Users\Chris\AppData\Roaming\swarm\data directory. Or is that a feature? Not sure, haven't played it before...

chhackett:
Disregard my previous comment about mouse mode. I did not realize mouse mode was only enabled for some menus. For example, the main menu doesn't respond to mouse clicks, but other UI elements do. So mouse mode does work.

byorgey:
> program expects the data files to be present in the C:\Users\Chris\AppData\Roaming\swarm\data directory. Or is that a feature? Not sure, haven't played it before...

That's normal.  You can get the latest data files from https://github.com/swarm-game/swarm/releases under swarm-data.zip.  That actually reminds me that if we are updating to support Windows we should probably include changes to https://swarm-game.github.io/installing/ to mention the proper instructions.  But we can do that separately.

Issue #53, opened 2021-09-24
Windows support


kajvan:
i am trying to use the command stack run but i get the error:

```
In the dependencies for brick-0.64:
    unix needed, but the stack configuration has no specified version  (latest matching version is 2.7.2.2)
needed due to swarm-0.1.0.0 -> brick-0.64

In the dependencies for vty-5.33:
    unix needed, but the stack configuration has no specified version  (latest matching version is 2.7.2.2)
needed due to swarm-0.1.0.0 -> vty-5.33
```

how can i fix this

byorgey:
Ah, my apologies.  Swarm depends on `brick`, which depends on `vty`, and `vty` is not supported on Windows.  So that means Swarm only works on Linux or OSX, at least for now.  I would have mentioned it in the README if I had realized it.

I do not know whether it might be possible to get it to build on Windows using something like Cygwin or the Windows Subsystem for Linux. I would be happy to receive any reports either way.

I will leave this issue open and change the name as a place to collect ideas and information re: supporting Windows.

abyo:
It is possible to get the game running with Windows Subsystem. Installation is really easy too those days.

```
wsl --install
```

Follow instructions (this will install ubuntu by default) then upgrade/update. Install ghc/ghcup.

```
sudo apt-get install ghc
curl -sSL https://get.haskellstack.org/ | sh
```

Install stack (if you didn't press yes before), cd in repo then stack.

```
ghcup install stack
stack run
```

byorgey:
@abyo Thanks, glad it works!  I am curious, do you really need to `apt-get install ghc` first?  You should just be able to get `ghcup` then use that to install `stack`.

abyo:
@byorgey I had an issue and found that installing ghc first resolved it, but I suppose you don't have too.

amalloy:
I ran it on WSL as well. ghcup worked fine, after I installed the prerequisite packages it asked me for.

byorgey:
Since it seems we have a viable path to building on Windows, I'm going to close this issue.

PR #1617, opened 2023-11-11
Update to support running on Windows

chhackett:
Added support for running on Windows, in command line or Powershell terminals.

Currently terminal emulators such as mintty, ConEmu, alacritty, etc are not supported.

Addresses issues #1607 and #53.

### DIFF
--- a/src/Swarm/App.hs
+++ b/src/Swarm/App.hs
@@ -111,18 +111,18 @@ appMain opts = do
             handleEvent e
 
       -- Setup virtual terminal
-      let mkVty =
+      let buildVty =
             case colorMode opts of
               Nothing    -> VS.mkVty V.defaultConfig
               Just cMode -> do
-                    platformSettings <- VS.defaultSettings
-                    VS.mkVtyWithSettings V.defaultConfig $ platformSettings { VS.settingColorMode = cMode }
-      vty <- mkVty
+                platformSettings <- VS.defaultSettings
+                VS.mkVtyWithSettings V.defaultConfig $ platformSettings {VS.settingColorMode = cMode}
+      vty <- buildVty
 
       V.setMode (V.outputIface vty) V.Mouse True
 
       -- Run the app.
-      void $ customMain vty mkVty (Just chan) (app eventHandler) s'
+      void $ customMain vty buildVty (Just chan) (app eventHandler) s'
 
 -- | A demo program to run the web service directly, without the terminal application.
 --   This is useful to live update the code using @ghcid -W --test "Swarm.App.demoWeb"@.
--- a/src/Swarm/App.hs
+++ b/src/Swarm/App.hs
@@ -113,7 +113,7 @@ appMain opts = do
       -- Setup virtual terminal
       let buildVty =
             case colorMode opts of
-              Nothing    -> VS.mkVty V.defaultConfig
+              Nothing -> VS.mkVty V.defaultConfig
               Just cMode -> do
                 platformSettings <- VS.defaultSettings
                 VS.mkVtyWithSettings V.defaultConfig $ platformSettings {VS.settingColorMode = cMode}
--- a/src/Swarm/App.hs
+++ b/src/Swarm/App.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE CPP #-}
 {-# LANGUAGE OverloadedStrings #-}
 
 -- |
@@ -18,6 +19,13 @@ import Data.IORef (newIORef, writeIORef)
 import Data.Text qualified as T
 import Data.Text.IO qualified as T
 import Graphics.Vty qualified as V
+#if defined(mingw32_HOST_OS) || defined(__MINGW32__)
+import Graphics.Vty.Platform.Windows.Settings qualified as VS
+import Graphics.Vty.Platform.Windows qualified as VS
+#else
+import Graphics.Vty.Platform.Unix.Settings qualified as VS
+import Graphics.Vty.Platform.Unix qualified as VS
+#endif
 import Swarm.Game.Failure (SystemFailure)
 import Swarm.Language.Pretty (prettyText)
 import Swarm.Log (LogSource (SystemLog), Severity (..))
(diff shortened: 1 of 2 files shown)


src/Swarm/App.hs
```diff
@@ -103,12 +111,18 @@ appMain opts = do
...
-      let buildVty = V.mkVty $ V.defaultConfig {V.colorMode = colorMode opts}
-      initialVty <- buildVty
-      V.setMode (V.outputIface initialVty) V.Mouse True
+      let mkVty =
+            case colorMode opts of
+              Nothing    -> VS.mkVty V.defaultConfig
+              Just cMode -> do
+                    platformSettings <- VS.defaultSettings
+                    VS.mkVtyWithSettings V.defaultConfig $ platformSettings { VS.settingColorMode = cMode }
+      vty <- mkVty
```
xsebek, line 120: This is odd, why do we need to run `mkVty` twice? (Here and below in `customMain`.)

I noticed this because it would be nicer to have `vty <-` instead of `let mkVty =`, but now I wonder if running it twice could cause any issues with the terminal. 🤔 

If it has to be done this way, please add a comment with explanation.

swarm.cabal
```diff
@@ -310,13 +310,18 @@ library
...
                       warp                          >= 3.2 && < 3.4,
                       witch                         >= 1.1.1.0 && < 1.3,
                       witherable                    >= 0.4 && < 0.5,
                       word-wrap                     >= 0.5 && < 0.6,
-                      yaml                          >= 0.11 && < 0.11.12.0,
+                      yaml                          >= 0.11 && < 0.11.12.0
+    if os(windows)
+      build-depends:  vty-windows                   >= 0.1.0.3 && < 1.0
+    else
+      build-depends:  vty-unix                      >= 0.1.0.0 && < 1.0
```
xsebek, line 324: Why not use only vty-crossplatform? 😕

xsebek (changes requested):
Thanks @chhackett, I look forward to playing Swarm on Windows. 🚀  

Could you please also update the CI file and add Windows to the test matrix?

swarm.cabal
```diff
@@ -254,7 +254,7 @@ library
                       astar                         >= 0.3 && < 0.3.1,
                       blaze-html                    >= 0.9.1 && < 0.9.2,
                       boolexpr                      >= 0.2 && < 0.3,
-                      brick                         >= 1.10 && < 1.11,
+                      brick                         >= 2.1.1 && < 3.0,
```
byorgey, line 257: Let's make this `< 2.2` instead of `< 3.0`, according to the PVP there's no guarantee we will be compatible with other 2.x releases.

The same comment applies to the upper bounds on `vty`, `vty-crossplatform`, and `vty-windows` / `vty-unix` below.

swarm.cabal
chhackett: The only reason for the explicit dependency on unix/windows libraries is because the ColorMode setting is not part of the default configuration anymore, but down in the unix/windows settings. 

That is an unfortunate consequence of the design choice we made in how to split up the unix/windows implementations.

src/Swarm/App.hs
chhackett: So it looks like it is called twice, but actually the function to construct a Vty instance is only called once. the 'mkVty' variable has type 'IO Vty' and is necessary because you are using the customMain function which requires a Vty instance, and an IO Vty (I don't know why...). 

Anyway, this is exactly what you were doing in this function before. I just changed mkVty to mkVtyWithSettings when the colorMode is available. Otherwise, use the old mkVty.

swarm.cabal
chhackett: Done!

chhackett:
> Thanks @chhackett, I look forward to playing Swarm on Windows. 🚀
> 
> Could you please also update the CI file and add Windows to the test matrix?

I'd be happy to do that, but I'm not sure what change I should make to the CI file (assuming you mean cabal.haskell-ci).

And I don't know where to find the test matrix. Sorry, I'm new to contributing to github projects.

I'm happy to do that, but I'm not sure what

chhackett:
@xsebek 
OK, I've been reading about the haskell github CI project, and I'm guessing you were hoping I would add a Windows section to the /.github/workflows/haskell-ci.yml file?

Unfortunately, as far as I can tell, there is no Windows version. Am I missing something?

swarm.cabal
chhackett: So now that I'm thinking about this, I wonder if we made a mistake by putting ColorMode in the UnixSettings/WindowsSettings data types. It is something that we could potentially add to the VtyUserConfig
since both implementations support it.
That would make one less reason for users to need to have an explicit dependency on the unix/windows plugins.
@jtdaugherty - what do you think?

p3rsik:
@chhackett I believe you would want to define new job for windows(as done [here](https://github.com/swarm-game/swarm/blob/aca8049789f7821a37d7fb875f26d0a9b70c0284/.github/workflows/haskell-ci.yml#L49)) with it's own steps and compiler matrix for windows.

chhackett:
@p3rsik @xsebek 
I would love to learn how to define a workflow job for windows. However, my time available for doing this kind of work is very limited. And right now I have my hands full working out some bugs in the vty-windows implementation. So I'm unlikely to get to this task for a while. So if anyone else has bandwidth and wants to take that on, please be my guest. 

And just to be clear, I'm not promising I'll ever get around to doing this. I contributed this PR in the spirit of being helpful. But I want to stay focused on improving the experience of using vty/brick on Windows, at least for now.

swarm.cabal
jtdaugherty: Looking back at the change history, this was an oversight. I moved the color mode *detection* code to `vty-unix`, but there was a lot going on with the split-up of `Config` that the color mode setting got moved in the shuffle. Since the `ColorMode` type lives in `vty`, it makes sense to me to put that in the Vty user config as well, even if the backend packages still do their own detection. I think the caveat would be that backends might need to ignore or override the setting if it's unsuitable, but that was already true before Vty 6.

jtdaugherty: I'll look into the implementation and see what this would involve.

xsebek:
I forgot it was so linux specific, we can look into it separately - let's just add an Issue for Windows CI.

xsebek (approved):

byorgey:
So it sounds like there may be changes coming which would allow us to depend only on `vty-crossplatform` instead of having a conditional dependency on one of `vty-unix` and `vty-windows`.  Should we wait for that, or should we go ahead and merge now and make those updates later?  @jtdaugherty , any thoughts on a potential timeline on your end?

jtdaugherty:
@byorgey I plan on looking into it more tonight. I started working on a patch earlier today, but it turns out that it's awkward to have a setting in `vty`'s `VtyUserConfig` that `vty` itself doesn't do anything with -- the platform packages will need to make use of it -- so I'm not yet clear on what a good design will turn out to be. If it isn't too much trouble, I'd say go ahead and move forward with what you have and then adjust later. I wouldn't want to hold up your work unless you find the patch unpalatable and would like to hold off in the hope of something better.

byorgey:
@jtdaugherty OK, thanks, no rush.  We'll go ahead and merge this, but I'll add an issue to remind us to possibly update in the future.  Conceptually, I'd love to eventually be able to depend only on `vty-crossplatform` and thus have a sort of static guarantee that we are limited to doing things that are supported on both platforms.  But I don't have a problem with the conditional platform-specific dependencies for now.

jtdaugherty:
Understood, and I definitely don't see depending on specific platform packages as a good thing, either, if you're going for cross-platform support. So I will aim for something that removes that requirement.

PR #1619, opened 2023-11-13
Don't auto-complete commands that require God capability outside creative mode

### LINKED ISSUE #1572: Do not suggest commands that require God capability outside creative mode
Players should ideally not get suggestions for not relevant commands.

But they should definitely not get suggestions for commands requiring `CGod` capability unless they are in creative mode.

For example `in` should complete with `inl` but not `instant`.

**Alternatives**

We could also use current robot capabilities or known commands.

p3rsik:
fixes #1572
Fixed auto-completion for non-creative mode


p3rsik:
Well, it failed because of hlint suggestion. Do I fix it, or is it ok?

byorgey:
Hi @p3rsik , thanks for the contribution!  Yes, please do fix the hlint suggestions.  If there are particular hlint suggestions that you think we should ignore, we can also discuss them and potentially add an exception in the `.hlint.yaml` file, but either way we cannot merge until the CI passes.

Also, would you mind taking the Nix flake stuff out of this PR?  It's better to separate such unrelated changes into separate PRs (though also see #80 and #169 for some previous discussion re: adding Nix support in particular).

p3rsik:
@byorgey there is a suggestion and a warning there. Suggestion is easily fixable, however warning is strange, since it recommends replacing `not el `elem` with `notElem el` and I don't know why it fires only in this pipeline. Is it okay to change it as well(won't it cause issues in other pipelines?) or do I leave it there.

Also, about Nix support, I see your point and can just remove this altogether, if you want. It's as simple as it can get and I can support it, but since core contributors aren't using it and are not willing to support it, I have no reason to force it onto you guys.

byorgey (approved):
Thanks, this looks great!  I especially appreciate that you made a new `CompletionContext` newtype instead of just passing a `Bool` to `tabComplete`, `Bool` arguments are the worst. :smile:

byorgey:
@p3rsik  , congrats on having your first PR accepted to Swarm!   We appreciate the contribution and we're really glad to welcome you as part of the [community](https://github.com/swarm-game/swarm/blob/main/COMMUNITY.md).  In a minute we will send you an invite to the repo (see https://github.com/swarm-game/swarm/blob/main/CONTRIBUTING.md#i-have-push-access-to-the-swarm-repository-now-what ).  If you use IRC, feel free to also join the `#swarm` channel on Libera.Chat.

Issue #1624, opened 2023-11-14
Fix or deprecate `stack build`


kostmo:
Since #1617, `cabal build` works but `stack build` no longer does.

Do we believe that `cabal` has parity with `stack` for the purposes of our project?  Then we should scrub/replace references to `stack` in our scripts and documentation.

byorgey:
I don't think we should be so quick to get rid of `stack`.  I use it for building swarm, and it's quite useful because anyone wanting to build swarm can be sure they will get a reproducible build, without worrying about whether they have the right version of GHC installed.

byorgey:
I didn't realize #1617 broke the stack build, but I will happily fix it.

kostmo:
See https://github.com/swarm-game/swarm/pull/1798#discussion_r1586949897 for a demonstration of a `stack` bug regarding rebuilding a target that does not occur with `cabal`:

> Try the following:
> 1. Add a newline to the end of `src/swarm-tournament/Swarm/Web/Tournament/Validate.hs`
> 2. Run `stack build swarm:exe:swarm`
> 3. Observe that it recompiles `swarm-tournament`, despite that library not being a dependency of the `swarm` executable.
> 
> Now do the same cycle, but with `cabal build swarm:exe:swarm` instead, and observe `Up to date` is printed.

PR #1626, opened 2023-11-15
replace stack commands with cabal

kostmo:
Towards #1624


byorgey:
I'm definitely against this change.  Using `cabal` means that whatever GHC version is installed will be used; I often change GHC versions via `ghcup`, and I don't want to have to worry about changing it back every time I want to build Swarm.

kostmo:
Sure, I actually prefer `stack` as well.

Issue #1623, opened 2023-11-14
Depend on only `vty-crossplatform`


byorgey:
> @jtdaugherty OK, thanks, no rush.  We'll go ahead and merge this, but I'll add an issue to remind us to possibly update in the future.  Conceptually, I'd love to eventually be able to depend only on `vty-crossplatform` and thus have a sort of static guarantee that we are limited to doing things that are supported on both platforms.  But I don't have a problem with the conditional platform-specific dependencies for now.

_Originally posted by @byorgey in https://github.com/swarm-game/swarm/issues/1617#issuecomment-1811238238_

See the rest of the discussion on #1617 for more context.  Currently, we conditionally depend on either `vty-unix` or `vty-windows`; ideally, if the `vty` packages are updated in an appropriate way, we will be able to depend only on `vty-crossplatform`.

byorgey:
Don't forget to remove `vty-unix` from `stack.yaml`!

jtdaugherty:
@byorgey @chhackett here's a summary of what I have so far in some local uncommitted changes. Let me know what you think:
* `vty`'s `VtyUserConfig` type will get a new field, `configPreferredColorMode :: Maybe ColorMode`.
* `vty`'s user configuration file parser will start to support a new field, `colorMode`, whose value is one of the `ColorMode` choices. It's optional, and if it's absent it's understood that the backend driver will detect and choose an appropriate color mode. If it's present, the backend driver should honor the configured color mode regardless of what its detection finds.
* `vty-unix`'s `UnixSettings` type will have its `settingColorMode` field removed.
* `vty-unix`'s `buildOutput` function will start to take a `VtyUserConfig` in order to respect the configured setting, although this is not an end-user-facing API so that shouldn't matter.

Once the various packages have been updated, this ought to allow users to configure the color mode (provided each backend package respects the setting!) and should allow you to get away from a dependency on a specific backend package.

byorgey:
@jtdaugherty Sounds good to me.

jtdaugherty:
This is now released in `vty-6.1` and `vty-unix-0.2.0.0`.

jtdaugherty:
It turns out that `vty-crossplatform` also needs to get updated so I'm working on that now. I forgot that it depends on the `buildOutput` functions from both platform packages for testing and the `vty-unix` `buildOutput` changed. The Windows version doesn't need to, necessarily, since I can deal with that with CPP.

jtdaugherty:
[...]

PR #1627, opened 2023-11-15
Fix the stack build

### LINKED ISSUE #1624: Fix or deprecate `stack build`

byorgey:
Closes #1624.

Note, however, that currently (until #1623) this will not work on Windows, since I explicitly added `vty-unix` as an `extra-dep` in the `stack.yaml` file.  However, I expect we will get #1623 soon and then we can remove it.


kostmo (approved):
Works on my machine.

Issue #1046, opened 2023-01-25
Optional failure condition for Objectives


kostmo:
From comment on #927:

> As an aside, it seems like it could make sense to generalize goal conditions a bit --- either (1) by having an optional second failure condition that runs every tick, or (2) by having conditions return something more general than just a boolean, *e.g.* a `unit + bool` where `inl ()` means "no conclusion", and `inr x` means the goal is done, with `x` determining whether it was completed or failed.  However, with (2) it would be a pain to update all existing conditions, so probably (1) would be the way to go.

byorgey:
Having objectives with two separate conditions (one success + one failure) sounds like a bad idea: we would probably end up duplicating code between them.  This could be mitigated with imports but it would probably still be wasteful to run both.  Instead, we could have a new `generalized condition` field with type `Cmd (Unit + Bool)`; if that is present, we would use it (with the meaning explained in the original comment above) instead of the usual `condition : Cmd Bool`.

Perhaps we could even allow the `condition` field to have *either* type, and we would use whichever one it checks at.  We could first try checking against the type `Cmd Bool`, and if that fails, we could try `Cmd (Unit + Bool)` instead.  The downside of this would be slightly more confusing error messages for scenario developers if they mess up the condition.

PR #1628, opened 2023-11-15
fishing scenario

kostmo:
This scenario makes use of several relatively recent features/commands:
* structure placement
* structure queries (for goal checking)
* density command (for goal checking)
* combustion
* tags
* goal prerequisites with boolean expressions

## Demo

    scripts/play.sh -i Challenges/Ranching/fishing.yaml --autoplay

image

### DIFF
--- a/data/scenarios/Challenges/Ranching/fishing.yaml
+++ b/data/scenarios/Challenges/Ranching/fishing.yaml
@@ -1,11 +1,11 @@
 version: 1
-name: Fishing trip
+name: Responsible Fishing
 author: Karl Ostmo
 description: |
-  Catch a the appropriate seafood to make lunch.
+  Catch the appropriate seafood to make lunch.
 
   If you haul any trash out of the lake, you
-  must throw it away in the enclosure to the southeast.
+  must throw it away in the enclosures to the southeast.
   No littering!
 creative: false
 seed: 0
--- a/data/scenarios/Challenges/Ranching/fishing.yaml
+++ b/data/scenarios/Challenges/Ranching/fishing.yaml
@@ -1,0 +1,544 @@
+version: 1
+name: Fishing trip
+author: Karl Ostmo
+description: |
+  Catch a the appropriate seafood to make lunch.
+
+  If you haul any trash out of the lake, you
+  must throw it away in the enclosure to the southeast.
+  No littering!
+creative: false
+seed: 0
+attrs:
+  - name: shark
+    bg: "#2040bb"
+objectives:
+  - teaser: Hot rod
+    goal:
+      - |
+        `equip` your `fishing tackle`{=entity}, i.e. a rod and its accoutrements.
+    condition: |
+      as base {
+        equipped "fishing tackle";
+      }
+  - teaser: Sushi chain
+    id: lots_of_sushi
+    goal:
+      - |
+        Make 16 `california roll`{=entity}s
+    condition: |
+      as base {
+        rollCount <- count "california roll";
+        return $ rollCount >= 16;
+      }
+    prerequisite: made_sushi
+  - teaser: Sushi chef
+    id: made_sushi
+    goal:
+      - |
+        Make a `california roll`{=entity}
+    condition: |
+      as base {
+        has "california roll"
+      }
+  - teaser: Waste disposal
+    prerequisite:
+      logic:
+        and:
+          - lots_of_sushi
+          - not: littered
+    goal:
+      - |
+        Inventory must be free of all `junk`{=tag}.
+      - |
+        Dispose of `junk`{=tag} only in designated `rubbish enclosure`{=structure}s.
+      - |
+        Certain items can be burned to make room.
+        Good thing you brought your `torch`{=entity}!
+      - |
+        Otherwise, when the enclosure becomes full, the trash
+        hauler will come to empty it.
+    condition: |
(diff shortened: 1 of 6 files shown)


data/scenarios/Challenges/Ranching/fishing.yaml
```diff
@@ -0,0 +1,510 @@
...
+
+  If you haul any trash out of the lake, you
+  must throw it away in the enclosure to the southeast.
+  No littering!
+creative: false
+seed: 0
+attrs:
+  - name: shark
+    bg: "#2040bb"
+objectives:
```
byorgey, line 15: The objectives are quite confusing.  After completing the scenario by disposing of everything properly, I was confused why I still had not achieved the optional "no littering" objective.  "But surely I did not litter!" I thought.  "Why does it still think I have not completed that objective?"  I thought perhaps everything had to be hauled away, so I spent some time doing more fishing until I had the perfect number of junk items so that the hauler could haul them away, but I still did not complete the objective.

It is only when I looked at this `.yaml` file that I realized that objective is "completed" when you *do* litter, and that completing it is actually *bad*!  Do we have a way yet to mark an objective as "bad"/negative and display it with a different color / in a different section?  I thought we had an issue for that but I can't find it now.  In any case maybe just some text in the objective description would help that says something like "Note, this is an "anti-objective": it will be marked "completed" if you litter, which will cause you to fail the scenario.  Don't litter!"

Even better, maybe the description for the No Littering goal can be moved into the Sushi chef goal description, and the No littering goal can be marked hidden.  I think that would be a lot less confusing.

byorgey (commented):
This was fun!  A few suggestions:

1. The way the objectives are set up is confusing; see my inline comments and suggestions below.
2. I never really switched out of "exploratory" mode and ended up solving the whole thing just typing a bunch of things manually at the REPL, without having to set up any kind of automation with `tagmembers` etc.  What about adding a final objective that ramps up the numbers so that you are forced to actually write some generic code lest it be incredibly tedious?  e.g. after you make one california roll, a new objective could be revealed that says you want to start your own sushi business so you need to make 32 california rolls (or something like that).

data/scenarios/Challenges/Ranching/fishing.yaml
kostmo: > Do we have a way yet to mark an objective as "bad"/negative and display it with a different color / in a different section? I thought we had an issue for that but I can't find it now.

Closest issue may be #1046.

Agreed that the goal presentation should be improved.  Will look into it.

data/scenarios/Challenges/Ranching/fishing.yaml
```diff
@@ -0,0 +1,510 @@
...
+        Also allows you to iterate over the list of junk, e.g.:
+      - |
+        `tagmembers "junk" 0`
+    capabilities: [hastag, tagmembers]
+  - name: torch
+    display:
+      attr: wood
+      char: 't'
+    description:
+      - Can set things on fire
```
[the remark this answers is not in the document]
kostmo: > Can this description be elaborated a bit? 

Will add instructions for `ignite`.

> As an aside, do we really still need a separate ignite command?

Previous discussion for reference: https://github.com/swarm-game/swarm/pull/1432#discussion_r1299015273

data/scenarios/Challenges/Ranching/fishing.yaml
```diff
@@ -0,0 +1,510 @@
...
+    .........................h.
+    .........................r.
+    .AA......................r.
+    .AA........wwwwww........r.
+    ......B...wwwwwwww.......r.
+    .ii.......wwswwwww.......r.
+    .ii......wwwwwwwww.......r.
+    .........wwwwwwww........r.
+    .cc........wwww..........r.
+    .cc......................r.
```
kostmo, line 505: Have to do placement of the `rubbish enclosure` by blind reckoning.

kostmo:
Added an objective to make 16 California rolls.

data/scenarios/Challenges/Ranching/fishing.yaml
```diff
@@ -0,0 +1,544 @@
...
+        end;
+
+      checkProhibited 0;
+  - teaser: No littering
+    id: littered
+    hidden: true
+    optional: true
+    goal:
+      - |
+        You may dispose of `junk`{=tag} only in either of the designated `rubbish enclosure`{=structure}s.
```
byorgey, line 88: This is indeed less confusing now, but instead of being confusing it is simply cryptic.  None of the first few goals say anything about not littering, and if you happen to place litter on the ground it just immediately says "sorry, the scenario is not winnable!".  If you choose to keep playing after that and look at the goals, you can learn that you were not supposed to litter.  But players might not think to keep playing the scenario after losing.  I still think some of the text about not littering should also be copied into the "sushi chef" goal.

kostmo: Here's what it looks like after rebasing atop #1735.  Would that suffice?

Screenshot from 2024-02-17 15-30-03

byorgey: Yes, that looks great!  Except it should say "Catch the" not "Catch a the".

kostmo:
ping for approval @byorgey

byorgey (approved):

Issue #1415, opened 2023-08-06
Render world map regions to various formats outside of an active game


kostmo:
One use of this would be able to view `Structures` (#1138) at various stages of composition.
Other uses:
* https://github.com/swarm-game/swarm/issues/353
* https://github.com/swarm-game/swarm/issues/493
* https://github.com/swarm-game/swarm/issues/1320

I believe @xsebek has already inquired toward this in https://github.com/jtdaugherty/brick/issues/405 .
A proof-of-concept is supplied here: https://github.com/jtdaugherty/brick/issues/447

byorgey:
@kostmo Can we close this now that we have #1632?  Or was there more you had in mind here?

PR #1650, opened 2023-11-21
demo pixi.js

kostmo:
Towards #1415.

Potential candidate for a web-based world display

Uses [pixi.js](https://pixijs.com/).

# Demo

1. Run swarm: `scripts/play.sh -i scenarios/Fun/horton.yaml --autoplay --speed 6`
1. Visit http://localhost:5357/play.html

### DIFF
--- a/src/swarm-util/Swarm/Util/OccurrenceEncoder.hs
+++ b/src/swarm-util/Swarm/Util/OccurrenceEncoder.hs
@@ -2,15 +2,11 @@
 -- SPDX-License-Identifier: BSD-3-Clause
 -- Description: Compress representation of traversable
 --
--- For structures with many repeating elements,
--- will store the expanded element representation
--- in an array so that the structure can be transmitted
--- with array indices as elements.
+-- Useful for compressing the representation of a
+-- structure that has many repeating elements
+-- for transmission (e.g. over the network).
 module Swarm.Util.OccurrenceEncoder (
-  Encoder,
-  encodeOccurrence,
-  getIndices,
-  emptyEncoder,
+  runEncoder,
 ) where
 
 import Control.Monad.Trans.State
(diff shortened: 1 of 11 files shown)


src/swarm-util/Swarm/Util/OccurrenceEncoder.hs
```diff
@@ -0,0 +1,47 @@
+-- |
+-- SPDX-License-Identifier: BSD-3-Clause
+-- Description: Compress representation of traversable
+--
+-- For structures with many repeating elements,
+-- will store the expanded element representation
+-- in an array so that the structure can be transmitted
+-- with array indices as elements.
```
byorgey, line 8: This description and comment did not make much sense to me.  What is a "traversable" in this context?  What is a "structure"?  What is an "expanded element representation"?  It seems to me that this module is doing something fairly straightforward: keep track of seen items with a unique index for each, allowing us to quickly look up the index of an item or assign a new index to an item we have not seen before.  I'm guessing that you were writing these descriptions with the *intended use case* in mind, but I think prior to that we should just write a simple description of what it actually does (and then optionally include a description of how it is used).

web/play.html
```diff
@@ -0,0 +1,32 @@
...
+        doFetch(button, appView, graphics, displayWidth, displayHeight);
+      }
+
+      window.onload=()=>{
+        startLoop();
+      }
+    </script>
+  </head>
+  <body>
+    <button id="restart-button" onclick="startLoop();" style="display: block;">Restart</button>
```
byorgey, line 30: When I run the web frontend I just see an image displaying a view of the world; I don't see any kind of "Restart" button.  I'm not sure I understand whether that is the expected behavior.

byorgey (approved):

web/play.html
kostmo: Yeah, the restart button will (or is supposed to, anyway) appear only if an error is encountered.

src/swarm-util/Swarm/Util/OccurrenceEncoder.hs
kostmo: Yes, the struggle with module documentation was a sign that the API was not good.  I've simplified it a bit.

PR #1652, opened 2023-11-22
group robot info into subrecord

kostmo:
Add another subrecord to `GameState` dedicated to robot fields.  This shrinks the size of `GameState` by 8 more fields.  It's getting reasonable now!

The `_viewCenter*` fields were put there too, for now, because the manually-defined `setter` of the `viewCenterRule` lens needs access to the `robotMap` field.


byorgey (approved):
:tada:

PR #1653, opened 2023-11-22
dedicated record for randomness

kostmo:
Reduces the number of fields in the toplevel `GameState` by one.


byorgey:
Good idea!

byorgey (approved):

kostmo:
@mergifyio refresh

PR #1665, opened 2023-11-26
Split `Step.hs`

kostmo:
Both `execConst` and `stepCESK` are huge case statements.  This refactor puts them in different modules.

Also many supporting functions for implementing commands in `execConst` are moved to their own modules.

Whereas `Step.hs` previously had `2931` lines, the new linecounts are:

| File | lines |
| --- | --- |
| `Step.hs` | 848 |
| `Arithmetic.hs` | 124 |
| `Const.hs` | 1687 |
| `Command.hs` | 417 |

The only tricky thing was that `execConst` is mutually recursive with `runCESK`.  So to get them into different modules, I pass a wrapper of `runCESK` as an argument to `execConst`.


byorgey:
I'm all for splitting out the `Step` module into multiple modules!  However, splitting two mutually recursive functions across modules by passing one as an argument to the other makes me slightly worried about performance, though I very much hope it is an unfounded worry.  Do we have any benchmarking results comparing before + after the refactoring?

kostmo:
> Do we have any benchmarking results comparing before + after the refactoring?

In 5ac3448 I have introduced some tooling:

    ./scripts/benchmark-against-parent.sh

The benchmarks are a bit noisy, but within 2-3 re-runs I can get it to "pass"---that is, none of the benchmarks take more than 3% longer.

The changes to `swarm.cabal` are by recommendation of the [`tasty-bench` docs](https://hackage.haskell.org/package/tasty-bench-0.3.5#comparison-against-baseline).

byorgey (approved):
Thanks! Works for me!

PR #1672, opened 2023-12-04
robot background passthrough

### LINKED ISSUE #1662: Robots and entities should take on underlying cell background unless overridden

kostmo:
Closes #1662

## Demo

Using:

    scripts/play.sh -i creative --seed 2

| Before | After |
| --- | --- |
| image | Screenshot from 2023-12-03 23-37-18 |

Also:

    scripts/play.sh --scenario data/scenarios/Testing/1034-custom-attributes.yaml --autoplay --speed 2 --hide-goal

image

### DIFF
--- a/src/swarm-scenario/Swarm/Game/Display.hs
+++ b/src/swarm-scenario/Swarm/Game/Display.hs
@@ -25,6 +25,7 @@ module Swarm.Game.Display (
   boundaryOverride,
   displayAttr,
   displayPriority,
+  displayObscured,
   invisible,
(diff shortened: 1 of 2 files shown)


byorgey (commented):
I am not sure why, but with this PR it seems like all robots now look like `Ω`.

What is the advantage of doing this *only* for robots (as opposed to for entities as well)?  It seems like we needed to add various special cases (`displayObscured`, etc.) to get it to work only for robots, rather than just generically changing the way `Display`s combine.

PR #1674, opened 2023-12-04
use strictly background color for terrain

kostmo:
Towards #1662

Prerequisite for #1672

To allow a robot-occupied cell to take on the underlying terrain color as the background color of its cell, the terrain color must be representable strictly as a "background".  However, currently, the `dirt`, `stone`, and `grass` terrains are represented by a half-shaded **foreground** glyph upon `black` background.

Currently the ["Medium Shade" unicode character](https://www.compart.com/en/unicode/U+2592) (`▒`) is used  to "blend" a somewhat bright color with a black background, resulting in a moderately dark color in the terminal for `dirt`, `stone`, and `grass`.  However, these same dark colors are not reproducible in the 240-color scheme without this foreground+background blending trick; the closest approximations as a background-only or foreground-only color come out quite a bit lighter.

## Visual comparison

Using:

    scripts/play.sh -i creative --seed 2 --autoplay

| Before | After |
| --- | --- |
| Screenshot from 2023-12-03 23-33-15 | Screenshot from 2023-12-03 23-32-36 |

## Possible approaches

So, we need to decide whether to:
1. Accept the new lighter colors
2. Choose new, alternative terrain colors that may be darker given the 240-color palette
3. Abandon support for 240 colors and assume "full" color terminals
4. Abandon efforts to passthrough terrain color as background of robot cells

### DIFF
--- a/src/Swarm/App.hs
+++ b/src/Swarm/App.hs
@@ -97,7 +97,7 @@ appMain opts = do
 
       let logP p = logEvent SystemLog Info "Web API" ("started on :" <> T.pack (show p))
       let logE e = logEvent SystemLog Error "Web API" (T.pack e)
-      let s' =
+      let s1 =
             s
               & runtimeState
                 %~ case eport of
@@ -121,8 +121,14 @@ appMain opts = do
 
       V.setMode (V.outputIface vty) V.Mouse True
 
+      let cm = V.outputColorMode $ V.outputIface vty
+      let s2 =
+            s1
+              & runtimeState
+                %~ (eventLog %~ logEvent SystemLog Info "Graphics" ("Color mode: " <> T.pack (show cm)))
+
       -- Run the app.
-      void $ customMain vty buildVty (Just chan) (app eventHandler) s'
+      void $ customMain vty buildVty (Just chan) (app eventHandler) s2
 
 -- | A demo program to run the web service directly, without the terminal application.
 --   This is useful to live update the code using @ghcid -W --test "Swarm.App.demoWeb"@.
--- a/src/Swarm/App.hs
+++ b/src/Swarm/App.hs
@@ -124,8 +124,7 @@ appMain opts = do
       let cm = V.outputColorMode $ V.outputIface vty
       let s2 =
             s1
-              & runtimeState
-                %~ (eventLog %~ logEvent SystemLog Info "Graphics" ("Color mode: " <> T.pack (show cm)))
+              & runtimeState . eventLog %~ logEvent SystemLog Info "Graphics" ("Color mode: " <> T.pack (show cm))
 
       -- Run the app.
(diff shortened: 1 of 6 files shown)


byorgey:
I don't think I like the new lighter colors, not just on aesthetic grounds but also functional grounds: if the goal is to pass through the background color behind entities and robots, having bright background colors is going to make it very difficult to see.

What would be the downside of abandoning 240-color support and assuming full color terminals?

kostmo:
> What would be the downside of abandoning 240-color support and assuming full color terminals?

Actually, this was simpler than I thought.  We had been using the [`rgbColor`](rgbColor) function; I just swapped it for [`linearColor`](https://hackage.haskell.org/package/vty-6.1/docs/Graphics-Vty-Attributes-Color.html#v:linearColor).  I reverse-engineered the precise full-color RGB triples for the half-shade darkened terrain using the Gimp color picker.

The new colors look like so:
Screenshot from 2023-12-08 17-58-44

byorgey:
Ah, that looks very nice indeed!

src/Swarm/App.hs
```diff
@@ -121,8 +121,14 @@ appMain opts = do
 
       V.setMode (V.outputIface vty) V.Mouse True
 
+      let cm = V.outputColorMode $ V.outputIface vty
+      let s2 =
+            s1
+              & runtimeState
+                %~ (eventLog %~ logEvent SystemLog Info "Graphics" ("Color mode: " <> T.pack (show cm)))
```
byorgey, line 128:
```suggestion
              & runtimeState . eventLog %~ logEvent SystemLog Info "Graphics" ("Color mode: " <> T.pack (show cm))
```

src/Swarm/Game/Terrain.hs
```diff
@@ -59,9 +59,9 @@ getTerrainWord = T.toLower . T.pack . init . show
 terrainMap :: Map TerrainType Display
 terrainMap =
   M.fromList
-    [ (StoneT, defaultTerrainDisplay '▒' (ATerrain "stone"))
-    , (DirtT, defaultTerrainDisplay '▒' (ATerrain "dirt"))
-    , (GrassT, defaultTerrainDisplay '▒' (ATerrain "grass"))
+    [ (StoneT, defaultTerrainDisplay ' ' (ATerrain "stone"))
```
byorgey, line 62: I was going to suggest that we remove the first `Char` argument from the `defaultTerrainDisplay` function.  But I suppose it could still be useful to define terrain types in the future which are displayed as some non-space character?

byorgey (approved):
Looks great!

src/Swarm/Game/Terrain.hs
kostmo: Agreed it should be removed.

PR #1675, opened 2023-12-05
split utils into sublibrary

kostmo:
Towards #1043

This is the first of potentially more granular sub-library splits.


byorgey (approved):
I think this looks good, and it proves that some of my worries were unfounded: everything stays in the same `.cabal` file so we still only have to upload to Hackage once, etc.

One thing to keep in mind is that we have to add `visibility: public` to any library components which we want to be publicly accessible.  I don't think that matters much for this utils sublibrary (no one is going to import `swarm` in order to use our utility modules) but it may definitely matter if we continue splitting e.g. language stuff into a sublibrary etc.

Issue #1268, opened 2023-05-18
Devices/entities with no recipe


byorgey:
This is a new umbrella issue for entities or devices without a recipe.  Eventually we should create recipes so the entities are craftable in classic mode.

- [x] `detonator` #1017 
- [x] `dozer blade` #2028 
- [x] `hourglass` #2023 
- [x] `olfactometer` #2028 
- [x] `rolex` #2028 

See #26 for an older version of this issue.

xsebek:
Just so we don't miss any, we could have integration tests that all entities are craftable from entities present in the classic world.

Those named here could be an exception marked with “expected fail”. Adding recipes would then give the good feeling of making tests green. 😄

xsebek:
I am surprised we don’t have hourglass made from sand, wood and glass, which we all have.

Not sure if that seemed too easy for `wait`?

byorgey:
> we could have integration tests that all entities are craftable from entities present in the classic world.

That sounds like a great idea!

> I am surprised we don’t have hourglass made from sand, wood and glass, which we all have.  Not sure if that seemed too easy for wait?

No, that sounds good to me!  I don't think it needs to be that difficult to obtain `wait`.

PR #1676, opened 2023-12-06
unit test for recipe coverage

kostmo:
Towards #1268

## Demo

    scripts/run-tests.sh --test-arguments '--pattern "Recipe coverage"'


test/unit/TestRecipeCoverage.hs
```diff
@@ -0,0 +1,48 @@
...
+      T.unpack $
+        T.unwords
+          [ "Missing recipes for:"
+          , commaList $ map quote $ Set.toList nonCoveredEntities
+          ]
+
+    -- Only include entities that grant a capability:
+    entityNames = Set.fromList . map (^. entityName) . concat . M.elems . entitiesByCap $ rs ^. stdEntityMap
+
+    recipesByOutputEntity = binTuples $ concatMap (\r -> map (,r) $ getOutputsForRecipe r) $ rs ^. stdRecipes
```
byorgey, line 46: Why bother constructing a map here by tupling with the recipe and using `binTuples`, when we're just going to later keep only the keys?  Why not just
```
outputEntities = S.fromList . concatMap getOutputsForRecipe $ rs ^. stdRecipes
nonCoveredEntities = Set.difference entityNames outputEntities
```

byorgey (approved):
Looks good, this is a great idea.

Sorry I've been unresponsive recently, I've been slammed with end-of-the-semester grading and student meetings.

PR #1683, opened 2023-12-13
pushable property

### LINKED ISSUE #1681: Introduce 'pushable' entity property
Currently, an entity may be `push`ed if that entity is `portable`:

https://github.com/swarm-game/swarm/blob/1bcd050780d51e70401749dc0bf40c542237e0db/src/Swarm/Game/Step/Const.hs#L190

For scenario authors to force use of the `push` command rather than `grab`, two workarounds are available:
* Make the entity `unwalkable`, so that one's robot may not occupy its cell to `grab` it
* Do not provide a device in the scenario that is capable of `grab`ing.

However, this is limiting; one may like to create a scenario with `walkable` entities that require `push`ing but do not permit `grab`-ing, without restricting availability of a `harvester`/`fast grabber`/`grabber`.

We can introduce a new `pushable` property of entities that allows `push` without necessarily allowing the entity to be `portable`.

kostmo:
Closes #1681

### DIFF
--- a/data/scenarios/Testing/00-ORDER.txt
+++ b/data/scenarios/Testing/00-ORDER.txt
@@ -53,3 +53,4 @@ Achievements
 1575-structure-recognizer
 1631-tags.yaml
 1634-message-colors.yaml
+1681-pushable-entity.yaml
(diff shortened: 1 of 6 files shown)


byorgey (approved):
Looks good to me!

PR #1686, opened 2023-12-14
dim sum restaurant

kostmo:
scripts/play.sh -i scenarios/Challenges/dimsum.yaml --autoplay

Screenshot from 2023-12-13 22-39-53

### DIFF
--- a/data/entities.yaml
+++ b/data/entities.yaml
@@ -1289,7 +1289,7 @@
     `watch : dir -> cmd unit` will mark an adjacent (in the specified direction) location of interest to monitor for placement or removal of items.
     A subsequent call to `wait` will be interrupted upon a change to the location.
   properties: [portable]
-  capabilities: [wakeself]
+  capabilities: [timerel, wakeself]
 
 - name: comparator
   display:
--- a/data/entities.yaml
+++ b/data/entities.yaml
@@ -1284,7 +1284,7 @@
     char: R
     attr: silver
   description:
-  - Enables robots to use the `watch` command.
+  - Enables robots to use the `watch` and `wait` commands.
   - |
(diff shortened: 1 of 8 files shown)


data/entities.yaml
```diff
@@ -1289,7 +1289,7 @@
     `watch : dir -> cmd unit` will mark an adjacent (in the specified direction) location of interest to monitor for placement or removal of items.
     A subsequent call to `wait` will be interrupted upon a change to the location.
   properties: [portable]
-  capabilities: [wakeself]
+  capabilities: [timerel, wakeself]
```
byorgey, line 1292: If we're going to add the `timerel` capability to the `watch` device (which is fine) then we should also mention it in the description (i.e. "Enables robots to use the `watch` and `wait` commands").

data/scenarios/Challenges/_dimsum/patron.sw
```diff
@@ -0,0 +1,24 @@
+def modifyCart = \cartType. \device.
```
byorgey, line 1: It irks me that we have to duplicate code like this.  I am very motivated to work on #495 .

data/scenarios/Challenges/_dimsum/solution.sw
```diff
@@ -0,0 +1,27 @@
+def doN = \n. \f. if (n > 0) {f; doN (n - 1) f} {}; end;
```
byorgey, line 1: Notice that the solution uses recursion, but the base robot does not have a `strange loop`.  This should not be allowed but currently we can get away with it because of #231 .  I forgot about this bug and assumed that I was not allowed to use recursion in my solution!  But we should add a `strange loop` and a `lambda` to the `base` inventory.

byorgey (approved):
Looks good!  My solution was very similar to the "official" solution except I thought I had to do it without using recursion --- see my comment. :smile:

PR #1688, opened 2023-12-14
Grant achievement for crafting a bitcoin

byorgey:
Grant `CraftedBitcoin` achievement when using the `make` command to craft a `bitcoin`,  not in creative mode, and not a system robot.

Towards #1435 .


src/Swarm/Game/Step/Const.hs
```diff
@@ -422,6 +423,9 @@ execConst runChildProg c vs s k = do
         -- take recipe inputs from inventory and add outputs after recipeTime
         robotInventory .= invTaken
         traverse_ (updateDiscoveredEntities . snd) (recipe ^. recipeOutputs)
+        -- Grant CraftedBitcoin achievement
+        when (name == "bitcoin" && not creative && not sys) $ grantAchievement CraftedBitcoin
```
kostmo, line 427: I think I would like to make `ValidInCreativeMode` and `ValidForSystemRobot` into properties of the achievements themselves.  Then the `grantAchievement` function would perform the verification (perhaps with the addition of a `HasRobotStepState` constraint).

kostmo (approved):

Issue #1689, opened 2023-12-14
Module guide


byorgey:
There are so many modules now that I often have real trouble finding the right place to make the changes I want to make.  It would be nice to have some kind of developer guide on the wiki which explains the organization of the codebase and gives some general hints about what can be found in various modules (though we don't necessarily need to list every single module).  The challenge, of course, would be keeping it up to date.

kostmo:
In addition to documentation, #1043 will surely help with this.

byorgey:
I started working on this on the wiki:

https://github.com/swarm-game/swarm/wiki/Module-guide

There's still quite a ways to go, and it's probably already a bit out of date...

PR #1690, opened 2023-12-15
short module descriptions

kostmo:
Towards #1689

The [`Description:` field](https://haskell-haddock.readthedocs.io/en/latest/markup.html#the-module-description) in a module's toplevel haddock will be displayed on the Contents page.

Screenshot from 2023-12-14 20-55-26


byorgey (approved):
Great idea!  I didn't know about that field!

PR #1697, opened 2023-12-30
web sublibrary

kostmo:
Towards #1043

Creates a new sub-library for the web service.  Also moved the `Swarm.App` module from the `swarm` library to the `swarm` executable.


swarm.cabal
```diff
@@ -309,22 +338,22 @@ library
     build-depends:    base                          >= 4.14 && < 4.19,
                       brick-list-skip               >= 0.1.1.2 && < 0.2,
                       AhoCorasick                   >= 0.0.4 && < 0.0.5,
-                      aeson                         >= 2 && < 2.2,
+                      aeson,
```
byorgey, line 341: Why are the version bounds removed from a bunch of libraries?

kostmo: The ones that I removed have specified bounds in the sub libraries depended upon, e.g. in `swarm-util`.

byorgey: OK, I understand not wanting to have duplicate bounds information specified, but wouldn't it be better to specify all the bounds in one canonical place, instead of having them spread out over a bunch of different sublibraries?  Or is there a reason that wouldn't work?

kostmo: Actually I assumed, for whatever reason, that the bounds must be taken from the furthest "upstream" dependency.  Since `swarm-util` is "upstream" from the `swarm` library in the DAG of sub-library dependencies, I figured that `swarm-util` would be the authoritative source of bounds.

But perhaps it works such that constraints anywhere within the dependency tree may be authoritative, and that sub-libraries that are *either* dependencies *or* dependents will use those constraints?  If that is the case, then I agree we should try to put as many version bounds in the "main" `swarm` sub-library as the canonical place.

But for now I have reverted to their original, redundantly-specified state to minimize this diff.

byorgey: Hmm, yeah, I don't actually know.  I agree reverting for now makes sense and we can look into it more later.

byorgey (approved):

PR #1699, opened 2023-12-31
snake game

kostmo:
Demos use of a string to maintain a queue of coordinates.

Screenshot from 2024-01-02 10-34-05

Issue #996, opened 2023-01-10
Add a programmatic way to get recipes


xsebek:
It would be neat to be able to query the recipe requirements in Swarm language.

This would simplify creating an automated factory where robots are sent to fetch recipes dynamically.

Another use case is writing a helper function to list all requirements for a recipe:

```
>>> showMeWhatINeed "branch"
- 2 branch:
  - 1 tree
```

## Solution

I would like to have a command like this one:
```
recipe: entity -> cmd (entity * CmdFold (int * entity))

type CmdFold a = (a -> cmd unit) -> cmd unit
```

The `CmdFold` is basically a substitution for a list and can be inlined.

### Manual solution

It's certainly possible to write this function by hand, but it will get outdated.

### Scripted solution

A function generated by a script is easier to keep up to date but still would not work in challenges that add new recipes.

Also, it takes away from the fun of discovering recipes.

### Returning a list

This would be definitely more useful as the result could be used in a pure context. E.g. not take extra ticks in `atomic`.

byorgey:
What should this command do if you run it on an entity you don't "know" yet?  Should it throw an exception?

kostmo:
I'd like to understand the `CmdFold` idea more.  Does this require functionality that is currently missing?  Does it relate to the `cons`/`nil` technique demonstrated [here](https://github.com/swarm-game/swarm/pull/1214/files#r1174451625)?

Another approach for iterating list members, which seems to work well enough, is that used by the [`waypoint` and `structure` commands](https://github.com/swarm-game/swarm/blob/f9ef0094f0bfbe57d589e44a355446a54566d4bd/src/Swarm/Language/Typecheck.hs#L776-L777).  That is, take an index as input, and return a tuple of `(list length, item at index)`.

An API for recipe members might also be used for querying the materials required for structures (#1579).

byorgey:
A `CmdFold a = (a -> cmd unit) -> cmd unit` represents a list of `a`s, in the sense that you can hand it any function of type `(a -> cmd unit)` you want, and the only thing it can do is call your function on some collection of `a` values and sequence the resulting commands.  So if you want to *e.g.* run some command for each ingredient of a recipe you would get the `CmdFold (int * entity)` and then pass it your `(int * entity -> cmd unit)` function, and your function would get run on each ingredient.

[...]

PR #1702, opened 2024-01-02
[WIP] caterpillar

kostmo:
Demo of functions-as-lists and their limitations

### DIFF
--- a/data/scenarios/Fun/00-ORDER.txt
+++ b/data/scenarios/Fun/00-ORDER.txt
@@ -1,3 +1,4 @@
 GoL.yaml
 logo-burst.yaml
 horton.yaml
+caterpillar.yaml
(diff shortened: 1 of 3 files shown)


byorgey:
How is this related to #1699 ?

kostmo:
> How is this related to #1699 ?

I had first tried to implement a queue for the snake's tail using [`cons` and `nil` functions](https://github.com/swarm-game/swarm/pull/1214/files#r1174451625), with inspiration from https://github.com/swarm-game/swarm/issues/996#issuecomment-1811574210 .  I then saw for myself how the swarm typechecker chokes when attempting to mutate the list.

I may turn this PR into a testcase for `cons`/`nil`, or maybe craft a different scenario in which the limited support for `cons`/`nil` (i.e. append-only lists) is still enough to solve a challenge.

PR #1705, opened 2024-01-03
sort module list

kostmo:
As title


byorgey (approved):
Is it worth adding something to CI (in a separate PR) to check that the module list is sorted?  (I'm not sure that it is, just thought it was worth mentioning.)

kostmo:
> Is it worth adding something to CI (in a separate PR) to check that the module list is sorted? (I'm not sure that it is, just thought it was worth mentioning.)

#1709

PR #1711, opened 2024-01-04
normalize scenarios

### LINKED ISSUE #845: Normalizing .yaml files

kostmo:
Closes #845

**All changes are non-significant whitespace.**  This can be verified with:

    git show --ignore-all-space --ignore-blank-lines

in which the remaining changes are only elimination of manual word-wrap in descriptions.

Normalization is accomplished with this command:

    scripts/normalize-all-scenarios.sh

This is an initial normalization pass that shall be a pre-requisite for #1713.


kostmo:
Need your stamp on this one too, in case you missed it @byorgey

byorgey (approved):
Oops, I did miss it, thanks for the reminder!

PR #1714, opened 2024-01-05
extract game engine sublibrary

kostmo:
Towards #1043

Extracts `Swarm.Game.*` modules to their own sublibrary.

There was already pretty good separation along this boundary; just had to move three functions into a new module `Swarm.Util.Content`.


src/swarm-engine/Swarm/Util/Content.hs
```diff
@@ -0,0 +1,56 @@
+-- |
+-- SPDX-License-Identifier: BSD-3-Clause
+module Swarm.Util.Content where
```
byorgey, line 3: Can we add a short comment explaining what kinds of things can be found in this module?

src/swarm-engine/Swarm/Util/Content.hs
```diff
@@ -0,0 +1,56 @@
...
+import Swarm.Game.Entity.Cosmetic.Assignment (terrainAttributes)
+import Swarm.Game.Scenario.Topography.Area qualified as EA
+import Swarm.Game.Scenario.Topography.Cell (PCell (..))
+import Swarm.Game.Scenario.Topography.EntityFacade
+import Swarm.Game.Terrain (TerrainType, getTerrainWord)
+import Swarm.Game.Universe
+import Swarm.Game.World
+import Swarm.Util.Erasable (erasableToMaybe, maybeToErasable)
+
+getMapRectangle ::
```
byorgey, line 20: While we're at it, some Haddock comments on these functions would be nice too.

byorgey (approved):
Should we at this point also edit the description in the `.cabal` file to talk about the sublibraries instead of saying "roughly the following layers"?

PR #1718, opened 2024-01-05
enforce scenario normalization

### LINKED ISSUE #1713: Enforce scenario normalization in CI
See https://github.com/swarm-game/swarm/issues/845#issuecomment-1309619014:
> If we do this, i.e. standardize on some normalized form, then I think going forward we would need to have something in our CI that requires it.

kostmo:
Closes #1713.


data/recipes.yaml
```diff
@@ -202,11 +202,10 @@
     - [10, solar panel]
   out:
     - [1, big furnace]
+## VEINS
```
byorgey, line 205: Putting this `## VEINS` comment here does not make sense, it is supposed to be a subsection of the `METAL` section.  Does the normalization script do this?

kostmo: That was me resolving the normalization change incorrectly.  I guess it's fixed now?

byorgey (approved):

PR #1721, opened 2024-01-08
whitelist walkable entities

kostmo:
A walkability "blacklist" had already been implemented in #1536.  The whitelist shall **only** permit walking on the specified entities; blank cells become unwalkable.

This feature would allow use of the `path` command to check for "connectivity" by way of a "trail" of entities.

## Use cases

* system robots or goal conditions can use this to check whether the player has connected two points with a road, cable, aqueduct, etc.
* monkeys, which can only move along `tree`s
* sea life, which should only be able to move in `water`

## Testing

    scripts/run-tests.sh --test-arguments '--pattern "walkable-entities"'

### DIFF
--- a/src/swarm-engine/Swarm/Game/Step/Path/Walkability.hs
+++ b/src/swarm-engine/Swarm/Game/Step/Path/Walkability.hs
@@ -12,7 +12,7 @@ import Swarm.Language.Capability
 
 data MoveFailureMode
   = -- | If the robot has a path Whitelist,
-    -- then the absence of an entity prevents walkability.
+    -- then the absence of an entity could prevent walkability (represented by `PathBlockedBy Nothing`).
     PathBlockedBy (Maybe Entity)
   | PathLiquid Entity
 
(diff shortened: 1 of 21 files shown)


src/swarm-engine/Swarm/Game/Step/Path/Walkability.hs
```diff
@@ -7,28 +7,35 @@ module Swarm.Game.Step.Path.Walkability where
...
-data MoveFailureMode = PathBlocked | PathLiquid
-
-data MoveFailureDetails
-  = MoveFailureDetails
-      -- | Occupies the destination cell
-      Entity
-      MoveFailureMode
+data MoveFailureMode
+  = -- | If the robot has a path Whitelist,
+    -- then the absence of an entity prevents walkability.
```
byorgey, line 15:
```suggestion
    -- then the absence of an entity could prevent walkability (represented by `PathBlockedBy Nothing`).
```

byorgey (approved):
Nice!

Issue #1663, opened 2023-11-26
Built robots should inherit display properties from parent


kostmo:
Most notably, the status quo is that children built by invisible system robots are automatically **visible**, and their "invisible" property cannot be set.  It doesn't quite work to use the `appear " "` command to hide them, because their background will still be "blank", and hence visible (see #1662).

I think that the only existing scenario that even entails a system robot that `build`s children is #1599.  Hence the relative recency of #1430 being discovered and fixed.

Also related:
* #1230
* #1664

## Implementation

This should be a simple change [here](https://github.com/swarm-game/swarm/blob/44c2e607b34f80b355ec681712b6dacfdf377b73/src/Swarm/Game/Step.hs#L1984).

PR #1722, opened 2024-01-09
Child robots only inherit their parent's `displayAttr`

### LINKED ISSUE #1693: Built robots all look like `Ω`
**Describe the bug**
Robots built by the `base` look like `Ω`.

**To Reproduce**
Start a classic game, and execute `build {move}`.  Observe that the new robot looks like `Ω`.

**Expected behavior**
The built robot should look like `^`.

**Screenshots**
build-omega

byorgey:
Fixes #1693.  It's not necessarily appropriate to copy the entire `Display` record.  So instead, we explicitly list the fields that should be inherited.  For now, it is only `displayAttr`, but we could add `invisible` in the future (see #1670, #1663).

Tested to make sure behavior is preserved with:
```
scripts/play.sh -i scenarios/Challenges/Ranching/beekeeping.yaml --autoplay
```


src/swarm-util/Swarm/Util/Lens.hs
```diff
@@ -40,3 +43,7 @@ makeLensesExcluding exclude =
         & lensField . mapped . mapped %~ \fn n ->
           if n `elem` exclude then [] else fn n
     )
+
+-- | Copy a given field from one record to another.
+inherit :: Lens' s a -> s -> (s -> s)
```
kostmo, line 48: Neat.

kostmo (approved):

PR #1724, opened 2024-01-11
`infinite improbability drive` device enabling `teleport` command

### LINKED ISSUE #1041: device to provide `teleport` capability
Some kind of space-time machine?  See some discussion at https://github.com/swarm-game/swarm/pull/406 .

byorgey:
* `tea plant` (which now spawns rarely in a certain biome of the classic world) can be harvested to yield...
* `tea leaves`, which can be combined with `water` to produce...
* `cup of tea`, which, along with `atomic vector plotter` (= `rubber band` + `typewriter` + `compass` + `counter`), are ingredients for...
* `infinite improbability drive`, which enables `teleport`.

When used by a privileged robot (system robot or in creative mode), the `teleport` command continues to function as before.  When used by an unprivileged robot, that is, via the `infinite improbability drive` device, a random entity spawns somewhere near the target location.

Closes #1041.


src/swarm-engine/Swarm/Game/Location.hs
```diff
@@ -205,6 +207,17 @@ manhattan (Location x1 y1) (Location x2 y2) = abs (x1 - x2) + abs (y1 - y2)
...
+-- | Get all the locations that are within a certain manhattan
+--   distance from a given location.
+--
+-- >>> getLocsInArea (P (V2 0 0)) 1
+-- [P (V2 0 0), P (V2 0 1), P (V2 0 (-1)), P (V2 1 0), P (V2 1 1), P (V2 1 (-1))]
+-- >>> map (\i -> length (getLocsInArea origin i)) [0..8]
+-- [1,5,13,25,41,61,85,113,145]
+getLocsInArea :: Location -> Int32 -> [Location]
+getLocsInArea loc r =
+  [loc .+^ V2 dx dy | x <- [0 .. 5], y <- [0 .. r - x], dx <- nub [x, -x], dy <- nub [y, -y]]
```
kostmo, line 219: > `x <- [0 .. 5]`

Where does the number `5` come from?

kostmo: May also be worth making mention of `Swarm.Game.Step.Const.genDiamondSides`.

byorgey: > Where does the number `5` come from?

Oops!  Good catch, I abstracted this from some specific code I wrote for teleport, and missed changing that 5 to `r`.  I wonder why the tests passed?  Maybe those doctests aren't getting run or something?

Thanks for the reference to `genDiamondSides`, I'll add a note.

kostmo (approved):

byorgey:
I verified manually that the doctests are now passing, but it looks like our CI is not running all of them.  I will open a new issue.

PR #1729, opened 2024-01-14
Use type family for robot CESK machine field

kostmo:
Towards #1715 and #1043.

This refactoring step is a prerequisite for #1719 to extricate references to the `CESK` module from the base `RobotR` definition.

In this PR:
* `Swarm.Game.CESK` is imported qualified to more easily track usages
* A new `RobotMachine` type family is added to parameterize the `_machine` field.
* `CESK` is a new parameter to `addTRobot`

### DIFF
--- a/src/swarm-engine/Swarm/Game/Robot.hs
+++ b/src/swarm-engine/Swarm/Game/Robot.hs
@@ -21,7 +21,7 @@ module Swarm.Game.Robot (
   TRobot,
 
   -- ** Runtime robot update
-  RobotUpdate (..),
+  C.RobotUpdate (..),
 
   -- * Robot context
   RobotContext,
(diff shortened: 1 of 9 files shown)


src/swarm-engine/Swarm/Game/Robot.hs
```diff
@@ -603,16 +616,13 @@ instance FromJSONE EntityMap TRobot where
       <*> liftE (v .:? "loc")
       <*> liftE (fmap getHeading $ v .:? "dir" .!= HeadingSpec zero)
       <*> localE (const defDisplay) (v ..:? "display" ..!= defDisplay)
-      <*> liftE (mkMachine <$> (v .:? "program"))
+      <*> liftE (v .:? "program")
```
byorgey, line 619: Yes, this is perfect!  It makes complete sense to me to store a `ProcessedTerm` in a `TemplateRobot` and only turn it into a CESK machine when instantiating it to a `ConcreteRobot`.  One of those things that seems completely obvious in retrospect.

byorgey (approved):

PR #1730, opened 2024-01-16
split module for CESK store

kostmo:
Towards #1715 and #1043.

A follow-on to #1729; this refactoring step is a prerequisite for #1719 to extricate references to the `CESK` module from the base `RobotR` definition.


byorgey:
So is the idea that by splitting out this module we can keep `Swarm.Game.CESK.Store` in the scenario construction library with `Robot`, and `Swarm.Game.CESK` would go in the `swarm-engine` library?  I don't like that, since, again, `Store` really has nothing to do with scenario construction and everything to do with the runtime system, and there's no reason a template robot should ever need to have a `Store`.  In other words I'm not sure this is the correct place to do the split.

Looking at the definition of `RobotContext`, I think perhaps the `Store` really doesn't belong there.  Everything else in the `RobotContext` stores information about defined names; notice the `At` instance, which returns the type, value, and requirements for a name (using the first three components of a `RobotContext`) but has nothing at all to do with the `defStore`.  Although it would be more work, I think perhaps the correct way to do this is to split the `defStore` out of the `RobotContext` and store it in the `Robot` record separately, but using a type family so that it stores `()` in a `TRobot` and an actual `Store` in a concrete `Robot`.  There is no reason a template robot should ever need to have a `Store`.

By the way, all this refactoring is annoying but I'm actually really happy with this process so far, and I really appreciate your work here.  Splitting into sublibraries like this is really forcing us to think through some things that are too tightly coupled, and it's resulting in refactored code that is cleaner and more sensible.

PR #1731, opened 2024-01-16
Remove superfluous trobotContext lens

### LINKED ISSUE #394: Inner build does not correctly require Drill

kostmo:
# Motivation

Want to remove dependence of `TRobot` on the `RobotContext` record.  However, a lens `trobotContext` had been added in #817 to fix #394.

# Test

    scripts/run-tests.sh --test-arguments '--pattern "394-build-drill"'

### DIFF
--- a/src/swarm-engine/Swarm/Game/Step/Const.hs
+++ b/src/swarm-engine/Swarm/Game/Step/Const.hs
@@ -1061,7 +1061,7 @@ execConst runChildProg c vs s k = do
         -- Construct the new robot and add it to the world.
         parentCtx <- use robotContext
         newRobot <-
-          zoomRobots . fmap (robotContext .~ parentCtx) . addTRobot (In cmd e s [FExec]) $
+          zoomRobots . addTRobotWithContext parentCtx (In cmd e s [FExec]) $
             mkRobot
               ()
               (Just pid)
--- a/src/swarm-engine/Swarm/Game/Step/Const.hs
+++ b/src/swarm-engine/Swarm/Game/Step/Const.hs
@@ -1061,7 +1061,7 @@ execConst runChildProg c vs s k = do
         -- Construct the new robot and add it to the world.
         parentCtx <- use robotContext
         newRobot <-
-          zoomRobots . addTRobot (In cmd e s [FExec]) . (trobotContext .~ parentCtx) $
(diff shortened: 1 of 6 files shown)


src/swarm-engine/Swarm/Game/Step/Const.hs
```diff
@@ -1061,7 +1061,7 @@ execConst runChildProg c vs s k = do
         -- Construct the new robot and add it to the world.
         parentCtx <- use robotContext
         newRobot <-
-          zoomRobots . addTRobot (In cmd e s [FExec]) . (trobotContext .~ parentCtx) $
+          zoomRobots . fmap (robotContext .~ parentCtx) . addTRobot (In cmd e s [FExec]) $
```
byorgey, line 1064: I think the reason this doesn't work is that it only modifies the `robotContext` of the `Robot` value returned by `addTRobot`, but it does not modify the original `Robot` value which `addTRobot` actually adds to the robot map (of which it then returns a copy).

byorgey: Maybe we could just have `addTRobot` take a `Maybe RobotContext` as an extra argument?  I don't know if that's actually a good idea, just random brainstorming...

byorgey (approved):
Ah, wonderful.  I like how this uncovers the fact that the only time we really needed the returned `Robot` was when we were actually setting the context, so now in the common case we can have `addTRobot` return `()` and get rid of a bunch of calls to `void`.

PR #1732, opened 2024-01-16
Use type family for RobotContext field

kostmo:
Continuation of #1731.

Towards #1715 and #1043.

This refactoring step is a prerequisite for #1719 to extricate references to the `CESK` module from the base `RobotR` definition.

# In this PR:

* Extracts the `RobotContext` type to a new module
* Introduces a type family for the `RobotContext` field


byorgey (approved):
Makes sense.  Sorry for the delay in looking at this, the semester just started and we had snow and super low temps that caused the college and public schools to be closed for a few days so it has been a crazy week :sweat_smile:

PR #1737, opened 2024-01-24
simplify infinite entity grabbing logic

kostmo:
Refactoring towards #1598

### DIFF
--- a/src/swarm-engine/Swarm/Game/Step/Const.hs
+++ b/src/swarm-engine/Swarm/Game/Step/Const.hs
@@ -1688,13 +1688,11 @@ execConst runChildProg c vs s k = do
     (omni || e `hasProperty` Pickable)
       `holdsOrFail` ["The", e ^. entityName, "here can't be", verbed <> "."]
 
-    -- Remove the entity from the world.
-    updateEntityAt loc (const Nothing)
-    flagRedraw
-
-    -- Immediately regenerate entities with 'infinite' property.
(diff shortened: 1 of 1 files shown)


src/swarm-engine/Swarm/Game/Step/Const.hs
```diff
@@ -1688,13 +1688,11 @@ execConst runChildProg c vs s k = do
...
 
-    -- Remove the entity from the world.
-    updateEntityAt loc (const Nothing)
-    flagRedraw
-
-    -- Immediately regenerate entities with 'infinite' property.
-    when (e `hasProperty` Infinite) $
-      updateEntityAt loc (const (Just e))
+    -- Entities with 'infinite' property are not removed
+    unless (e `hasProperty` Infinite) $ do
```
byorgey, line 1692: I think the only difference between these is that previously, any robot with a `watch` registered on an infinite entity that is harvested would be woken up, but with the refactored code they will not be woken up, is that correct?  We probably don't rely on that anywhere, and I'm fine with it either way, just thought it was worth mentioning.

byorgey (approved):

src/swarm-engine/Swarm/Game/Step/Const.hs
kostmo: > with the refactored code they will not be woken up, is that correct?

That's right, and I believe more appropriate.

PR #1738, opened 2024-01-24
single entity update for swap

kostmo:
Towards #1598

### DIFF
--- a/src/swarm-engine/Swarm/Game/Step/Const.hs
+++ b/src/swarm-engine/Swarm/Game/Step/Const.hs
@@ -104,6 +104,11 @@ data RobotFailure = ThrowExn | Destroy | IgnoreFail
 --   to a location.
 type MoveFailureHandler = MoveFailureMode -> RobotFailure
 
+-- | Whether to remove the entity in the world inside the 'doGrab' function
+-- or leave it to be done by other code.
+data GrabRemoval = DeferRemoval | PerformRemoval
+  deriving (Eq)
+
 -- | Interpret the execution (or evaluation) of a constant application
 --   to some values.
 execConst ::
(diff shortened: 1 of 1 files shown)


src/swarm-engine/Swarm/Game/Step/Const.hs
```diff
@@ -1672,8 +1672,9 @@ execConst runChildProg c vs s k = do
 
   -- The code for grab and harvest is almost identical, hence factored
   -- out here.
-  doGrab :: (HasRobotStepState sig m, Has Effect.Time sig m) => GrabbingCmd -> m Entity
-  doGrab cmd = do
+  -- Optionally defer removal from the world, for the case of the Swap command.
+  doGrab :: (HasRobotStepState sig m, Has Effect.Time sig m) => GrabbingCmd -> Bool -> m Entity
+  doGrab cmd deferRemoval = do
```
byorgey, line 1677: Can we make a simple enumeration type for use as the second argument to `doGrab` instead of using `Bool`?  Looking at the call sites of `doGrab` it is impossible to know what the `True` or `False` arguments mean.  Even when one does know what the argument is about it is hard to remember which is `True` and which is `False`. It would be much more perspicuous as `doGrab Swap' DeferRemoval` and `doGrab Grab' RemoveNow`, or something similar.

byorgey (approved):
Makes sense!  This should definitely make `swap` a bit more efficient.

src/swarm-engine/Swarm/Game/Step/Const.hs
kostmo: Yeah, you can say [Boolean blindness](https://github.com/quchen/articles/blob/master/algebraic-blindness.md#boolean-blindness) and I'll get what you mean :)

Issue #1741, opened 2024-01-24
Improve module placement within scenario/engine sublibraries


kostmo:
Following up on #1719:

> * `Swarm.Game.ScenarioInfo`, `Swarm.Game.Tick`, and `Swarm.Game.Scenario.Status` could probably be moved from `swarm-scenario` to `swarm-engine`.

PR #1742, opened 2024-01-24
move activity counts into swarm-engine

kostmo:
Towards #1741


src/swarm-engine/Swarm/Game/Robot/Concrete.hs
```diff
@@ -60,6 +66,7 @@ instance ToSample Robot where
       mkRobot
```
byorgey, line 66: It seems like this is the only place we use `mkRobot` to create a `Robot`, as opposed to a `TRobot`.  I wonder if we should just specialize `mkRobot` to only produce `TRobot`, which would cut down on all the annoying `()` arguments we have to pass to it right now.  Then here, we could create `sampleBase` by composing `mkRobot` and `instantiateRobot`?  Would that work?

byorgey (approved):

src/swarm-engine/Swarm/Game/Robot/Concrete.hs
kostmo: Good idea, that works.

I wonder if at this point it would make sense to consolidate the three fields that are `()` for `TRobot` into their own record that itself would be `()` for `TRobot`, since no `TRobot` ever needs to access those fields individually.

PR #1744, opened 2024-01-25
apply hlint 3.8 suggestions

kostmo:
As title


byorgey (approved):

byorgey:
Is there a way we can ensure we are running version 3.8 of hlint in our CI?

PR #1746, opened 2024-01-25
Remove Benchmark dependence on AppState and TUI

kostmo:
`stack bench` is now independent of the TUI and `AppState`.


test/bench/Benchmark.hs
```diff
@@ -134,10 +137,15 @@ initRobot prog loc =
 mkGameState :: ProcessedTerm -> (Location -> TRobot) -> Int -> IO GameState
 mkGameState prog robotMaker numRobots = do
   let robots = [robotMaker (Location (fromIntegral x) 0) | x <- [0 .. numRobots - 1]]
-  Right initAppState <- runExceptT classicGame0
```
byorgey, line 137: Was this the only place we were using this `classicGame0` definition?  Can we get rid of it now?

byorgey (approved):

test/bench/Benchmark.hs
kostmo: It is still used by unit tests:
https://github.com/swarm-game/swarm/blob/669163384e62093d01536c342a7446859468f9ca/test/unit/Main.hs#L46

Perhaps will remove that too at some point.

PR #1747, opened 2024-01-25
volume command

kostmo:
Measures the volume of an enclosed space.
A useful alternative to the `path` command for goal checking.

## Demo

    scripts/play.sh -i data/scenarios/Testing/1747-volume-command.yaml --autoplay --speed 2

### DIFF
--- a/src/swarm-engine/Swarm/Game/Step/Flood.hs
+++ b/src/swarm-engine/Swarm/Game/Step/Flood.hs
@@ -1,0 +1,65 @@
+-- |
+-- SPDX-License-Identifier: BSD-3-Clause
+--
+-- Implementation of the 'Swarm.Language.Syntax.Volume' command for robots.
+module Swarm.Game.Step.Flood (
+  floodFill,
+) where
+
+import Control.Effect.Lens
+import Control.Monad (filterM)
+import Data.HashSet (HashSet)
+import Data.HashSet qualified as HashSet
+import Swarm.Game.Location
+import Swarm.Game.Step.RobotStepState
+import Swarm.Game.Step.Util (checkMoveFailureUnprivileged)
+import Swarm.Game.Step.Util.Inspect (getNeighborLocs)
+import Swarm.Game.Universe
+
+data FloodParms = FloodParms
+  { theSubworld :: SubworldName
+  , maxVisits :: Int
+  }
+
+floodRecursive ::
+  HasRobotStepState sig m =>
+  HashSet Location ->
+  [Location] ->
+  FloodParms ->
+  m (Maybe Int)
+floodRecursive visited pending st =
+  case pending of
+    nextLoc : otherLocs ->
+      if visitedCount > maxVisits st
+        then return Nothing
+        else checkNeighbors nextLoc otherLocs
+    [] -> return $ Just visitedCount
+ where
+  visitedCount = HashSet.size visited
+  checkNeighbors nextLoc otherLocs = do
+    candidateNeighbors <- listNeighbors $ Cosmic (theSubworld st) nextLoc
+    let visitableNeighbors = filter (not . (`HashSet.member` visited)) candidateNeighbors
+        -- It's cheaper to prepend the "visitableNeighbors" list because
+        -- it should in general be a shorter list than the pending queue.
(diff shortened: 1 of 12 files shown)


src/swarm-engine/Swarm/Game/Step/Flood.hs
```diff
@@ -0,0 +1,65 @@
...
+import Swarm.Game.Step.Util (checkMoveFailureUnprivileged)
+import Swarm.Game.Step.Util.Inspect (getNeighborLocs)
+import Swarm.Game.Universe
+
+data FloodParms = FloodParms
+  { theSubworld :: SubworldName
+  , maxVisits :: Int
+  }
+
+floodRecursive ::
```
byorgey, line 24: I note that because you *prepend* the `visitableNeighbors` to the `pending` list, you are essentially treating it like a stack, so this is doing a DFS rather than a BFS.  When the robot is enclosed, this doesn't make much difference---the same cells end up being discovered either way.  I guess the difference is that in cases where the robot is *not* enclosed, the DFS visits cells in basically a straight line until it has visited the max number, rather than visiting all the cells in a square region around the robot.  I don't know how much this matters.

Note I'm not saying you should append `visitableNeighbors` to the end of `pending`, that would indeed be very inefficient.  If we really want a BFS we would have to switch to queue data structure for `pending` --- `Data.Sequence` works great as a queue in practice (even though in theory it incurs an extra log factor).

byorgey (commented):
This command doesn't seem all that useful to me, but I'm guessing you must have some particular use case(s) in mind?

kostmo:
> some particular use case(s) in mind?

Yes.  I don't have a particular use case for player robots, but I envision some interesting gameplay mechanics involving "enclosure" when used by system robots/goal conditions.  The integration test scenario is one example.  In contrast with the `gated-paddock` scenario, which used the `path` command to test for a route to a surrounding "ocean" to determine enclosure by a fence, this `volume` command does not require an extra surrounding entity layer, so long as there exists a maximum expected enclosure size.

A few scenario ideas:

1. Caterpillar robots that turn into butterflies when snugly enclosed in a cocoon
2. Defeating opponents by trapping them
3. Excavating a burrow or mine of sufficient size as a goal (in combination with portals)

byorgey:
Ah, OK, yes, those sound nice!

src/swarm-engine/Swarm/Game/Step/Flood.hs
```diff
@@ -0,0 +1,110 @@
...
+  checkNeighbors nextLoc otherLocs = do
+    isWalkable <- null <$> checkMoveFailureUnprivileged cosmicLoc
+    let candidateNeighbors =
+          if isWalkable
+            then map (view planar) $ getNeighborLocs cosmicLoc
+            else []
+        visitableNeighbors = filter (not . (`HashSet.member` visited tracking)) candidateNeighbors
+
+        -- It's cheaper to prepend the "visitableNeighbors" list because
+        -- it should in general be a shorter list than the "pending" list.
```
byorgey, line 83: This comment seems a little misleading --- the reason for prepending the `visitableNeighbors` is because we're treating `pending` as a stack rather than a queue, not because it is more efficient (though I agree it will be).  Regardless of which list is shorter, prepending the `visitableNeighbors` will take constant time per new neighbor, and hence $O(n)$ overall since each vertex occurs as a new neighbor at most once. Whereas appending `visitableNeighbors` will take time linear in the number of vertices in the queue, but the same vertices might be in the queue for multiple iterations, so we can end up doing more than $O(n)$ work appending.

byorgey (approved):

src/swarm-engine/Swarm/Game/Step/Flood.hs
kostmo: > the reason for prepending the `visitableNeighbors` is because we're treating `pending` as a stack rather than a queue, not because it is more efficient (though I agree it will be)

Well, I originally went about this intending to implement BFS, until I got to this line and realized that **appending** to the `pending` list would be expensive (and I couldn't be bothered to use `Seq` instead of `[]`).  So I decided to **prepend** instead, which turned it into DFS.

PR #1749, opened 2024-01-26
autopopulate spellchecker

kostmo:
Builds upon #1587.

Extract all symbol names that are not native to the current project and insert them into our own custom spell checking dictionary's "words" list.

The premise is that symbols that *are* native to our project should be spellchecked, but foreign symbols that constitute unrecognized dictionary words are presumably intentionally spelled that way.

# Convention

Manually-added words (i.e. for names in code that we've written for this project) will go into `.vscode/settings.json`.
The automatically generated word list from third-party packages goes into `cspell.json`.

# Usage

    scripts/spellcheck/autopopulate-spellchecker.sh


byorgey (approved):
Looks fine to me!

PR #1751, opened 2024-01-27
check achievement robot requirements

### LINKED ISSUE #1726: Parameterize achievements on validity for creative mode + system robots
> I think I would like to make `ValidInCreativeMode` and `ValidForSystemRobot` into properties of the achievements themselves.  Then the `grantAchievement` function would perform the verification (perhaps with the addition of a `HasRobotStepState` constraint).

_Originally posted by @kostmo in https://github.com/swarm-game/swarm/pull/1688#discussion_r1427394090_

kostmo:
Closes #1726

There are now two entry points to obtain a `GameplayAchievement`:
* `grantAchievementForRobot`
* `grantAchievement`

The latter enforces that the achievement is applicable to the current game mode (i.e. creative), while the former makes an additional check for validity of system robots.

This is an improvement over the status quo, but note that it is still currently possible to make a coding error in which an achievement specifies a requirement of **not** being a system robot, but the site at which the achievement is granted only calls `grantAchievement`, bypassing the robot validity check.


src/swarm-engine/Swarm/Game/Step/Util/Command.hs
```diff
@@ -150,18 +151,42 @@ onTarget rid act = do
...
+  Has (State GameState) sig m =>
+  GameplayAchievement ->
+  m Bool
+checkGameModeAchievementValidity a = do
+  creative <- use creativeMode
+  return $ not creative || gameplayModeRequired == ValidInCreativeMode
+ where
+  ValidityConditions _ gameplayModeRequired = getValidityRequirements a
+
 grantAchievement ::
```
byorgey, line 176: Can you add a comment to `grantAchievement` explaining that one probably wants to call `grantAchievementForRobot` instead of calling `grantAchievement` directly?

byorgey (approved):
Looks good to me.  I can imagine ways to eliminate the possibility of calling `grantAchievement` when one should be calling `grantAchievementForRobot` (via some extra type parameter(s) on `GameplatAchievement`), but it's probably not worth it.

src/swarm-engine/Swarm/Game/Step/Util/Command.hs
kostmo: Added in #1785.

PR #1752, opened 2024-01-27
extract scene renderer to its own binary

### LINKED ISSUE #1715: Extract a sublibrary for map generation that is independent of game state

kostmo:
Closes #1715

Extract scene renderer to its own binary with `swarm-scenario` as its only dependency.

This was mainly enabled by moving the `Landscape` sub-record out of the `swarm-engine` sublibrary and into `swarm-scenario`, and refactoring rendering functions to only depend on `Landscape` rather than the entire `GameState`.

# Demo

    stack build --fast swarm:swarm-scene && stack exec swarm-scene -- data/scenarios/Fun/horton.yaml --png -w 180 -h 150

output


byorgey (approved):
Nice!

PR #1755, opened 2024-01-29
simplify spellchecker words script

kostmo:
Found the `tr` command to replace the extra `split-module-names.hs` script.


byorgey (approved):
Ah yes, very nice!

PR #1756, opened 2024-01-30
preview rendered world with inotify

kostmo:
Opens a live-reloading preview of the world in VS Code.

The renderer has been modified to optionally render a blank image instead of crashing upon invalid YAML.

## Prerequisites:
Install inotify tools:

    sudo apt install inotify-tools

## Usage:

    scripts/preview-world-vscode.sh data/scenarios/Fun/horton.yaml

Once the VS Code editor tabs are opened, one can press <kbd>CTRL</kbd> + <kbd>\\</kbd> (backslash) with the image selected to split the editor pane horizontally.
One may then navigate to the left-pane's copy of the image preview with <kbd>CTRL</kbd> + <kbd>PageUp</kbd>, and then <kbd>CTRL</kbd> + <kbd>w</kbd> will close the redundant image preview.

## Screenshot

Screenshot from 2024-01-29 18-53-55

### DIFF
--- a/scripts/preview-world-vscode.sh
+++ b/scripts/preview-world-vscode.sh
@@ -1,0 +1,37 @@
+#!/bin/bash -xe
+
+# Opens a live-reloading preview of the world
+#
+# Prerequisites:
+# --------------
+# Install inotify-wait:
+#
+#    sudo apt install inotify-tools
+#
+# Usage:
+# --------------
+# Once the VS Code editor tabs are opened, one can press
+# CTRL+\ (backslash) with the image selected to split the
+# editor pane horizontally.
+# One may then navigate to the left-pane's copy of the image
+# preview with CTRL+PageUp, and then
+# CTRL+w will close the redundant image preview.
+
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+cd $SCRIPT_DIR/..
+
+
+SCENARIO_PATH=${1?"Usage: $0 SCENARIO_PATH"}
+
+IMG_WIDTH=200
+IMG_HEIGHT=150
+
+IMG_OUTPUT_PATH=output.png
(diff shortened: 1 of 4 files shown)


scripts/preview-world-vscode.sh
```diff
@@ -0,0 +1,37 @@
...
+IMG_WIDTH=200
+IMG_HEIGHT=150
+
+IMG_OUTPUT_PATH=output.png
+RENDER_IMG_COMMAND="stack exec swarm-scene -- $SCENARIO_PATH --fail-blank --dest $IMG_OUTPUT_PATH --png --width $IMG_WIDTH --height $IMG_HEIGHT"
+
+stack build --fast swarm:swarm-scene
+
+$RENDER_IMG_COMMAND
+code --reuse-window $SCENARIO_PATH && code --reuse-window $IMG_OUTPUT_PATH
```
byorgey, line 35: Do we need to make this specific to VS Code?  It seems like other than these calls to `code` there is nothing specific to VS Code here.  I could imagine wanting to use this but just having the scenario code open in Emacs and the world preview image open in a browser or Emacs buffer.

byorgey (approved):

scripts/preview-world-vscode.sh
kostmo: I'm open to making this generic w.r.t. the editor, but VS Code is the only tool I use, so I wouldn't know what changes to make.

FWIW, VS Code auto-refreshes an image display if the image file changes, but a browser window would not do so automatically.

byorgey: I guess I had in mind something like refactoring this script into two: one editor-independent script which generically sets up the automatic image regeneration with `inotify`, and a second wrapper script which calls the first and also launches VS Code. Then anyone could use the first script with whatever tools they want. I may try doing that myself.

Sure, I'm aware a browser window would not automatically refresh without some kind of plugin to make it do so, it was just an example.

PR #1757, opened 2024-01-31
Refactor `ObjectiveCompletion` to use lenses

### LINKED ISSUE #1037: Simplify ObjectiveCompletion state modification with lenses
> It seems to me like `addCompleted`, `addUnwinnable`, `setIncomplete`, etc. could be made much simpler using lenses.  The main complication is that setting `completed` must also update the `completedIDs`, but that could be handled using the technique explained here: https://byorgey.wordpress.com/2021/09/17/automatically-updated-cached-views-with-lens/ .

Original comment by @byorgey here:
https://github.com/swarm-game/swarm/pull/927/files#r1084692447

byorgey:
Closes #1037.

The main changes are in `Swarm.Game.Scenario.Objective`.
- No longer export the `CompletionBuckets` type, nor the constructor or record accessors of the `ObjectiveCompletion` type
- Export folds such as `incompleteObjectives` (a `Fold` is like a read-only `Traversal`, i.e. it has multiple targets but does not allow setting/modification), as well as `allObjectives`.
- Export some custom functions such as `addCompleted` which properly maintains the internal invariants of `ObjectiveCompletion`.

The rest of the changes are simply adapting to use these new lenses + folds instead of directly accessing `CompletionBucket` values or using the underscored record field projections.

This simplifies the code in `Swarm.Game.Scenario.Objective` and hopefully makes the code safer (previously, it was possible in theory to violate `CompletionBuckets` invariants by directly modifying its fields).


src/swarm-engine/Swarm/Game/Scenario/Objective/WinCheck.hs
```diff
@@ -22,45 +23,38 @@ import Swarm.Game.Scenario.Objective.Logic as L
 
 -- | We have "won" if all of the "unwinnable" or remaining "incomplete" objectives are "optional".
 didWin :: ObjectiveCompletion -> Bool
-didWin oc = all _objectiveOptional $ incomplete buckets <> unwinnable buckets
- where
-  buckets = completionBuckets oc
+didWin oc =
+  and (oc ^.. incompleteObjectives . objectiveOptional)
+    && and (oc ^.. unwinnableObjectives . objectiveOptional)
```
kostmo, line 28: This new formulation is a bit less concise; in the original, we assemble a list of objectives and check that `all` of them are optional.  In the new code, we have to perform such a check twice independently.

Can we do something like ```incompleteObjectives `concatFold` unwinnableObjectives``` first?

kostmo (approved):
Nice improvement.  Just one nit to address.

src/swarm-engine/Swarm/Game/Scenario/Objective/WinCheck.hs
byorgey: Oh, excellent point, thanks.  I think I tried something to simplify this at one point but it failed in a strange way (the game thought every scenario was instantly won no matter what).  But that was before I added `concatFold`.

The new version using `concatFold` is much shorter (and doesn't repeat code).

PR #1762, opened 2024-02-11
Update to depend on `lsp-2.4.0.0`

### LINKED ISSUE #1350: Update to `lsp-2.0`
`lsp-2.0` and `lsp-types-2.0` are now included in Stackage LTS-21, but they apparently introduce [a bunch of breaking changes](https://hackage.haskell.org/package/lsp-types-2.0.0.1/changelog).  Someone should take a close look at what will be necessary to update to the latest versions of these dependencies.

byorgey:
Update to use the `lsp-2.4` API.  Closes #1350.

Initially I hoped that any `lsp-2.x` would work.  However, the `SeverDefinition` record changed in `2.2` so I initially set that as a lower bound.  But then it turns out that `2.4` changed which module it is importing `Rope` from; since we work with ropes in the `Hover` module it matters since we have to import the matching module.  Updating to the new `Rope` type also required some changes as the API provided by the new `Rope.Mixed` module is a bit different than the old module, so we would not even be able to easily put in CPP to conditionally depend on the right rope module depending on the `lsp` version.  Finally, this means dropping support for GHC 9.0 since `lsp-2.4` does not support it.

Along the way I also fixed a minor issue related to showing type information returned by the LSP server, so that it uses `prettyTypeLine` to display the type on a single line (in my editor, when the type does not use a single line it gets cut off).  For comparison see also #1610. 

This refactoring was a big pain because a lot of things (names of types and constructors, locations of exports, etc...) changed from 1.x to 2.x, but there was not much in the way of documenting what had changed. =(   I am pretty sure that all functionality has been preserved but I would appreciate independent confirmation.

This is also a prerequisite for updating other dependencies such as the `base` version (I will open a follow-up PR soon) since the old `lsp-1.x` versions do not allow many newer versions of various dependencies.


kostmo (approved):
Thanks for slogging through this.

Approving but I won't have time to try it out for a while.

byorgey:
@kostmo No worries, thanks!  Everything I've tried seems to work, so I'll merge and of course we can open an issue if you discover any regressions later.

byorgey:
Have to merge manually since the mergify configuration changed.

PR #1766, opened 2024-02-17
Scenario description in goals dialog

### LINKED ISSUE #1735: Display toplevel scenario description inside Goals dialog
Currently, the scenario description is only visible in the scenario selection menu.

I propose that it should also be visible from within a currently playing scenario---either on the top of the Goals dialog, or in some new information dialog.

kostmo:
Closes #1735.

# Demo

    stack exec swarm -- -i data/scenarios/Challenges/bridge-building.yaml

## Before

Screenshot from 2024-02-17 15-27-32

## After

Screenshot from 2024-02-17 15-25-24

### DIFF
--- a/src/Swarm/TUI/View.hs
+++ b/src/Swarm/TUI/View.hs
@@ -621,7 +621,9 @@ drawModal s = \case
         ]
   DescriptionModal e -> descriptionWidget s e
   QuitModal -> padBottom (Pad 1) $ hCenter $ txt (quitMsg (s ^. uiState . uiMenu))
-  GoalModal -> GR.renderGoalsDisplay (s ^. uiState . uiGameplay . uiGoal)
+  GoalModal ->
+    GR.renderGoalsDisplay (s ^. uiState . uiGameplay . uiGoal) $
+      view scenarioDescription . fst <$> s ^. uiState . scenarioRef
   KeepPlayingModal ->
     padLeftRight 1 $
       displayParagraphs $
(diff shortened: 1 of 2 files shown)


src/Swarm/TUI/View/Objective.hs
```diff
@@ -31,19 +33,25 @@ makeListWidget (GoalTracking _announcements categorizedObjs) =
...
+    if hasMultiple
+      then
+        vBox
+          [ hBox
+              [ leftSide
+              , padLeft (Pad 2) goalElaboration
+              ]
+          , footer
+          ]
+      else goalElaboration
```
byorgey, line 53:
```suggestion
  goalsWidget
    | hasMultiple =
          vBox
          [ hBox
              [ leftSide
              , padLeft (Pad 2) goalElaboration
              ]
          , footer
          ]
    | otherwise = goalElaboration
```
I find `if-then-else` ugly, and this seems an easy place to get rid of it.

byorgey (approved):

src/Swarm/TUI/View.hs
```diff
@@ -621,7 +621,9 @@ drawModal s = \case
         ]
   DescriptionModal e -> descriptionWidget s e
   QuitModal -> padBottom (Pad 1) $ hCenter $ txt (quitMsg (s ^. uiState . uiMenu))
-  GoalModal -> GR.renderGoalsDisplay (s ^. uiState . uiGoal)
+  GoalModal ->
+    GR.renderGoalsDisplay (s ^. uiState . uiGoal) $
+      view scenarioDescription . fst <$> s ^. uiState . scenarioRef
```
byorgey, line 626: Untested, but would something like this work?

```suggestion
      s ^? uiState . scenarioRef . _Just . _1 . scenarioDescription
```

kostmo: Eh, I'm not really a fan of the `_Just` and `_1` lenses when they can be avoided.  I recognize they are not technically "partial functions" here, but the explicit indexing and `Just` are evocative of partiality in a way I find unpalatable :stuck_out_tongue:

byorgey: OK, that's fine!

PR #1767, opened 2024-02-18
Decompose UIState record

kostmo:
Introduced 3 new sub-records in `UI.hs`:
* `UIGameplay`
* `UITiming`
* `UIInventory`

This allows some functions (e.g. `populateInventoryList`) to operate on a narrower scope of state.

Also uses `Brick.zoom` in a few more places to mitigate the longer lens chains.


byorgey (approved):
:+1:

PR #1789, opened 2024-03-07
Allow blaze-html-0.9.2

byorgey:
Not sure why we had such a restrictive set of bounds on `blaze-html`.  It used to be `< 0.9.2`; I changed it to `< 0.10` which should be fine according to the PVP, and it seems to build fine with version `0.9.2.0`.

### DIFF
--- a/swarm.cabal
+++ b/swarm.cabal
@@ -710,7 +710,7 @@ executable swarm
   build-depends:
     -- Imports shared with the library don't need bounds
     base,
-    blaze-html >=0.9.1 && <0.9.2,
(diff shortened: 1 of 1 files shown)


kostmo (approved):

byorgey:
@kostmo Thanks, sorry I've been relatively unresponsive recently, I've been sick with the flu. :face_with_thermometer:  What should I be focused on reviewing next?  #1777 ?

kostmo:
> What should I be focused on reviewing next? #1777 ?

No worries! Yes, #1777 is ready.

PR #1790, opened 2024-03-11
debugger robot runpool

kostmo:
Addresses https://github.com/swarm-game/swarm/pull/1736#issuecomment-1980029206


src/swarm-engine/Swarm/Game/Step.hs
```diff
@@ -245,45 +257,58 @@ singleStep ss focRID robotSet = do
...
             then do
               when (newR ^. activityCounts . tickStepBudget == 0) $
+                -- Our robot ran out of its step budget for this tick,
+                -- so we will have to continue on to the other robots.
                 temporal . gameStep .= RobotStep (SAfter focRID)
               return False
             else do
               -- continue to newly focused
-              singleStep SBefore focRID postFoc
+              singleStep SBefore focRID $ augmentRunPool postFoc
```
kostmo, line 290: Here is the only behavior change made to the debugger.

src/swarm-engine/Swarm/Game/Step.hs
```diff
@@ -174,15 +174,40 @@ type HasGameStepState sig m = (Has (State GameState) sig m, Has (Lift IO) sig m,
...
     forM_ mr (stepOneRobot rn)
  where
   stepOneRobot :: HasGameStepState sig m => RID -> Robot -> m ()
-  stepOneRobot rn rob = tickRobot rob >>= insertBackRobot rn
+  stepOneRobot rn rob = tickRobot rob >>= registerUpdatedRobotState rn
+
+-- | This function must be utilized in every loop that iterates over
+-- robots and invokes 'stepRobot' (i.e. steps a robot's CESK machine).
+getExtraRunnableRobots :: Has (State GameState) sig m => TickNumber -> m (IS.IntSet -> IS.IntSet)
+getExtraRunnableRobots time = do
```
kostmo, line 195: TODO: Add a comment about "guarding" the `union` operation with a `null` check; even though a "branchless" union may be semantically equivalent with shorter code, it would end up doing `O(log N)` operation ever time, rather than only when needed.

PR #1791, opened 2024-03-12
fix watch wakeup test

kostmo:
This improves the test introduced in #1736 in a few ways:
* Place the system robot's code inline in the `program` field
    * even though the system bot had been using `instant`, the `run` command itself incurs a delay.  So this ensures the system robot is ready to monitor the cell immediately
* System robot uses exclusively `Intangible` commands in its monitoring loop
    * This allows removal of the `instant` command.
    * `Surveil`, which does not appear in any scenario yet, is made `Intangible`.
* Remove the initial `wait 2` in the player robot's solution.

Tested to ensure that this still exercises the problem that #1736 set out to solve, by commenting out this line (causing the scenario to fail).

https://github.com/swarm-game/swarm/blob/01ae0e45d75c30619af76ae2438dc1b92ef61e08/src/swarm-engine/Swarm/Game/State/Robot.hs#L396

## Demo

    scripts/play.sh -i data/scenarios/Testing/1598-detect-entity-change.yaml --autoplay --speed 1


byorgey (approved):
Makes sense to me.

Issue #1797, opened 2024-04-08
Competitive play


kostmo:
### Discussed in https://github.com/swarm-game/swarm/discussions/1691

<div type='discussions-op-text'>

<sup>Originally posted by **kostmo** December 15, 2023</sup>
For reference, some ideas pertaining to competition were presented in #967.

I'll describe a style of multiplayer participation that could be fun.

# Submitting scenarios and solutions

In competitive play, there should be a way to submit a scenario (a `.yaml` file along with any other resource dependencies such as `.sw` or `.world` files) that is not inspectable by other players.

Likewise, there should be a way to submit solutions that remain private until all participants have submitted their own.

A competition should target a specific Git revision SHA1 of Swarm itself.

# Honor-system vs. secure competitions

In an honor-system competition, it may be sufficient for a locally-running Swarm program to fetch a scenario from a server at runtime, so that its only representation is "in memory".  This avoids forms of "casual" or "accidental" cheating.  However, it will nevertheless be possible to run a locally-modified `swarm` binary that could dump the fetched scenario data to file on demand.  Therefore, a "secure" competition would require the `swarm` binary to run on a remote server.

Supporting remote execution would require protections against intentional or accidental infinite loops in swarm-lang code.  In competition, the "gamemaster" might specify a maximum number of ticks that a supplied solution is allowed to run before being terminated.  Distinguishing between an inefficient solution or adversarial input is not necessary.

An external wall-clock timeout would serve as a fallback protection against bugs in the `swarm` interpreter itself.

## Local execution pros and cons

Pros:
* may be perfectly fine in a casual setting among a familiar group of competitors
* less complex in terms of infrastructure
    * no significant compute cost or guardrails required
* retains the TUI interface
    * supports interactive play

Cons:
* still entails some extra infrastructure, e.g.:
    * a central server to exchange scenarios and solutions
    * a remote swarm binary to evaluate submitted solutions
* not sufficient for secure/serious competitions

## Security mitigations for local execution

Most importantly for local execution, competitors should not have access to the solution program text for a scenario they have fetched from the server.  This can be easily mitigated by having the server simply blank the `solution` field (after having tested the official solution and obtained its metrics) before transmitting the serialization of the scenario to the player.

Program code of system robots, other NPCs, or goal conditions should also not be exposed.  The CESK representation may be sufficiently obfuscated for the purpose of competition; so long as the original program text is not transmitted, a player would have to have written a "decompiler" to glean any helpful information from the in-memory programs.

[...]

PR #1798, opened 2024-04-08
Upload and parse scenarios

kostmo:
Towards #1797

Hosts an online repository of scenarios, against which solutions may be submitted.  This is the foundational layer that may support more structured "tournaments", scenario ranking, or other social activity.

# Demo

## Live server

http://swarmgame.net/list-games.html

One can use the [`submit.sh`](https://github.com/swarm-game/swarm/pull/1798/files#diff-450877e3442a0ec1c5cbe964808a263d67f1e680d3aa3c3bf9ae6f51eca682fb) script and see valid uploads reflected live on the website.

### Automated tests

These are database-agnostic.

    scripts/run-tests.sh swarm:test:tournament-host

### Manual tests

These test database interactions.  It requires first setting up a local Postgres server.

1. Start `tournament/scripts/demo/server-native.sh` in one console
2. Run `tournament/scripts/demo/client/test-cases/local/good-submit.sh` in another

# Features

* Upload and validates scenarios
* Download scenarios with solution redacted
* Submit, validate, execute, and score solutions

# Key components

* Servant server
* Hosted on AWS in a Docker container
* Stores to a Postgres database in Amazon RDS
* Shares some code with the integration tests for evaluating scenarios and solutions

The production database uses IAM to manage logins.  The web app uses the AWS API to fetch a "token" which can be used to log in instead of a password.  This avoids having to store a password on the server.

# TODO
- [ ] User authentication (GitHub OpenID?)


app/tournament/Main.hs
```diff
@@ -0,0 +1,99 @@
...
+webPort =
+  optional $
+    option
+      auto
+      ( long "port"
+          <> metavar "PORT"
+          <> help ("Set the web service port (or disable it with 0). Default to " <> show defaultPort <> ".")
+      )
+
+gameVersion :: Parser Sha1
```
byorgey, line 33: I don't understand why there is a flag for setting the game version.  Shouldn't it just be the hash of whatever commit the server was compiled from?  Otherwise, couldn't you just set it to something arbitrary that doesn't actually match the version that is really being used to validate and run scenarios etc.?

src/swarm-tournament/Swarm/Web/Tournament/Validate.hs
```diff
@@ -0,0 +1,259 @@
...
+      . ExceptT
+      . runThrow
+      $ evalAccum (mempty :: Seq SystemFailure) initGameStateConfig
+
+  let scenarioInputs = gsiScenarioInputs $ initState gsc
+  scenarioObject <- initScenarioObject scenarioInputs content
+  liftIO $ scenarioToGameState scenarioObject emptyLaunchParams gsc
+
+-- |
+-- NOTE: Borrowed from @testSolution'@ in @tests/integration/Main.hs@
```
byorgey, line 204: Could we refactor to pull out a common function instead of duplicating code?

tournament/scripts/demo/client/submit.sh
```diff
@@ -0,0 +1,33 @@
...
+# $1 = hostname (and optional port)
+# $2 = path to scenario file
+# $3 = path to solution file
+
+HOST=$1
+SCENARIO_FILEPATH=$2
+SOLUTION_FILEPATH=$3
+
+# Example:
+#   tournament/scripts/demo/client/submit.sh localhost:8008 data/scenarios/Challenges/dimsum.yaml data/scenarios/Challenges/_dimsum/solution.sw
```
byorgey, line 13: The first time I tried this script it didn't work.  After poking at it for a while I realized the reason is that the scenario file and solution file paths must be either absolute paths, or relative to the repository root.  Would be good to add a note here to that effect.

byorgey (commented):
This is amazing!  Now I know why I hadn't heard anything from @kostmo in a while. :wink: 

Besides a few minor comments/questions, I think the biggest issue is that we need to configure things so that by default, doing a `cabal build` does *not* build the tournament server, so it does not need to download and build servant, postgresql-simple, etc. Most people building `swarm` from source will not need all this so we should save them the trouble; anyone who actually wants the server can enable a flag.  I think we should be able to do this by creating a `flag` in the `.cabal` file and marking relevant libraries + executable stanzas as `buildable: false` conditional on the flag.  Or maybe there is a better way to accomplish this with modern cabal, I'm not sure.

I did not test actually building the server myself but I was able to test uploading a solution to one of the scenarios listed on swarmgame.net.

tournament/web/favicon.ico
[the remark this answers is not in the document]
kostmo: Yes, it's just a placeholder for now.

app/tournament/Main.hs
kostmo: Indeed, it's supposed to be the Git hash that the server was compiled from: see [this line](https://github.com/swarm-game/swarm/pull/1798/files#diff-da43e5f2be51b502f88653e55b25b54be14c32ad581bb5fa2354b2bad0eb84d8R7), which is fed into the Docker image [here](https://github.com/swarm-game/swarm/pull/1798/files#diff-d75b1260f266d4d00102549606dd159d662db8342f9d595baddeebbf2545444fR94), and then supplied as the `--version` CLI option.

swarm.cabal
```diff
@@ -429,6 +429,56 @@ library swarm-web
     -- See discussion in #415
     StrictData
 
+library swarm-tournament
```
kostmo, line 432: Replying to the toplevel review comment here, so that the discussion can be properly threaded.

> I think the biggest issue is that we need to configure things so that by default, doing a `cabal build` does *not* build the tournament server, so it does not need to download and build servant, postgresql-simple, etc. Most people building `swarm` from source will not need all this so we should save them the trouble; anyone who actually wants the server can enable a flag.  I think we should be able to do this by creating a `flag` in the `.cabal` file and marking relevant libraries + executable stanzas as `buildable: false` conditional on the flag.  Or maybe there is a better way to accomplish this with modern cabal, I'm not sure.

Recall that the main `swarm` executable already includes `servant`, [from way back in 2022](https://github.com/swarm-game/swarm/pull/623/files#diff-42de5b60f981e95037ce731549b8f3d075c21bfa84407f76f56de17d5374c3deR150).

`postgresql-simple`, however, is a new dependency.  I think the simplest way to avoid building `postgresql-simple` is to pass an explicit target argument to cabal:
```
cabal build swarm:exe:swarm
```

As a convenience, we may provide a wrapper script (named `build-game.sh`) that makes the above invocation.

P.s. there seems to be a bug in `stack` that builds components/executables that are not requested on the command line.  So I think our documentation should explicitly recommend `cabal` in any places where it doesn't already.

tournament/scripts/demo/client/submit.sh
kostmo: Added a note.

swarm.cabal
byorgey: > Recall that the main `swarm` executable already includes `servant`, [from way back in 2022](https://github.com/swarm-game/swarm/pull/623/files#diff-42de5b60f981e95037ce731549b8f3d075c21bfa84407f76f56de17d5374c3deR150).

Ah, good point.  But avoiding `postgresql-simple` when not necessary is especially important since it has external library dependencies.  When I tried building I got an error about some thing that wasn't installed.  Getting it to work would have required me to go figure out which OS package to install that would provide whatever thing was missing.  But in any case, I think you're right that specifying a particular target like `swarm:exe:swarm` is the way to go, I had forgotten about that.

> P.s. there seems to be a bug in stack that builds components/executables that are not requested on the command line.

Really? It works for me.  e.g. when I do `stack build swarm:exe:swarm` it only builds the libraries needed for the main `swarm` executable.  By contrast just doing `stack build` builds other stuff like `swarm-scene` and `swarm-docs`.

byorgey (approved):

swarm.cabal
kostmo: > Really? It works for me. e.g. when I do `stack build swarm:exe:swarm` it only builds the libraries needed for the main `swarm` executable. By contrast just doing `stack build` builds other stuff like `swarm-scene` and `swarm-docs`.

Try the following:
1. Add a newline to the end of `src/swarm-tournament/Swarm/Web/Tournament/Validate.hs`
2. Run `stack build swarm:exe:swarm`
3. Observe that it recompiles `swarm-tournament`, despite that library not being a dependency of the `swarm` executable.

Now do the same cycle, but with `cabal build swarm:exe:swarm` instead, and observe `Up to date` is printed.

byorgey: Hmm, you're right.

> I think our documentation should explicitly recommend `cabal` in any places where it doesn't already.

I think I'm convinced.  It just means anyone who wants to build from source has to pay more attention to what version of GHC they are using.  But these days with `ghcup` that's pretty trivial.

byorgey: Created #1820 to track this.

PR #1799, opened 2024-04-10
Refactor hierarchy of state inputs

kostmo:
Towards #1797

Combines `TerrainEntityMaps` and `WorldMap` into a single record, `ScenarioInputs`.

Establishes a Matroska-style arrangement of records to represent various scopes of state, eliminating several redundant fields.


byorgey (approved):
Looks good to me, seems like a nice change.

Issue #154, opened 2021-10-03
Recursive types


xsebek:
Add a `rec` (like [`Fix`](https://hackage.haskell.org/package/data-fix-0.3.2/docs/Data-Fix.html)) operator which would open the door to any recursive data structure, like list, set or map.

This would also add two reserved function names:
- `roll : f (rec f) -> rec f`
- `unroll : rec f -> f (rec f)`

Example from IRC (lets assume `a + b = L a | R b`, and functions `getLeft`, `getRight`):
```haskell
type listf a b = () + a * b end
type list a = rec (listf a) end

def nil  : list a           = roll (L ()) end
def cons : a -> list a      = \x. \xs. roll (R (x,xs)) end
def head : list a -> a      = \xs. fst (getRight (unroll xs)) end
def tail : list a -> list a = \xs. snd (getRight (unroll xs)) end

> let l = cons 1 (cons 2 (cons 3 nil)) in head l, tail l
(1, cons 2 (cons 3 nil)) : int * rec 
```

## Special case for `list`

As the list is particularly useful it might be provided to players early and internally stored as `rec`, with special rules for parsing and pretty-printing. Something like `[1,2,3]` and hardcoding the type synonym `list`.
In gameplay, this could be introduced as a "stable form" of something more advanced. Having the appropriate device installed would load the basic list functions into the dictionary. Curious players could then type `> cons` and wonder what it will later be useful for. :slightly_smiling_face:  

## Related:
- would require #153 to get the `rec` applied to correct types
- #46 to make it useful (though there is the magical `streamf a = (int, () -> a)`)
- #495  (so less savvy players could `load data.list.sw` which they would find on Wiki)

xsebek:
@byorgey the equirecursive types look cool, but the [paper about them in F-omega system](https://doi.org/10.1145/2837614.2837660) seems a bit too powerful for Swarm. :open_mouth: 

I like the `rec` binding a type variable suggestion, as that seems easier to implement in the current type system. :+1: 

If I understand correctly, it is not even blocked on type synonyms (#153), though it would be annoying to use without them. :smile:

byorgey:
> @byorgey the equirecursive types look cool, but the [paper about them in F-omega system](https://doi.org/10.1145/2837614.2837660) seems a bit too powerful for Swarm. :open_mouth:

Well, we don't have System F-omega, just System F.

> If I understand correctly, it is not even blocked on type synonyms (#153), though it would be annoying to use without them.

Agreed.

[...]

PR #1802, opened 2024-04-24
Replace `unification-fd` with a custom implementation of unification

### LINKED ISSUE #1661: Custom implementation of unification
Right now we use the `unification-fd` package to implement [unification](https://en.wikipedia.org/wiki/Unification_(computer_science)) of types during type checking, [as described here](https://byorgey.wordpress.com/2021/09/08/implementing-hindley-milner-with-the-unification-fd-library/).  However, I think perhaps it's time to replace our use of `unification-fd` with our own custom implementation of unification.  This would have a number of benefits:

- `unification-fd` is quite old and it has some idiosyncracies that have been awkward to work around.  For example, it insists on working in the context of a monad transformer stack that has an error transformer on top of its own `IntBindingT` transformer.  If we implement our own we can integrate it with our existing effects framework instead.
- Implementing our own unification means the possibility to generate better custom error messages.
- `unification-fd` can only do pure structural unification.  However, several features we are contemplating might need more custom processing:
    - Type synonyms (#153) would require expanding type synonyms to their definitions whenever they are encountered during unification
    - Recursive types (#154) require being able to unroll recursive types one step when they are encountered during unification.
        - With equirecursive types, we also would have to track a set of encountered pairs during unification so that we can cut off the recursion when seeing a given pair again
        - Even with isorecursive types, it is too fragile to require that a recursive type be available in the right mode (check or infer) when encountering every `roll` or `unroll`; we need to be able to solve for unification variables matched with recursive types, which means being able to unroll them as needed during unification.

On the other hand there would also be some downsides:

- `unification-fd` is quite sophisticated and optimized; it is almost surely way faster than anything we might implement ourselves.  However, I am hoping that this will not make too much of a difference as long as our new implementation is "good enough".  And even if it isn't, we might be able to incorporate some optimization techniques from `unification-fd` into our own implementation.
- It represents a significant refactoring effort.

byorgey:
Closes #1661; towards #154.

`unification-fd` is very powerful, and extremely fast, but it was written a long time ago and its age shows.  It was not possible to incorporate it into our effects system in a nice way, necessitating the use of concrete monad transformers in the typechecking code.  In addition it is impossible to customize, and we have been contemplating new type system features such as #153 and #154 that turn out to require hooking into the way the unification algorithm works (see https://github.com/swarm-game/swarm/issues/154#issuecomment-1822877487 for more details).

This PR thus removes the dependency on `unification-fd` and implements our own version of unification.  It is not quite as fast as `unification-fd` but I consider the slowdown acceptable in order to gain e.g. recursive types.  And of course there is also room to optimize it.

The custom `UTerm` from `unification-fd` is replaced with the standard `Free` (free monad) construction from the `free` package, and the custom `Fix` from `unification-fd` is replaced with the one from `data-fix`.

We also get rid of the `unifyCheck` function, which used to be a quick short-circuiting way to check whether two types definitely did not unify or might unify, allowing us to give better error messages more quickly.  Now, the `=:=` unification operator itself just does this.


src/swarm-lang/Swarm/Effect/Unify/Fast.hs
```diff
@@ -0,0 +1,232 @@
...
+  unUnificationC >>> evalState idS >>> evalState (FreshVarCounter 0) >>> runThrow
+
+------------------------------------------------------------
+-- Unification
+
+-- The idea here (using an explicit substitution as a sort of
+-- "functional shared memory", instead of directly using IORefs), is
+-- based on Dijkstra et al. Unfortunately, their implementation of
+-- unification is subtly wrong; fortunately, a single integration test
+-- in the Swarm test suite failed, leading to discovering the bug.
```
kostmo, line 123: Which test, out of curiosity?

kostmo (approved):
Out of my depth, but exciting nonetheless!

src/swarm-lang/Swarm/Effect/Unify/Fast.hs
byorgey: It was `Challenges/Sliding Puzzles/_sliding-puzzle/solution.sw` that failed to typecheck, specifically the `getInitialBlankDestination` function.  Something about the way it involved inferring the type of a higher-order function that took tuple projections as an argument tickled the type inference algorithm in just the right way to trigger the bug.  So thanks for writing that code! :smile:

PR #1807, opened 2024-04-28
Change `appear` command to take an optional attribute

### LINKED ISSUE #1230: Command for changing robot display attribute (e.g. color)
We could either create a new command, or modify the existing `appear` command to accept another (optional) argument.
The type signature might change to:

    appear : text -> unit + text -> cmd unit

The `unit + text` represents either the new attribute name or no attribute override.

There is little usage of the `appear` command so far, so changing it should be easy.

byorgey:
Also, fix a bug---it didn't work before if an appearance string of length 5 was given.

Closes #1230.

I don't know how to write an automated test for this, but you can see it working by *e.g.*:

- Start a creative game
- Execute e.g. `appear "DNESW" (inr "rock")`
- Observe that the base now looks like a rock-colored `N`
- Now do some `turn` commands (including `turn down`) and observe the appearance changing to e.g. `S` when facing south


byorgey:
Scratching my head over this one.  It was working for me, and now it doesn't work.

EDIT: never mind, I was just being dumb, but a bug in the way we report fatal errors was introducing extra confusion.

byorgey:
@kostmo I think I've finally worked out all the kinks here.  Note that there were several places where robot code in scenarios was using `appear` (*e.g.* the bees in the beekeeping scenario, the gophers in the gopher scenario, *etc.*).  I updated all of them simply by providing `inl ()` as a second argument.  However, please feel free to give them more appropriate attributes if you wish; you can just push any changes directly to this branch.

stack.yaml
```diff
@@ -17,3 +17,4 @@ extra-deps:
 - AhoCorasick-0.0.4
 resolver: lts-22.10
 
+system-ghc: true
```
kostmo, line 20: Does this solve a particular problem?

kostmo (approved):

stack.yaml
byorgey: Oops, I have that in my personal `stack.yaml` as a workaround for some bug or other (I forget what), I didn't mean to commit it.  I'll revert that change.

byorgey:
@kostmo I assume by your approval that you didn't want to change any of the arguments to `appear` in scenarios, so I'll go ahead and merge this.  Of course we can always change those later as well.

kostmo:
> @kostmo I assume by your approval that you didn't want to change any of the arguments to `appear` in scenarios, so I'll go ahead and merge this. Of course we can always change those later as well.

Yes I may revisit later.

Issue #742, opened 2022-10-10
Terraforming


byorgey:
Just a placeholder for now: it should be possible, somehow, to change the terrain.  See #205.

xsebek:
The natural solution would be to add a `terraform` command that would require God's capability. :innocent: 

I am not sure what is the best structure to represent tiles or if we even need to optimize it if can not measure it realistically yet. I suppose we could run one tick with a mutable array and get back a normal one by using `runSTArray`. :thinking:

xsebek:
I think a good performance test would be to run a line of robots across the screen and change the terrain of every visible location.

We could run that a few times, left to right and back again, like windshield wipers.

Compare that with the same run with the terraform replaced by `noop` and see if it is even noticeable to the player.

byorgey:
Using actual mutable arrays for the terrain seems problematic; see the discussion at https://github.com/swarm-game/swarm/issues/30 .  Another possibility would be to do something similar to entities: we have a base, immutable array storing initial terrain values, and then a map of "overrides".  To look up the terrain value for a particular cell you first look it up in the override map, and then fall back to the base array if you don't find it.

xsebek:
I think with terraforming the common use case will be doing sweeping terraforming of an area once and then leaving it as is.

See my EDIT2 - if we collect the terrain updates at the end of the tick, we can update the pure immutable array once. So if you have a lot of robots changing the terrain at the same time it will be more efficient but the data can stay immutable. This will be particularly useful for `As` which must not change the original immutable array and in fact, it will never even get to it because it does not run in ticks.

xsebek:
As another thought, we should add a counterpart function `scanTerrain: cmd (unit + text)`. Should be trivial to add. :slightly_smiling_face:

byorgey:
I think updating the immutable terrain array once per tick is probably too frequent.  Imagine you just have one robot walking back and forth changing the terrain - it can only affect one square per tick and you would be regenerating the entire tile every tick.  But I can imagine buffering a bunch of terrain updates and then regenerating the immutable tile array once enough updates have accumulated.

xsebek:
I bet it would not be so bad, especially if the compiler can figure out that we will not need the original array anymore. But we will have to try measure it and see. 🙂

byorgey:
I think it will be bad :smile: But yes, we will have to benchmark it!

PR #1815, opened 2024-05-02
biomes

### LINKED ISSUE #1642: Whitelist biomes for growable entities
A new, optional Entity property `biomes` could be added to restrict the terrain in which a `Growable` entity may actually grow.
This would be the first non-cosmetic use case for terrain, i.e. in which terrain can affect gameplay.  c.f. #205.

For backward compatibility, an empty/absent `biomes` list means there is no restriction on where the entity can grow.

This might also serve as mechanism to limit the spread of growth (https://github.com/swarm-game/swarm/issues/1533#issuecomment-1722511849).

This may motivate a refactoring of terrain to allow custom scenario-defined terrain types: #1641

kostmo:
Closes #1642

### DIFF
--- a/data/schema/entity.json
+++ b/data/schema/entity.json
@@ -82,7 +82,7 @@
                     "dirt"
                 ]
             },
-            "description": "A list of properties of this entity."
+            "description": "A list of terrains that support growth by this entity. Empty list means no growth restrictions."
         },
         "yields": {
             "default": null,
--- a/data/schema/entity.json
+++ b/data/schema/entity.json
@@ -73,6 +73,17 @@
             "$ref": "combustion.json",
             "description": "Properties of combustion."
         },
+        "biomes": {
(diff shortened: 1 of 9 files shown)


data/schema/entity.json
```diff
@@ -73,6 +73,17 @@
...
+        "biomes": {
+            "default": [],
+            "type": "array",
+            "items": {
+                "type": "string",
+                "examples": [
+                    "dirt"
+                ]
+            },
+            "description": "A list of properties of this entity."
```
byorgey, line 85: Description should be updated.

byorgey (approved):
Looks good to me, just a couple minor things to fix.

Might be interesting to incorporate this into standard entities in the classic open world (but perhaps only once we address #742 ).  Would make farming a bit more interesting!

Issue #502, opened 2022-06-27
Make seeds real entities


byorgey:
Originally proposed by @xsebek .  See the discussion at #490 .

byorgey:
I think I still like this idea.  Here are some more concrete thoughts:
- For each `growable` entity named `X`, we automatically generate another entity named `X seed` with default attributes.
- ...*unless* there is *already* an entity named `X seed`, in which case we just leave it alone.  This way you can define seeds with custom attributes, if you want.
- There probably should be some way to customize how often seeds drop.  For example, some entities might drop a lot of seeds every time they are harvested, where others might only occasionally drop one seed.
- We might want to add a new command `plant` for planting seeds as opposed to `place` which would simply place a seed entity on the ground with no special behavior.  This is analogous to the reason we added `harvest` vs `grab`, so that you can control when you do and don't want growth to happen.

xsebek:
Good idea, you might be placing seeds just to give them to another robot. :+1:

xsebek:
If we order the unions right, then the generated seeds will be overwritten with explicit seeds. :slightly_smiling_face:

byorgey:
Now that we have a `sow` command and more robust growth mechanisms (#1817), I'd like to revisit this.  For a very brief recap of the context here:

- It needs to be possible to simply `place` and `grab` things without causing any growth to happen (see the long discussion at #490 for motivation).
- Growing/farming stuff should be possible for some entities, but with a bit more effort.
- In #490 three alternatives were considered, and we ended up implementing option 2, namely, adding a `harvest` command which acts like `grab` but causes an entity to immediately start regrowing (#503).
    - This seemed like the simplest sensible option, and indeed, I think it probably was the right choice at the time.
    - However, it has some weird issues: in particular, if you repeatedly `place` and then `harvest` on the same cell, you end up with a bunch of different seed robots all growing at the same time.
    - It also seems kind of unrealistic.

I'm still brainstorming here, but concretely, I would propose the following:
- We currently have `yields` which specifies what entity will be obtained from either `grab` or `harvest`.
- I propose generalizing `yields` to optionally contain a more complex record, something like this:
  ```
  yields:
    grab: oak tree
    harvest:
      - products: [(1, log), (1, branch)]
        weight: 2
      - products: [(1, log), (2, branch)]
        weight: 2
      - products: [(1, log), (1, branch), (2, acorn)]
        weight: 1
  ```
[...]

PR #1817, opened 2024-05-03
spreadable plant growth 

### LINKED ISSUE #1533: Growth that spreads
As a continuation of #1355 (which addressed only fire/combustion), should implement a new `spread` attribute for `Growable` entities.

It was proposed in #1355 that there may be some common base between spreading weeds and fire, but having implemented the latter, there is less overlap than anticipated.

kostmo:
Closes #1533

# Demo
```
scripts/test/run-tests.sh --test-options '--pattern "sow"'
```
or
```
scripts/play.sh -i data/scenarios/Testing/1533-sow-command.yaml --autoplay
```
Screenshot from 2024-05-03 19-12-08

## Other changes

* Introduced the `ChildInheritance` type to specify how the `Display` attribute is inherited by built child robots.
* Introduce the `mature` property of `growth`.  For example, it doesn't make much sense to plant an `acorn` and call the resulting, fully-grown plant also an `acorn`.  Instead, an `acorn` matures into an `oak`.

### DIFF
--- a/data/scenarios/Testing/1533-sow-seed-maturation.yaml
+++ b/data/scenarios/Testing/1533-sow-seed-maturation.yaml
@@ -69,7 +69,7 @@ entities:
       char: 'a'
     description:
       - Seed of an `oak`{=entity}
-    properties: [known, growable]
+    properties: [known, pickable, growable]
     growth:
       duration: [200, 300]
       mature: oak
--- a/data/scenarios/Testing/1533-sow-seed-maturation.yaml
+++ b/data/scenarios/Testing/1533-sow-seed-maturation.yaml
@@ -1,0 +1,101 @@
+version: 1
+name: Seed maturation
+description: |
+  A seed grows into a different entity, which again
+  yields the original seed when harvested.
+
+  The 'mature' sub-property of the 'growth' property
+  facilitates the distinction between seeds and the
+  entities those seeds produce. For example, it doesn't make much sense
+  to plant an `acorn`{=entity} and call the resulting,
+  fully-grown plant also an `acorn`{=entity}.
+
+  In combination with 'yields', it
+  effectively precludes the grown entity (as it exists in the world)
+  from entering into the player's inventory.
+creative: false
+objectives:
+  - id: plant_acorn
+    goal:
+      - |
+        Plant an `acorn`{=entity}
+    condition: |
+      as base {
+        x <- has "acorn";
+        return $ not x;
+      }
+  - goal:
+      - |
+        Go to the `oak`{=entity}
+    condition: |
+      as base {
+        ishere "oak";
+      }
+  - goal:
+      - |
+        Pick another `acorn`{=entity}
+    prerequisite: plant_acorn
+    condition: |
+      as base {
+        has "acorn";
+      }
+solution: |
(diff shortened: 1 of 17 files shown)


byorgey:
@kostmo I know this is still a draft, but in a nutshell, what's the difference between `place` and `sow`?

kostmo:
> @kostmo I know this is still a draft, but in a nutshell, what's the difference between `place` and `sow`?

`sow` initiates the growth lifecycle of a `Growable` entity (starting from a seed robot), whereas `place` just inserts the entity in its final state (an entity).

Stepping back, I needed to choose which "event" should instigate the action of "spreading" plants to nearby cells.  I decided that "spreading" should be a continuation of the natural growth cycle (the `seedProgram`).  Consequently, `harvest` and `sow` are the two player actions that can result in spreading, since both cause a plant to grow from a seed.

As implemented in the new version of `seedProgram`, some time after a seed (robot) becomes a mature plant (entity), a number of additional robots shall spawn, and (with some delay) each create their own robots that run `seedProgram`.  The `sow` command serves as a way to start the `seedProgram` again for these new plants.  `sow`, at this point, is primarily intended for this system robot use case; I don't actually have a concrete player-initiated use case in mind for `sow` yet.  Though it does naturally complement the `harvest` command.

In retrospect, I suppose simple recursion could be used to restart the `seedProgram` for these child plants.  But, looking ahead, I could imagine we might want to have other scenario-defined events spawn the growth of new crops, in which case a system bot would need the `sow` command as an entry point to the `seedProgram`.

Viewed another way, defining a new `Const` appears to be the only means for code reuse when I want to call one TH-defined program from another TH-defined program.  I vaguely recall encountering this conundrum when implementing `ignite` (#1432) as well.

For code reuse, it briefly occurred to me that it would be nice to be able to `mappend` one `ProcessedTerm` to another, though I'm not sure if doing so is conceptually sound.  Though it seems analogous to how a sequence of REPL commands are executed?

byorgey:
Makes sense, thanks!

I don't think it makes sense to `mappend` two `ProcessedTerm`s.  Yes, REPL commands can be executed/evaluated in sequence, but a sequence of REPL inputs is not itself a valid term.

byorgey (approved):
This looks good to me.  I definitely want to keep thinking about this more in the context of the classic open-world game and #502 , but I'll open another issue to discuss.

data/scenarios/Testing/1533-sow-seed-maturation.yaml
```diff
@@ -0,0 +1,101 @@
...
+      - treads
+    inventory:
+      - [1, acorn]
+entities:
+  - name: acorn
+    display:
+      char: 'a'
+    description:
+      - Seed of an `oak`{=entity}
+    properties: [known, growable]
```
byorgey, line 72: This doesn't matter all that much since it's just a testing scenario, but I just realized it's possible to get yourself stuck by executing `place "acorn"` instead of `sow "acorn"`; once you have placed it on the ground you can't pick it back up.  In this scenario I don't know of any reason why the `acorn` shouldn't be `pickable`?

byorgey:
Added a comment at #502 with some current thoughts.

PR #1819, opened 2024-05-03
Fix child robot privileges

### LINKED ISSUE #1664: Children of system robots do not bypass device requirements
Normally, system robots are not required to possess devices in order to use any particular command.

However, a robot built by a system robot becomes subject to the requirement (see worker bee code in #1599).

Also related: #1430

kostmo:
Fixes #1664

### DIFF
--- a/src/swarm-engine/Swarm/Game/Step/Const.hs
+++ b/src/swarm-engine/Swarm/Game/Step/Const.hs
@@ -1485,7 +1485,7 @@ execConst runChildProg c vs s k = do
   checkRequirements parentInventory childInventory childDevices cmd subject fixI = do
     currentContext <- use $ robotContext . defReqs
     em <- use $ landscape . terrainAndEntities . entityMap
-    creative <- use creativeMode
+    privileged <- isPrivilegedBot
     let -- Note that _capCtx must be empty: at least at the
         -- moment, definitions are only allowed at the top level,
         -- so there can't be any inside the argument to build.
@@ -1542,10 +1542,10 @@ execConst runChildProg c vs s k = do
         -- already has.
         missingChildInv = reqInv `E.difference` childInventory
 
-    if creative
+    if privileged
       then
         return
-          ( -- In creative mode, just equip ALL the devices
+          ( -- When 'privileged', just equip ALL the devices
             -- providing each required capability (because, why
             -- not?). But don't re-equip any that are already
             -- equipped.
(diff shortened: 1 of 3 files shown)


src/swarm-engine/Swarm/Game/Step/Const.hs
```diff
@@ -1542,7 +1542,7 @@ execConst runChildProg c vs s k = do
         -- already has.
         missingChildInv = reqInv `E.difference` childInventory
 
-    if creative
+    if privileged
```
byorgey, line 1545: Should all privileged robots always be able to create devices + entities out of thin air like this?

Right now, if robot A is a system robot and tries to create a child robot B which (say) executes `move`, it won't be able to if it doesn't have `treads` in its inventory.  This change represents one way to fix this, which is to simply allow robot A to conjure `treads` out of thin air.  However, in a way this seems unnecessary, since robot B will also be a system robot and hence will not actually need the `treads`.

kostmo: > Should all privileged robots always be able to create devices + entities out of thin air like this?

I think so.

I suppose the alternative would be to just allow the `build` command to succeed, without equipping anything?  I'd be fine with that as well.  One goal of this change for me, is to minimize "forking" of behavior between "creative" mode and "system" robots.  I wouldn't want it to be that in "creative" mode, all of the requisite devices are equipped, but for "system" robots (not in creative mode), we bypass the `build` command's restrictions by a different code path.

byorgey: > I suppose the alternative would be to just allow the build command to succeed, without equipping anything?

Right, that's the alternative I had in mind.  In a sense this alternative feels more "correct" from an in-game logic point of view.  System robots can just do whatever they want; they don't need to equip devices on other system robots they build, since those other robots can also just do whatever they want.  However:

> One goal of this change for me, is to minimize "forking" of behavior between "creative" mode and "system" robots.

That's a very good point.  Let's keep things simple for now, and only make things more complex if really necessary at some point in the future.

src/swarm-engine/Swarm/Game/Step/Const.hs
```diff
@@ -1542,7 +1542,7 @@ execConst runChildProg c vs s k = do
...
       then
         return
           ( -- In creative mode, just equip ALL the devices
```
byorgey, line 1548: Let's fix this comment so it doesn't just refer to creative mode.

byorgey (approved):

PR #1826, opened 2024-05-07
structure grid expansion

### LINKED ISSUE #1780: Grow map bounds to accommodate placed structures
Currently, "structures" may be placed only inside the bounds of their parent map rectangle.  Otherwise, they will be cropped to the parent bounds.

This is problematic if we want to place structures at great distances from each other; currently, it would require that an explicit cell grid of great size be included in the YAML file.

# Design

The truncating behavior that is the status quo should be preserved as the default.
A new property of the `placement` object in YAML named `truncate` could default to `True`.  When overridden to `False`, structures that are overlaid atop a base structure will enlarge its area instead of being truncated.

Importantly, the "origin" must be made consistent in the base structure, when multiple substructures are to be overlaid upon it.  I think that the current behavior is that the upper left cell is treated as the origin.  If one of the substructure placements happens to enlarge the area to the left, this could move the origin and mess up subsequent placements.

kostmo:
Closes #1780.

This change allows "child" structures to be placed outside of the bounds of their "parent" map.  Until now, any child structures that exceed the bounds of their parent were simply truncated.

# Demo
```
scripts/play.sh -i data/scenarios/Testing/1780-structure-merge-expansion/structure-composition.yaml
```

### DIFF
--- a/src/swarm-scenario/Swarm/Game/Scenario/Topography/Structure/Assembly.hs
+++ b/src/swarm-scenario/Swarm/Game/Scenario/Topography/Structure/Assembly.hs
@@ -20,11 +20,13 @@ import Data.Map qualified as M
 import Data.Set qualified as Set
 import Data.Text (Text)
 import Data.Text qualified as T
+import Linear.Affine
 import Swarm.Game.Location
 import Swarm.Game.Scenario.Topography.Area
 import Swarm.Game.Scenario.Topography.Navigation.Waypoint
 import Swarm.Game.Scenario.Topography.Placement
 import Swarm.Game.Scenario.Topography.Structure
+import Swarm.Game.Scenario.Topography.Structure.Overlay
 import Swarm.Language.Syntax.Direction (directionJsonModifier)
 import Swarm.Util (commaList, quote, showT)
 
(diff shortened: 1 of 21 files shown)


src/swarm-scenario/Swarm/Game/Scenario/Topography/Structure/Assembly.hs
```diff
@@ -38,18 +40,21 @@ overlaySingleStructure ::
...
     MergedStructure overlayArea overlayPlacements overlayWaypoints <-
       mergeStructures inheritedStrucDefs (WithParent p) $ structure ns
 
     let mergedWaypoints = inputWaypoints <> map (fmap $ placeOnArea overlayArea) overlayWaypoints
         mergedPlacements = inputPlacements <> map (placeOnArea overlayArea) overlayPlacements
-        mergedArea = overlayGrid inputArea pose overlayArea
+        mergedArea =
+          if shouldTruncate
+            then overlayGridTruncated (gridContent inputArea) pose overlayArea
+            else overlayGridExpanded (gridContent inputArea) pose overlayArea
```
kostmo, line 53: TODO: use condition to select the function name, then make a single application of that function

byorgey:
@kostmo Starting to take a look at this; can you fill in the description with an elevator-pitch version of what this PR is doing?

src/swarm-scenario/Swarm/Game/Scenario/Topography/Structure/Overlay.hs
```diff
@@ -0,0 +1,161 @@
...
+      , "located at"
+      , show p
+      ]
+
+data OverlayPair a = OverlayPair
+  { _base :: a
+  , _overlay :: a
+  }
+
+getBottomRightCorner :: PositionedGrid a -> Location
```
byorgey, line 40: Seems like this code (`getBottomRightCorner`, `getNorthwesternExtent`, `getSoutheasternExtent`, etc.) could be made simpler by introducing some kind of `Rectangle` abstraction which consists of two `Location`s, an upper left and a bottom right (or perhaps a `Location` + `AreaDimension`).  Then define a `Semigroup` instance on `Rectangles` which takes the max + min appropriately, i.e. `r1 <> r2` is the smallest rectangle that contains both `r1` and `r2`.  Then instead of `getBottomRightCorner` we can just have a `getRectangle` function which gets the `Rectangle` corresponding to a `PositionedGrid`, etc.

kostmo: @byorgey added a `SubsumingRect` for this.

byorgey (approved):
Looks great!

PR #1828, opened 2024-05-10
Support GHC 9.8, bump upper bounds, and update CI

byorgey:
Make sure Swarm builds with the most recent versions of dependencies. The only one that is still lagging is `filepath`, but `servant` doesn't support `filepath-1.5` so Swarm can't either.

The only real update needed was due to some breaking changes in `aeson-2.2`, but the necessary changes were very minor.

Also update our CI: `haskell-ci` recently dropped support for `hlint`, in favor of projects defining their own standalone `hlint` workflows.


src/swarm-lang/Swarm/Language/Direction.hs
```diff
@@ -50,7 +50,7 @@ data AbsoluteDir = DEast | DNorth | DWest | DSouth
   deriving (Eq, Ord, Show, Read, Generic, Data, Hashable, Enum, Bounded)
 
 directionJsonModifier :: String -> String
-directionJsonModifier = map C.toLower . L.tail
+directionJsonModifier = map C.toLower . L.drop 1
```
kostmo, line 53: Nice workaround for partiality

kostmo (approved):

src/swarm-lang/Swarm/Language/Direction.hs
byorgey: Yeah, that's an easy one.  There's nothing like `drop` to replace `head`, but https://github.com/haskell/core-libraries-committee/blob/main/guides/warning-for-head-and-tail.md had some good ideas, including the idea to replace uses of `head` with a locally-defined version that pattern-matches and calls `error` explicitly.  That avoids the warning and also generates a better error message in case it ever does actually crash.

byorgey:
Whew, glad to finally get this done, I've been working on it all day!  Updates like this always end up taking longer than I think...

PR #1833, opened 2024-05-10
structure placement order

kostmo:
It is more intuitive in scenario design for "later" things to be placed "on top of" earlier things.

There were only a small number of scenarios for which placement order was relevant, so I fixed them.


byorgey (approved):
> It is more intuitive in scenario design for "later" things to be placed "on top of" earlier things.

I agree!

PR #1834, opened 2024-05-10
Split out `swarm-doc` and `swarm-tui` sublibraries

### LINKED ISSUE #1043: Split Swarm components into libraries

byorgey:
Split the final remaining default library into `swarm-doc` and `swarm-tui`.  Closes #1043 .


byorgey:
@kostmo I'd be curious to hear what you think about the problem explained in the comment in https://github.com/swarm-game/swarm/pull/1834/commits/e18a329c87273614d4417837bb5a83de273c4f62 (i.e. how to deal with a circular dependency between `swarm-web` and `swarm-doc`).

kostmo:
> @kostmo I'd be curious to hear what you think about the problem explained in the comment in [e18a329](https://github.com/swarm-game/swarm/commit/e18a329c87273614d4417837bb5a83de273c4f62) (i.e. how to deal with a circular dependency between `swarm-web` and `swarm-doc`).

I'm fine with commenting that out for now.  We can consider the web API docs a separate category of documentation from the other swarm docs, so it makes sense to exclude it from the swarm-docs library.

kostmo (approved):

PR #1836, opened 2024-05-11
topography sublibrary

kostmo:
Towards #1043.

The eventual goal of this sublibrary split is to have a self contained library that can compose 2D grids of arbitrary content (perhaps colored pixels, or boolean values).  This could be useful outside of the `swarm` game.

I would also like to write unit tests for the structure recognizer that are independent of the `Entity` type.

# Major Changes

## Direction module
* Moved `Swarm.Language.Syntax.Direction` to `swarm-util`, since both `swarm-lang` and `swarm-topology` depend on it, but not on each other.
* Removed the re-export of direction things from `Swarm.Language.Syntax`

## Structure module

The `Swarm.Game.Scenario.Topography.Structure` module has been split into two:
* `Swarm.Game.Scenario.Topography.Structure`
* `Swarm.Game.Scenario.Topography.Structure.Type`

The former retains the YAML parsing logic.  The latter is agnostic of `Enitiy` type and the palette .
At some future point, I might want to move the YAML parsing to this sublibrary while still retaining independence of `Entity` type.

## Structure recognizer

The structure recognizer is independent of the content of Cells (i.e. it does not need to know what an `Entity` is), except:
1. during initialization
2. when retrieving the original cell content after recognition

Type parameters for three kinds of data have been added to the recognizer:
1. `Cell`/`PCell`
2. `Entity`
3. `EntityName`

Eventually it may be possible to eliminate one or two of these type parameters, with some refactoring.


src/swarm-topography/Swarm/Game/Scenario/Topography/Structure/Recognition.hs
```diff
@@ -12,11 +12,11 @@ import Swarm.Game.Scenario.Topography.Structure.Recognition.Log
 import Swarm.Game.Scenario.Topography.Structure.Recognition.Registry
 import Swarm.Game.Scenario.Topography.Structure.Recognition.Type
 
-data StructureRecognizer = StructureRecognizer
-  { _automatons :: RecognizerAutomatons
-  , _foundStructures :: FoundRegistry
+data StructureRecognizer b en a = StructureRecognizer
```
byorgey, line 15: I think there should be a comment here explaining what the different type parameters represent.

src/swarm-topography/Swarm/Game/Scenario/Topography/Structure/Recognition/Registry.hs
```diff
@@ -29,35 +29,35 @@ import Data.Map.NonEmpty (NEMap)
...
 import Swarm.Game.Scenario.Topography.Structure.Recognition.Type
+import Swarm.Game.Scenario.Topography.Structure.Type qualified as Structure
 import Swarm.Game.Universe (Cosmic)
 import Swarm.Util (binTuples, deleteKeys)
 
 -- | The authoritative source of which built structures currently exist.
-data FoundRegistry = FoundRegistry
-  { _foundByName :: Map StructureName (NEMap (Cosmic Location) StructureWithGrid)
-  , _foundByLocation :: Map (Cosmic Location) FoundStructure
+data FoundRegistry b a = FoundRegistry
```
byorgey, line 38: Again, I think we should add a comment here explaining what the type parameters `b` and `a` generally represent.

byorgey (approved):

PR #1840, opened 2024-05-11
actually move all `swarm-tui` modules into a `swarm-tui` subfolder

byorgey:
#1834 created a `swarm-tui` sublibrary for the remaining source modules that weren't in a sublibrary of their own, but I forgot to actually move them all into a `swarm-tui` folder.


swarm.cabal
```diff
@@ -674,7 +674,7 @@ library swarm-tui
     swarm:swarm-scenario,
     swarm:swarm-util,
 
-  hs-source-dirs: src
+  hs-source-dirs: src/swarm-tui
```
kostmo, line 677: Add a `cabal-gild` line just above the `exposed-modules:` line, like so:
https://github.com/swarm-game/swarm/blob/d6c0fb006dc589fb8d95107c795fbe155e335f70/swarm.cabal#L545-L546

Then run the `scripts/normalize/cabal.sh` script, which will ensure that all of the files are in the expected directory.

kostmo (changes requested):

swarm.cabal
kostmo: Fir this sublibrary, need to change `discover src/swarm-doc` to `discover src/swarm-tui`.

byorgey: Good catch, done!

kostmo (approved):

PR #1845, opened 2024-05-13
Insert parsed comments back into the AST and pretty-print with comments

### LINKED ISSUE #1467: Make `--format` preserve comments
> I noticed it drops comments, but we can fix that later.

_Originally posted by @xsebek in https://github.com/swarm-game/swarm/issues/1459#issuecomment-1692394374_

byorgey:
This PR does three main things:
1. Insert parsed comments into the AST
2. Extend pretty-printing to include comments
3. Extend the `format` subcommand with a few additional options and move the code to `Swarm.Language.Format`.

The pretty-printed code-with-comments is OK but not great.  It does a reasonable job with comments in standard-ish places; for example, it turns
```
// This function increments a number
def incr : int -> int =
  \n. n + 1
end

/* This command does some stuff.  It is super
   cool and important. */
def stuff : cmd unit =
  move;
  move;
  move;  // the third move is important
  move;
end
```
into
```

// This function increments a number
def incr: int -> int = \n. n + 1 end;

/* This command does some stuff.  It is super
   cool and important. */
def stuff: cmd unit =
  move;
  move;
  move // the third move is important
  ;
  move
end
```
which is good other than the fact that it moves the inline comment after `move;` to before the semicolon.

However, it turns this:
```
// This function does a cool math thing
def foo : int -> int =     // This is an optional type signature
  // pre
  \n. n + 1    // add one
end

/* This is a
   block comment which
   spans multiple lines */

def bar : int -> int   // Another type signature, = on the next line
  = \x. foo /* very important to use foo here */ (foo x)   // very cool implementation
end

def baz : cmd unit =
  move;
  move;
  turn left;   // don't forget to turn left!
  move
end

// And one last thing
```
into this:
```

// This function does a cool math thing
def foo: int -> int 
  = \n.
  n + 1 // add one

end // This is an optional type signature
;

/* This is a
   block comment which
   spans multiple lines */
def bar: int -> int 
  = \x.
  foo /* very important to use foo here */ (
    foo x // very cool implementation

  )
end // Another type signature, = on the next line
```
which has several obvious problems.  I think I know what the problem is in most cases; it will just require more engineering and special cases to get the output to look nicer, but I honestly don't really want to spend more time on this right now.  I'm hoping we can merge this as is (since it is still better than the status quo, namely, deleting all comments) and continue to improve it in the future.

The important point is that I ran the code formatter on every single `.sw` file in the repository and then re-ran the test suite; all the tests passed. So at least `swarm format` does not seem to break anything even if the output does not look the best.

Closes #1467 .


src/swarm-lang/Swarm/Language/Syntax.hs
```diff
@@ -990,6 +998,76 @@ constInfo c = case c of
...
+-- SrcLoc
+------------------------------------------------------------
+
+data SrcLoc
+  = NoLoc
+  | -- | Half-open interval from start (inclusive) to end (exclusive)
+    SrcLoc Int Int
+  deriving (Eq, Ord, Show, Data, Generic, FromJSON, ToJSON)
+
+instance Semigroup SrcLoc where
```
kostmo, line 1011:
```suggestion
-- | subsumes the span within the extrema of both inputs
instance Semigroup SrcLoc where
```

src/swarm-lang/Swarm/Language/Syntax.hs
```diff
@@ -990,6 +998,76 @@ constInfo c = case c of
...
+  NoLoc <> l = l
+  l <> NoLoc = l
+  SrcLoc s1 e1 <> SrcLoc s2 e2 = SrcLoc (min s1 s2) (max e1 e2)
+
+instance Monoid SrcLoc where
+  mempty = NoLoc
+
+-- | Check whether one SrcLoc starts before another one.
+srcLocBefore :: SrcLoc -> SrcLoc -> Bool
+srcLocBefore (SrcLoc a _) (SrcLoc b _) = a <= b
```
kostmo, line 1021: Justify choice of `<=` vs `<`?

src/swarm-lang/Swarm/Language/Syntax.hs
```diff
@@ -990,6 +998,76 @@ constInfo c = case c of
...
+--   comments come textually before the node and some come after.
+data Comments = Comments
+  { _beforeComments :: Seq Comment
+  , _afterComments :: Seq Comment
+  }
+  deriving (Eq, Show, Generic, Data, ToJSON, FromJSON)
+
+makeLenses ''Comments
+
+instance Semigroup Comments where
```
kostmo, line 1063: Can/should this instance be derived?

src/swarm-lang/Swarm/Language/LSP/Hover.hs
```diff
@@ -60,8 +60,8 @@ showHoverInfo ::
 showHoverInfo _ p vf@(VirtualFile _ _ myRope) =
   case readTerm' content of
     Left _ -> Nothing
-    Right (Nothing, _) -> Nothing
-    Right (Just stx, _) -> Just $ case processParsedTerm stx of
+    Right Nothing -> Nothing
```
kostmo, line 63: Use `fmap`?

kostmo (approved):

src/swarm-lang/Swarm/Language/Syntax.hs
byorgey: I added a comment to clarify.  The fact is it doesn't really matter in practice, since we only ever use this to compare `SrcLoc`s between comments and code, which cannot possibly start at the same index.

src/swarm-lang/Swarm/Language/Syntax.hs
byorgey: We could if we used a package such as https://hackage.haskell.org/package/generic-monoid ; I'm not sure it's worth it.

PR #1849, opened 2024-05-13
Update sublibrary graph generation script & cabal description

byorgey:
As a follow-up to #1834:

- Update the sublibrary graph generation script so it works with the no-longer-linear graph
- Regenerate the actual image
- Update the `.cabal` description to match


scripts/gen/render-sublibrary-dependencies.sh
```diff
@@ -16,4 +16,4 @@ cd $SCRIPT_DIR/../..
 rm -r dist-newstyle/cache
 cabal build --dry-run
 
-cabal-plan --hide-global --hide-builtin dot --tred --root swarm | twopi -Tsvg -o docs/image/sublibrary-graph.svg
\ No newline at end of file
+cabal-plan --hide-global --hide-builtin dot --tred --root swarm | sed 's/"swarm-[^:]*:\(lib\|exe\):/"/g' | sed 's/LR/BT/' | dot -Tsvg -o docs/image/sublibrary-graph.svg
```
kostmo, line 19: Add a comment to explain what the two sed passes are doing

swarm.cabal
```diff
@@ -16,12 +16,14 @@ description:
...
-  * swarm game engine
-  * swarm TUI
-  * swarm app
+  * swarm-util: miscellaneous utilities
+  * swarm-lang: parsing, typechecking, etc. for the Swarm language
+  * swarm-scenario: scenario descriptions, parsing, & processing
+  * swarm-engine: game simulation
+  * swarm-doc: generating documentation
+  * swarm-tui: textual user interface
+  * swarm-web: web interface
```
kostmo, line 25: Would it be appropriate to explain here why the web interface depends on `swarm-tui`?

kostmo (approved):

swarm.cabal
byorgey: Why *does* the web interface depend on `swarm-tui`?  I'm not sure I know the answer to that off the top of my head.

In any case I'm not sure this is the place to explain that, it seems like too much detail for a comment in the `.cabal` description.  I've started working on #1689 which seems like a better place for that (and we can add a link to the module guide to the `.cabal` description).

byorgey: Sure thing, sorry, I didn't realize it was waiting for a review.

PR #1851, opened 2024-05-15
Add `format --v0.5` option to port code from older syntax

byorgey:
This is a followup on top of #1583 which turns `swarm format` into a tool for porting existing Swarm code into the newest syntax, via an extra `--v0.5` argument.  In particular, this PR:

- Generalizes the parser to take a configuration record, which among other things contains the language version being parsed.
- Adds code to allow the parser to parse either the current syntax or one version ago (when types did not start with capital letter) depending on the version in the configuration.
    - The idea is to have the parser always support the current version and one older version, so we can always upgrade version n to version n+1.
- Adds a new flag `--v0.5` to the `format` subcommand which causes the input to be parsed in v0.5 mode.  However, the output of `format` will always use the latest syntax.  Thus, `swarm format --v0.5` reads code in v0.5 format and prints it in the latest format, so this can be used to automatically port existing `.sw` files.

This PR also makes a few minor improvements to pretty-printing.

### DIFF
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -12,6 +12,7 @@ import Options.Applicative
 import Swarm.App (appMain)
 import Swarm.Language.Format
 import Swarm.Language.LSP (lspMain)
+import Swarm.Language.Parser.Core (LanguageVersion (..))
 import Swarm.TUI.Model (AppOpts (..), ColorMode (..))
 import Swarm.TUI.Model.UI (defaultInitLgTicksPerSecond)
 import Swarm.Version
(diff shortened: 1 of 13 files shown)


kostmo (approved):
Looks like this PR is a superset of #1583?  For reviewability, the base branch should be set to `feature/uppercase-types` so we don't review the same subset of changes twice.

byorgey:
> For reviewability, the base branch should be set to `feature/uppercase-types` so we don't review the same subset of changes twice.

Oh, that makes sense, thanks, that's what I was missing.  I guess I haven't done stacked PRs that much and I wasn't sure how to do it.

PR #1856, opened 2024-05-19
Implement GitHub authentication

kostmo:
## Authentication flow

1. Users are represented by a GitHub username (primary key) and an "authentication cookie" in the SQLite database.
2. Site prompts user to login when the cookie is nonexistent or does not match any user in the database.
3. Upon clicking the login link, redirects user to the GitHub login page.  GitHub sends a code to our callback url.  We use that code to look up the username of the person who is logging in.  We generate and store a new cookie in the database row for that username and set the cookie value on the user's client.
4. As long as the client keeps sending the cookie value known to the server, all uploads/activity will be attributed to their GitHub username.

## New features

* Login/Logout
* All uploaded content is attributed to an authenticated GitHub user
* Separate pages for scenario lists and solution lists
* Download a solution file
* Unit tests exercise cookies/login flow

### DIFF
--- a/tournament/web/script/list-solutions.js
+++ b/tournament/web/script/list-solutions.js
@@ -1,0 +1,90 @@
+function insertTableRows(myTableBody, entries) {
+    for (const entry of entries) {
+        const rowItem = document.createElement("tr");
+        const cellVals = [
+            mkLink(entry.solutionSubmitter, "https://github.com/" + entry.solutionSubmitter),
+            regularSpan(entry.submissionTime),
+            regularSpan(entry.submissionScore.solutionCharacterization.scenarioSeed),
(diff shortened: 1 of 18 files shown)


tournament/web/script/list-solutions.js
```diff
@@ -0,0 +1,90 @@
...
+            if (response.status == 401) {
+
+                response.text().then(login_url => {
+                    const login_message = "Please log in";
+                    const a = document.createElement("a");
+                    a.setAttribute('href', login_url);
+
+                    a.appendChild(document.createTextNode(login_message));
+                    document.body.insertBefore(a, myTable);
+                });
```
kostmo, line 79: TODO: This code branch is obsolete

byorgey (approved):
Looks good to me!

@kostmo , when you get a chance, could you take a look at #1583 and #1851 ?

PR #1863, opened 2024-05-23
Use nubOrd instead of nub

nitinprakash96:
`nub` has a complexity of `O(n^2)` while `nubOrd` runs in `O(n logn)`. The only gotcha is that is needs an `Ord` instance.
But we don't have to worry about it in this PR since we're using lists.

### DIFF
--- a/src/swarm-scenario/Swarm/Game/Location.hs
+++ b/src/swarm-scenario/Swarm/Game/Location.hs
@@ -42,9 +42,9 @@ module Swarm.Game.Location (
 
 import Control.Arrow ((&&&))
 import Data.Aeson (FromJSONKey, ToJSONKey)
+import Data.Containers.ListUtils (nubOrd)
 import Data.Function (on, (&))
 import Data.Int (Int32)
-import Data.List (nub)
 import Data.Map (Map)
 import Data.Map qualified as M
 import Data.Yaml (FromJSON (parseJSON), ToJSON (toJSON))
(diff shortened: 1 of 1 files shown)


src/swarm-scenario/Swarm/Game/Location.hs
```diff
@@ -218,7 +218,7 @@ euclidean p1 p2 = norm (fromIntegral <$> (p2 .-. p1))
 --   See also @Swarm.Game.Step.Const.genDiamondSides@.
 getLocsInArea :: Location -> Int32 -> [Location]
 getLocsInArea loc r =
-  [loc .+^ V2 dx dy | x <- [0 .. r], y <- [0 .. r - x], dx <- nub [x, -x], dy <- nub [y, -y]]
+  [loc .+^ V2 dx dy | x <- [0 .. r], y <- [0 .. r - x], dx <- nubOrd [x, -x], dy <- nubOrd [y, -y]]
```
byorgey, line 221: Yes, in general `nub` is $O(n^2)$ and `nubOrd` is $O(n \lg n)$, but that only matters when $n$ is large.  In this specific case $n = 2$ (`nub` is just being used to turn `[x, -x]` into a singleton list when `x = 0`).  On such small lists `nub` is actually faster since `nubOrd` has the overhead of constructing a `Set`.  ([Here is a writeup of some benchmarks](https://andreaspk.github.io/posts/2019-02-01-nub-benchmarks.html) which suggest that `nub` is fastest for lists of length < 20.)

nitinprakash96: Oh neat. Today I learnt something 😄
Thanks!

PR #1865, opened 2024-05-27
Type synonyms

### LINKED ISSUE #153: Add type synonyms (or just generalize to System Fω)

byorgey:
Adds type synonym declarations to the language, like so:
```
tydef Maybe a = Unit + a end
def lookdown : Cmd (Maybe Text) = scan down end

> lookdown
it2 : Maybe Text = inl ()
```

`tydef` behaves very similarly to `def` except that it defines a new (parameterized) type instead of a new term.

Note that higher-kinded types are not yet supported; for example we cannot say `tydef Foo f = Unit + f Int` (in fact this would be a syntax error since type variables cannot be at the head of an application).  However there is nothing stopping us from adding that in the future; I just wanted to keep things simple for now.

If you know of any scenario code that would be a good candidate for using type synonyms, let me know --- it would be nice to update a few scenarios as a better stress test of type synonyms before merging this.

Closes #153.

### DIFF
--- a/editors/vim/swarm.vim
+++ b/editors/vim/swarm.vim
@@ -1,4 +1,4 @@
-syn keyword Keyword def end let in require
+syn keyword Keyword def tydef end let in require
 syn keyword Builtins self parent base if inl inr case fst snd force undefined fail not format chars split charat tochar key
 syn keyword Command noop wait selfdestruct move backup volume path push stride turn grab harvest sow ignite place ping give equip unequip make has equipped count drill use build salvage reprogram say listen log view appear create halt time scout whereami waypoint structure floorplan hastag tagmembers detect resonate density sniff chirp watch surveil heading blocked scan upload ishere isempty meet meetall whoami setname random run return try swap atomic instant installkeyhandler teleport as robotnamed robotnumbered knows
 syn keyword Direction east north west south down forward left back right
--- a/src/swarm-lang/Swarm/Language/Parser/Term.hs
+++ b/src/swarm-lang/Swarm/Language/Parser/Term.hs
@@ -7,7 +7,7 @@
 module Swarm.Language.Parser.Term where
 
 import Control.Lens (view, (^.))
-import Control.Monad (guard)
+import Control.Monad (guard, join)
 import Control.Monad.Combinators.Expr
 import Data.Foldable (asum)
 import Data.List (foldl')
@@ -22,8 +22,10 @@ import Swarm.Language.Parser.Record (parseRecord)
 import Swarm.Language.Parser.Type
 import Swarm.Language.Syntax
 import Swarm.Language.Types
+import Swarm.Util (findDup)
 import Text.Megaparsec hiding (runParser)
 import Text.Megaparsec.Char
+import Witch (into)
 
 -- Imports for doctests (cabal-docspec needs this)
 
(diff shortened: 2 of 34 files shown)


editors/vim/swarm.vim
```diff
@@ -1,4 +1,4 @@
-syn keyword Keyword def end let in require
+syn keyword Keyword def tydef end let in require
```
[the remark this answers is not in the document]
kostmo: I guess it's here, but I'm not acquainted with the mechanics:
https://github.com/swarm-game/swarm/blob/4a6f413c34dea226fede7d81ceb791cdce1e1c97/editors/vscode/syntaxes/swarm.tmLanguage.json#L13

src/swarm-lang/Swarm/Language/Parser/Term.hs
```diff
@@ -109,6 +114,21 @@ sLet x ty t1 = SLet (lvVar x `S.member` setOf freeVarsV t1) x ty t1
...
 
+-- | Create a polytype from a list of variable binders and a type.
+--   Ensure that no binder is repeated, and all type variables in the
+--   type are present in the list of binders (/i.e./ the type contains
+--   no free type variables).
+bindTydef :: [Var] -> Type -> Parser Polytype
+bindTydef xs ty
+  | Just repeated <- findDup xs = fail $ "Duplicate variable on left-hand side of tydef: " ++ into @String repeated
+  | not (S.null free) =
+      fail $
```
kostmo, line 125: Maybe we want to create a `failWords` utility method that has the convenience of builtin `unwords` like `failT`, but yields a string output.

src/swarm-lang/Swarm/Language/Types.hs
```diff
@@ -429,5 +432,84 @@ pattern UTyCmd ty = UTyConApp TCCmd [ty]
...
+--   number of arguments must match up; we don't worry about what
+--   happens if the lists have different lengths since in theory that
+--   can never happen.
+expandTydef :: Has (Reader TDCtx) sig m => Var -> [UType] -> m UType
+expandTydef userTyCon tys = do
+  mtydefInfo <- Ctx.lookupR userTyCon
+  tdCtx <- ask @TDCtx
+  -- In theory, if everything has kind checked, we should never encounter an undefined
+  -- type constructor here.
+  let tydefInfo = fromMaybe (error $ "Encountered undefined type constructor " ++ into @String userTyCon ++ " in expandTyDef (tdCtx = " ++ show tdCtx ++ ")") mtydefInfo
```
kostmo, line 483: Maybe better to use our `Text` helpers like `parens`, then convert to `String` at the end?

kostmo (approved):

editors/vim/swarm.vim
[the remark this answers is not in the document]
byorgey: Oh, OK, no worries, for some reason I thought you had created the VSCode configuration, but I see now I was misremembering.

src/swarm-lang/Swarm/Language/Parser/Term.hs
byorgey: Good idea!  It turns out we can just use `failT` itself here, since it already works in any `MonadFail`, which I had forgotten.

PR #1870, opened 2024-05-27
Update `FromJSON TickNumber` instance to match `ToJSON` instance

### LINKED ISSUE #1868: Save files fail to load
**Describe the bug**
Save files generated by previous invocations of the swarm executable fail to parse with error messages like:
```
Error in $.status.contents[1][1].attemptmetrics.durationmetrics.elapsedticks: parsing Swarm.Game.Tick.TickNumber(TickNumber) failed, expected Object, but encountered Number
```

**To Reproduce**
1. Run `swarm` and start a previously unplayed scenario.
2. Quit `swarm`.
3. Run `swarm` again and observe the parse error in Messages.  Also observe that data about the previous run of the scenario is not present in the New Game menu.

byorgey:
Fixes #1868.

### DIFF
--- a/src/swarm-engine/Swarm/Game/Tick.hs
+++ b/src/swarm-engine/Swarm/Game/Tick.hs
@@ -12,8 +12,8 @@ import Data.Aeson qualified as A
 import Data.Int (Int64)
 import GHC.Generics (Generic)
 import Prettyprinter (Pretty (..))
-import Swarm.Util.WindowedCounter (Offsettable (..))
 import Swarm.Util.JSON (optionsUnwrapUnary)
+import Swarm.Util.WindowedCounter (Offsettable (..))
 
 -- | A newtype representing a count of ticks (typically since the
 --   start of a game).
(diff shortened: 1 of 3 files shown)


src/swarm-engine/Swarm/Game/Tick.hs
```diff
@@ -17,11 +17,14 @@ import Swarm.Util.WindowedCounter (Offsettable (..))
...
 --   start of a game).
 newtype TickNumber = TickNumber {getTickNumber :: Int64}
-  deriving (Eq, Ord, Show, Read, Generic, FromJSON)
+  deriving (Eq, Ord, Show, Read, Generic)
 
 instance ToJSON TickNumber where
   toJSON = A.genericToJSON (A.defaultOptions {A.unwrapUnaryRecords = True})
 
+instance FromJSON TickNumber where
+  parseJSON = A.genericParseJSON (A.defaultOptions {A.unwrapUnaryRecords = True})
```
kostmo, line 26: Maybe want to define `(A.defaultOptions {A.unwrapUnaryRecords = True})` at the toplevel and share it?

kostmo (approved):

src/swarm-engine/Swarm/Game/Tick.hs
byorgey: Sure, good idea.  Do you think we might want to use it in other modules as well (i.e. should we put it in some kind of util module, or just define it right here)?

kostmo: Util module sounds good.  Perhaps name it `optionsUnwrapUnary`.

PR #1873, opened 2024-05-28
Represent types with applications of type constructors

byorgey:
We used to represent types with a different constructor for each different sort of type, e.g. there was a constructor `TyFunF` which took two types as arguments, etc.  This refactoring creates a new type `TyCon` which has things like `TCFun` and `TCBase` and represents various types as an application of a `TyCon` to some arguments.  For example function type `t1 -> t2` would be represented no longer as `TyFunF t1 t2` but instead as `TyConApp TCFun [t1, t2]`.

This is slightly more roundabout, and it does make parsing slightly trickier, but it greatly simplifies and shortens code for e.g. unification, extracting free variables, etc. (because what used to be many essentially identical cases now turn into a single case).  It also paves the way for #1865, so we have a way to represent new type constructors defined by the user.  E.g. if the user defined `tydef Maybe a = Unit + a` then we would represent the type `Maybe Int` as something like `TyConApp (TCUser "Maybe") [TyBase BInt]`.

This also means that types like `Int Unit` are no longer a parse error, so we need kind checking to rule out applications of type constructors to the wrong number of arguments; this PR adds such kind checking as well.  Having some sort of kind checking is unavoidable when allowing the user to define their own new type constructors (with an arbitrary number of arguments).

### DIFF
--- a/src/swarm-lang/Swarm/Language/Typecheck.hs
+++ b/src/swarm-lang/Swarm/Language/Typecheck.hs
@@ -467,75 +467,69 @@ data InvalidAtomicReason
 ------------------------------------------------------------
 -- Type decomposition
 
--- | Decompose a type that is supposed to be a delay type.  Also take
---   the term which is supposed to have that type, for use in error
+-- | Decompose a type that is supposed to be the application of a
+--   given type constructor to a single type argument. Also take the
+--   term which is supposed to have that type, for use in error
 --   messages.
-decomposeDelayTy ::
+decomposeTyConApp1 ::
   ( Has Unification sig m
   , Has (Throw ContextualTypeErr) sig m
   , Has (Reader TCStack) sig m
   ) =>
+  TyCon ->
   Syntax ->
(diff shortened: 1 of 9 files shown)


src/swarm-lang/Swarm/Language/Typecheck.hs
```diff
@@ -467,85 +486,89 @@ data InvalidAtomicReason
...
+------------------------------------------------------------
+-- Type inference / checking
 
 -- | Top-level type inference function: given a context of definition
 --   types and a top-level term, either return a type error or its
 --   type as a 'TModule'.
 inferTop :: TCtx -> Syntax -> Either ContextualTypeErr TModule
 inferTop ctx = runTC ctx . inferModule
 
+-- Note that we choose to do kind checking inline as we go during
```
kostmo, line 562: Should this comment be surfaced to Haddock?

kostmo (approved):

PR #1877, opened 2024-05-29
Make log error messages ephemeral

### LINKED ISSUE #1291: Syntax/type errors should be ephemeral
Currently, if you `run` a `.sw` file that has syntax or type errors, the error messages go into your log, and hence stick around forever.  This is not ideal since (1) who cares about past error messages, and (2) when you load a `.sw` file again it is hard to know whether you have fixed the problems since the error messages are still there in your log.

Somehow the errors should be shown to you but disappear later.  Maybe we could make a new kind of "ephemeral" log message that gets deleted at some point, perhaps the next time anything gets added to the log.  We could make `run` generate an ephemeral `OK, loaded` kind of message as well, so it would replace any errors from the previous `run` attempt.

byorgey:
Closes #1291.  The idea is that log entries produced by `log` or `say` will always be shown; but other log messages (error messages, status reports, etc.) are "ephemeral", meaning they will be shown as long as they are the most recent log entry, but will no longer be shown once a new log entry is added.  The motivation is that it is often unhelpful to see one's log full of error messages.

- All the log messages are still retained in the `RobotLog`; we just filter out some entries when displaying the log in the info panel.
- All unfiltered log entries are still shown in the Messages dialog.
- Also added a `run: OK.` log entry for successful execution of `run`, so that it in the common situation when you are trying repeatedly to `run` a file and then fixing type errors, you can immediately tell when you are successful since the error will disappear to be replaced by an `OK` message.

To test this, start a new game and try executing some commands that cause an error (e.g. `run` a non-existent file, or execute `move` with the base robot, etc.) as well as executing `log` or `say`.  Observe that `log` and `say` entries persist, whereas error messages are shown only until a new log entry is added.

### DIFF
--- a/src/swarm-tui/Swarm/TUI/View.hs
+++ b/src/swarm-tui/Swarm/TUI/View.hs
@@ -934,6 +934,7 @@ colorLogs e = case e ^. leSource of
     Said -> robotColor rid
     Logged -> notifAttr
     RobotError -> colorSeverity (e ^. leSeverity)
+    CmdStatus -> notifAttr
  where
   -- color each robot message with different color of the world
   robotColor = indexWrapNonEmpty messageAttributeNames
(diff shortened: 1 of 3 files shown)


src/swarm-tui/Swarm/TUI/View.hs
```diff
@@ -1443,15 +1444,26 @@ drawRobotLog :: AppState -> Widget Name
...
  where
   logEntries = s ^. gameState . to focusedRobot . _Just . robotLog
+  n = Seq.length logEntries
+
+  -- Only show the most recent entry, and any entries which were
+  -- produced by "say" or "log" commands.  Other entries (i.e. errors
+  -- or command status reports) are thus ephemeral, i.e. they are only
+  -- shown when they are the most recent log entry, but hidden once
+  -- something else is logged.
+  logEntriesToShow = logEntries ^.. traversed . ifiltered shouldShow
```
kostmo, line 1458: If you promote this to a toplevel function, `getLogEntriesToShow`, then this comment can be surfaced to haddock.  It also encapsulates other local definitions, like `n` and `shouldShow`.

kostmo (approved):

PR #1880, opened 2024-05-29
Always use parentheses when pretty-printing sum and product types

### LINKED ISSUE #1625: Type parentheses dropped in markdown rendering
Given the following markdown text (from [this `.yaml` file](https://github.com/swarm-game/swarm/blob/f9ef0094f0bfbe57d589e44a355446a54566d4bd/data/scenarios/Fun/horton.yaml#L55)):

    `path : (unit + int) -> ((int * int) + text) -> cmd (unit + dir)`

It is rendered as:
Screenshot from 2023-11-14 16-52-38

The dropped parentheses make it unclear whether the second argument is supposed to be a product type within a sum type `((int * int) + text)` or a sum type within a product type `(int * (int + text))`.

byorgey:
Fixes #1625.

Before: `path : Unit + Int -> Int * Int + Text -> Cmd (Unit + Dir * Int)`
After: `path : (Unit + Int) -> ((Int * Int) + Text) -> Cmd (Unit + (Dir * Int))`

We essentially consider sum, product, and arrow to all have the same precedence level; but arrows still associate to the right.

### DIFF
--- a/src/swarm-lang/Swarm/Language/Pretty.hs
+++ b/src/swarm-lang/Swarm/Language/Pretty.hs
@@ -156,19 +156,20 @@ instance ((UnchainableFun t), (PrettyPrec t)) => PrettyPrec (TypeF t) where
   prettyPrec _ (TyVarF v) = pretty v
   prettyPrec _ (TyRcdF m) = brackets $ hsep (punctuate "," (map prettyBinding (M.assocs m)))
   -- Special cases for type constructors with special syntax.
+  -- Always use parentheses around sum and product types, see #1625
   prettyPrec p (TyConF TCSum [ty1, ty2]) =
-    pparens (p > 1) $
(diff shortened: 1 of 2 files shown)


kostmo (approved):
Perhaps another test case is `((a * a) * (a * a))`, as in the [`density` command](https://github.com/swarm-game/swarm/wiki/Commands-Cheat-Sheet#Resonate) (which is not rendered properly on the wiki).

byorgey:
It actually is technically rendered properly on the website, since type product associates to the right (so we can write `(1,2,3) : Int * Int * Int` as syntax sugar for `(1,(2,3)) : Int * (Int * Int)`).  However I agree it is confusing, and with this PR it is indeed rendered with more parens.  I will add that as a test case.

PR #1881, opened 2024-05-30
Refactoring: Parse JSON using ApplicativeDo + RecordWildCards

### LINKED ISSUE #1878: Refactor JSON parsing code to use `ApplicativeDo` + `RecordWildCards`?
See https://www.haskellforall.com/2024/05/prefer-do-notation-over-applicative.html

nitinprakash96:
Taking a stab at refactoring `FromJSON` instances. Initially while skimming through the swarm code, I found multiple occurrences of using applicatives for `FromJSON` instances but was hesitant to bring it up. In my experience, it's almost always ergonomic to use `do` notation with `RecordWildCards`.

Closes #1878

### DIFF
--- a/src/swarm-scenario/Swarm/Game/Scenario/Topography/Placement.hs
+++ b/src/swarm-scenario/Swarm/Game/Scenario/Topography/Placement.hs
@@ -1,4 +1,5 @@
 {-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE RecordWildCards #-}
 
 -- |
 -- SPDX-License-Identifier: BSD-3-Clause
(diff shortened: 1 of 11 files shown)


src/swarm-scenario/Swarm/Game/Scenario/Topography/Placement.hs
```diff
@@ -82,10 +83,9 @@ data Placement = Placement
...
-    shouldTruncate <- v .:? "truncate" .!= True
-    p <-
-      Pose
-        <$> v .:? "offset" .!= origin
-        <*> v .:? "orient" .!= defaultOrientation
-    return $ Placement sName shouldTruncate p
+    src <- v .: "src"
+    truncateOverlay <- v .:? "truncate" .!= True
+    offset <- v .:? "offset" .!= origin
+    orient <- v .:? "orient" .!= defaultOrientation
```
nitinprakash96, line 89: This makes it seem like `Pose` shoud've had a `FromJSON` instance 😄

byorgey: It probably should, feel free to add one!

byorgey (approved):
This looks great!  I was already mostly convinced by the blog post I linked, but actually seeing the diff, this style looks so much better.  Much more readable and maintainable.

Is this every `FromJSON` instance in the codebase, or were there other ones that you didn't touch because they seemed tricky/less obvious/etc.?

nitinprakash96:
@byorgey I missed a few but have added them in my latest commit. Note that a lot of the `FromJSON` instances are pretty much straightforward and using `genericParseJSON` so they don't really need to be touched.

Issue #1869, opened 2024-05-27
Test round-tripping of save + achievement data files in CI


byorgey:
> More generally, we should add something to our CI that makes sure we can round-trip save files + achievement data.

_Originally posted by @byorgey in https://github.com/swarm-game/swarm/issues/1868#issuecomment-2133907943_

kostmo:
Related: #1507

byorgey:
I could imagine testing round trips in both directions, which would both be useful.

1. Generate achievement/save files from scratch, then test that we can parse the results back in, and that the results match the values used to generate them.  This would guard against errors or omissions in serialization.  Ideally, when we add new features we would make sure to update these tests to exercise the new features.
2. Keep some "canonical" save + achievement files stored in the repo somewhere, and test that we can parse and then re-serialize them, resulting in identical files as the originals.  This would guard against breakage that fails to parse existing files.

PR #1882, opened 2024-05-31
Fix FromJSON instance for better error msgs

nitinprakash96:
Before:
```
ghci> let s = "[{ \"radius\": 2, \"density\": \"x\"}]"
ghci> eitherDecodeStrict s :: Either String GrowthSpread
Left "Error in $: parsing Growth failed, expected Object, but encountered Array
```

The error msg above is misleading because we weren't trying to parse `Growth` datatype but `GrowthSpread` instead.

After:
```
ghci> let s = "[{ \"radius\": 2, \"density\": \"x\"}]"
ghci> eitherDecodeStrict s :: Either String GrowthSpread
Left "Error in $: parsing GrowthSpread failed, expected Object, but encountered Array"
```

### DIFF
--- a/src/swarm-scenario/Swarm/Game/Entity.hs
+++ b/src/swarm-scenario/Swarm/Game/Entity.hs
@@ -192,7 +192,7 @@ data GrowthSpread = GrowthSpread
   deriving (Eq, Ord, Show, Read, Generic, Hashable, ToJSON)
 
 instance FromJSON GrowthSpread where
-  parseJSON = withObject "Growth" $ \v -> do
+  parseJSON = withObject "GrowthSpread" $ \v -> do
     spreadRadius <- v .: "radius"
     spreadDensity <- v .: "density"
     pure GrowthSpread {..}


nitinprakash96:
A couple of things that I've noticed:
- names in `withObject`, `withText` etc do not correspond 1-1 with the datatype. This can be misleading when dealing with error msgs like in the description of the PR.
- Manually written `FromJSON` instances should have manually written `ToJSON` instances. This enables round trip tests.
Here is an example of not doing so:
```
ghci> let s = "{\"radius\": 2, \"density\": 0.3}"
ghci> eitherDecodeStrict s :: Either String GrowthSpread
Right (GrowthSpread {spreadRadius = 2, spreadDensity = 0.3})
ghci>
ghci> let spread = GrowthSpread 2 0.3
ghci> encode spread
"{\"spreadDensity\":0.3,\"spreadRadius\":2}" -- different than what the input was
```
Having written it manually gives us more control over the JSON in question. Is there a reason why we do not do so?

byorgey (approved):

byorgey:
@nitinprakash96 , no, there's no reason to have a manually written `FromJSON` instance but an autogenerated `ToJSON` instance.  If we can't round-trip the encoding then that is definitely a bug.  In fact I just fixed another such bug recently, see #1870.  I also added a ticket to test round-tripping of JSON instances in our CI, since currently there's nothing to catch things like this: #1869.  Currently, we don't really use the `ToJSON` instance for anything like `GrowthSpread` in a scenario description (we only read them from the data files in `data/scenarios`).  But we will eventually need them once we implement #50.

nitinprakash96:
Makes sense. Let me see if I can get round tripping to work in the test suite.

PR #1890, opened 2024-06-02
Extend syntax highlighting in emacs + Vim config

### LINKED ISSUE #1883: Update editor language modes for uppercase types
All the editor modes (e.g. `swarm.el` for emacs, etc.) still refer to types with lowercase names, which means syntax highlighting won't work anymore.  This should have been part of #1583 but I didn't think of it at the time.

nitinprakash96:
Changes include:
- Refactor `swarm-mode.el`
  - Everything doesn't need to be in one `let` binding inside a `setq` function. I've split
    into `defvar`s depending on the type. Eg: `swarm-mode-commands-regexp` defines
    all the commands in swarm game with `regex-opt` on them.
- Add syntax highlighting for operators like `->`, `=`, `+`, `:` etc. This is Haskell-ish.
- Make syntax highlighting respect case sensitivity.
- Add `autoload` macro to `define-derived-mode` since it defines the syntax table and syntax related variables.
- Add `commentary` section in top level comments as emacs convention.
- Update Vim config in order for syntax highlighting to be case sensitive for types.

Before:
CleanShot 2024-06-02 at 20 55 47@2x

After:
CleanShot 2024-06-02 at 20 52 12@2x

In order to test locally:
1. Open `swarm-mode.el` in emacs. Then `M-x eval-buffer`
2. Then open up any `.sw` file and `M-x swarm-mode`


editors/emacs/swarm-mode.el
```diff
@@ -7,151 +7,170 @@
...
+    (modify-syntax-entry ?\/ ". 124b" synTable)
+    (modify-syntax-entry ?* ". 23" synTable)
+    (modify-syntax-entry ?\n "> b" synTable)
+    synTable)
+  "Syntax table for `swarm-mode'.")
 
-(setq swarm-mode-syntax-table
-      (let ( (synTable (make-syntax-table)))
+(defvar swarm-mode-operators-regexp
+  (regexp-opt '(":" "->" "=" "<-" "+" "*" "-") t)
```
nitinprakash96, line 28: I think $ can be added here as well.

byorgey: Yes, `$` can be added I think.

editors/emacs/swarm-mode.el
```diff
@@ -7,151 +7,170 @@
...
-               "forward"
-               "left"
-               "back"
-               "right"
-             ))
-             (x-types '("int" "text" "dir" "bool" "cmd" "void" "unit" "actor"))
+(defvar swarm-mode-types-regexp
+  (concat
+   "\\b"
+   (regexp-opt '("Int" "Text" "Dir" "Bool" "Cmd" "Void" "Unit" "Actor") t)
```
byorgey, line 151: How about just making this match on *any* word that starts with an uppercase letter?  If I use `tydef` to make my own type (which is required to start with a capital letter) it would be nice if my new type name is also highlighted like a type.

Technically it is possible to also create term-level variables that start with an uppercase letter, in which case those would be highlighted incorrectly, but in practice I don't think that's a big deal since I assume everyone will typically name their term variables starting with lowercase letters.

byorgey (commented):
This looks great.  Just a few things that should be addressed:
1. Please remove the "Closes #1883" comment in the description, since I don't want to close that issue until we have updated the other officially supported editors as well.
2. This does not currently pass the integration test suite (`cabal test swarm:swarm-integration`) because the test suite checks for the presence of specific text in the editor configuration files (to make sure we update them when e.g. adding new primitive commands or keywords).  I guess our CI is not quite configured correctly because it thought it did not have to run that test suite when only this file was edited. (Really, we ought to move the editor configuration checks out into their own test suite...) In any case, I think the discrepancy is just down to different indentation---you should be able to update `swarm-doc/Swarm/Doc/Keyword.hs` to account for the change.

nitinprakash96:
@byorgey Thanks for catching that. I've addressed the comments. I've also updated the vim config for highlighting types to work. Surprisingly, the regex from emacs config worked in there as well.

Although note that instead of modifying `Keywords.hs`, I've resorted to keep the old style in `swarm-mode.el` because it seemed like too much of a change for not so ideal solution. Ideally speaking, just matching on the commands and built in functions should be enough instead of matching on the entire file. I'll probably come up with something in the future.

editors/emacs/swarm-mode.el
```diff
@@ -7,151 +7,170 @@
...
+   "\\b[A-Z][a-zA-Z_]*\\b"
+  "Regexp that recognizes types (all uppercase strings are supposed to be types) in swarm language.")
 
-             (x-keywords-regexp (regexp-opt x-keywords 'words))
-             (x-builtins-regexp (regexp-opt x-builtins 'words))
-             (x-commands-regexp (regexp-opt x-commands 'words))
-             (x-types-regexp (regexp-opt x-types 'words)))
+(defvar swarm-mode-keywords-regexp
+  (concat "\\b" (regexp-opt '("def" "tydef" "end" "let" "in" "require") t) "\\b")
+  "Regexp that recognizes types in swarm language.")
```
byorgey, line 158: Should this say "keywords" instead of "types"?

byorgey (approved):
Looks good to me!

editors/emacs/swarm-mode.el
nitinprakash96: Right!

PR #1894, opened 2024-06-03
Recursive types

### LINKED ISSUE #154: Recursive types

byorgey:
Implements recursive types.  Example:
```
tydef List a = rec l. Unit + a * l end

def nil : List a = inl () end
def cons : a -> List a -> List a = \x. \l. inr (x, l) end

def foldr : (a -> b -> b) -> b -> List a -> b = \f. \z. \xs.
  case xs
    (\_. z)
    (\c. f (fst c) (foldr f z (snd c)))
end

def map : (a -> b) -> List a -> List b = \f.
  foldr (\y. cons (f y)) nil
end
```
As discussed at #154, this uses the syntax `rec x. F(x)` to define the type `x` which is a solution to `x = F(x)`.  However, unlike the discussion there, I ended up going with *equirecursive* types (so a recursive type is *equal to* its unfolding). For example, as you can see above, if we have a value of type `List a` (defined as `rec l. Unit + a * l`), then it actually has the equivalent type `Unit + a * List a`, so we can simply do a `case` on it, without having to `unroll` first.  This actually turned out to be *easier* to implement (and it is cooler).

There are multiple built-in functions that conceptually return a list but currently do something different (like return a fold, or take an index and return a single element, etc.)  We should consider changing them to actually return lists, but that should definitely be in a separate PR.

Closes #154.

### DIFF
--- a/src/swarm-lang/Swarm/Language/Parser/Lex.hs
+++ b/src/swarm-lang/Swarm/Language/Parser/Lex.hs
@@ -164,7 +164,7 @@ primitiveTypeNames = "Cmd" : baseTypeNames
 
 -- | List of keywords built into the language.
 keywords :: [Text]
-keywords = T.words "let in def tydef end true false forall require requirements"
+keywords = T.words "let in def tydef end true false forall require requirements rec"
 
 -- | List of reserved words that cannot be used as variable names.
 reservedWords :: Set Text
--- a/editors/emacs/swarm-mode.el
+++ b/editors/emacs/swarm-mode.el
@@ -154,7 +154,7 @@
   "Regexp that recognizes types (all uppercase strings are supposed to be types) in swarm language.")
 
 (defvar swarm-mode-keywords-regexp
-  (concat "\\b" (regexp-opt '("def" "tydef" "end" "let" "in" "require") t) "\\b")
+  (concat "\\b" (regexp-opt '("def" "tydef" "rec" "end" "let" "in" "require") t) "\\b")
   "Regexp that recognizes keywords in swarm language.")
 
 (defvar swarm-font-lock-keywords
--- a/editors/vim/swarm.vim
+++ b/editors/vim/swarm.vim
@@ -1,4 +1,4 @@
-syn keyword Keyword def tydef end let in require
+syn keyword Keyword def tydef rec end let in require
 syn keyword Builtins self parent base if inl inr case fst snd force undefined fail not format chars split charat tochar key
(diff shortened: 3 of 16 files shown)


byorgey:
I just realized I need to rule out types like `rec t. t` since any value may have that type, and it breaks type soundness.  Should not be too hard to address, but in the meantime feel free to take a look at what's already in the PR and try it out.

xsebek:
@byorgey Ha, thanks, I will try this out! 👍 

I noticed you needed the VSCode syntax updated and I see the `rec` keyword will also need to be added. 🙂

byorgey:
This should be ready now.  In theory we ought to have a device that provides the capability to use recursive types, but that will require some refactoring of the requirements analysis code so I opted to leave it for another PR: see #1898.

src/swarm-lang/Swarm/Language/Parser/Lex.hs
```diff
@@ -163,7 +163,7 @@ primitiveTypeNames = "Cmd" : baseTypeNames
 
 -- | List of keywords built into the language.
 keywords :: [Text]
-keywords = T.words "let in def tydef end true false forall require requirements"
+keywords = T.words "let in def tydef end true false forall require requirements rec"
```
nitinprakash96, line 166: This should probably be followed up with updating the editor configurations (`swarm-mode.el`, `swarm.vim` etc)?

byorgey: Good catch.  I will update that shortly.

nitinprakash96 (approved):

PR #1895, opened 2024-06-04
Update VSCode plugin

xsebek:
* Update VSCode plugin NPM package
* Update VSCode syntax
  * Rewrite the `definition` rule to make it simpler and work with multiline types
    * Use shared `#types` group with added support for `rec`
  * Refactor groups to use ordered include `patterns` and unordered `repository` of named subgroups
  * Add escape sequences to strings
* Fix a typo in a regular expression that broke syntax highlighting
* Generate the syntax JSON file from YAML
---

* Closes #1887

PR #1896, opened 2024-06-04
Fix non-exhaustive pattern in LSP

xsebek:
* fix an LSP exception on top level comment

Found while testing #1895.


byorgey (approved):
Oh, good catch, thanks!

PR #1902, opened 2024-06-05
More recipe for rubber

### LINKED ISSUE #1892: Should be possible to produce `rubber` in a `big furnace`
The recipe for `rubber` requires a `furnace` but it should be possible to use a `big furnace` as well.  See also #480.

nitinprakash96:
Should close #1892

### DIFF
--- a/data/recipes.yaml
+++ b/data/recipes.yaml
@@ -198,7 +198,7 @@
   out:
     - [1, furnace]
 - in:
-    - [10, rock]
+    - [100, rock]
     - [10, solar panel]
   out:
     - [1, big furnace]
@@ -625,6 +625,12 @@
   required:
     - [1, furnace]
 - in:
+    - [1, sand]
+  out:
+    - [1, glass]
+  required:
+    - [1, big furnace]
+- in:
     - [1, glass]
     - [1, mithril]
   out:
@@ -674,7 +680,6 @@
     - [1, furnace]
 - in:
     - [1, LaTeX]
-    - [1, log]
   out:
     - [1, rubber]
   required:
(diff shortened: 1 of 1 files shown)


data/recipes.yaml
```diff
@@ -672,6 +672,13 @@
     - [1, rubber]
   required:
     - [1, furnace]
+- in:
+    - [1, LaTeX]
+    - [1, log]
```
byorgey, line 677: To be consistent with the other recipes using `furnace` vs `big furnace`, this should not require a `log`.  (*e.g.* take a look at recipes for `copper wire`, `copper pipe`, and `iron plate`).  Generally the idea is that a `furnace` is easier to construct, but a `big furnace` doesn't need `log`s for fuel (and there are a couple things you can only make with a `big furnace`, e.g. `silicon` and `PhD thesis`).

Also, while we're at it, I just noticed that we should probably add a similar recipe for making `glass` with a `big furnace`.

As an aside, to make the game more balanced we might also consider requiring more `rock`s to construct a `big furnace` (say, 100).  I have recently been having fun once again developing strategies for the classic game, and I've realized that it's pretty easy to scrounge together 10 rocks instead of 5 and probably worth it to sacrifice 10 of your solar panels at the start just so that you don't have to worry about the supply of logs in the early game.  Requiring 100 rocks would make a `big furnace` effectively out of reach until you develop drilling technology (at which point you will definitely want to build one to be able to process mined `quartz` into `silicon`).

nitinprakash96: I see. Sounds good.
I've added big furnace based recipe for making glass. Also, increased the number of rocks required to make big furnace.

byorgey (approved):
Looks good!

PR #1906, opened 2024-06-06
Add device to provide `rectypes` capability

### LINKED ISSUE #1898: Device providing ability to use recursive types
#1894 adds a `CRectype` capability, but it is not used anywhere in the code, and currently it is always possible to use recursive types.  Since we went with equirecursive types, there is no special *term-level* syntax required to use recursive types, and we currently do not do requirements analysis on types.

We should create a device that provides the ability to use recursive types (and restrict their use otherwise).  This will necessitate adding requirements analysis for types (which is probably a good idea anyway, *e.g.* it should not be possible to mention a sum type unless you have an `ADT calculator` equipped).

byorgey:
Closes #1898.  Adds the `hyperloop` device (recipe: `ADT calculator` + `strange loop`) which provides the ability to use recursive types (#1894).  Also refactors requirements analysis code and adds the ability for types to trigger capability requirements.

### DIFF
--- a/data/entities.yaml
+++ b/data/entities.yaml
@@ -1034,7 +1034,7 @@
       of values is just a fancy number.  For example, the type `Bool`{=type} is
       just a fancy version of the number 2, where the two things happen to be
       labelled `false` and `true`.  There are also types `Unit`{=type} and
-      `void`{=type} that correspond to 1 and 0, respectively.
+      `Void`{=type} that correspond to 1 and 0, respectively.
     - |
       The product of two types is a type of pairs, since, for example,
       if `t`{=type} is a type with three elements, then there are 2 * 3 = 6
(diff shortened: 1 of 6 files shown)


data/entities.yaml
```diff
@@ -1053,6 +1053,21 @@
       3) (\x. 2*x) (\y. 3*y) == 9`.
   properties: [pickable]
   capabilities: [arith, sum, prod]
+- name: hyperloop
+  display:
+    attr: device
+    char: '∞'
+  description:
+    - |
```
nitinprakash96, line 1061: I'm not sure if this is just on my machine but the rendering for this entity is off?
CleanShot 2024-06-06 at 22 35 13@2x

Is this intentional?

byorgey: It's not intentional, and in fact I think it is related to #1794.  Somehow each of those code snippets `rec t. T(t)` and `t = T(t)` is being pretty-printed onto multiple lines.

nitinprakash96: Quite possibly.

nitinprakash96 (approved):

data/entities.yaml
byorgey: I have been trying to understand where this happens in the code and I have not been able to figure it out.  But in any case the issue is independent of this PR.  I created #1910 to track it.

PR #1915, opened 2024-06-07
Fix variable lookup during requirements analysis

### LINKED ISSUE #1914: Local variables pick up requirements of global vars with the same name
**Describe the bug**
Requirements analysis is apparently deciding that locally-bound variable (e.g. the argument of a lambda) have the same requirements as globally-defined things with the same name.

**To Reproduce**
```
> def m = move end
> requirements \m. m
```
Note that the reported requirements includes `treads`, which are definitely required to execute `move` but should not be required for the identity function `\m .m`.

**Additional context**
This must have been introduced by the refactoring in #1906 (which I really should have tested more thoroughly!).

byorgey:
When I refactored requirements analysis during #1906 I didn't properly update the variable case: the code continued to look up the variable in `ctx` which used to be the name of the local context being passed around; after the refactoring, however, `ctx` was the name of the initial context passed to the top-level `requirements` function, and the local context was passed around as a `Reader` effect.  Thus, the lambda case (for example) was correctly deleting the bound variable from the context, but then when checking the variable inside the body of the lambda we looked up its requirements in the initial, global context instead of the modified local one.

This PR fixes the issue and also adds some test cases, one of which was failing before this fix.

Fixes #1914.

### DIFF
--- a/test/unit/TestRequirements.hs
+++ b/test/unit/TestRequirements.hs
@@ -1,0 +1,49 @@
+{-# LANGUAGE OverloadedStrings #-}
+{-# OPTIONS_GHC -fno-warn-orphans #-}
+
+-- |
+-- SPDX-License-Identifier: BSD-3-Clause
+--
+-- Swarm requirements analysis tests
+module TestRequirements where
+
+import Control.Lens (view)
+import Data.Set qualified as S
+import Data.Text (Text)
+import Swarm.Language.Capability
+import Swarm.Language.Context qualified as Ctx
+import Swarm.Language.Pipeline
+import Swarm.Language.Requirement
+import Test.Tasty
+import Test.Tasty.HUnit
+import TestUtil (check)
+
+testRequirements :: TestTree
+testRequirements =
+  testGroup
+    "Requirements analysis"
+    [ testGroup
+        "Basic capabilities"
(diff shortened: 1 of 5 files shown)


test/unit/TestRequirements.hs
```diff
@@ -0,0 +1,55 @@
...
+checkRequirements code expect = check code (expect . view processedRequirements)
+
+requiresCap :: Text -> Capability -> Assertion
+requiresCap code cap = checkRequirements code ((cap `S.member`) . capReqs)
+
+check :: Text -> (ProcessedTerm -> Bool) -> Assertion
+check code expect = case processTerm code of
+  Left err -> assertFailure $ "Term processing failed: " ++ from @Text err
+  Right Nothing -> assertFailure "Term is empty"
+  Right (Just pt) -> assertBool "Predicate was false!" (expect pt)
```
xsebek, line 55: Is this not used anywhere else? I would expect this to be in `TestUtil`.

xsebek (approved):
Nice unit tests! 👍

test/unit/TestRequirements.hs
byorgey: Good idea, I'll move it there.

PR #1917, opened 2024-06-09
Fix Lazy IO opening too many files in integration tests

xsebek:
When running on macOS, the integration tests open too many file handles,
even when you do not need to run any swarm file tests.
```sh
> cabal run -O0 swarm-integration -- -p editors
swarm-integration: data/scenarios/Challenges/_bucket-brigade/hauler.sw: openFile: resource exhausted (Too many open files)
```
The solution is to get the file paths only and safely open them as needed.
Surprisingly, this even makes the integration test code simpler.

* discovered while fixing failing tests in #1895


byorgey:
`479-atomic-race` needs a longer timeout, I have seen spurious failures with that one occasionally.

byorgey (approved):
:+1:

PR #1919, opened 2024-06-09
Longer timeout for 479-atomic-race test

byorgey:
I have seen the `479-atomic-race` integration test timeout several times recently (most recently in #1917).  Increase the timeout for that test from the default 1s to 2s.


xsebek (approved):

byorgey:
Four things in the Mergify queue at the same time, this must be a new record! :smile:

PR #1924, opened 2024-06-10
Make structures parseable independently of entity definitions

kostmo:
This refactoring makes structures parseable independently of `Entity` and `Robot `definitions.

We can define a `Palette` that maps to an arbitrary type, for example `RGBColor` instead of `Cell Entity`.

## Testing
```
scripts/test/run-tests.sh standalone-topography
```


src/swarm-topography/Swarm/Game/Scenario/Topography/ProtoCell.hs
```diff
@@ -0,0 +1,41 @@
...
+  deriving (Eq, Show)
+
+instance (FromJSONE e a) => FromJSONE e (SignpostableCell a) where
+  parseJSONE x =
+    withObjectE "SignpostableCell" objParse x
+      <|> (SignpostableCell Nothing <$> parseJSONE x)
+   where
+    objParse v =
+      SignpostableCell
+        <$> liftE (v .:? "waypoint")
```
byorgey, line 40: After #1881 I think we are going to prefer the style using `do`-notation with named fields and `RecordWildcards` to assemble the final result, like so (untested):
```
objParse v = do
  waypointCfg <- liftE (v .:? "waypoint")
  standardCell <- v ..: "cell"
  pure SignpostableCell {..}

**Review by byorgey — APPROVED** [review]
Looks good to me!

Being able to define palettes of arbitrary types is something I thought about a long time ago when designing the world DSL.  I'd be interested to think more about how we can make use of this from within the world DSL.

File: src/swarm-topography/Swarm/Game/Scenario/Topography/Structure.hs [hunk 2]
```diff
@@ -12,38 +11,82 @@ module Swarm.Game.Scenario.Topography.Structure where
...
-      <*> v
-        ..: "structure"
-
-instance FromJSONE (TerrainEntityMaps, RobotMap) (PStructure (Maybe Cell)) where
+    structure <- v ..: "structure"
+    liftE $ do
+      name <- v .: "name"
+      recognize <- v .:? "recognize" .!= mempty
+      description <- v .:? "description"
+      return $ NamedArea {..}
```
kostmo, line 84: Using the new style

src/swarm-topography/Swarm/Game/Scenario/Topography/ProtoCell.hs
kostmo: Agreed, I like the new style, and used it [here](https://github.com/swarm-game/swarm/pull/1924/files#r1641329560).  I'll fix this one, too.

PR #1927, opened 2024-06-12
Refactor some functions to use case

byorgey:
I used to define functions with multiple clauses in a style like
```
f pat1 = ...
f pat2 = ...
f pat3 = ...
```
but since then have been convinced (by [this blog post](https://ro-che.info/articles/2014-05-09-clauses) and my own experience) that it is superior for multiple reasons to prefer using `case` (or `\case`), like
```
f = \case
  pat1 -> ...
  pat2 -> ...
  pat3 -> ...
```
This PR refactors several functions that were still using the old style.


nitinprakash96 (approved):
LGTM

PR #1933, opened 2024-06-15
nonempty grid rows

### LINKED ISSUE #1843: Use `NonEmpty` and separate `EmptyGrid` constructor for `Grid`
With a grid definition like `newtype Grid = Grid [[a]]`, there is more than one way to represent an "empty grid":
* `Grid []`
* `Grid [[]]`
* `Grid [[], [], [], [], [], ...]`

If instead we use `NonEmpty` for both rows and columns, then there can be exactly one "normalized" representation for an empty grid:
```
data Grid a = EmptyGrid | Grid (NonEmpty (NonEmpty a))
```

kostmo:
Closes #1843

Towards #1981

### DIFF
--- a/src/swarm-scenario/Swarm/Game/Scenario/Topography/WorldDescription.hs
+++ b/src/swarm-scenario/Swarm/Game/Scenario/Topography/WorldDescription.hs
@@ -18,9 +18,9 @@ import Swarm.Game.Entity
 import Swarm.Game.Land
 import Swarm.Game.Location
 import Swarm.Game.Scenario.RobotLookup
-import Swarm.Game.Scenario.Topography.Area (Grid (EmptyGrid))
 import Swarm.Game.Scenario.Topography.Cell
 import Swarm.Game.Scenario.Topography.EntityFacade
+import Swarm.Game.Scenario.Topography.Grid (Grid (EmptyGrid))
 import Swarm.Game.Scenario.Topography.Navigation.Portal
 import Swarm.Game.Scenario.Topography.Navigation.Waypoint (
   Parentage (Root),
--- a/src/swarm-scenario/Swarm/Game/Scenario/Topography/WorldDescription.hs
+++ b/src/swarm-scenario/Swarm/Game/Scenario/Topography/WorldDescription.hs
@@ -18,7 +18,7 @@ import Swarm.Game.Entity
 import Swarm.Game.Land
 import Swarm.Game.Location
 import Swarm.Game.Scenario.RobotLookup
-import Swarm.Game.Scenario.Topography.Area (emptyGrid)
+import Swarm.Game.Scenario.Topography.Area (Grid (EmptyGrid))
 import Swarm.Game.Scenario.Topography.Cell
 import Swarm.Game.Scenario.Topography.EntityFacade
 import Swarm.Game.Scenario.Topography.Navigation.Portal
@@ -84,7 +84,7 @@ integrateArea ::
 integrateArea palette initialStructureDefs v = do
   placementDefs <- v .:? "placements" .!= []
   waypointDefs <- v .:? "waypoints" .!= []
-  rawMap <- v .:? "map" .!= emptyGrid
+  rawMap <- v .:? "map" .!= EmptyGrid
   (initialArea, mapWaypoints) <- paintMap Nothing palette rawMap
   let unflattenedStructure =
         Structure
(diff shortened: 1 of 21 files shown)


src/swarm-topography/Swarm/Game/Scenario/Topography/Area.hs
```diff
@@ -94,8 +112,13 @@ getAreaDimensions cellGrid =
 computeArea :: AreaDimensions -> Int32
 computeArea (AreaDimensions w h) = w * h
 
+-- |
+-- Warning: size must be nonzero,
```
byorgey, line 116: This function seems unnecessarily partial.  Why not have a special case to first check whether `w` or `h` is zero, and if so return `EmptyGrid`?

src/swarm-scenario/Swarm/Game/Scenario/Topography/WorldDescription.hs
```diff
@@ -18,7 +18,7 @@ import Swarm.Game.Entity
 import Swarm.Game.Land
 import Swarm.Game.Location
 import Swarm.Game.Scenario.RobotLookup
-import Swarm.Game.Scenario.Topography.Area (emptyGrid)
+import Swarm.Game.Scenario.Topography.Area (Grid (EmptyGrid))
```
byorgey, line 21: It seems like a bit of a code smell to import a constructor of `Grid` just to be able to make an empty grid, especially if the eventual goal is to make a smart constructor.  I think the constructors of `Grid` should not be exported at all.  That way this module (and others like it) won't have to be updated if the constructors ever change again. We could either (1) continue to export `emptyGrid` as before, or (2) (slightly fancier) make an instance of `AsEmpty` and then import `pattern Empty` from `Control.Lens.Empty`.
```
instance AsEmpty Grid where
  _Empty = prism' (const EmptyGrid) (\case {EmptyGrid -> Just (); _ -> Nothing})
```

src/swarm-topography/Swarm/Game/Scenario/Topography/Rasterize.hs
```diff
@@ -5,13 +5,15 @@
 module Swarm.Game.Scenario.Topography.Rasterize where
 
 import Codec.Picture
+import Data.List.NonEmpty qualified as NE
 import Data.Vector qualified as V
 import Swarm.Game.Scenario.Topography.Area
 
 -- | Converts linked lists to vectors to facilitate
 -- random access when assembling the image
 gridToVec :: Grid a -> V.Vector (V.Vector a)
```
byorgey, line 14: I think this function should move to `Swarm.Game.Scenario.Topography.Area` (especially if we stop exporting the constructors of `Grid`).

src/swarm-topography/Swarm/Game/Scenario/Topography/Structure/Overlay.hs
```diff
@@ -68,14 +68,14 @@ computeMergedArea (OverlayPair pg1 pg2) =
 zipGridRows ::
   Alternative f =>
   AreaDimensions ->
-  OverlayPair (Grid (f a)) ->
+  OverlayPair [[f a]] ->
```
byorgey, line 71: What is the rationale for changing `Grid (f a)` to `[[f a]]` here?  It seems like a step backward in terms of generality.

src/swarm-topography/Swarm/Game/Scenario/Topography/Structure/Overlay.hs
```diff
@@ -124,11 +124,11 @@ padSouthwest ::
   Alternative f =>
   V2 Int32 ->
   OverlayPair (Grid (f a)) ->
-  OverlayPair (Grid (f a))
+  OverlayPair [[f a]]
```
byorgey, line 127: Same question here.

byorgey (approved):

src/swarm-topography/Swarm/Game/Scenario/Topography/Structure/Overlay.hs
kostmo: Yeah, I had attempted `Grid (f a)` at first, but encountered a bug in `padSouthwest` that was causing structures to get blanked out.  I didn't have time to fully investigate it.  Created #2004 to follow up.

src/swarm-scenario/Swarm/Game/Scenario/Topography/WorldDescription.hs
kostmo: For now there are still a number of places where the `Grid` constructor is used.  Hiding exports of the constructors will be a prerequisite for completing #1981.

PR #1936, opened 2024-06-15
Stop printing context after a definition.

### LINKED ISSUE #1336: Stop printing type error context when reaching "while defining..."
When checking a file full of definitions, technically all the definitions are joined by (possibly implicit) semicolons.  So when you get a type error in a definition, it currently prints something like

```
Error blah blah

- While checking the definition of foo
- While checking the left-hand side of a semicolon
- While checking the right-hand side of a semicolon
- While checking the right-hand side of a semicolon
- While checking the right-hand side of a semicolon
- While checking the right-hand side of a semicolon
```
This is obviously ridiculous and provides no useful information.  We should stop printing the context as soon as we print out a "While checking the definition of ..." since those can only occur at the top level, and the combination of the line and column info plus the name of the function is plenty to uniquely identify the location of the error.

Related: #1314, #1297

noahyor:
Fixes #1336.


byorgey (approved):
:+1:

byorgey:
@noahyor looks good, feel free to add the `merge me` label once you are happy with it.

Issue #1660, opened 2023-11-25
Mutable cells


byorgey:
It's a little hard to believe we don't already have an issue for this, but I couldn't find one.  It would be fun to add mutable reference cells to the language, so you could do some imperative programming when convenient.  Our CESK machines already track a mutable store, so this really just requires adding some surface language features that allow accessing it directly.

Proposal
-------------

In particular, we could add the following primitives:

- `Ref A` is the type of references to a mutable cell holding a value of type `A`
- `mkref : a -> cmd (ref a)` creates a new mutable cell holding the given value
- `read : ref a -> cmd a` reads the current value of a mutable cell
- `write : ref a -> a -> cmd unit` writes a new value to a mutable cell

We could optionally use special syntax for `read` and `write`, *e.g.* `!r` for `read r` and `r := x` for `write r x`.

Additional considerations
-------------------------------------

- We should put a bit of thought into what device(s) will provide the capability to create and use mutable cells.
- Within the world we've created, I'm not sure it makes sense to have *shared* mutable state.  Or at least, it would only make sense in certain restricted situations which we can work out in the future.  So passing a `ref a` value to another robot (e.g. via #94 ) should result in the cell being copied, not in a shared reference to the same mutable cell.
- Once we implement immutable arrays (#98), we could get mutable arrays simply by making an immutable array full of references.

xsebek:
We might want to keep `read` and `write` for reading values from paper. How about `deref` and `updateRef`? 🙂

byorgey:
> We might want to keep `read` and `write` for reading values from paper. How about `deref` and `updateRef`? 🙂

Oh, good catch.  When I first wrote this issue we didn't have paper.  I forgot that we now do already have a function defined the name `read` (and `print` instead of `write`).

PR #1949, opened 2024-06-17
Get rid of memoizing delay + switch to a strict store

### LINKED ISSUE #1948: Reevaluate use of `Store`/references for recursive variables
I forget why I did it this way, and I am not sure that it is really necessary.  It certainly introduces extra complication (*e.g.* the need to wrap free variables in `force` during elaboration).  We should either get rid of it, or add some clear comments explaining why it is necessary.

byorgey:
Closes #1948; see that issue for a much more in-depth discussion.

Depends on merging #1928 first.

A lot of this PR consists in deleting code that is either (1) ugly or (2) overly clever! :partying_face: 

The short version is that the `Store` used to incorporate some laziness + memoization: when a cell was first allocated, it was an unevaluated thunk; it then got evaluated the first time it was referenced.  However, this wasn't really needed to handle recursive definitions (which is the only thing we were using it for).  Getting rid of it means we can get rid of a lot of weird ugly code needed to wrap free variables in extra calls to `force` and so on.

The new and improved `Store` just stores `Value`s, period.  A special `VBlackhole` value was added, to be used while evaluating a recursive `let`.

Note that `VRef` is no longer really used, but I left it there for use in implementing #1660 .  Once we have mutable references we can use them + delay/force to implement lazy cells.

### DIFF
--- a/src/swarm-engine/Swarm/Game/CESK.hs
+++ b/src/swarm-engine/Swarm/Game/CESK.hs
@@ -60,8 +60,8 @@ module Swarm.Game.CESK (
   Store,
   Addr,
   emptyStore,
-  MemCell (..),
   allocate,
+  resolveValue,
   lookupStore,
   setStore,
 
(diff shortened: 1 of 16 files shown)


xsebek (approved):
This is making CESK much more approachable, I might even understand it one day. 😄

Great job @byorgey!

PR #1950, opened 2024-06-17
Generate operators for all editors

nitinprakash96:
Currently, operator names are generated only for vscode only using the command `cabal run swarm:swarm-docs -- editors --code`.
With this PR, I intend to bring that behaviour to all the editors.

Changes include:
- `cabal run swarm:swarm-docs -- editors` command now supports `vim` as well.
- `operatorNames` can generate operator list catering to all the editors supporting Swarm.
- Update operator list in `swarm-mode.el`, `swarm.vim` and `swarm.tmLanguage.yaml`.

How to test emacs syntax:
- Open `editors/emacs/swarm-mode.el` in emacs.
- Then `M-x eval-buffer`
- Open up any of the `.sw` file under `examples`.
- Then `M-x swarm-mode`

How to test vim syntax:
- Copy swarm.vim to vim directory using `cp editors/vim/swarm.vim ~/.vim/syntax/sw.vim`
- Setup auto detect in vim. `echo 'autocmd BufRead,BufNewFile *.sw set filetype=sw' > ~/.vim/ftdetect/sw.vim`
- Open up any of the `.sw` files under `examples`. (Also ensure that you have syntax on in vim. `ESC :syntax on`)

### DIFF
--- a/editors/emacs/swarm-mode.el
+++ b/editors/emacs/swarm-mode.el
@@ -25,7 +25,30 @@
   "Syntax table for `swarm-mode'.")
 
 (defvar swarm-mode-operators-regexp
-  (regexp-opt '(":" "->" "=" "<-" "+" "*" "/" "-") t)
+  (regexp-opt
+   '(
+     "-"
+     "=="
+     "!="
+     "<"
+     ">"
+     "<="
+     ">="
+     "||"
+     "&&"
+     "+"
+     "-"
+     "*"
+     "/"
+     "\\"
+     ":"
+     "^"
+     "++"
+     "$"
+     "->"
+     "<-"
+     "=")
+   t)
   "Regexp that recognizes operators for swarm language.")
 
 (defvar swarm-mode-commands-regexp
--- a/editors/emacs/swarm-mode.el
+++ b/editors/emacs/swarm-mode.el
@@ -43,6 +43,7 @@
      "^"
      "++"
      "$"
+     ":"
      )
    t)
   "Regexp that recognizes operators for swarm language.")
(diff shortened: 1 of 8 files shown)


editors/emacs/swarm-mode.el
```diff
@@ -25,7 +25,30 @@
...
+     "/"
+     "\\"
+     ":"
+     "^"
+     "++"
+     "$"
+     "->"
+     "<-"
+     "=")
+   t)
```
nitinprakash96, line 51: It seems that the list of operators are not in sync across editors. I'd like a second pair of eyes to check if this covers all or not.

xsebek: * What is `\\`? Is that lambda?
* There is duplicate `-` because one is unary, but that is OK.
* `++` will not matter I think, because `+` will be parsed twice instead. Should not be noticeable though.

It's probably best idea to test in emacs, but it looks good to me. 👍

xsebek: Why add it manually when you can copy the output of `cabal run swarm:swarm-docs -- editors --emacs`?

nitinprakash96: I need to get some input on what the operators are for now. 
The list generated by `operatorNames` in it's current implementation does not cater to emacs (and possibly vim). Emacs fails to parse the generated list because of escaping where it does not need it.

But ultimately I'll be using that command once I figure out the requirements by emacs and vim.

nitinprakash96: @xsebek My recent commit ensures that there's no manual labour involved. All the operators can now simply be copied to respective config files by running the `cabal run` command.

byorgey (approved):
LGTM!

PR #1953, opened 2024-06-17
Automatically insert matching close brackets at REPL

### LINKED ISSUE #1952: Typing open bracket at the REPL should insert matching close bracket
It would be super handy if typing any kind of open bracket (paren, square bracket, curly brace) at the REPL caused a matching close bracket to be immediately inserted to the right of the cursor!  I would love love love this feature.  It would both (1) make it easier to end up with matching brackets and (2) keep the current input valid more of the time.  I don't know why I never thought of this before.

byorgey:
Closes #1952.  Whenever you type `(`, `[`, or `{` at the REPL, a matching close bracket is automatically inserted.

I suppose the next level would be to have it delete both brackets if you hit backspace immediately after inserting an open bracket, but I'll leave that for another PR.  Doesn't seem that important anyway.

### DIFF
--- a/src/swarm-tui/Swarm/TUI/Controller.hs
+++ b/src/swarm-tui/Swarm/TUI/Controller.hs
@@ -353,9 +353,9 @@ handleMainEvent ev = do
         then -- ignore repeated keypresses
           continueWithoutRedraw
         else -- hide for two seconds
-          do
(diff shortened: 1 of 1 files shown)


noahyor (approved):
Great! I love this feature!

nitinprakash96 (approved):
LGTM

PR #1957, opened 2024-06-18
Highlight error part in REPL

### LINKED ISSUE #1956: Highlight only the problem in REPL
> It's annoying when you are in the middle of typing something like
> ```Haskell
> build {myComplexFunction ... }
> ```
> and you need to fill in the ..., but you can't remember exactly what the type of myComplexFunction is, but Swarm is no help because the entire thing does not typecheck so it is just highlighted in red.

@byorgey in #1955

I want the function `renderREPLPrompt` to split the text into invalid and valid parts and color them separately.
The solution is to also have error location in REPL model and to use `processParsedTerm` in `validateREPLForm`.

We could use other ascii art tricks, for example write `ˆˆˆ` on the line under.

A more complex solution to solving type checking problems is in #1955.

xsebek:
Closes #1956


xsebek:
@byorgey could you please have a look? It only works to a degree, and I suspect there are bugs:

* This looks OK:
  Screenshot 2024-06-18 at 21 29 34
* When no location is given it should highlight whole line:
  Screenshot 2024-06-18 at 21 30 06
* But it seems to suspiciously extend more then it should, so likely made a mistake somewhere:
  Screenshot 2024-06-18 at 21 30 57

Feel free to push to this branch if you find anything; I am out of ideas for today. 😄

src/swarm-tui/Swarm/TUI/Model/Repl.hs
```diff
@@ -275,6 +278,7 @@ data REPLState = REPLState
   { _replPromptType :: REPLPrompt
   , _replPromptEditor :: Editor Text Name
   , _replValid :: Bool
+  , _replErrorSrcLoc :: SrcLoc
```
byorgey, line 280: Right now we have `replValid :: Bool` and `replErrorSrcLoc :: SrcLoc`; but as far as I understand there can only be a `SrcLoc` other than `NoLoc` when `replValid` is `False`.  So why not combine them into a single field like `replValid :: Either SrcLoc ()` ?

src/swarm-tui/Swarm/TUI/Controller.hs
```diff
@@ -1,6 +1,9 @@
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE QuasiQuotes #-}
 {-# LANGUAGE RecordWildCards #-}
+{-# OPTIONS_GHC -Wno-unrecognised-pragmas #-}
+
+{-# HLINT ignore "Use fewer imports" #-}
```
byorgey, line 6: What is this HLINT pragma for?

byorgey (commented):
The part of the prompt being highlighted does not seem correct.  It always seems to extend from the beginning of the error all the way to the end of the line, no matter what.  For example, when I entered `build {(place 2); move }` everything from `2` to the closing `}` is red.

src/swarm-tui/Swarm/TUI/Controller.hs
xsebek: That's what HLS does if you by mistake click "ignore".

src/swarm-tui/Swarm/TUI/Model/Repl.hs
xsebek: Sure, it's almost isomorphic and should be easy to refactor. 👍

xsebek:
@byorgey, after reading your comment again, I found the arithmetic bug that extended the range. 👍

src/swarm-tui/Swarm/TUI/Controller.hs
```diff
@@ -1303,13 +1317,18 @@ validateREPLForm s =
     CmdPrompt _
       | otherwise ->
           let ctxs = Contexts (topCtx ^. defTypes) (topCtx ^. defReqs) (topCtx ^. tydefVals)
-              result = processTerm' ctxs uinput
-              theType = case result of
-                Right (Just pt) -> Just (pt ^. processedSyntax . sType)
-                _ -> Nothing
+              (theType, errSrcLoc) = case readTerm' defaultParserConfig uinput of
+                Left err ->
+                  let ((x1, _y1), (x2, _y2), _msg) = showErrorPos err
```
byorgey, line 1322: Aren't these (line, column)?  i.e. should we be ignoring `x1` and `x2` and using `y1` and `y2` instead?

src/swarm-tui/Swarm/TUI/View.hs
```diff
@@ -1540,11 +1540,18 @@ renderREPLPrompt focus theRepl = ps1 <+> replE
...
   prompt = theRepl ^. replPromptType
   replEditor = theRepl ^. replPromptEditor
-  color = if theRepl ^. replValid then id else withAttr redAttr
+  color t =
+    if theRepl ^. replValid
+      then txt t
+      else case theRepl ^. replErrorSrcLoc of
+        NoLoc -> withAttr redAttr (txt t)
+        SrcLoc s e ->
+          let (validL, (invalid, validR)) = T.splitAt (e - s) <$> T.splitAt s t
```
byorgey, line 1549: A common situation that I think we need a special case for is when there is a syntax error right at the very end.  This frequently happens when you have mismatched parens or a missing close quote etc.  Currently, in that case nothing gets highlighted at all, since the `invalid` part after the start location of the error (as well as `validR`) is the empty string.

Perhaps we can check whether the `invalid` part is the empty string (or only whitespace) and in that case highlight the entire REPL input?

byorgey: And yes, I was going to comment that this should be `e - s` instead of just `e` but you figured it out before me. :smile:

xsebek: Good idea, done.

src/swarm-tui/Swarm/TUI/Controller.hs
xsebek: It seems so. I copied the code that was using this function and assumed the convention is that `y` is lines.

xsebek:
Should work well now. Please have a look.

I hit up through my history and it seemed nicely highlighted.

byorgey (approved):
Looks great!  All the examples I tried seemed nicely highlighted too.  Thanks for putting this together, it's a very nice feature!

nitinprakash96 (approved):
Works nicely!

PR #1963, opened 2024-06-19
Delete stack.yaml

### LINKED ISSUE #1960: Delete `stack.yaml`
Stack currently cannot compile swarm on windows and I think it is currently not recommended.

However, we still have a stack.yaml, which is very annoying for me because VS Code defaults to stack, so I have to delete it manually every time I pull, and then explicitly not stage the deletion when I commit.

noahyor:
Delete stack.yaml

Closes #1960.


byorgey (approved):

xsebek:
@noahyor why not get rid of the HIE stack file and make the cabal one default?

noahyor:
Done.

noahyor:
We should also change the mergify rules accordingly.

nitinprakash96:
Since `hie.yaml` now defaults to cabal cradle, should it be committed to repository? If not, then I think we should still keep `hie.yaml` in `.gitignore`.

byorgey:
You're right, let's remove `hie.yaml` and add it to `.gitignore`.

PR #1965, opened 2024-06-19
Allow Cabal to recognize swarm as a library.

### LINKED ISSUE #1964: Cabal does not recognize swarm as a library.
Can be fixed by adding a blank main library to swarm.cabal.

noahyor:
Allow Cabal to recognize swarm as a library.

Closes #1964.


noahyor:
Why is docspec failing with such a strange error message? It compiled just fine the first time through.

xsebek:
@noahyor what does this `library` do?

I am no wiser from the Cabal message. Did you try running Docspec yourself?

swarm.cabal
```diff
@@ -117,6 +117,9 @@ common ghc2021-extensions
     TypeApplications
     TypeOperators
 
+-- Allow Cabal to recognize swarm as a library.
```
byorgey, line 120: It would be nice to add a comment here with a link to the Cabal issue you found that explains why this is a necessary workaround for now.

byorgey:
No, you can import modules from the other libraries like `swarm-lang` that are marked as public.  The point is that because of a Cabal bug it will not let you import anything at all from a package unless it has a default "main" library.

byorgey:
It seems like `docspec` does not like having an empty library component?  I'm not sure...

nitinprakash96:
Ah I see. Thanks a lot! I wasn't aware of that bug.

PR #1971, opened 2024-06-22
Reset scroll position when switching modals

### LINKED ISSUE #1901: Scrolling of modal windows affects each other
**Describe the bug**
If you scroll in one modal window (e.g. Recipes or Commands), close it, then open a different one, it will be scrolled to the same position as the previous window.  Sometimes, if one window has a lot of content and you are scrolled to the bottom, then close it and open another window, it will be blank until you start to scroll it, at which point it will suddenly jump to the content.

**To Reproduce**
You can reproduce this with either a lot of content in one of the windows (e.g. Commands), or just by making the terminal smaller.

nitinprakash96:
Closes #1901 

When switching description modals or info panel in the bottom left corner, the viewport used to retail the scrolled state.
This PR attempts to reset the scroll position when we switch modals.

### DIFF
--- a/src/swarm-tui/Swarm/TUI/Controller.hs
+++ b/src/swarm-tui/Swarm/TUI/Controller.hs
@@ -1470,23 +1470,27 @@ handleInventoryListEvent ev = do
 handleInventorySearchEvent :: BrickEvent Name AppEvent -> EventM Name AppState ()
 handleInventorySearchEvent = \case
   -- Escape: stop filtering and go back to regular inventory mode
-  EscapeKey ->
+  EscapeKey -> do
+    resetViewport infoScroll
     Brick.zoom (uiState . uiGameplay . uiInventory) $ do
       uiInventoryShouldUpdate .= True
       uiInventorySearch .= Nothing
   -- Enter: return to regular inventory mode, and pop out the selected item
   Key V.KEnter -> do
+    resetViewport infoScroll
     Brick.zoom (uiState . uiGameplay . uiInventory) $ do
       uiInventoryShouldUpdate .= True
       uiInventorySearch .= Nothing
     gets focusedEntity >>= maybe continueWithoutRedraw descriptionModal
   -- Any old character: append to the current search string
-  CharKey c ->
+  CharKey c -> do
+    resetViewport infoScroll
     Brick.zoom (uiState . uiGameplay . uiInventory) $ do
       uiInventoryShouldUpdate .= True
       uiInventorySearch %= fmap (`snoc` c)
   -- Backspace: chop the last character off the end of the current search string
   BackspaceKey -> do
+    resetViewport infoScroll
     Brick.zoom (uiState . uiGameplay . uiInventory) $ do
       uiInventoryShouldUpdate .= True
       uiInventorySearch %= fmap (T.dropEnd 1)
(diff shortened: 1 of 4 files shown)


byorgey (commented):
This resets the viewports on *every* robot panel event, but there are some events where we do not want to reset: for example, hitting `Enter` to pop out the description in a modal; hitting `m` to `make` an item; or even just hitting some random key that doesn't do anything.  I think we should (1) only reset the `infoScroll` when actually switching to a different item in the inventory list, and (2) only reset the `modalScroll` when opening a new modal.

src/swarm-tui/Swarm/TUI/Controller.hs
```diff
@@ -1510,6 +1511,7 @@ makeEntity e = do
 descriptionModal :: Entity -> EventM Name AppState ()
 descriptionModal e = do
   s <- get
+  resetViewport modalScroll
```
nitinprakash96, line 1514: Looks like description modals are handled a little differently than other modals. I wanted to reset the viewport in just one place (`openModal`) but that function doesn't really cover description.

src/swarm-tui/Swarm/TUI/Controller.hs
```diff
@@ -1469,23 +1470,27 @@ handleInventoryListEvent ev = do
 handleInventorySearchEvent :: BrickEvent Name AppEvent -> EventM Name AppState ()
 handleInventorySearchEvent = \case
   -- Escape: stop filtering and go back to regular inventory mode
-  EscapeKey ->
+  EscapeKey -> do
+    resetViewport infoScroll
```
nitinprakash96, line 1474: I'm not sure of the desired behavior in this case. If a user searches in the inventory list, scrolls a little bit in the info panel and then presses `Esc` in the inventory list, then do we want to reset the viewport? Or should we keep the state to where the user scrolled before pressing `Esc`?

src/swarm-tui/Swarm/TUI/Controller.hs
```diff
@@ -1458,6 +1458,7 @@ handleInventoryListEvent ev = do
   -- However, this does not work since we want to skip redrawing in the no-list case!
 
   mList <- preuse $ uiState . uiGameplay . uiInventory . uiInventoryList . _Just . _2
+  resetViewport infoScroll
```
byorgey, line 1461: This looks to me like it will still reset the viewport on *every* keypress in the inventory list, which would include *e.g.* hitting `m` to `make` the currently focused item or hitting some random key that has no effect.  We should only reset the viewport when actually moving to a different item.

src/swarm-tui/Swarm/TUI/Controller.hs
byorgey: I think in this case we should just keep the state to where the user scrolled before pressing `Esc`.

src/swarm-tui/Swarm/TUI/Controller.hs
```diff
@@ -1469,23 +1470,27 @@ handleInventoryListEvent ev = do
...
   -- Escape: stop filtering and go back to regular inventory mode
-  EscapeKey ->
+  EscapeKey -> do
+    resetViewport infoScroll
     Brick.zoom (uiState . uiGameplay . uiInventory) $ do
       uiInventoryShouldUpdate .= True
       uiInventorySearch .= Nothing
   -- Enter: return to regular inventory mode, and pop out the selected item
   Key V.KEnter -> do
+    resetViewport infoScroll
```
byorgey, line 1480: I don't think we need to reset it here either.

src/swarm-tui/Swarm/TUI/Controller.hs
nitinprakash96: You were right. The reason I failed to catch this bug in my testing was because it didn't occur when hitting defined key press events. Just the random ones which I never tested earlier 🤦

byorgey (approved):
Looks good, thanks!

PR #1972, opened 2024-06-22
Apply bindings before returning unification mismatch error

### LINKED ISSUE #1888: Poor type error message for `Int` used in function context
```
> (\f. f 3) 2
1:11: Type mismatch:
  From context, expected `2` to have a type like `_`,
  but it actually has type `Int`
```
It really ought to say "expected 2 to have a function type".  "A type like `_`" makes no sense.

byorgey:
Closes #1888.  When we discovered a type mismatch, we were not fully substituting everything we had learned about the types (via `applyBindings`) before reporting the error.  For example, in #1888 we had `2 : u1` and a unification constraint `u1 = Int -> u2`, but we were just reporting the type of `2` as `u1` rather than first substituting `Int -> u2` for `u1`.

I think we have had this problem for a while, but the recent rewrite of the unification engine possibly made the effects more noticeable.

This incidentally improved a couple more error messages in the test suite. =)

### DIFF
--- a/src/swarm-lang/Swarm/Language/Typecheck.hs
+++ b/src/swarm-lang/Swarm/Language/Typecheck.hs
@@ -127,6 +127,19 @@ type Sourced a = (Source, a)
 
 -- | A "join" where an expected thing meets an actual thing.
 newtype Join a = Join (Source -> a)
+  deriving (Functor)
+
+instance Foldable Join where
+  foldMap :: Monoid m => (a -> m) -> Join a -> m
+  foldMap f j = f a1 <> f a2
+   where
+    (a1, a2) = getJoin j
+
+instance Traversable Join where
+  traverse :: Applicative f => (a -> f b) -> Join a -> f (Join b)
+  traverse f j = joined <$> f a1 <*> f a2
+   where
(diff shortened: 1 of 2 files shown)


nitinprakash96 (approved):

src/swarm-lang/Swarm/Language/Typecheck.hs
```diff
@@ -127,6 +127,19 @@ type Sourced a = (Source, a)
 
 -- | A "join" where an expected thing meets an actual thing.
 newtype Join a = Join (Source -> a)
+  deriving (Functor)
+
+instance Foldable Join where
+  foldMap :: Monoid m => (a -> m) -> Join a -> m
```
xsebek, line 133: Is this used anywhere?

xsebek (approved):
Traversals are impressive feats of magic. 😲

src/swarm-lang/Swarm/Language/Typecheck.hs
byorgey: Not directly, but `Foldable` is a prerequisite of `Traversable`.

PR #1977, opened 2024-06-23
Add Custom Widget Names.

### LINKED ISSUE #1976: Add a `CustomName Text` Constructor to `Name`
In my quest to create [Termite](https://github.com/noahyor/termite), I need to be able to create a Brick Editor. However, this requires a name constructor to be provided, and I cannot use my own type, or else Swarm Embedding will not function correctly. I also cannot use one of Swarm's because it is undefined which you will get if you try to perform a widget lookup by name, which will also break Swarm.

I propose that we add a `CustomName Text` constructor to the `Name` type (in Swarm.TUI.Model.Name) to allow external/custom widgets to be made.

noahyor:
Add Custom Widget Names.

Closes #1976.

### DIFF
--- a/src/swarm-tui/Swarm/TUI/Model/Name.hs
+++ b/src/swarm-tui/Swarm/TUI/Model/Name.hs
@@ -8,6 +8,8 @@
 -- case matching.
 module Swarm.TUI.Model.Name where
 
+import Data.Text (Text)
+
 data WorldEditorFocusable
   = BrushSelector
   | EntitySelector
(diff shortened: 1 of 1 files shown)


src/swarm-tui/Swarm/TUI/Model/Name.hs
```diff
@@ -110,4 +112,6 @@ data Name
     REPLViewport
   | -- | A clickable button in a modal dialog.
     Button Button
+  | -- | A custom widget name.
```
byorgey, line 115: Might be nice to elaborate a bit here to explain what this is for, especially since we will not be using this directly.

byorgey (approved):

src/swarm-tui/Swarm/TUI/Model/Name.hs
noahyor: I hope this is satisfactory.

PR #1979, opened 2024-06-23
Customizable keybindings

### LINKED ISSUE #523: Customizable keybindings
It always takes me a second to think about the right key shortcut to get to the panel I want, and it's annoying.

Personally, I think I would find it more natural if the key shortcuts were arranged *physically* on the keyboard in a way that mirrors the physical layout of the panels.  For example, I would like
```
M-w : inventory       M-e : world
M-s : info panel       M-d : REPL
```
However, I doubt everyone will agree!  Some people might use a different keyboard layout.  Some people might simply prefer different keys.

The most general solution would be a way to configure keyboard shortcuts, i.e. some kind of "configuration" menu that would let you set whatever shortcuts you like (and, in time, other things as well).  Configuration should be stored in `$XDG_CONFIG_HOME` so that it would be persistent across runs of the app.

### LINKED ISSUE #1962: Map editor shortcut conflicts with move-to-end-of-line shortcut
It has happened to me several times that while in creative mode, I have hit Ctrl-E to move to the end of the REPL input and was susprised when it opened up the map editor instead.  I think we should pick a different keyboard shortcut for the map editor.

xsebek:
* create a new `SwarmEvent` enumeration
  * define configuration names (`swarmEvents`)
  * define default keybindings (`defaultSwarmBindings`)
* add `KeyConfig` and `KeyDispatcher`s to `AppState`
  * load custom bindings from an INI file
* migrate key event controller code to use `onEvent`
  * call `handleKey` with handlers
* show the shortcuts in TUI
* check for conflicts between parent/child handler (right now `mainHandler` and the rest)
* generate the INI file (`keybinding --init`) with commented-out (`;`) settings
* allow `brick` version 2.4 to get uppercase keybindings fix

---

* Closes #523
* Closes #1962


src/swarm-tui/Swarm/TUI/Model/StateUpdate.hs
```diff
@@ -91,6 +94,35 @@ import Swarm.TUI.View.Structure qualified as SR
...
+        T.putStrLn $ "Handlers with the '" <> BK.ppBinding b <> "' binding:"
+        forM_ hs $ \h -> do
+          let trigger = case BK.kehEventTrigger $ BK.khHandler h of
+                ByKey k -> "triggered by the key '" <> BK.ppBinding k <> "'"
+                ByEvent e -> "triggered by the event '" <> fromMaybe "<unknown>" (BK.keyEventName swarmEvents e) <> "'"
+              desc = BK.handlerDescription $ BK.kehHandler $ BK.khHandler h
+          T.putStrLn $ "  " <> desc <> " (" <> trigger <> ")"
+      exitFailure
+
+initKeyHandlingState :: IO KeyEventHandlingState
```
byorgey, line 121: Could be nice to give these types like `Has (Lift IO) sig m => m KeyEventHandlingState`.  Then you can call them without `sendIO`, and they become a bit more flexible (e.g. maybe later we realize they need some other effect).

src/swarm-tui/Swarm/TUI/Model.hs
```diff
@@ -139,6 +146,24 @@ logEvent src sev who msg el =
  where
   l = LogEntry (TickNumber 0) src sev who msg
 
+data KeyEventHandlingState = KeyEventHandlingState
+  { _keyConfig :: KeyConfig SwarmEvent
+  , _keyHandlers :: EventHandlers
+  }
+
+keyConfig :: Lens' KeyEventHandlingState (KeyConfig SwarmEvent)
```
byorgey, line 154: Should be possible to just `makeLenses ''KeyEventHandlingState`, I don't think these need to be written by hand.

xsebek: It is not because of the TH splice `KeyEventHandlingState` would not see `AppState`.

I could move the lenses to the end of file or some other trick, I guess...

byorgey: Ah, yeah, sometimes I have to reorder things to make the staging work out.  Defining these by hand is also  fine, they are pretty simple.  Maybe just put a comment explaining why they have to be defined by hand.

xsebek:
It works! 🥳  Add an INI file:

```bash
$ cat /Users/xsebek/.config/swarm/config.ini
[keybindings]
debug CESK = Ctrl-u
```

Print the changed keybindings:
```bash
cabal run -O0 swarm -- keybindings --markdown
```
---

### Keybindings

#### main
| Keybinding | Event Name | Description |
| ---------- | ---------- | ----------- |
| `C-q` | `quit` | Open quit game dialog |
| `F1` | `view help` | View Help screen |
| `F2` | `view robots` | View Robots screen |
| `F3` | `view recipes` | View Recipes screen |
| `F4` | `view commands` | View Commands screen |
| `F6` | `view structures` | View Structures screen |
| `C-g` | `view goal` | View scenario goal description |
| `M-h` | `hide robots` | Hide robots for a few ticks |
| `C-u` | `debug CESK` | Show active robot CESK machine debugging line |

#### repl
| Keybinding | Event Name | Description |
| ---------- | ---------- | ----------- |
| `C-c`, `Esc` | `cancel running program` | Cancel running base robot program |
| `M-p` | `toggle piloting mode` | Toggle piloting mode |
| `M-k` | `toggle custom key handling` | Toggle custom key handling mode |


The configuration file is at:
/Users/xsebek/.config/swarm/config.ini

src/swarm-tui/Swarm/TUI/Model/StateUpdate.hs
xsebek: Done, I will use it to plug into existing SystemFailure handling. 👍 

I am not sure whether I should use Throw or Accum, maybe it's better to only have the failures as warnings...

byorgey: I think `Throw` is fine, if the user's keybinding config cannot be parsed I think they would want to fix that before running the game.  Especially since right now warnings are a bit hidden (they go into `Messages` but I never think to look there).

xsebek:
After a lot of shuffling code back and forth, the main game handlers are customizable:

---

# Keybindings

# main
| Keybinding | Event Name | Description |
| ---------- | ---------- | ----------- |
| `Esc` | (non-customizable key) | Close open modal |
| `C-q` | `quit` | Open quit game dialog |
| `F1` | `view help` | View Help screen |
| `F2` | `view robots` | View Robots screen |
| `F3` | `view recipes` | View Recipes screen |
| `F4` | `view commands` | View Commands screen |
| `F5` | `view messages` | View Messages screen |
| `F6` | `view structures` | View Structures screen |
| `C-g` | `view goal` | View scenario goal description |
| `M-h` | `hide robots` | Hide robots for a few ticks |
| `M-d` | `debug CESK` | Show active robot CESK machine debugging line |
| `C-p` | `pause` | Pause or unpause the game |
| `C-o` | `run single tick` | Run game for a single tick |
| `C-x` | `increse TPS` | Increase game speed by one tick per second |
| `C-z` | `decrease TPS` | Descrease game speed by one tick per second |
| `M-w` | `focus World` | Set focus on the World panel |
| `M-e` | `focus Robot` | Set focus on the Robot panel |
| `M-r` | `focus REPL` | Set focus on the REPL panel |
| `M-t` | `focus Info` | Set focus on the Info panel |
| `C-v` | `creative mode` | Toggle creative mode |
| `C-e` | `world editor` | Toggle world editor mode |
| `M-,` | `toggle REPL` | Collapse/Expand REPL panel |

# repl
| Keybinding | Event Name | Description |
| ---------- | ---------- | ----------- |
| `C-c`, `Esc` | `cancel running program` | Cancel running base robot program |
| `M-p` | `toggle piloting mode` | Toggle piloting mode |
| `M-k` | `toggle custom key handling` | Toggle custom key handling mode |

# world
| Keybinding | Event Name | Description |
| ---------- | ---------- | ----------- |
| `c` | `view base` | View the base robot |
| `f` | `show fps` | Show frames per second |
| `k`, `Up` | `move view north` | Scroll world view in the north direction |
| `l`, `Right` | `move view east` | Scroll world view in the east direction |
| `j`, `Down` | `move view south` | Scroll world view in the south direction |
| `h`, `Left` | `move view west` | Scroll world view in the west direction |

# robot
| Keybinding | Event Name | Description |
| ---------- | ---------- | ----------- |
| `Enter` | (non-customizable key) | Show entity description |
| `m` | `make entity` | Make the selected entity |
| `0` | `show zero inventory entites` | Show entities with zero count in inventory |
| `;` | `cycle inventory sort` | Cycle inventory sorting type |
| `:` | `switch inventory direction` | Switch ascending/descending inventory sort |
| `/` | `search inventory` | Start inventory search |

xsebek:
@byorgey @kostmo, While this is not finished yet, I would appreciate it if you could test this out.

The original code could have behaved differently because the conditions allowed events to be handled by subhandlers.

If we really want that behavior, then those handlers have to stay in the old not-as-declarative code.
For example, `Esc` is used everywhere, and I made the mistake of putting it into the main handler.
But there could be more obscure bugs or just plain typos that I made.

---
By the way, this refactoring makes the Controller significantly smaller:
```
 src/swarm-tui/Swarm/TUI/Controller.hs | 827 ++---------------------------------------------
 27 insertions(+), 800 deletions(-)
```
It is necessary because the handlers are stored in the Model, which is imported into the Controller, and a cycle would form.

byorgey:
Just some small initial feeback: so far I'm just trying it out without having looked at the code yet (since that's what players will do).  I found `swarm keybindings` quickly, which told me where the config file should go, but I don't know what format is expected.  However, I see you have some todos for documentation and generating an initial `.ini` file, so maybe you have that under control!

xsebek:
> ```shell
> $ cat /Users/xsebek/.config/swarm/config.ini
> [keybindings]
> debug CESK = Ctrl-u
> ```

@byorgey, exactly. The INI file looks like that, with **Event Name**s as keys and values being parsed by [`parseBindingList`](https://hackage.haskell.org/package/brick-2.3.1/docs/Brick-Keybindings-Parse.html#v:parseBindingList).

I think showing by example will be much clearer than trying to explain the format. 🙂

src/swarm-scenario/Swarm/Game/ResourceLoading.hs
```diff
@@ -83,17 +98,11 @@ getDataFileNameSafe asset name = do
     then return fp
     else throwError $ AssetNotLoaded (Data asset) fp $ DoesNotExist File
 
--- | Get a nice message suggesting to download @data@ directory to 'XdgData'.
-dataNotFound :: FilePath -> IO LoadingFailure
```
byorgey, line 87: Were we not using this file before?  I can't seem to find anywhere you moved it to.

byorgey:
I accidentally wrote a keybinding in uppercase, like `Ctrl-T`, and it turns out this means it parses successfully, with no errors or warnings, but simply does not work --- both `Ctrl-t` and `Ctrl-Shift-t` do nothing.  I suspect it turns into a keybinding which is impossible to ever see, because `Ctrl-Shift-t` is not the same as `Ctrl-T`.  Well, I don't know, perhaps it is possible to get `Ctrl-T` by turning on caps lock and then typing `Ctrl-t`?  In any case it seems like some kind of warning here would be nice.

src/swarm-tui/Swarm/TUI/Controller/UpdateUI.hs
```diff
@@ -0,0 +1,250 @@
+{-# LANGUAGE OverloadedStrings #-}
+
+-- |
+-- SPDX-License-Identifier: BSD-3-Clause
+--
+-- The main TUI update logic that is called from other controller parts.
+module Swarm.TUI.Controller.UpdateUI (
```
byorgey, line 7: I love that you have also split out the `Controller` module into submodules --- it was definitely one of the most in need of splitting up.

byorgey (commented):
I haven't looked at absolutely all the code yet, but so far it seems to work great.  I was able to remap the key shortcuts for the four panels which is something I have wanted for a very long time. :grin:

byorgey:
> I think showing by example will be much clearer than trying to explain the format in detail.

Agreed, I think generating an initial `.ini` file with all the default keybindings commented out will be perfect.

src/swarm-scenario/Swarm/Game/ResourceLoading.hs
xsebek: When the new error system was implemented, this function stopped being used.

We should add the suggestion to the new warnings.

It might also be a good idea to use a tool to find dead code so we can identify it sooner.

xsebek: Split out as #2002.

src/swarm-tui/Swarm/TUI/Model/KeyBindings.hs
```diff
@@ -0,0 +1,105 @@
...
+keybindingINI :: Ord k => KeyConfig k -> [(Text, [KeyEventHandler k m])] -> Text
+keybindingINI kc sections =
+  T.unlines $
+    "[keybindings]\n"
+      : "; Uncomment the assignment and set comma separated list"
+      : "; of keybindings or \"unbind\" on the right. See:"
+      : "; https://hackage.haskell.org/package/brick/docs/Brick-Keybindings-Parse.html#v:parseBinding\n"
+      : map (keyBindingEventINI kc) handlersData
+ where
+  handlersData = concatMap (mapMaybe handlerData . snd) sections
```
byorgey, line 80: Could we print some section headers in comments in the generated `.ini` file as well?  Looking at the generated file it's impossible to know which keybindings are active in which substates.  For example there is a line that says `show fps = f` but there's no indication that only works when the world panel is focused.  I'm imagining a header something like `;;; World keybindings (only active when the world panel is focused) ;;;`.

src/swarm-tui/Swarm/TUI/Model/KeyBindings.hs
```diff
@@ -0,0 +1,105 @@
...
+  pure $ case bindings of
+    Left e -> prettyText e
+    Right bs -> showTable kPrint (bs ^. keyConfig) sections
+ where
+  showTable = \case
+    MarkdownPrint -> keybindingMarkdownTable
+    TextPrint -> keybindingTextTable
+    IniPrint -> keybindingINI
+  sections =
+    [ ("main", mainEventHandlers)
```
byorgey, line 65: Might be nice to have these headers be a bit more informative, e.g. `main keybindings (always active)`, `Robot panel keybindings`, etc.

src/swarm-tui/Swarm/TUI/Model/Event.hs
```diff
@@ -0,0 +1,223 @@
...
+  ShowZeroInventoryEntitiesEvent -> [bind '0']
+  CycleInventorySortEvent -> [bind ';']
+  SwitchInventorySortDirection -> [bind ':']
+  SearchInventoryEvent -> [bind '/']
+
+-- ----------------
+-- Helper methods
+
+allKeyEvents :: (Ord e, Bounded e, Enum e) => (e -> Text) -> KeyEvents e
+allKeyEvents f = keyEvents $ map (\e -> (f e, e)) [minBound .. maxBound]
```
byorgey, line 220: For `[minBound .. maxBound]` we have `listEnums` from `Swarm.Util`.

Also `\e -> (f e, e)` is equivalent to `f &&& id` if you import `Control.Arrow ((&&&))`.  I personally find `f &&& id` clearer than `\e -> (f e, e)` but YMMV.

src/swarm-tui/Swarm/TUI/Model/Event.hs
```diff
@@ -0,0 +1,223 @@
...
+  | ShowZeroInventoryEntitiesEvent
+  | CycleInventorySortEvent
+  | SwitchInventorySortDirection
+  | SearchInventoryEvent
+  deriving (Eq, Ord, Show, Enum, Bounded)
+
+robotPanelEvents :: KeyEvents RobotEvent
+robotPanelEvents = allKeyEvents $ \case
+  MakeEntityEvent -> "make entity"
+  ShowZeroInventoryEntitiesEvent -> "show zero inventory entites"
```
byorgey, line 203:
```suggestion
  ShowZeroInventoryEntitiesEvent -> "show zero inventory entities"
```

src/swarm-tui/Swarm/TUI/Controller/Util.hs
```diff
@@ -117,3 +157,41 @@ resetViewport :: ViewportScroll Name -> EventM Name AppState ()
...
+  [KeyEventHandler e2 (EventM Name AppState)]
+allHandlers eEmbed f = map handleEvent1 [minBound .. maxBound]
+ where
+  handleEvent1 e1 = let (n, a) = f e1 in onEvent (eEmbed e1) n a
+
+runBaseTerm :: (MonadState AppState m) => Maybe TSyntax -> m ()
+runBaseTerm = maybe (pure ()) startBaseProgram
+ where
+  -- The player typed something at the REPL and hit Enter; this
+  -- function takes the resulting ProcessedTerm (if the REPL
```
byorgey, line 188:
```suggestion
  -- function takes the resulting term (if the REPL
```
Oops, looks like I forgot to remove the reference to `ProcessedTerm` here in #1928 .  Would you mind including this change?

src/swarm-tui/Swarm/TUI/Controller/Util.hs
```diff
@@ -117,3 +157,41 @@ resetViewport :: ViewportScroll Name -> EventM Name AppState ()
...
+  c <- use $ gameState . creativeMode
+  when c a
+
+-- | Create a list of handlers with embedding events and using pattern matching.
+allHandlers ::
+  (Ord e2, Enum e1, Bounded e1) =>
+  (e1 -> e2) ->
+  (e1 -> (Text, EventM Name AppState ())) ->
+  [KeyEventHandler e2 (EventM Name AppState)]
+allHandlers eEmbed f = map handleEvent1 [minBound .. maxBound]
```
byorgey, line 180: For `[minBound .. maxBound]` use `Swarm.Util.listEnums`.

src/swarm-tui/Swarm/TUI/Controller/EventHandlers/Main.hs
```diff
@@ -0,0 +1,155 @@
...
+import Swarm.TUI.Model.Event (MainEvent (..), SwarmEvent (..))
+import Swarm.TUI.Model.Goal
+import Swarm.TUI.Model.UI
+import System.Clock (Clock (..), TimeSpec (..), getTime)
+
+-- | Main keybindings event handler while running the game itself.
+--
+-- See 'Swarm.TUI.Controller.handleMainEvent'.
+mainEventHandlers :: [KeyEventHandler SwarmEvent (EventM Name AppState)]
+mainEventHandlers = allHandlers Main $ \case
```
byorgey, line 35: So just to make sure I understand the process, let's say we wanted to add a new action which can be bound to a key by the user, but doesn't have a keybinding by default.  It looks like all we would have to do is add a constructor to one of the enumeration types in `Swarm.TUI.Model.Event`, and then update a few functions (such as here) wherever the compiler yells at you about missing patterns.  Is that it, or is there anything else?

app/Main.hs
```diff
@@ -41,6 +46,7 @@ cliParser =
         [ command "format" (info (Format <$> parseFormat) (progDesc "Format a file"))
         , command "lsp" (info (pure LSP) (progDesc "Start the LSP"))
         , command "version" (info (pure Version) (progDesc "Get current and upstream version."))
+        , command "keybindings" (info (ListKeybinding <$> printKeyMode) (progDesc "List the keybindings"))
```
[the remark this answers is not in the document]
xsebek: Great, I was confused that it did not show help. 👍 Strange that other commands don't have the `helper`.

Btw. I found this worked:
```
swarm --help keybindings
```

app/Main.hs
```diff
@@ -125,11 +137,22 @@ showVersion = do
   up <- getNewerReleaseVersion gitInfo
   either (hPrint stderr) (putStrLn . ("New upstream release: " <>)) up
 
+printKeybindings :: KeybindingPrint -> IO ()
```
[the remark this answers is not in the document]
xsebek: Not a strong reason, I just wanted to have the caller decide on how output should be done.

It is comparable to `showVersion` above, but if you prefer, I can move it.

app/Main.hs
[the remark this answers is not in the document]
byorgey: > Strange that other commands don't have the `helper`.

Perhaps we should add it!  Though actually `format` does have it, and `lsp` and `version` don't have any options so it doesn't matter so much for them.

app/Main.hs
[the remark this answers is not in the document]
byorgey: OK, no, that makes sense.  Let's keep it here.

byorgey:
By the way, key shortcuts are shown in two places in the UI: in the bar between the REPL and world panels, and also in the F1 help modal.  I think in the F1 modal we should mention something about how to customize keybindings.

app/Main.hs
```diff
@@ -125,11 +137,22 @@ showVersion = do
...
+printKeybindings p = do
+  kb <- showKeybindings p
+  T.putStrLn kb
+  when (p /= IniPrint) $ do
+    (iniExists, ini) <- getSwarmConfigIniFile
+    let iniState = if iniExists then "is" else "can be created"
```
byorgey, line 146: If the `config.ini` doesn't already exist, in addition to saying "can be created at..." maybe it could also explain how to create it, i.e. mention `swarm keybindings --ini > ...`  For that matter, why not add a flag like `swarm keybindings --init` which creates a default `config.ini` in the correct location (if one does not already exist)?

src/swarm-tui/Swarm/TUI/Controller/EventHandlers/Main.hs
xsebek: Yes, if it is part of existing handlers, you would add it to `Swarm.TUI.Model.Event` and leave the default list of bindings empty.

There are two less common situations that require more work:
* If the event has a condition _and_ is also bound in sub-handlers, then you add it to the old-style handler.
* If you need a new dispatcher (for example, to add the main scenarios menu), then it has to be added to more places as described in `Swarm.TUI.Controller.EventHandlers`.

src/swarm-tui/Swarm/TUI/Controller/Util.hs
xsebek: Thanks for the tip, I suspected this was a common pattern already. 👍

src/swarm-tui/Swarm/TUI/Controller/Util.hs
xsebek: Could have also survived while I was rebasing, no problem. 🙂

src/swarm-tui/Swarm/TUI/Model/KeyBindings.hs
xsebek: They will turn into markdown sections, so it should be short, but thanks for the ideas. 👍

src/swarm-tui/Swarm/TUI/Model/KeyBindings.hs
xsebek: Sure, I was just lazy when copy-pasting parts from the markdown printer implementation in Brick. 🙂

src/swarm-tui/Swarm/TUI/Controller/EventHandlers/Main.hs
byorgey: Would be good to make sure this info ends up in a comment somewhere obvious, for anyone who wants to add some handlers in the future.

app/Main.hs
xsebek: Yes, I was thinking about it as well. 👍 

Otherwise, if you wanted to add new keybindings, then you would have to use this trick:
```bash
swarm keybindings --ini | tee ~/.config/swarm/config.ini
#                         ^^^ does not truncate the file
```

src/swarm-tui/Swarm/TUI/Controller/EventHandlers/Main.hs
xsebek: I am not sure there is such an obvious place while `Controller` is still used for some handlers.

Maybe we could use the note reference in all relevant handlers? 🤔

app/Main.hs
xsebek: ✅

src/swarm-tui/Swarm/TUI/Controller/UpdateUI.hs
xsebek: It was a lot of work; I wish HLS had refactorings to move and rename stuff between files, like JetBrains IDEs. 😅

src/swarm-tui/Swarm/TUI/Controller/EventHandlers/Main.hs
byorgey: > Maybe we could use the note reference in all relevant handlers? 🤔

Yes, that's a good idea.

src/swarm-tui/Swarm/TUI/Controller/UpdateUI.hs
byorgey: Agreed!

src/swarm-tui/Swarm/TUI/Model.hs
xsebek: I finally moved them to the end of the file.

byorgey:
I just discovered a regression: on `main`, hitting `Esc` at the REPL has no effect.  With this PR, hitting `Esc` makes the REPL display `...` (as if the base is executing a program) and it's impossible to get back the normal prompt (even with Ctrl-C).

src/swarm-tui/Swarm/TUI/Model/Event.hs
```diff
@@ -0,0 +1,225 @@
...
+  PauseEvent -> [ctrl 'p']
+  RunSingleTickEvent -> [ctrl 'o']
+  IncreaseTpsEvent -> [ctrl 'x']
+  DecreaseTpsEvent -> [ctrl 'z']
+  FocusWorldEvent -> [meta 'w']
+  FocusRobotEvent -> [meta 'e']
+  FocusREPLEvent -> [meta 'r']
+  FocusInfoEvent -> [meta 't']
+  ToggleCreativeModeEvent -> [ctrl 'v']
+  ToggleWorldEditorEvent -> [ctrl 'e']
```
byorgey, line 131:
```suggestion
  ToggleWorldEditorEvent -> []
```

I suggest addressing #1962 by having this *not* bound to a key by default.  I think the world editor is still relatively experimental anyway.

src/swarm-tui/Swarm/TUI/Controller/Util.hs
byorgey: BREAKING NEWS: use `enumerate` instead. :facepalm:  #2003

xsebek:
@byorgey I tried setting Esc as alternative to Ctrl-C to cancel running computation.

Does Ctrl-C break the game on main?

byorgey:
> Does Ctrl-C break the game on main?

Aha, it does!  So this is not your fault.  I've created #2007.

xsebek:
Finally, I managed to show the custom keybindings in the UI:
Screenshot 2024-07-05 at 22 04 52

src/swarm-tui/Swarm/TUI/Controller/Util.hs
xsebek: Rebased and done again! ✅

xsebek:
I think the custom keybindings are now ready. 😁 

@byorgey @kostmo and anyone else interested, please give this a try. 

I would also welcome your code review. There is not that much new code; it was mostly just moved from Controller.
Notable new modules are:
* `Swarm.TUI.Model.Event`
* `Swarm.TUI.Controller.EventHandlers`
* `Swarm.TUI.Model.KeyBindings`

---

@jtdaugherty has promptly fixed uppercase keybindings parsing. You can test that, too, with Brick `master`. 🙂

jtdaugherty:
Nice!!

jtdaugherty:
The keybinding case normalization fix is now released in `brick-2.4`.

app/Main.hs
```diff
@@ -30,6 +34,7 @@ commitInfo = case gitInfo of
 
 data CLI
   = Run AppOpts
+  | ListKeybinding Bool KeybindingPrint
```
kostmo, line 37: Add haddock for this `Bool` parameter?  Or alternatively define a self-documenting Bool-isomorphic type

src/swarm-tui/Swarm/TUI/Controller.hs
```diff
@@ -1017,37 +461,14 @@ resetREPL t r replState =
...
-                invalidateCacheEntry REPLHistoryCache
-    MetaChar 'k' -> do
-      when (isJust (s ^. gameState . gameControls . inputHandler)) $ do
-        curMode <- use $ uiState . uiGameplay . uiREPL . replControlMode
-        (uiState . uiGameplay . uiREPL . replControlMode) .= case curMode of Handling -> Typing; _ -> Handling
-
+    -- Pass to key handler (allows users to configure bindings)
+    -- See Note [how Swarm event handlers work]
+    VtyEvent (V.EvKey k m)
+      | isJust (B.lookupVtyEvent k m keyHandler) -> do
```
kostmo, line 470: Can get rid of this `do`

src/swarm-tui/Swarm/TUI/Controller/EventHandlers.hs
```diff
@@ -0,0 +1,131 @@
...
+     in T.intercalate "\n" (hsm : hss)
+
+-- | Take two dispatchers (that do not have conflict themselves) and find conflicting keys between them.
+conflicts :: SwarmKeyDispatcher -> SwarmKeyDispatcher -> [(Binding, [KeyHandler SwarmEvent (EventM Name AppState)])]
+conflicts d1 d2 = combine <$> badGroups
+ where
+  l1 = keyDispatcherToList d1
+  l2 = keyDispatcherToList d2
+  gs = groupBy ((==) `on` fst) $ sortBy (compare `on` fst) (l1 <> l2)
+  badGroups = filter ((1 <) . length) $ mapMaybe NE.nonEmpty gs
```
kostmo, line 118: Use the `NonEmpty` version of [`groupWith`](https://hackage.haskell.org/package/base-4.20.0.1/docs/Data-List-NonEmpty.html#v:groupWith)?
```suggestion
  gs = NE.groupWith fst $ sortBy (compare `on` fst) (l1 <> l2)
  badGroups = filter ((1 <) . length) gs
```

src/swarm-tui/Swarm/TUI/Controller/EventHandlers.hs
```diff
@@ -0,0 +1,131 @@
...
+      throwLoadingFailure $
+        "Error: some key events have the same keys bound to them.\n"
+          : handlerErrors collisions
+    Right d -> return d
+  buildSubDispatcher parentName parentDispatcher name handlers = do
+    d <- buildDispatcher handlers
+    let collisions = conflicts parentDispatcher d
+    if null collisions
+      then return d
+      else
```
kostmo, line 95: May as well have the `conflicts` function return `Maybe (NonEmpty ...)` so that the `NonEmpty`-ness remains attested downstream
```suggestion
    let maybeCollisions = conflicts parentDispatcher d
    case maybeCollisions of
      Nothing -> return d
      Just collisions -> ...
```

src/swarm-tui/Swarm/TUI/Controller/EventHandlers.hs
```diff
@@ -0,0 +1,131 @@
...
+
+  throwLoadingFailure = throwError . CustomFailure . T.intercalate "\n"
+  handlerErrors collisions = flip map collisions $ \(b, hs) ->
+    let hsm = "Handlers with the '" <> BK.ppBinding b <> "' binding:"
+        hss = flip map hs $ \h ->
+          let trigger = case BK.kehEventTrigger $ BK.khHandler h of
+                ByKey k -> "triggered by the key '" <> BK.ppBinding k <> "'"
+                ByEvent e -> "triggered by the event '" <> fromMaybe "<unknown>" (BK.keyEventName swarmEvents e) <> "'"
+              desc = BK.handlerDescription $ BK.kehHandler $ BK.khHandler h
+           in "  " <> desc <> " (" <> trigger <> ")"
```
kostmo, line 108: Good opportunity to use the [`squote`](https://github.com/swarm-game/swarm/blob/9b1b6985fa51ca396358cb77c7d496f807eafae8/src/swarm-util/Swarm/Util.hs#L430) and [`parens`](https://github.com/swarm-game/swarm/blob/9b1b6985fa51ca396358cb77c7d496f807eafae8/src/swarm-util/Swarm/Util.hs#L442) functions

src/swarm-tui/Swarm/TUI/Controller/EventHandlers.hs
```diff
@@ -0,0 +1,131 @@
...
+     in T.intercalate "\n" (hsm : hss)
+
+-- | Take two dispatchers (that do not have conflict themselves) and find conflicting keys between them.
+conflicts :: SwarmKeyDispatcher -> SwarmKeyDispatcher -> [(Binding, [KeyHandler SwarmEvent (EventM Name AppState)])]
+conflicts d1 d2 = combine <$> badGroups
+ where
+  l1 = keyDispatcherToList d1
+  l2 = keyDispatcherToList d2
+  gs = groupBy ((==) `on` fst) $ sortBy (compare `on` fst) (l1 <> l2)
```
kostmo, line 117: Use `sortOn`
```suggestion
  gs = groupBy ((==) `on` fst) $ sortOn fst (l1 <> l2)
```

src/swarm-tui/Swarm/TUI/Model/Event.hs
```diff
@@ -0,0 +1,227 @@
...
+--                 REPL EVENTS
+-- ----------------------------------------------
+
+data WorldEvent
+  = ViewBaseEvent
+  | ShowFpsEvent
+  | MoveViewNorthEvent
+  | MoveViewEastEvent
+  | MoveViewSouthEvent
+  | MoveViewWestEvent
```
kostmo, line 171: This can simplify some code that is cardinal-direction-specific:
```suggestion
data WorldEvent
  = ViewBaseEvent
  | ShowFpsEvent
  | MoveViewDirectionEvent AbsoluteDir
```

src/swarm-tui/Swarm/TUI/View.hs
```diff
@@ -1060,27 +1074,56 @@ drawKeyMenu s =
...
   keyCmdsFor (Just (FocusablePanel WorldPanel)) =
-    [ ("←↓↑→ / hjkl", "scroll") | canScroll
+    [ ( T.intercalate "/" $
+          map
+            keyW
+            [ SE.MoveViewWestEvent
+            , SE.MoveViewNorthEvent
+            , SE.MoveViewEastEvent
+            , SE.MoveViewSouthEvent
+            ]
```
kostmo, line 1089: Code like this can be simplified with `enumerate` after refactoring to use the `MoveViewDirectionEvent AbsoluteDir`

src/swarm-tui/Swarm/TUI/Model/KeyBindings.hs
```diff
@@ -0,0 +1,137 @@
...
+ where
+  showTable = \case
+    MarkdownPrint -> keybindingMarkdownTable
+    TextPrint -> keybindingTextTable
+    IniPrint -> keybindingINI
+
+keySections :: [(Text, [KeyEventHandler SwarmEvent (EventM Name AppState)])]
+keySections =
+  [ ("Main game (always active)", mainEventHandlers)
+  , ("REPL panel ", replEventHandlers)
```
kostmo, line 74: Extra trailing space?

src/swarm-tui/Swarm/TUI/Controller/EventHandlers.hs
xsebek: Cool, this is even nicer than `groupOn` from `Extra`.

src/swarm-tui/Swarm/TUI/Controller/EventHandlers.hs
xsebek: Good idea, shame that it needs an import so I can't just apply the suggestion.

src/swarm-tui/Swarm/TUI/Controller/EventHandlers.hs
xsebek: I support proving with types, but here you run into `brick` not using `NonEmpty` and I want to share code for it.

Unless you suspect `brick` will send us an empty list, I do not see the benefit. The code would have to use `NE.toList` after `NE.nonEmpty`, which seems like a more verbose way to say `null`. 🤔

src/swarm-tui/Swarm/TUI/Controller/EventHandlers.hs
xsebek: I wish we had f-strings `f"event '{event}'"`, they would make this much easier to read.

At least I can drop `fromMaybe`.

src/swarm-tui/Swarm/TUI/Model/KeyBindings.hs
xsebek: Nice catch. 👍

src/swarm-tui/Swarm/TUI/Model/Event.hs
xsebek: This makes me handwrite `Enum` and `Bounded` and add a few imports, but I admit I wanted to do this too.

The usage becomes much more elegant, I just wish Haskell supported this out of the box.

app/Main.hs
xsebek: I can not come up with a good name for the type "initialize the INI file or not," so doc it is. 😄

src/swarm-tui/Swarm/TUI/Controller/EventHandlers/Main.hs
```diff
@@ -0,0 +1,155 @@
...
+  ViewRecipesEvent -> ("View Recipes screen", toggleDiscoveryNotificationModal RecipesModal availableRecipes)
+  ViewCommandsEvent -> ("View Commands screen", toggleDiscoveryNotificationModal CommandsModal availableCommands)
+  ViewMessagesEvent -> ("View Messages screen", toggleMessagesModal)
+  ViewStructuresEvent -> ("View Structures screen", toggleDiscoveryModal StructuresModal (structureRecognition . automatons . originalStructureDefinitions))
+  ViewGoalEvent -> ("View scenario goal description", viewGoal)
+  HideRobotsEvent -> ("Hide robots for a few ticks", hideRobots)
+  ShowCESKDebugEvent -> ("Show active robot CESK machine debugging line", showCESKDebug)
+  PauseEvent -> ("Pause or unpause the game", whenRunning safeTogglePause)
+  RunSingleTickEvent -> ("Run game for a single tick", whenRunning runSingleTick)
+  IncreaseTpsEvent -> ("Increase game speed by one tick per second", whenRunning . modify $ adjustTPS (+))
```
byorgey, line 48: These should actually say "double game speed" and "halve game speed", not "increase/decrease by one tick per second".

src/swarm-tui/Swarm/TUI/Controller.hs
xsebek: Done. 👍

byorgey (approved):
This is looking great!  Thanks for the hard work on this really nice feature.

byorgey:
@xsebek I assume you were already going to do this, but just to be sure, before merging we should make sure to allow `brick-2.4` in the `.cabal` file.  I don't think we need to *require* `brick-2.4` (if anyone reports running into issues with lower vs upper keybindings we can just direct them to ensure they are building with `brick-2.4`; I doubt it will come up).

src/swarm-tui/Swarm/TUI/Controller/EventHandlers/Main.hs
xsebek: Ups, you are right. I saw the `+` and forgot it adds to log2 of speed.

src/swarm-tui/Swarm/TUI/Controller/EventHandlers.hs
byorgey: https://hackage.haskell.org/package/fmt ?

src/swarm-tui/Swarm/TUI/Model/Event.hs
kostmo: I [asked about this on Discourse](https://discourse.haskell.org/t/enumerating-all-values-of-a-nested-sum-type/9907), and got some good suggestions:
* [`universeGeneric`](https://hackage.haskell.org/package/universe-1.2.3/docs/Data-Universe.html#v:universeGeneric) from the `universe` package
* the [`finitary`](https://hackage.haskell.org/package/finitary-2.1.2.0/docs/Data-Finitary.html) package

src/swarm-tui/Swarm/TUI/Model/Event.hs
```diff
@@ -0,0 +1,236 @@
...
+  SearchInventoryEvent -> [bind '/']
+
+-- ----------------
+-- Helper methods
+
+allKeyEvents :: (Ord e, Bounded e, Enum e) => (e -> Text) -> KeyEvents e
+allKeyEvents f = keyEvents $ map (f &&& id) enumerate
+
+allBindings :: (Bounded e, Enum e) => (e -> [Binding]) -> [(e, [Binding])]
+allBindings f = map (\e -> (e, f e)) [minBound .. maxBound]
```
kostmo, line 236: Potentially:
```suggestion
allBindings :: (Bounded e, Enum e) => (e -> [Binding]) -> NonEmpty (e, [Binding])
allBindings f = NE.map (\e -> (e, f e)) enumerateNonEmpty
```
I've gone through the exercise of plumbing it through in #2013.

xsebek: @kostmo could you please explain how this will help us? Maybe I am missing something, but it seems like you prove the list has at least one element and then immediately discard that with `NE.toList`.

Here, `brick` is happy to take a list with any number of elements, and zero elements seem perfectly well-defined, so I do not see the point.

I did miss `enumerate` though, that definitely should be here.

kostmo: Agreed, the current Brick consumer does not require nonemptiness. I tend to "hold on to the proofs" as long as possible for the benefit of refactoring.  But this is sometimes in tension with the YAGNI principle.  So I leave it to you to decide.

kostmo (approved):

src/swarm-tui/Swarm/TUI/Model/Event.hs
xsebek: Thanks, @kostmo; there are some good tips in that discussion. 👍 

I think I was considering `finitary` before, but there was not as much need. Did not know about `universeGeneric`.

I will split that as a separate Refactoring Issue.

src/swarm-tui/Swarm/TUI/Model/Event.hs
xsebek: In this case, let's keep it simple. But thanks for showing the non-empty version; I will keep it in mind. 👍

xsebek:
Hooray! 🚀

Thanks, @byorgey and @kostmo, for your reviews and support. 😁

Issue #1981, opened 2024-06-24
smart constructor for Grid


kostmo:
Should have a "smart constructor" that enforces the invariant that all rows are equal length.

Later, if we create a separate `EmptyGrid` constructor, this can also ensure that the alternative has nonempty rows.

PR #1982, opened 2024-06-24
remove 'unGrid' accessor

kostmo:
Towards #1981


byorgey (approved):
Looks good to me.

I guess I hadn't taken a good look at some of the new functions in `src/swarm-topography/Swarm/Game/Scenario/Topography/Area.hs`. It occurs to me that we could replace several of them with relatively simple uses of things from `lens` (e.g. `allMembers`, `mapRows`, `mapIndexedMembers`).  I don't know if that's something you'd be interested in.  If so I could make a PR.

PR #1988, opened 2024-06-25
Fix world DSL coordinate bug

byorgey:
The special variables `x` and `y` in the world DSL did not refer to the correct thing --- in the interpreter apparently I forgot to convert from `Coords` to `Location` and just used the row, column in `Coords` as the `x` and `y`.  I have also fixed up all the scenarios which used `x` and `y` in their world description so that they now look identical to before, by replacing every use of `x` with `-y` and every use of `y` with `x` (though in some cases where the result would be algebraically equal I did not literally do this replacement, *e.g.* `(x + y) % 2` is equal to `((-y) + x) % 2`, so I left it alone).

### DIFF
--- a/data/scenarios/Challenges/Ranching/beekeeping.yaml
+++ b/data/scenarios/Challenges/Ranching/beekeeping.yaml
@@ -430,7 +430,7 @@ world:
     in
       overlay
       [ {grass}
-      , mask (flowers && (x + y) % 3 == 0) {wildflower}
+      , mask (flowers && (x - y) % 3 == 0) {wildflower}
       , mask (rubble && (x + y) % 2 == 0) {rock}
       , mask rock {rock}
       , mask trees {tree}
--- a/data/scenarios/Challenges/gallery.yaml
+++ b/data/scenarios/Challenges/gallery.yaml
@@ -241,8 +241,8 @@ world:
   dsl: |
     overlay
     [ {terrain: stone}
-    , if (x/5 + y/5) % 2 == 0 then {terrain: dirt} else {blank}
-    , if (((x + 3) % 20)/11 + ((y + 3) % 20)/11) == 0 then {terrain: grass} else {blank}
+    , if (x/5 + (-y)/5) % 2 == 0 then {terrain: dirt} else {blank}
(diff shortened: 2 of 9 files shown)


kostmo:
There are actually a few scenarios that use `x` and `y`:
```
find data/scenarios -name "*.yaml" -exec sh -c "yq .world.dsl \"{}\" | grep --color=always '\b[xy]\b'" \; -print
```
shows
```
, if (x/5 + y/5) % 2 == 0 then {terrain: dirt} else {blank}
, if ((x + 3) % 19)/12 + (y % 19)/12 == 0 then {terrain: grass} else {blank}
data/scenarios/Testing/1535-ping/1535-in-range.yaml
, if (x/4 + y/4) % 2 == 0 then {terrain: dirt} else {blank}
, if ((x + 3) % 19)/12 + (y % 19)/12 == 0 then {terrain: grass} else {blank}
data/scenarios/Testing/1535-ping/1535-out-of-range.yaml
, if (x + y) % 2 == 0 then {erase} else {blank}
data/scenarios/Testing/1320-world-DSL/erase.yaml
  , mask (flowers && (x + y) % 3 == 0) {wildflower}
  , mask (rubble && (x + y) % 2 == 0) {rock}
  , mask (outerShore && (x + y) % 2 == 0) {reed}
data/scenarios/Challenges/Ranching/beekeeping.yaml
, if (x/5 + y/5) % 2 == 0 then {terrain: dirt} else {blank}
, if (((x + 3) % 20)/11 + ((y + 3) % 20)/11) == 0 then {terrain: grass} else {blank}
data/scenarios/Challenges/gallery.yaml
, if (x + y / 2) % 5 == 0 then {dirt, wavy water} else {blank}
data/scenarios/Challenges/wave.yaml
  , mask (y < -80 || y > 80 || x < -60 || x > 60) (overlay [{water}])
data/scenarios/Fun/horton.yaml
```

In fact I do recall some bewilderment about the horizontal and vertical directions while authoring `wave.yaml`, but I suppose I brushed it off :)

byorgey:
@kostmo Oh, right, of course!  I just observed that all the tests still passed and assumed that meant nothing was using them.  I'll go through and fix these up so the worlds look the same as before.

byorgey:
@kostmo should be all fixed up now!

kostmo (approved):

PR #1990, opened 2024-06-25
Add `imap` to world DSL, with examples

### LINKED ISSUE #1584: Implement more world manipulation functions in world DSL
We should add more functions such as:

- `reflectV`
- `reflectH`
- `offset`/`translate`

See original design discussion at https://github.com/swarm-game/swarm/issues/1320#issuecomment-1588464512 .

byorgey:
Adds a new primitive `imap`, aka index map, primitive to the world DSL.  `imap` has type `World int -> World int -> World a -> World a`.  Think of it like `(Coords -> Coords) -> World a -> World a`, i.e. given a coordinate mapping, it creates a new world by looking up the cell at the transformed coordinates in the given base world.  However, since there are no lambdas we cannot directly give it that type; instead, the first `World int` represents a function `Coords -> int` which gives an x coordinate, and the second gives the y coordinate.  All told, `imap wx wy wa` is like `\c -> wa (wx c, wy c)`.  For example, `imap (-x) y w` is a reflection of `w` across the y-axis.

Adds a description of `imap` to the language reference, as well as adding a few examples.

Also removes the `rot` and `reflect` primitives, since they can now be simply implemented in terms of `imap`.

Depends on merging #1989 first.  Closes #1584.

### DIFF
--- a/src/swarm-scenario/Swarm/Game/World/Syntax.hs
+++ b/src/swarm-scenario/Swarm/Game/World/Syntax.hs
@@ -10,7 +10,6 @@ module Swarm.Game.World.Syntax (
   RawCellVal,
   CellTag (..),
   CellVal (..),
-  Rot (..),
   Var,
   Axis (..),
   Op (..),
@@ -65,16 +64,6 @@ instance PrettyPrec CellVal where
         ++ [(Just CellEntity, e ^. entityName) | EJust (Last e) <- [ent]]
         ++ map ((Just CellRobot,) . view trobotName) rs
 
-data Rot = Rot0 | Rot90 | Rot180 | Rot270
-  deriving (Eq, Ord, Show, Bounded, Enum)
-
-instance PrettyPrec Rot where
-  prettyPrec _ = \case
-    Rot0 -> "rot0"
-    Rot90 -> "rot90"
-    Rot180 -> "rot180"
-    Rot270 -> "rot270"
-
 type Var = Text
 
 data Axis = X | Y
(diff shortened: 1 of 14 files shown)


src/swarm-scenario/Swarm/Game/World/Syntax.hs
```diff
@@ -83,7 +72,7 @@ data Axis = X | Y
 instance PrettyPrec Axis where
   prettyPrec _ = \case X -> "x"; Y -> "y"
 
-data Op = Not | Neg | And | Or | Add | Sub | Mul | Div | Mod | Eq | Neq | Lt | Leq | Gt | Geq | If | Perlin | Reflect Axis | Rot Rot | Mask | Overlay | Abs
+data Op = Not | Neg | And | Or | Add | Sub | Mul | Div | Mod | Eq | Neq | Lt | Leq | Gt | Geq | If | Perlin | Mask | Overlay | Abs | IMap
```
xsebek, line 75: Were there no worlds using reflection and rotation?

xsebek (approved):
LGTM. I really like the clearing example. 🌲 🌲 🌲

src/swarm-scenario/Swarm/Game/World/Syntax.hs
byorgey: Correct.  Reflection wasn't even fully implemented, as it turns out.  And rotation is not very interesting until you have things like structures you can refer to from within the language.

PR #1995, opened 2024-06-27
Remove grid overlay truncation behavior

kostmo:
In the implementation of #1826, I decided to preserve the old truncating behavior as the default.  But after some consideration, there's no real use case for the old behavior, so let's take it out to simplify the code.


byorgey (approved):
I am always in favor of PRs that delete more lines than they add. :grin:

PR #1996, opened 2024-06-27
Remove duplicate transformers dependency

### LINKED ISSUE #1151: [Editor] Use hoistMaybe
See code comment

### LINKED ISSUE #1993: Remove redundant `transformers` dependency
In a couple different places in `swarm.cabal` we have `transformers` listed twice with different version bounds:
```
    transformers >=0.5 && <0.7,
    transformers >=0.5.6.2 && <0.6.2.0,
```
Does anyone know why this is?  Is there a specific reason we need version 0.5.6.2?

Relatedly, if we can get away with requiring `0.6`, then we can close #1151.

nitinprakash96:
Closes #1993


nitinprakash96:
What a surprise, builds fail for GHC <= 9.48. 🤦 Failed to catch this because I was developing on 9.6.5

byorgey:
Yeah, I guess we need to wait until we drop support for GHC 9.4.

nitinprakash96:
One other _possible but perhaps slightly less desirable_ can be to have our own `hoistMaybe` function but can be wrapped in CPP pragma `#if !MIN_VERSION_transformers(0,6,0) ... #endif`. That at least allows us to get rid of duplicate transformer dependency and keep supporting older GHC version.
But that might be an overkill if we'd like to go ahead with dropping support for 9.4 once adding 9.10

byorgey:
True, but I agree, using CPP seems a bit overkill here.

Note we can still fix #1993 right now, using `transformers >= 0.5 && < 0.7`.

byorgey (approved):

PR #2000, opened 2024-06-28
Save current REPL input with down arrow

### LINKED ISSUE #713: Save invalid inputs in REPL history with down arrow
The basic problem is that the only way to get a REPL input added to the history is after it successfully parses and type checks and is actually executed.  It's always frustrating when I type in a long REPL input but then realize I need to do something else first.  For example, suppose I type a long input but then realize that it uses some functions that are defined in `my_stuff.sw` but I forgot to load `my_stuff.sw`.  In that case my input doesn't even type check.  Or maybe my long input does type check but I realize that there's some other command I need to execute first in preparation for it.  The reason this is so frustrating is that the only solution is to clear the input, type the other command first, and then retype the long input from scratch, because there is no way to save it in between.

I guess the way the REPL currently works is actually pretty similar to the way shells or, say, ghci work.  But somehow it seems frustrating with Swarm, perhaps because it feels natural/encouraged to type longish commands at the REPL.

I really like the way irssi handles this: at any time you can hit the "down" arrow key, which adds the current input to the history and clears it.  So for example if you are in the middle of typing a longer message but want to pause in the middle in order to respond to something else quickly, you can just hit "down", type and send the new message, and then hit "up" a couple times to retrieve your partially complete message and continue composing it.

byorgey:
Closes #713.

Also does a bit of refactoring:
- Decompose `REPLHistItem` into a `REPLHistItemType` and some `Text` instead of having multiple constructors all containing a `Text`
- Add some helpers in `Controller.Util` to cut down on duplicated code

### DIFF
--- a/src/swarm-tui/Swarm/TUI/Controller.hs
+++ b/src/swarm-tui/Swarm/TUI/Controller.hs
@@ -425,7 +425,7 @@ quitGame :: EventM Name AppState ()
 quitGame = do
   -- Write out REPL history.
   history <- use $ uiState . uiGameplay . uiREPL . replHistory
-  let hist = mapMaybe getREPLEntry $ getLatestREPLHistoryItems maxBound history
+  let hist = mapMaybe getREPLSubmitted $ getLatestREPLHistoryItems maxBound history
   liftIO $ (`T.appendFile` T.unlines hist) =<< getSwarmHistoryPath True
 
   -- Save scenario status info.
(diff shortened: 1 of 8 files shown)


xsebek (approved):
If I approve this, I'll have to deal with the merge conflicts. 😅 

But it looks nice. 👍

byorgey:
> If I approve this, I'll have to deal with the merge conflicts. 😅

Haha, I actually thought about that.  I think I'll wait to merge this until after your keybindings PR - it might he easier for me to resolve the conflicts in a smaller PR, and I still have to fix up the test suite anyway.

byorgey:
> Congratulations on the two thousandth PR/Issue! 🥳 

Oh, nice! :tada:  I knew we were coming up on 2000 but hadn't noticed this was it.

xsebek:
@byorgey, thanks for waiting for me. Good luck with the rebase. 😅

PR #2003, opened 2024-06-30
Use 'enumerate' from the 'extra' package

kostmo:
Use [`enumerate`](https://hackage.haskell.org/package/extra-1.7.16/docs/Data-List-Extra.html#v:enumerate) instead of defining `listEnums` ourselves.

Also discovered a "safer" implementation of nonempty enumerations, which lets of remove the `Data.List.NonEmpty.fromList` hlint exclusion.

### DIFF
--- a/src/swarm-util/Swarm/Language/Syntax/Direction.hs
+++ b/src/swarm-util/Swarm/Language/Syntax/Direction.hs
@@ -25,11 +25,11 @@ import Data.Char qualified as C (toLower)
 import Data.Data (Data)
 import Data.Hashable (Hashable)
 import Data.List qualified as L (drop)
+import Data.List.Extra (enumerate)
 import Data.Text hiding (filter, length, map)
+import Data.Text qualified as T
 import GHC.Generics (Generic)
-import Swarm.Util qualified as Util
 import Swarm.Util.JSON (optionsMinimize)
-import Witch.From (from)
 
 ------------------------------------------------------------
 -- Directions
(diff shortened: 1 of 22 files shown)


src/swarm-util/Swarm/Language/Syntax/Direction.hs
```diff
@@ -126,7 +126,7 @@ instance ToJSON Direction where
 -- | Direction name is generated from the deepest nested data constructor
 -- e.g. 'DLeft' becomes "left"
 directionSyntax :: Direction -> Text
-directionSyntax d = from $ directionJsonModifier $ case d of
+directionSyntax d = T.pack $ directionJsonModifier $ case d of
```
byorgey, line 129: I don't like using `from` either, but how about `into @Text`?

src/swarm-util/Swarm/Util.hs
```diff
@@ -156,13 +156,10 @@ cycleEnum e
...
-listEnums :: (Enum e, Bounded e) => [e]
-listEnums = [minBound .. maxBound]
-
--- | Members of the Bounded class are guaranteed to
--- have at least one element.
+-- | See
+-- https://hackage.haskell.org/package/relude-1.2.1.0/docs/Relude-Enum.html#v:universeNonEmpty
 listEnumsNonempty :: (Enum e, Bounded e) => NonEmpty e
-listEnumsNonempty = NE.fromList listEnums
+listEnumsNonempty = minBound :| drop 1 enumerate
```
byorgey, line 162: Nice idea!  How about renaming this to `enumerateNonEmpty` for consistency?

byorgey (approved):

src/swarm-util/Swarm/Language/Syntax/Direction.hs
kostmo: Well, `T.pack` is shorter than that :smile:

PR #2005, opened 2024-07-01
Focus REPL panel on click

### LINKED ISSUE #1486: Clicking anywhere in REPL panel should focus it
> > I noticed I can no longer click in the middle of the REPL to focus on it, only the edges work.
>
> Good catch.  I looked into this for a while but can't figure it out.  It seems like after making the REPL into a scrollable viewport, click events are no longer generated for any clicks in the interior of the panel.  I'm wondering if it's a brick bug.

_Originally posted by @byorgey in https://github.com/swarm-game/swarm/pull/1481#discussion_r1313865515_

byorgey:
Closes #1486.

I'm not entirely sure I understand why this is necessary but it seems like for anything with a brick `viewport` it's not enough to just rely on wrapping it in `clickable`.  We have to actually set their focus explicitly depending on the identity of the clicked viewport.

### DIFF
--- a/src/swarm-tui/Swarm/TUI/Controller.hs
+++ b/src/swarm-tui/Swarm/TUI/Controller.hs
@@ -428,13 +428,14 @@ handleMainEvent ev = do
           EC.activateWorldEditorFunction y
         _ -> return ()
       flip whenJust setFocus $ case n of
-        -- Adapt click event origin to their right panel.
-        -- For the REPL and the World view, using 'Brick.Widgets.Core.clickable' correctly set the origin.
-        -- However this does not seems to work for the robot and info panel.
-        -- Thus we force the destination focus here.
+        -- Adapt click event origin to the right panel.  For the world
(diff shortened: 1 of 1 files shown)


nitinprakash96 (approved):

xsebek (approved):
Thanks for fixing this @byorgey. 👍 I read this code but it did not occur to me it was missing a line for REPL.

Perhaps this should be mentioned in the brick documentation.
Or it's a bug and clickable should apply through viewport.

byorgey:
> Perhaps this should be mentioned in the brick documentation.
> Or it's a bug and clickable should apply through viewport.

I'm really not sure!  Maybe someone should try to make a minimal example in order to ask about it on the brick issue tracker.  But I don't think I have the motivation to do that.

xsebek:
@byorgey on further thought, maybe this is correct.

If you have a clickable named widget and inside it another named widget, then the name in event will be the inner one.

Maybe the other panels just don't have inner named widgets.

PR #2008, opened 2024-07-01
On C-c, only cancel base CESK if it is currently running

### LINKED ISSUE #2007: Ctrl-C while base is not running results in frozen REPL
**Describe the bug**
Hitting Ctrl-C while the base is running a computation cancels it and returns control to the REPL, as it should.  However, hitting Ctrl-C while the base is *not* running a computation freezes the REPL: it displays `...` (as if the base were running a computation) and it is not possible to recover (even by hitting Ctrl-C again).  The remainder of the UI is still responsive.

**To Reproduce**
Start a new game, and immediately hit Ctrl-C.

**Expected behavior**
Hitting Ctrl-C while the base is currently idle should have no effect.

**Additional context**
This was most likely caused by #1928, though I have not yet confirmed that.

byorgey:
Fixes #2007.  After #1928, we handle exceptions a little differently, so it no longer works to simply set the base's CESK machine to an `Up` (exception) state no matter what.  Instead we have to first check if the base is running, and if so set it to `Up` as before; if not, do nothing (except clear the REPL prompt).

### DIFF
--- a/src/swarm-tui/Swarm/TUI/Controller.hs
+++ b/src/swarm-tui/Swarm/TUI/Controller.hs
@@ -1024,7 +1024,8 @@ handleREPLEvent x = do
     -- Handle Ctrl-c here so we can always cancel the currently running
     -- base program no matter what REPL control mode we are in.
     ControlChar 'c' -> do
-      gameState . baseRobot . machine %= cancel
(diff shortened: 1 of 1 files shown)


nitinprakash96 (approved):
Works fine!

xsebek (approved):

PR #2013, opened 2024-07-06
nonempty allBindings

kostmo:


xsebek:
As discussed, let's not use `NonEmpty` in this case.

PR #2017, opened 2024-07-08
Replace minBound to maxBound with enumerate

xsebek:
* remove the last two instances of `[minBound .. maxBound]`


byorgey (approved):
Thanks!

PR #2021, opened 2024-07-09
move more recognizer logic into topography sublibrary

kostmo:
Builds upon #1836.

This PR moves functions that are independent of `Entity` type into a new module within `swarm-topography`, alongside most of the other structure recognizer code.  Notably, this allows removal of the `AhoCorasick` dependency from the `swarm-scenario` sublibrary.


byorgey (approved):
Looks good to me!

PR #2028, opened 2024-07-13
Rolex, Olfactometer, and Dozer Blade Recipes

noahyor:
Moving towards finishing #1813 and #1268.

PR #2025, opened 2024-07-11
Add recipe for hourglass

noahyor:
Closes #2023.

PR #2030, opened 2024-07-13
Recipe coverage graph test

xsebek:
* Use the recipe graph to find unreachable entities instead of only looking for recipes for devices.
  * Improves on #1676.
* Add yielded entities in the recipe graph
* Add a test for each entity.

Before:
```
   Recipe coverage
    Ensure all devices have recipes (#1268):                                            FAIL (expected: Need to come up with more recipes)
      test/unit/TestRecipeCoverage.hs:36:
      Missing recipes for: "Elmer's glue", "binoculars", "blueprint", "caliper", "decoder ring", "dozer blade", "hourglass", "lambda", "linotype", "olfactometer", "rolex", "tape drive", and "wedge" (expected failure)
```
After
```
  Ensure all entities have recipes
    3D printer:                   OK
    ADT calculator:               OK
    Elmer's glue:                 FAIL (expected: More recipes needed (#1268))
      test/integration/TestRecipeCoverage.hs:34:
      Can not make "Elmer's glue" from starting entities. (expected failure)
    [... 100 more entities ...]
```


noahyor:
Looks good, but I don't really know enough about this to review it.

xsebek:
I am almost tempted to create a test for each entity - that way, when an entity is added without a recipe we would know.

That's what originally made me check out the test. Then I realized it is a simpler version of a recipe graph and only for devices, so I refactored it and got a few extra entities in the list.

xsebek:
And so I generated 100 more tests:
    
Screenshot 2024-07-13 at 20 43 24

@noahyor you said in #2025 that you expected the tests to fail - if this is merged before #2028, then they will. 😉

noahyor (approved):
I still think it's weird that it says the `lambda` doesn't have a recipe.

noahyor:
> @noahyor you said in #2025 that you expected the tests to fail - if this is merged before #2028, then they will. 😉


Well, there you go!

xsebek:
> I still think it's weird that it says the lambda doesn't have a recipe.

Well, it does not. (It is only _used_ to make curry.) That's why the test that @kostmo wrote reported it.

But with a recipe graph, the entities found in the world and base inventory are taken as the starting point, and everything else has to be craftable from them. So lambda is no longer reported. 🙂

xsebek:
Ha, it seems I did not fetch `main`:
```
hourglass:                                              OK (unexpected: More recipes needed (#1268))
   (unexpected success: More recipes needed (#1268))
    Use -p '/hourglass/' to rerun this test only.
```

byorgey:
`tea leaves` are obtainable from the world --- notice that they are `yield`ed by a `tea plant`.

test/integration/TestRecipeCoverage.hs
```diff
@@ -0,0 +1,74 @@
...
+    , "caliper"
+    , "cup of tea"
+    , "decoder ring"
+    , "dozer blade"
+    , "infinite improbability drive"
+    , "linotype"
+    , "olfactometer"
+    , "rolex"
+    , "tape drive"
+    , "tea leaves"
```
byorgey, line 69: We need to take into account entities which are listed under the `yields` field of existing world entities.

test/integration/TestRecipeCoverage.hs
```diff
@@ -0,0 +1,74 @@
...
+nonCoveredList :: [EntityName]
+nonCoveredList =
+  map
+    T.toCaseFold
+    [ "Elmer's glue"
+    , "ash"
+    , "binoculars"
+    , "blueprint"
+    , "caliper"
+    , "cup of tea"
```
byorgey, line 61: `cup of tea` should be covered once `tea leaves` are correctly included.

xsebek:
Thanks, @byorgey; it seems the recipe graph was not updated with the `yields` feature.

The infinite improbability drive is now properly shown as the hardest to obtain entity:
graphviz

byorgey:
Huh, strange, I thought I did update it a little while ago.  Maybe I am thinking of something else.

byorgey (approved):

src/swarm-doc/Swarm/Doc/Gen.hs
```diff
@@ -224,9 +229,37 @@ recipesToDot baseRobot classicTerm emap recipes = do
   mapM_ (uncurry (---<>)) (recipesToPairs recipeReqOut recipes)
   -- --------------------------------------------------------------------------
   -- also draw an edge for each entity that "yields" another entity
-  let yieldPairs = mapMaybe (\e -> (e ^. entityName,) <$> (e ^. entityYields)) . Map.elems $ entitiesByName emap
+  let yieldPairs = mapMaybe (\e -> (e ^. entityName,) <$> (e ^. entityYields)) . toList $ rgAllEntities graphData
   mapM_ (uncurry (.->.)) (both getE <$> yieldPairs)
```
xsebek, line 233: @byorgey I think you updated the graph display to Dot, but not the levels computation.

If Dot could lay out the graph well, the levels would not even be needed. But because we have so many connections, the levels at least group together nodes in order.

PR #2034, opened 2024-07-14
simplify recognizer more

kostmo:
* Eliminates redundant `EntityName` type parameter
* Removes `Ord` constraint, uses `HashMap`/`HashSet`


src/swarm-topography/Swarm/Game/Scenario/Topography/Structure/Recognition/Type.hs
```diff
@@ -22,23 +22,25 @@ import Control.Arrow ((&&&))
...
 import Linear (V2 (..))
 import Swarm.Game.Location (Location)
 import Swarm.Game.Scenario.Topography.Area
-import Swarm.Game.Scenario.Topography.Placement (StructureName)
-import Swarm.Game.Scenario.Topography.Structure (NamedGrid)
 import Swarm.Game.Universe (Cosmic, offsetBy)
 import Swarm.Language.Syntax.Direction (AbsoluteDir)
 import Text.AhoCorasick (StateMachine)
 
+type OriginalName = Text
```
xsebek, line 42: Could you please add some documentation to `OriginalName` and `NamedOriginal`?

I am not familiar with the code, but I wonder why `StructureName` needed to be switched out.

src/swarm-topography/Swarm/Game/Scenario/Topography/Structure/Recognition/Prep.hs
```diff
@@ -116,3 +115,15 @@ mkEntityLookup nameFunc grids =
       map swap $
         catMaybes $
           zipWith (\idx -> fmap (PositionWithinRow idx r,)) [0 ..] rowMembers
+
+-- * Util
+
+-- | Place the second element of the tuples into bins by
+-- the value of the first element.
+binTuplesHM ::
```
byorgey, line 123: How about moving this to `swarm-util`?

byorgey (approved):
Nice improvement/simplification!

src/swarm-topography/Swarm/Game/Scenario/Topography/Structure/Recognition/Prep.hs
kostmo: Let's do so if we get one more usage in the codebase.  As is, having this implementation separate would facilitate splitting the recognizer off to its own package at some point.

PR #2035, opened 2024-07-14
extract state initialization to new module

kostmo:
Towards #707

Reduces `Swarm.Game.State` by 200 lines and better encapsulates non-exported initialization functions.


byorgey (approved):
:+1:

PR #2036, opened 2024-07-14
Wait for windows

### LINKED ISSUE #2032: Mergify doesn't wait for Windows CI
Mergify merged and deleted a branch before Windows CI had finished.

xsebek:
Closes #2032.

Add windows to mergify

### DIFF
--- a/.github/.mergify.yml
+++ b/.github/.mergify.yml
@@ -24,6 +24,7 @@ queue_rules:
           - check-success=Haskell-CI - Linux - ghc-9.6.5
           - check-success=Haskell-CI - Linux - ghc-9.4.8
           - check-success=Haskell-CI - Linux - ghc-9.2.8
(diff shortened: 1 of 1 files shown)


byorgey (approved):
Thanks! Not sure how I missed this!

noahyor (approved):
I feel like this is good practice for now, but the windows CI takes >20 min sometimes!

byorgey:
> I feel like this is good practice for now, but the windows CI takes >20 min sometimes!

For some reason it seems like the caching does not work very well for the Windows build --- it seems like every time there is a cache miss and it has to build every single dependency from scratch.  I think this accounts for the discrepancy in build times but I have no idea what to do about it.

xsebek:
@byorgey I noticed it as well, but I have not figured it out yet.

The solution is probably to copy the Haskell-CI pattern. Recent discussion: https://discourse.haskell.org/t/github-cache-action/9926/8

Issue #2015, opened 2024-07-08
Move `src`, `test`, and `app` directories into a `code` directory


kostmo:
This is to simplify queries for `.hs` files, whether it be in IDEs or in the terminal, using `find`.

As is, the simplest `find` command includes unwanted build artifacts:
```
$ find . -name *.hs
./src/swarm-lang/Swarm/Language/Typed.hs
./src/swarm-lang/Swarm/Language/LSP/VarUsage.hs
[...]
./dist-newstyle/build/x86_64-linux/ghc-9.6.4/swarm-0.5.0.0/l/swarm-lang/noopt/build/swarm-lang/autogen/PackageInfo_swarm.hs
./dist-newstyle/build/x86_64-linux/ghc-9.6.4/swarm-0.5.0.0/l/swarm-lang/noopt/build/swarm-lang/autogen/Paths_swarm.hs
[...]
./app/doc/Main.hs
./app/tournament/Main.hs
[...]
./.stack-work/dist/x86_64-linux-tinfo6/ghc-9.6.4/build/swarm-tournament/autogen/PackageInfo_swarm.hs
./.stack-work/dist/x86_64-linux-tinfo6/ghc-9.6.4/build/swarm-tournament/autogen/Paths_swarm.hs
[...]
```
whereas these could be excluded easily if all of the code lived underneath a single `code` directory:
```
$ find code -name *.hs
```

xsebek:
There is [`fd`](https://github.com/sharkdp/fd) (a Rust `find` alternative) that solves this with better defaults:
```bash
fd --glob '*.hs'  # or fd '.*\.hs'
```
```
app/Main.hs
app/Swarm/App.hs
app/doc/Main.hs
app/scene/Main.hs
app/tournament/Main.hs
images/logo/MkLogo.hs
src/swarm-doc/Swarm/Doc/Command.hs
[...]
test/unit/TestUtil.hs
```
I would not mind moving the folders closer, though I would prefer cleaning up the top level. For example, it should be possible to move some of the CI configuration files to `.github`.

byorgey:
Can you elaborate on when and why we would want to use such commands?

To me, this seems to be optimizing for something that's not a big deal to me, but would make my day-to-day experience navigating the codebase slightly worse.

byorgey:
We could probably easily move everything in `app` into `src`.  I know that doesn't completely address the concern but it's something that occurred to me.

kostmo:
[...]

PR #2037, opened 2024-07-14
Move top-level Github/VSCode configuration files

xsebek:
* move Mergify, Restyled, Haskell-CI and Pull Request template to `.github`
  * regenerate Haskell-CI file with the new path and new Haskell CI version
* move CSpell from VSCode [Code Spell Checker](https://open-vsx.org/extension/streetsidesoftware/code-spell-checker) extension to `.vscode`

This is a complementary top-level cleanup to #2015.


byorgey (approved):

noahyor:
This will not work as of yet because `.vscode/` is in the `.gitignore`

noahyor:
However, we probably should exclude files other than the CSpell json file.

xsebek:
@noahyor I think it works because I force-added it to the commit, but it's better to have an exception for it in `.gitignore`.

PR #2041, opened 2024-07-14
Pretty print readable unicode characters

### LINKED ISSUE #1943: Unicode characters are escaped in pretty-printed `Text` values
**Describe the bug**
The `format` command escapes the Unicode character `∀`.

**To Reproduce**
Type in REPL:
```
> tydef L a = rec l. a * 1 end
> format {(undefined) : L a}
it : Text = "{undefined : \8704 a. L a}"
```

**Expected behavior**
The `format` command should show the ASCII `forall` or Unicode `∀`.

**Additional context**
This could be related to escaped characters in strings, like `\n`.

Discovered while testing https://github.com/swarm-game/swarm/pull/1928#discussion_r1641906543.

nitinprakash96:
Closes #1943

### DIFF
--- a/swarm.cabal
+++ b/swarm.cabal
@@ -191,7 +191,7 @@ library swarm-lang
     terminal-size >=0.3 && <1.0,
     text,
     text-rope >=0.2 && <0.3,
-    unicode-show,
+    unicode-show >=0.1 && <0.2,
     vector,
     vty,
     witch,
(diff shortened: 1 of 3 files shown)


swarm.cabal
```diff
@@ -191,6 +191,7 @@ library swarm-lang
     terminal-size >=0.3 && <1.0,
     text,
     text-rope >=0.2 && <0.3,
+    unicode-show,
```
byorgey, line 194: Since this is the only place we have this listed, let's include bounds, like `unicode-show >= 0.1 && <0.2`.

byorgey (commented):
Could we also add a unit test or two to ensure this works as expected?

byorgey (approved):
LGTM!

PR #2042, opened 2024-07-14
Fix restyled & mergify

noahyor:
Fixes restyled & mergify


xsebek (approved):
Thanks, sorry about that. 😅

PR #2049, opened 2024-07-15
Make logging intangible

### LINKED ISSUE #1991: Consider making the `log` command intangible
Currently, the `log` command is specified as a tangible command, meaning that robots can only run it at most once per tick.  However, I am not sure why we decided this in the first place.  The general idea is that tangible commands are ones which could have an observable effect to other robots, but there is no way for robots to observe some other robot's log entries (not without doing something like `salvage` first).  The one exception is that if you `view` another robot you can see its log, but that is only observable within the UI, and you cannot programmatically react to it.

noahyor:
Closes #1991.

### DIFF
--- a/data/scenarios/Testing/475-wait-one.yaml
+++ b/data/scenarios/Testing/475-wait-one.yaml
@@ -13,7 +13,7 @@ objectives:
         ishere "lambda"
       } { return false }
 solution: |
-  log "I will win next tick!"; place "lambda";
+  log "I will win next tick!"; wait 1; place "lambda";
 robots:
   - name: base
     loc: [0, 0]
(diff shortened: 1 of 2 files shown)


byorgey (approved):

data/scenarios/Testing/475-wait-one.yaml
```diff
@@ -30,7 +31,7 @@ robots:
       - logger
       - clock
     program: |
-      log "I shall sleep"; wait 1; log "I have awoken"
+      log "I shall sleep"; wait 2; log "I have awoken"
```
xsebek, line 34: Does this test still make sense, now that we wait for two ticks and  newly in solution for one? 🤔

xsebek (approved):

PR #2050, opened 2024-07-15
Add restore-keys logic for windows

### LINKED ISSUE #2040: Fix Windows CI caching
**Describe the bug**

Windows build rebuilds dependencies and does not seem to be cached.

**To Reproduce**

Open a PR that changes Haskell file and check Windows CI build log.

**Expected behavior**

The dependencies should be cached from prior runs.

**Additional context**

See discussion in #2036.

nitinprakash96:
Closes #2040


.github/workflows/haskell-ci-windows.yml
```diff
@@ -52,17 +52,21 @@ jobs:
...
           path: ${{ steps.setup-haskell-cabal.outputs.cabal-store }}
-          key: ${{ runner.os }}-${{ matrix.ghc }}-${{ hashFiles('cabal.project.freeze') }}
+          key: ${{ matrix.os }}-${{ matrix.ghc }}-${{ hashFiles('cabal.project.freeze') }}
+          restore-keys: |
+            ${{ matrix.os }}-${{ matrix.ghc }}-${{ hashFiles('cabal.project.freeze') }}
+            ${{ matrix.os }}-${{ matrix.ghc }}-
 
       - name: Install dependencies
+        if: steps.cache.outputs.cache-hit != 'true'
         run: cabal build all --only-dependencies
```
xsebek, line 66: Is this step even necessary?

xsebek (approved):

.github/workflows/haskell-ci-windows.yml
nitinprakash96: Yes. The idea is to have cache contain built dependencies only. This way even if the project itself fails to compile, we wouldn't have to worry about re-compiling dependencies all over again.

xsebek: But won't the next step compile dependencies if the cache is not present?

nitinprakash96: It will. But it will also compile swarm in the same step. And if compilation fails for swarm then on subsequent push, the dependencies will be compiled all over again.
Building dependencies in separate step will at least let up avoid such a case.

noahyor (approved):
I hope this fixes our Windows CI caching problems!

xsebek (changes requested):
I am very confused by the condition skipping the dependencies if there is a cache miss.

Please add a comment to the script explaining what is it trying to workaround.

.github/workflows/haskell-ci-windows.yml
xsebek: @nitinprakash96 is the condition there for testing and you dont intend to merge it?

nitinprakash96:
@xsebek we're not skipping the dependencies if there is a cache miss. Quite the opposite the actually.  If there is a cache miss, then we build the dependencies.

xsebek (approved):
@nitinprakash96 sorry I got confused.

I still think the condition should not be there - cabal will do the check in next step, so it does not save time. 

I am not sure if the restore key fallback could lead to dependencies being built anyway?
If not then feel free to merge.

PR #2051, opened 2024-07-15
Fix release URL asset path in Haskell release action

byorgey:
The `create_release` job worked fine.  The `build_artifact` job, however, after downloading the `release_url` artifact, seemed to be looking for it in the wrong place --- it tried to access `release_url/release_url.txt`, but the corrsponding zip file contains only the file `release_url.txt` with no folder.

This PR fixes the job to look for the `release_url.txt` file in the correct (I hope) place.


noahyor (approved):
Fingers crossed!

PR #2052, opened 2024-07-15
set required cabal-version to 3.6

byorgey:
Hackage does not like `cabal-version: 3.8`, it requires at most 3.6.  I don't know if there was a specific reason we were using 3.8, but it seems to work fine with 3.6 specified.


noahyor (approved):
Hopefully this works!

PR #2056, opened 2024-07-17
Add place to put local scripts

### LINKED ISSUE #2055: Add a local scripts directory which is ignored by git
This would add something like `scripts/local/` and would be included in the `.gitignore`

noahyor:
Closes #2055.


byorgey (commented):
I'm not quite sure I understand the purpose here.  If you just want to make some personal Swarm-related scripts, you can put them in a separate folder somewhere not under the Swarm repo root, and add them to your PATH.

PR #2060, opened 2024-07-19
[EXPERIMENTAL] weeder config test

kostmo:
Compare to: #2059


.github/workflows/cabal.haskell-ci
```diff
@@ -13,6 +13,8 @@ docspec: True
 -- Build haddocks. Accepts booleans or version range.
 haddock: True
 
+weeder: < 9.10.7
```
kostmo, line 16: Demonstration of a lexically "smaller" value (due to the leading `1`), which is converted to its numeric form as "greater" value.

Weeder gets run for all of the GHC versions (``,).

PR #2062, opened 2024-07-20
Show the current TPS if it is slower than the target.

### LINKED ISSUE #1561: Display UPS/FPS
A very useful tool for players conscious about performance in game is displayed UPS/FPS.

Here is [a nice explanation from helpful player of Factorio](https://www.reddit.com/r/factorio/comments/7c7ofu/comment/dpoysym/?utm_source=share&utm_medium=web2x&context=3):
> FPS = Frames Per Second. This number is sacrificed first if UPS is below 60.
> UPS = Updates Per Second, the number of times the game engine updates all entities on the map.

Because the terminal does not need to update all the time, we might want to add instead:
> TPS = Ticks Per Second. That is the actual number of ticks executed, not the target value we show in UI now.

This would allow even non-developers to notice and quantify performance problems and make life much easier for developers hunting those problems down.

noahyor:
Closes #1561.

### DIFF
--- a/src/swarm-tui/Swarm/TUI/View.hs
+++ b/src/swarm-tui/Swarm/TUI/View.hs
@@ -591,14 +591,14 @@ drawTPS :: AppState -> Widget Name
 drawTPS s = hBox (tpsInfo : rateInfo)
  where
   tpsInfo
-    | l >= 0 = hBox [str (show n), txt " ", txt (number n "tick"), txt " / s"]
+    | l >= 0 = hBox [str (show n), txt " ", tpsIndicator s, txt (number n "tick"), txt " / s"]
+    -- | l >= 0 = hBox [str (show n), txt " ", txt (number n "tick"), txt " / s"]
     | otherwise = hBox [txt "1 tick / ", str (show n), txt " s"]
 
   rateInfo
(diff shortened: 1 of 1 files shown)


src/swarm-tui/Swarm/TUI/View.hs
```diff
@@ -591,24 +591,39 @@ drawTPS :: AppState -> Widget Name
...
+tpsIndicator s
+  = hBox [txt "(", str $ show tps, txt ") "]
+  -- = str (show $ round tps )
+
+  -- | (tps / target) < 0.9 = str (show $ round tps )
+  -- | otherwise = emptyWidget
+  where
+    tpf = s ^. uiState . uiGameplay . uiTiming . uiTPF
+    fps = s ^. uiState . uiGameplay . uiTiming . uiFPS
+    tps = fps / tpf
```
byorgey, line 623: Shouldn't `tps = tpf * fps`?  ticks/frame * frames/sec = ticks/sec.  `fps / tpf` would be in units of square frames per tick-second or something like that.

noahyor: oops.. I guess I didn't do my math right

PR #2064, opened 2024-07-21
Use surveil in move tutorial

xsebek:
This should make the move tutorial run faster by making most of the robots `surveil` the door to their section.


data/scenarios/Tutorials/move.yaml
```diff
@@ -322,13 +398,25 @@ robots:
...
+      main [action="P", entity="flower", room=2]);
+      m
   - name: 3P flower
     system: true
-    program: run "scenarios/Tutorials/move_system.sw"
+    program: |
+      def main = \a. return noop end
+      m <- instant (run "scenarios/Tutorials/move_surveil.sw";
+      main [action="P", entity="flower", room=3]);
+      m
```
xsebek, line 419: @byorgey this is how `instant` has to be used to not pay for tangible `run`. The result is the rest of the computation which waits and places the wall.

byorgey:
For me, it's super laggy up until you move to the first flower, then very snappy after that.  By "super laggy" I mean e.g. it takes several seconds to respond at all when I start typing `move; move`; even after that it still takes up to half a second to respond to each keypress.

I actually don't quite understand what's going on here.  Sometimes, when I start the tutorial it pauses for several seconds *before* showing the goals dialog, and then is reasonably fast after that.  But sometimes, it shows the goal dialog right away, and then once I exit that it sits there not responding for several seconds.  The first feels annoying but acceptable; the second feels pretty bad because it looks like it is ready for you to start typing but it is not.

xsebek:
@byorgey I removed a few busy waiting robots and put the check robot to sleep if you are far from the door.

I don't think I can easily optimize this further; most of the time is spent parsing the same file 60 times:
Screenshot 2024-07-22 at 1 27 26

I could make a robot that runs the file and instantly builds new robots, which would teleport to their place.
That would only parse once, but have new problems:
* timing problems - build takes another tick, so the first corridor would have to be longer
* it would use manually written position lists with no declarative map
* I don't want to write that; I would rather work on `import`

byorgey:
Hmm, it definitely feels more responsive now (@DanRyba253 also posted some timing data in Discord suggesting the same thing).  However, it seems slightly broken: when I execute `move; move` to go to the first flower, the corridor to the second flower opens up, but there is no goal popup dialog.  After I move to the second flower, two goal dialogs pop up in succession.

byorgey:
And yes, as for `import`, definitely don't spend effort trying to optimize the repeated file reading, we'll work on that separately.

data/scenarios/Tutorials/move.yaml
```diff
@@ -13,7 +13,7 @@ objectives:
     condition: |
       r <- robotNamed "check";
       w <- as r {count "Win"};
-      return (w >= 2)
+      return (w >= 1)
```
xsebek, line 16: @byorgey, thanks for pointing it out; I made a typo while changing the conditions to use one robot.

byorgey (approved):
Looks good now, this seems a strict improvement over the previous version.

I still don't understand why sometimes I get a long pause *before* the initial goal modal is shown, and sometimes I get a long unresponsive pause *after* dismissing the goal modal.  I have not been able to find any particular way to consistently get one behavior or the other. But whatever the reason I don't think it's the fault of this scenario.

PR #2067, opened 2024-07-21
fix markdown link

byorgey:


xsebek (approved):
Markdown always parses. 😄

PR #2071, opened 2024-07-22
Consolidate code in `WorldDescription`

### LINKED ISSUE #2069: Consolidate structure assembly code between `WorldDescription.hs` and `Structure.hs`
See:
https://github.com/swarm-game/swarm/blob/6a3f84e514c2c8cf50e084bdf29368145b7993b2/src/swarm-scenario/Swarm/Game/Scenario/Topography/WorldDescription.hs#L85-L94
vs.
https://github.com/swarm-game/swarm/blob/6a3f84e514c2c8cf50e084bdf29368145b7993b2/src/swarm-topography/Swarm/Game/Scenario/Topography/Structure.hs#L105-L115

kostmo:
Closes #2069

Simplifies the code in `WorldDescription.hs` in three ways:
* extract many `liftE` invocations to a single location
* use a new `genStructure` function consolidated with `Structure.hs`
* remove a now-redundant `ul :: Location` field from `WorldDescription`

Also used this PR to remove some re-exports from `Scenario.hs`.


src/swarm-topography/Swarm/Game/Scenario/Topography/Structure.hs
```diff
@@ -100,19 +100,26 @@ instance FromJSON (Grid Char) where
           fail "Grid is not rectangular!"
         return g
 
+genStructure ::
```
byorgey, line 103: Does the `gen` stand for `generic`?  At first I thought it meant `generate` which didn't make much sense.

byorgey (approved):

src/swarm-topography/Swarm/Game/Scenario/Topography/Structure.hs
kostmo: Renamed to `parseStructure`.

PR #2079, opened 2024-07-24
ensure non-overlapping mask char

kostmo:
We want to ensure that the character used for the `mask` does not also occur as an entry in the `palette`.

# Testing
Added new test:
```
scripts/play.sh -i data/scenarios/Testing/_Validation/2077-mask-char.yaml
```
# Also in this PR

To simplify the test, we make use of the fact that `mask` is now also supported at the toplevel in addition to within a `structure` definition.

### DIFF
--- a/data/scenarios/Testing/_Validation/2077-mask-char.yaml
+++ b/data/scenarios/Testing/_Validation/2077-mask-char.yaml
@@ -1,0 +1,14 @@
+version: 1
+name: Mask char palette overlap
+description: |
+  Ensure mask char is not used as a palette key
+robots: []
+world:
+  mask: q
+  palette:
+    '.': [grass]
+    'q': [dirt]
+  upperleft: [0, 0]
+  map: |
+    ..
+    ..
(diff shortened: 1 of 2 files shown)


data/schema/world.json
```diff
@@ -56,6 +56,10 @@
             "type": "boolean",
             "description": "Whether players are allowed to scroll the world map."
         },
+        "mask": {
```
byorgey, line 59: Wait, I'm confused. Is this something we already had, but forgot to add to the schema (and shouldn't that be impossible, since we validate all the scenarios against the schema)? Or is this a new field you're adding in this PR?

kostmo: Use of the `"mask"` property at the toplevel became possible as of #2071.  Eventually it would be nice to re-use the `structure.json` schema as a `$ref` inside the `world.json` schema to avoid duplication.  But I think that may require putting it under a new key, which would entail re-organizing all existing scenario files (probably doable with automation).

xsebek:
@kostmo, since this is not tied to an issue, could you please explain this in the description?

I am unsure what a mask in the world means - what can you see through the transparent cell below the map? 👀

kostmo:
> what can you see through the transparent cell below the map?

This exposes the DSL underneath.  Demo in #2085.

data/schema/world.json
kostmo: Ok, I have now moved the schema update to #2085, which will have to be merged before this PR.

byorgey (approved):

PR #2080, opened 2024-07-28
Add CLI option to start the game paused

### LINKED ISSUE #2076: Option to start the game paused

Sometimes (#2064), I need to debug the first tick of the game.
That is hard because the game starts in a running state.

Add command line option `--paused` to initialize the game state with the `runState == ManualPaused`.

I would not mind writing `--debug=startPaused` from #1324.

Starting with the game paused could be useful for competitive play, but probably not a good default.

xsebek:
* add CLI option:
  ```
  -p,--paused              Pause the game at start.
  ```
* closes #2076

---

You can test it with:
```sh
cabal run swarm -O0 -- --scenario Tutorials/move --paused
```
After one step (<kbd>Ctrl</kbd>+<kbd>O</kbd>) the Goal dialog will show up along with the rest of the UI.

### DIFF
--- a/app/game/Main.hs
+++ b/app/game/Main.hs
@@ -57,6 +57,7 @@ cliParser =
  where
   appOpts :: Parser AppOpts
   appOpts = do
+    let repoGitInfo = gitInfo
     userSeed <- seed
     userScenario <- scenario
     scriptToRun <- run
@@ -66,7 +67,6 @@ cliParser =
     cheatMode <- cheat
     colorMode <- color
     userWebPort <- webPort
-    repoGitInfo <- pure gitInfo
     return $ AppOpts {..}
 
   input :: Parser FormatInput
--- a/app/game/Main.hs
+++ b/app/game/Main.hs
@@ -66,8 +66,8 @@ cliParser =
     cheatMode <- cheat
     colorMode <- color
     userWebPort <- webPort
-    let repoGitInfo = gitInfo
-    return AppOpts {..}
+    repoGitInfo <- pure gitInfo
+    return $ AppOpts {..}
 
   input :: Parser FormatInput
   input =
--- a/app/game/Main.hs
+++ b/app/game/Main.hs
@@ -1,10 +1,12 @@
+{-# LANGUAGE ApplicativeDo #-}
 {-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE RecordWildCards #-}
 {-# LANGUAGE TemplateHaskell #-}
 
 -- |
 -- SPDX-License-Identifier: BSD-3-Clause
-module Main where
+module Main (main) where
 
 import Control.Monad (when)
 import Data.Foldable qualified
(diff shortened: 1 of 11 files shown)


app/game/Main.hs
```diff
@@ -44,25 +46,29 @@ cliParser :: Parser CLI
...
+    userScenario <- scenario
+    scriptToRun <- run
+    pausedAtStart <- paused
+    autoPlay <- autoplay
+    speed <- speedFactor
+    cheatMode <- cheat
+    colorMode <- color
+    userWebPort <- webPort
+    let repoGitInfo = gitInfo
+    return AppOpts {..}
```
xsebek, line 70: Strange, it seems GHC 9.2 and 9.4 did not understand the [`ApplicativeDo`](https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/applicative_do.html) I wanted to use here.

byorgey: Hmm, maybe they don't deal properly with `ApplicativeDo` + record wildcards?

byorgey: That is unfortunate.  Perhaps just put it back the way it was, but add a comment (+ an issue?) to update the code once we drop support for GHC 9.4?

byorgey (commented):
It seems a bit strange to me that if you use the `-p` option without `--scenario`, it starts the menu but then auto-pauses whatever is the first scenario you choose from the menu.  I guess it makes sense though.

app/game/Main.hs
byorgey: Oh! Try using `pure` instead of `return`, perhaps?

This page seems to indicate this style has been possible for a while: https://chshersh.com/recordwildcards

src/swarm-tui/Swarm/TUI/View.hs
```diff
@@ -555,7 +555,9 @@ drawClockDisplay :: Int -> GameState -> Widget n
 drawClockDisplay lgTPS gs = hBox . intersperse (txt " ") $ catMaybes [clockWidget, pauseWidget]
  where
   clockWidget = maybeDrawTime (gs ^. temporal . ticks) (gs ^. temporal . paused || lgTPS < 3) gs
-  pauseWidget = guard (gs ^. temporal . paused) $> txt "(PAUSED)"
+  pauseWidget =
+    let runState = gs ^. temporal . runStatus
+     in guard (gs ^. temporal . paused) $> txt (if runState == AutoPause then "(AUTO-PAUSED)" else "(PAUSED)")
```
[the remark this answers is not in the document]
xsebek: I used it for testing, but it is a useful distinction.

The auto-paused game will **unpause** when you close the modal.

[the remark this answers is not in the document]
xsebek: Maybe I could split that into a separate PR and make it possible to switch from an auto to a manual pause.

app/game/Main.hs
xsebek: The error message is wrong - the problem is in the line above containing `let`. As GHC docs note:

> _In particular, slight variations such as `return . Just $ x` or `let x = e in return x` would not be recognised._

Later versions of GHC seem to have gotten better at this.

xsebek:
> It seems a bit strange to me that if you use the `-p` option without `--scenario`, it starts the menu but then auto-pauses whatever is the first scenario you choose from the menu. I guess it makes sense though.

@byorgey, it's not just the _first_ scenario you open; it changes the default for all scenarios to start paused. 🙂 

I hope it is an understandable and useful behavior.

Could you please also have a look at the added `Bool` parameters? I am not entirely happy with them, but I followed compilation errors and connected the CLI to GameState. I wonder if `pause` even belongs in GameState, but the wrapper functions make it hard to tell where it is used.

noahyor:
> I wonder if pause even belongs in GameState, but the wrapper functions make it hard to tell 
> where it is used.

I think it should be in the UI state in preparation for asynchronous UI (pr/issue soon).

byorgey:
> it's not just the _first_ scenario you open; it changes the default for all scenarios to start paused

Ah, that makes sense.

> I wonder if `pause` even belongs in GameState

On the contrary, I think it *does* belong in `GameState`, and thinking about asynchronous UI shows more clearly why.  To answer the question of what belongs in the `GameState` and what belongs in the `UIState`, ask yourself the question: is this something specific to this particular UI, or is it something inherent to the game which would be needed even if we put a different UI on top (web-based, GUI-based, etc.)?  Especially if we want to have the game running in one thread and the UI running in another thread, then the game itself needs to keep track of whether it is currently paused, so that it can know whether to step independently of the UI telling it so.  For example, the game may run for several ticks during a single frame, but if an objective is completed during one of those ticks, the game needs to immediately auto-pause without waiting for the UI to tell it that it should do so, which could come several ticks late.  (Actually I am unsure whether the game currently does this correctly or not.)

xsebek:
Thanks, @byorgey for thinking that through - that explanation could well be a note in Temporal Game State. 👍

PR #2082, opened 2024-07-28
Clarify speed option help text

### LINKED ISSUE #2078: Improve help text for `--speed` option
> Speed n means 2^n ticks per second.  We should improve the help text to make that clearer.  Right now it says "Initial game speed multiplier" which seems ambiguous at best.

_Originally posted by @byorgey in https://github.com/swarm-game/swarm/issues/2076#issuecomment-2245127116_

xsebek:
* update CLI option help: 
  ```
  -m,--speed N             Initial game speed. Counted as 2^N ticks per second,
                           default is 4 meaning 16.0 t/s.
  ```
* closes #2078

### DIFF
--- a/app/game/Main.hs
+++ b/app/game/Main.hs
@@ -120,6 +120,7 @@ cliParser =
       , "meaning"
       , show $ 2 ** fromIntegral @Int @Double defaultInitLgTicksPerSecond
       , "t/s."
+      , "(Negative values are allowed, e.g. -3 means 1 tick per 8 sec.)"
       ]
   cheat :: Parser Bool
   cheat = switch (long "cheat" <> short 'x' <> help "Enable cheat mode. This allows toggling Creative Mode with Ctrl+v and unlocks \"Testing\" scenarios in the menu.")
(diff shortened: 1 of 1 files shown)


app/game/Main.hs
```diff
@@ -111,7 +111,16 @@ cliParser =
...
-  speedFactor = option auto (long "speed" <> short 'm' <> value defaultInitLgTicksPerSecond <> help "Initial game speed multiplier")
+  speedFactor = option auto (long "speed" <> short 'm' <> metavar "N" <> value defaultInitLgTicksPerSecond <> help speedFactorHelp)
+  speedFactorHelp =
+    unwords
+      [ "Initial game speed."
+      , "Counted as 2^N ticks per second, default is"
+      , show defaultInitLgTicksPerSecond
+      , "meaning"
+      , show $ 2 ** fromIntegral @Int @Double defaultInitLgTicksPerSecond
+      , "t/s."
```
byorgey, line 122: Maybe also mention that a negative value of N is allowed?

byorgey (approved):

app/game/Main.hs
xsebek: Like so?
```suggestion
      , "(Negative value is allowed, e.g. -4 means 1 tick per 8 sec.)"
```

byorgey: Sounds good, with the minor nitpick that I would say either `Negative values are allowed` or `A negative value is allowed` (personally, I prefer the first).

PR #2084, opened 2024-07-28
remove pandoc dependency from main game

### LINKED ISSUE #2083: `build-game.sh` should not require `pandoc`
Currently, `swarm-web` depends on `swarm-doc`, which uses `pandoc`.  Since the `swarm` executable depends on `swarm-web`, we end up compiling `pandoc` just to play the game.

We probably should be able to remove the `swarm-doc` dependency from `swarm-web`.

kostmo:
Closes #2083

Rendering a matrix of command attributes was originally implemented in #1658.  But this generated the data to drive the table at runtime, which pulled in an unfortunate Pandoc dependency.

To work around this, let's just pre-generate the data offline.

# Demo

1. `scripts/play.sh`
2. Visit http://localhost:5357/command-matrix.html

### DIFF
--- a/web/data/commands.json
+++ b/web/data/commands.json
@@ -1,0 +1,4189 @@
+{
+  "entries": [
+    {
+      "argTypes": [
+        {
+          "TyConF": [
+            {
+              "TCCmd": []
+            },
+            [
+              {
+                "TyConF": [
+                  {
+                    "TCBase": "BUnit"
+                  },
+                  []
+                ]
+              }
+            ]
+          ]
+        }
+      ],
+      "cmd": "Noop",
+      "derivedAttrs": {
+        "hasActorTarget": false,
+        "modifiesEnvironment": false,
+        "modifiesRobot": false,
+        "movesRobot": false,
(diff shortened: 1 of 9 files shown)


web/data/commands.json
kostmo, line 1: @byorgey on second thought, maybe we should commit this to the `swarm-game.github.io` repo instead?

xsebek: @kostmo, that sounds great to me; I did not even know about this matrix. 👍

byorgey: @kostmo Perhaps, but I think that can go in a separate PR.

byorgey (approved):

Ondrej-Sebek-TFS:
@kostmo Will this work in release as the file is not specifically mentioned in cabal?

If this temporal state persists too long an integration test to keep it up to date could be useful.

I am not against merging it, just wary about potential issues if we forget about it. 😉

xsebek:
@kostmo Will this work in release as the file is not specifically mentioned in cabal?

If this temporal state persists too long an integration test to keep it up to date could be useful.

I am not against merging it, just wary about potential issues if we forget about it. 😉

PR #2094, opened 2024-08-04
Split debug options

### LINKED ISSUE #1324: better name for `--cheat` CLI option
There may be a name that better reflects the capabilities afforded by the `--cheat` option.  Possible alternatives:
* `--omnipotent`
* `--superuser`

Also related: #1270

xsebek:
* replace `_uiCheatMode :: Bool` with `_uiDebugOptions :: (Set DebugOption)`
  * replace `uiHideGoals` with `uiIsAutoPlay` and debug option `ShowGoalDialogsInAutoPlay` 
* add command line flag `--debug` and make `--cheat` an alias for `--debug=creative` only 

```
  -x,--cheat               Enable cheat mode. This is an alias for --debug=creative
  -d,--debug OPTS          Use 'all' or a comma separated list of options:
                             * creative           allow toggling creative mode on/off
                             * editor             allow toggling the world editor mode on/off
                             * cesk               allow toggling the CESK debug view on/off
                             * all_robots         list all robots (including system robots) in the robot panel
                             * robot_id           list robot IDs in the robot panel
                             * hidden_goals       show hidden objectives in the goal dialog
                             * autoplay_goals     show goal dialogs when running in autoplay
                             * testing            show Testing folder in scenarios menu
```

It even has shell completions:
```bash
swarm --zsh-completion-script `which swarm` > ~/.oh-my-zsh/completions/_swarm
swarm --debug  # press Tab
# all  all_robots  autoplay_goals  cesk  creative  editor  hidden_goals  robot_id  testing
```

* closes #1324

### DIFF
--- a/app/game/Main.hs
+++ b/app/game/Main.hs
@@ -138,17 +138,28 @@ cliParser =
   debug :: Parser (Set.Set DebugOption)
   debug = Set.unions <$> many debugOption
   debugOption :: Parser (Set.Set DebugOption)
-  debugOption = option debugOptionList (long "debug" <> short 'd' <> metavar "OPTS" <> hidden <> helpDoc debugOptionHelp)
+  debugOption =
+    option
+      debugOptionList
+      ( long "debug"
+          <> short 'd'
+          <> metavar "OPTS"
+          <> hidden
+          <> helpDoc debugOptionHelp
+          <> completeWith debugCompletions
+      )
+  allDebug = [minBound .. maxBound]
+  debugCompletions = "all" : map debugOptionName allDebug
   debugOptionList :: ReadM (Set.Set DebugOption)
   debugOptionList = eitherReader $ \case
-    "all" -> pure $ Set.fromAscList [minBound .. maxBound]
+    "all" -> pure $ Set.fromAscList allDebug
     opts -> Set.fromList <$> readDebugOptionList opts
   debugOptionHelp :: Maybe Doc
   debugOptionHelp =
     Just . nest 2 . vsep $
       "Use 'all' or a comma separated list of options:"
         : [ fillBreak 20 ("*" <+> pretty name) <+> pretty desc
-          | o <- [minBound .. maxBound]
+          | o <- allDebug
           , let name = debugOptionName o
           , let desc = debugOptionDescription o
           ]
--- a/app/game/Main.hs
+++ b/app/game/Main.hs
@@ -68,9 +68,10 @@ cliParser =
     autoPlay <- autoplay
     speed <- speedFactor
     debugOptions <- debug
+    cheatMode <- cheat
     colorMode <- color
     userWebPort <- webPort
-    return $ AppOpts {..}
+    return $ (\ao -> ao {debugOptions = Set.union cheatMode debugOptions}) $ AppOpts {..}
 
   input :: Parser FormatInput
   input =
@@ -135,7 +136,7 @@ cliParser =
       ]
 
   debug :: Parser (Set.Set DebugOption)
-  debug = Set.unions <$> (([Set.singleton ToggleCreative] <$ cheat) <|> many debugOption)
+  debug = Set.unions <$> many debugOption
   debugOption :: Parser (Set.Set DebugOption)
   debugOption = option debugOptionList (long "debug" <> short 'd' <> metavar "OPTS" <> hidden <> helpDoc debugOptionHelp)
   debugOptionList :: ReadM (Set.Set DebugOption)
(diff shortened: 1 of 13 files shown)


app/game/Main.hs
```diff
@@ -130,8 +133,30 @@ cliParser =
       , "t/s."
       , "(Negative values are allowed, e.g. -3 means 1 tick per 8 sec.)"
       ]
+
+  debug :: Parser (Set.Set DebugOption)
+  debug = Set.unions <$> (([Set.singleton ToggleCreative] <$ cheat) <|> many debugOption)
```
xsebek, line 138: @byorgey, this does not allow combinations of cheat and debug, like `swarm --cheat --debug=robot_id `.

I am not sure if there is a way to do it with Applicative and Alternative. If not, and we wanted to support it, I would have to put `cheatMode` back to app options and resolve it in state initialisation.

xsebek: OK, I found a way but I am not happy about it. 😅 Nicer ideas welcome.

app/game/Main.hs
```diff
@@ -64,10 +67,11 @@ cliParser =
     pausedAtStart <- paused
     autoPlay <- autoplay
     speed <- speedFactor
+    debugOptions <- debug
```
byorgey, line 70: I suggest something like this:

```suggestion
    initialDebugOptions <- debug
    let debugOptions = Set.union cheatMode initialDebugOptions
```

Then we can get rid of the lambda + record update in the final `return`.

app/game/Main.hs
```diff
@@ -130,8 +134,41 @@ cliParser =
...
+    option
+      debugOptionList
+      ( long "debug"
+          <> short 'd'
+          <> metavar "OPTS"
+          <> hidden
+          <> helpDoc debugOptionHelp
+          <> completeWith debugCompletions
+      )
+  allDebug = [minBound .. maxBound]
```
byorgey, line 151: Use `Data.List.Extra.enumerate` instead of `[minBound .. maxBound]`

src/swarm-tui/Swarm/TUI/Controller/EventHandlers/Main.hs
```diff
@@ -149,7 +150,7 @@ isRunning = do
 whenRunning :: EventM Name AppState () -> EventM Name AppState ()
 whenRunning a = isRunning >>= \r -> when r a
 
-whenCheating :: EventM Name AppState () -> EventM Name AppState ()
-whenCheating a = do
-  s <- get
-  when (s ^. uiState . uiCheatMode) a
+whenDebug :: DebugOption -> EventM Name AppState () -> EventM Name AppState ()
+whenDebug d a = do
+  debug <- use $ uiState . uiDebugOptions . icontains d
```
byorgey, line 155: Why is `icontains` necessary?  Wouldn't `contains` work just as well?

byorgey (approved):
Looks great!

src/swarm-tui/Swarm/TUI/Controller/EventHandlers/Main.hs
xsebek: It would, I somehow saw `icontains` first and did not question the naming. 😅

app/game/Main.hs
xsebek: I might have underestimated the powers of ApplicativeDo after last issues. I will try this. 👍

xsebek: Nope, this is not accepted by ApplicativeDo.
```
app/game/Main.hs:62:13: error:
    • No instance for (Monad Parser) arising from a use of ‘return’
```

src/swarm-tui/Swarm/TUI/Controller/EventHandlers/Main.hs
xsebek: Hah, it did spare me ambiguity caused by `Swarm.Game.Entity.contains`.

app/game/Main.hs
byorgey: Oh, right, because it's referring to `cheatMode` which is the result of a previous action.  Actually, I now understand why `let debug = Set.union cheatMode ...` is not accepted either---for the same reason.

byorgey: Now I don't even understand why the existing code works.

xsebek: @byorgey I imagine GHC constucts a suitable `a` that puts everything in the right place:

```Haskell
(\c d o -> A (c + d) o) <$> c <*> d <*> o
```

PR #2098, opened 2024-08-04
Replace `ReadableIORef` with IO

xsebek:
* replaces the `ReadableIORef a` with an `IO a`, by pre-applying `readIORef` to the `IORef a`

The result of `IO a` is immutable `a`, so this is even safer than a newtype that needs a carefully chosen export list.

Basically this removes one file and some boilerplate from `Web.hs`.


kostmo:
As I recall, the point of the `ReadableIORef` newtype was to hide access to the `writeIORef` function to convey that in the context of the webserver, the game state should be considered read-only.

xsebek:
@kostmo I understand the intent, but IMO the boilerplate is not worth it.

The readonly aspect is repeatedly mentioned in comments and it is impossible to modify a _value_ so it's just as safe.

If anything, a closure is a safer wrapper, as it would take even more effort to get the `IORef` from it (compared to say `coerce`).

kostmo:
As things are today, one thread (the main game thread) *writes* to the shared state, and another thread (the web thread) *reads* from the shared state.  So technically, if one were to somehow accidentally *write* to the shared state from the web thread, in the worst case the programmer has probably merely wasted their time; the authoritative game state is still hosted in the game thread, and would eventually overwrite the web-thread's errant writes.  Or perhaps it may manifest as a race condition between web operations that read from the state vs. web operations that write to it.

I want to make sure we're not tearing down [Chesterton's Fence](https://en.wiktionary.org/wiki/Chesterton%27s_fence) with this PR.  I was originally compelled to create this newtype wrapper because I had not yet fully understood the data flow, and I myself attempted to write to the shared state from the web thread (i.e. wasted my own time).

> it is impossible to modify a _value_ so it's just as safe.

Hypothetically, if some code in the main game thread was actually *reading* the shared state, it seems to me that the `ReadableIORef` wrapper would indeed protect against interfering writes from the web thread.  If we leave the `ReadableIORef` in place as it is, could you explain how it could be circumvented such that the web thread could write to the shared state?

byorgey:
Just to make sure I'm understanding correctly:

- We currently have `ReadableIORef` which is just like `IORef` except read-only.
- If we were to replace that with a plain `IORef` it would indeed be less safe, because then it would be possible to write to it from the web thread even though that is not supposed to happen.
- However, that is not what this PR does: it instead replaces the `ReadableIORef a` with an `IO a`, by pre-applying `readIORef` to the `IORef a`.  So all the webserver has is an `IO a`, and the only thing it can do with this is to run it, extracting the current value of type `a`.  This is just as safe as a `ReadableIORef` --- perhaps even more so, because its safety does not depend on carefully not exporting certain things from some utility module, and it cannot be circumvented with `coerce`.

app/game/Swarm/App.hs
```diff
@@ -91,7 +90,7 @@ appMain opts = do
       eport <-
         Swarm.Web.startWebThread
           (userWebPort opts)
-          (mkReadonly appStateRef)
+          (readIORef appStateRef)
```
byorgey, line 93: Would be nice to add a comment here explaining that we are providing read-only access to the `IORef` by pre-applying `readIORef`, resulting in an `IO AppState` value which cannot do anything except provide the current value every time it is executed.

src/swarm-web/Swarm/Web.hs
```diff
@@ -147,7 +145,7 @@ swarmApiMarkdown =
 -- ------------------------------------------------------------------
 
 mkApp ::
-  ReadableIORef AppState ->
+  IO AppState ->
```
byorgey, line 148:
```suggestion
  -- | Read-only access to the current AppState
  IO AppState ->
```

byorgey (commented):
Not sure I understand the CI failures.

kostmo:
> Just to make sure I'm understanding correctly:
> 
> * We currently have `ReadableIORef` which is just like `IORef` except read-only.
> * If we were to replace that with a plain `IORef` it would indeed be less safe, because then it would be possible to write to it from the web thread even though that is not supposed to happen.
> * However, that is not what this PR does: it instead replaces the `ReadableIORef a` with an `IO a`, by pre-applying `readIORef` to the `IORef a`.  So all the webserver has is an `IO AppState`, and the only thing it can do with this is to run it, extracting the current `AppState`.  This is just as safe as a `ReadableIORef` --- perhaps even more so, because its safety does not depend on carefully not exporting certain things from some utility module, and it cannot be circumvented with `coerce`.
> 
> I think it's a really nice idea. @kostmo, I am not sure I understand your objections, and I wonder if you have misunderstood what this PR does (which I why I wrote this summary). If not, apologies, and I'll try harder to understand.

Thanks for explaining! Indeed I had misunderstood the PR, and I agree that it's a good idea now.

byorgey (approved):

kostmo:
@xsebek I am impressed by this simplification, and it made me think, is there an analogous protection we could add to the game thread that enforces that the state is only **written** to, and never read?  I at first thought we would need something like a `WriteOnlyIORef` newtype wrapper, but perhaps we could "pre-apply" the `writeIORef` function, and instead pass in a `a -> IO ()`?

byorgey:
> we could "pre-apply" the `writeIORef` function, and instead pass in a `a -> IO ()`?

Yes, I think that works!

xsebek:
Thanks for talking this over @byorgey and @kostmo. I will use this to clarify the issue description. 👍

If I read this correctly the CI thinks swarm the executable does not need to depend on swarm utils.

PR #2099, opened 2024-08-08
Demo standalone colored structures

kostmo:
Demonstrates a cell type that stores hex colors, along with some refactoring.

## Test

```
scripts/test/run-tests.sh standalone-topography
```

# Changes

* Introduces a `ToPixel` class that consolidates conversion to RGBA color
* Moved some rendering code from `Swarm.Game.World.Render` to `Swarm.Game.Scenario.Topography.Rasterize`
* Adds a new structure rendering test


src/swarm-scenario/Swarm/Game/World/Render.hs
```diff
@@ -38,8 +40,34 @@ import Swarm.Language.Pretty (prettyString)
...
+    False -> PixelRGBA8 0 0 0 255
+    True -> PixelRGBA8 255 255 255 255
+
+data ColorableCell
+  = OneBit OneBitColor
+  | Hex HexColor
+
+instance ToPixel ColorableCell where
+  toPixel (OneBit x) = toPixel x
+  toPixel (Hex x) = toPixel x
```
kostmo, line 60: @byorgey it seems this instance should be derivable.  Do you know how?

byorgey: Hmm, no, not any simple way at least.  I can imagine very overcomplicated things involving `Generic` but definitely not worth it unless we had many of these instances.

data/test/standalone-topography/rainbow.yaml
```diff
@@ -0,0 +1,36 @@
...
+        'i': "#0000ff"
+        'v': "#ff00ff"
+      map: |
+        roygbiv
+mask: '.'
+palette:
+  'r':
+    structure:
+      name: stripe
+    cell: "#ffffff"
```
byorgey, line 19: What does the `cell: "#ffffff"` mean?

byorgey (approved):

src/swarm-scenario/Swarm/Game/World/Render.hs
xsebek: @kostmo I suppose you could use [`DerivingVia`](https://kowainik.github.io/posts/deriving#via) if you have another similarly shaped instance around. 🤔

byorgey: "Similarly shaped" isn't enough, they have to be `Coercible`.

src/swarm-scenario/Swarm/Game/World/Render.hs
```diff
@@ -38,8 +40,34 @@ import Swarm.Language.Pretty (prettyString)
 import Swarm.Util (surfaceEmpty)
 import Swarm.Util.Content
 import Swarm.Util.Erasable (erasableToMaybe)
+import Swarm.Util.Yaml
 import System.IO (hPutStrLn, stderr)
 
+newtype OneBitColor = OneBitColor Bool
```
xsebek, line 46: @kostmo could you please add an explicit export list, so that it's easier to see if any definitions are internal?

data/test/standalone-topography/rainbow.yaml
```diff
@@ -0,0 +1,36 @@
...
+  ......r.........
+  .......r........
+  ........r.......
+  .........r......
+  ..........r.....
+  ...........r....
+  ............r...
+  .............r..
+  ..............r.
+  ...............r
```
xsebek, line 36: I tried what a curve will do:
```suggestion
  ......rrrrr......
  ....rr.....rr....
  ...r.........r...
  ..r...........r..
  .r.............r.
  r...............r
```
With the structure rotated (manually, `orient.up` is only in placement) it gives this:
image

data/test/standalone-topography/rainbow.yaml
xsebek, line 1: Is there a way to view this in swarm, like we can the testing scenarios?

xsebek (approved):
@kostmo neat, I did not know we have this colorful feature. 👍 

Is this refactoring and adding test? If so could you update the description please?

data/test/standalone-topography/rainbow.yaml
kostmo: In this file, the type of the cells are hexadecimal color codes.  Generically, for "structure placement" palette entries, it is mandatory to specify a default value for the cell in the case that the upper-left coordinate of the structure is itself transparent.  In this case I arbitrarily chose white as the default.

data/test/standalone-topography/rainbow.yaml
kostmo: Within the swarm TUI, the rendering is tightly coupled to the `Entity`-based [`Cell`](https://github.com/swarm-game/swarm/blob/08045793b110f19a0389be476a9f865525c3781f/src/swarm-scenario/Swarm/Game/Scenario/Topography/Cell.hs#L49) type.  Though for the purpose of the [cursor info widget](https://github.com/swarm-game/swarm/blob/08045793b110f19a0389be476a9f865525c3781f/src/swarm-tui/Swarm/TUI/View.hs#L473) we do use [`getDisplay`](https://github.com/swarm-game/swarm/blob/08045793b110f19a0389be476a9f865525c3781f/src/swarm-tui/Swarm/TUI/Editor/Model.hs#L36) with an `EntityFacade`.

I could imagine some web-based tooling for viewing and composing `Entity`-independent structures.

PR #2103, opened 2024-08-09
Update integer list example

xsebek:
* reformat using `swarm format` 
* use `tydef`
* use `format`
* use accumulator functions

### DIFF
--- a/example/list.sw
+++ b/example/list.sw
@@ -1,3 +1,4 @@
+
 /*******************************************************************/
 /*                       LIST ENCODING IN INT                      */
 /*                                                                 */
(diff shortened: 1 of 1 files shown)


xsebek:
@byorgey what do you think of the reformatting? 🙂 I think it is pretty good, though it does merge comments.

example/list.sw
```diff
@@ -50,265 +46,250 @@ vvvvvvvvvvvv       vvvvvvvvvvvvvvvvvvvvvvvvv     vvv
...
+  = \i.
   let nextPart = shiftL 7 i in
-  if (nextPart == 0)
-    { ((2 * i), 1) } /* last part */
-    { let p = to1numA nextPart in
-      ( 1 /* header isEnd bit */ + setLenPart i + shiftRH 1 (fst p)
-      , 1 + snd p
-      )
-    }
+  if (nextPart == 0) {(2 * i, 1 /* last part */)} {
```
byorgey, line 101: Hmm, the reformatting moved the `last part` comment inside some parens...

byorgey (approved):
I think it looks pretty good.

In some sense we don't need this any more since we now have actual lists, but it's still nice as an extended example.  Might be nice to add a comment explaining this.

example/list.sw
xsebek: I moved the comment before the expression, but the formatted inserts an empty line before it.
I guess that is better, but will needs some tweaks if we want to use it for all files.

byorgey: Thinking about this more, I am actually surprised that it does not print `{(2 * i, 1)} /* last part */`.  I thought it was supposed to do that; I will have to think about whether this is a bug.

byorgey: Yeah, getting auto-formatted comments to look good in all situations is surprisingly tough.  But I would be very happy to collect more examples of where it breaks down to see if we can continue tweaking the output.

PR #2104, opened 2024-08-09
Only check for escaped skolem variables among those just generated

### LINKED ISSUE #2101: Strange type checking error "Skolem variable would escape its scope"
**Describe the bug**

Typechecking the following program gives a strange error:
```
Skolem variable s1 would escape its scope
```
because it uses a renamed type variable it is not immediately obvious what this is even about.

**To Reproduce**

```Haskell
def for_each_bit : Int -> Int -> (Int -> Int -> Cmd a) -> Cmd Unit
  = \i. \xs. \act.
  if (xs == 0) {} {
    let ht = (xs - 2 * (xs / 2), xs / 2) in
    act i (fst ht);
    for_each_bit (i + 1) (snd ht) act
  }
end
```

**Expected behavior**

The error should use the original type variable name `a` and a less formal language to explain the issue.

In this case there might not even be a reason for the error and it should typecheck.

**Screenshots**
It is interesting that only a part of the function body does not typecheck:
Screenshot 2024-08-09 at 5 49 28 PM

byorgey:
Fixes #2101.

- Modifies `skolemize` to return the list of generated variable names along with the substituted type (and remove an unnecessary call to `error`!)
- Then in `noSkolems` only check for relevant, recently-generated skolem variables instead of for any skolem variables at all.

### DIFF
--- a/test/unit/TestLanguagePipeline.hs
+++ b/test/unit/TestLanguagePipeline.hs
@@ -656,6 +656,15 @@ testLanguagePipeline =
             "def at cmd type"
             (valid "def x = 3 end; move; return (x+2)")
         ]
+    , testGroup
+        "nested let/def/annot #2101"
+        [ testCase
+            "nested polymorphic def/let"
+            (valid "def id : a -> a = \\y. let x = 3 in y end")
+        , testCase
+            "nested polymorphic def/annot"
+            (valid "def id : a -> a * Int = \\y. (y, 3 : Int) end")
+        ]
     ]
  where
   valid = flip process ""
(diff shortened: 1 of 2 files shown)


test/unit/TestLanguagePipeline.hs
```diff
@@ -656,6 +656,15 @@ testLanguagePipeline =
...
         ]
+    , testGroup
+        "nested let/def/annot #2101"
+        [ testCase
+            "nested polymorphic def/let"
+            (valid "def id : a -> a = \\y. let x = 3 in y end")
+        , testCase
+            "nested polymorphic def/annot"
+            (valid "def id : a -> a * Int = \\y. (y, 3 : Int) end")
+        ]
```
byorgey, line 667: These minimal test cases were failing before the fix.

xsebek (approved):
This looks reasonable to me. 👍 

I am glad that the test examples here will parse, those would surprise me even more than the example I have given.

PR #2108, opened 2024-08-11
Reply to Web API run with result

### LINKED ISSUE #1426: Return real result in Web API `/code/run`
Currently we don't wait for the evaluation to finish:
```
❯ curl --header "Content-Type: text/plain;charset=utf-8" --data "move" localhost:5357/code/run
Sent
```

Getting back the result would make it possible to control Swarm remotely... or at least from different window.

xsebek:
* make Web API `code/run` pass `webReply :: WebInvocationState -> IO ()` to TUI
  * send InProgress or Rejected
  * once REPLDone send Complete
* add remote REPL shell script
* closes #1426

This uses the "IO wrapper" trick from #2098 - thus we can be sure that TUI will be the producer and Web API the consumer. 

Example:
```bash
cabal run swarm -O0 -- --scenario blank  # in another terminal
curl -XGET --header "Content-Type: text/plain;charset=utf-8" localhost:5357/code/run --data "1 + 1"
```
```JSON
{"InProgress":[]}
{"Complete":"it1 : Int = 2"}
```

You can also test it with the shell script, that tries to act as a REPL and strips the JSON unless it gets an error:
```
❯ scripts/remote-repl.sh
> move
> scan down
it0 : Unit + Text = inl ()
> grab // exception written to log is not shown here
> grob
{
  "ParseError": "1:1: Unbound variable grob\n"
}
>
```


kostmo (approved):
Works for me!  Very cool.

For the test procedure, should specify that one must run a "creative" game to obtain the expected output.

The servant "streaming" API is new to me.  Perhaps add a comment to introduce readers to how it is being used here.

byorgey (approved):
Neat!

xsebek:
@kostmo @byorgey I added a big note that tries to explain this streaming thing. I don't have any prior experience with streaming, so please take it with a grain of salt.

It might even be overkill here, but I wanted to show that the program was parsed and the base is running it:
Screenshot 2024-08-15 at 12 34 13 AM

I think this approach would really pay off when streaming logs. 🤔

PR #2110, opened 2024-08-11
[EXPERIMENTAL] new zoom function

kostmo:


src/swarm-engine/Swarm/Game/State.hs
```diff
@@ -490,6 +493,20 @@ contentAt (Cosmic subworldName loc) = do
     return (maybe BlankT terrainName terrObj, maybeEnt)
   return $ fromMaybe (BlankT, Nothing) val
 
+
+-- | Perform an action requiring a 'Robots' state component in a
+--   larger context with a 'GameState'.
+zoomRecognizer ::
```
kostmo, line 499: @byorgey do you know why this `zoomRecognizer` function doesn't work?  I get `Overlapping instances for Control.Effect.Sum.Member` and `Could not deduce ‘Control.Effect.Sum.Member` errors.

I wanted to replace some boilerplate like `discovery . structureRecognition . foundStructures` with just `foundStructures`.  But it seems to be incompatible with the `%=` operator.

As it happens, if I instead define a `zoomDiscovery` function that dives only one level deep into the `GameState` (i.e. `use discovery` instead of `use $ discovery . structureRecognition`), then the callsites work fine with `%=` (though at the callsite I have to specify `structureRecognition . foundStructures` instead of just `foundStructures`).

kostmo: BTW, this is not urgent---I've found an alternative refactoring that works.

byorgey:
For posterity, the errors both had to do with the fact that GHC could not infer one or both type parameters to `StructureRecognizer`.  I expect both could have been solved with some kind of explicit `TypeApplication`, though off the top of my head I'm not sure exactly what that would have looked like in these cases.

PR #2112, opened 2024-08-12
More generic structure recognition

kostmo:
Builds upon #1836.

Most importantly in this PR, the `Swarm.Game.Scenario.Topography.Structure.Recognition.Tracking` module is made generic in its `Entity` parameter, and is now able to be moved from the `swarm-engine` sublibrary to the `swarm-topology` sublibrary.

I've also introduced an intermediate `RecognitionState` record inside `StructureRecognizer` to distinguish between the stateful and read-only elements.

The `AhoCorasick` dependency is now reduced to only one sublibrary.

### DIFF
--- a/src/swarm-engine/Swarm/Game/State/Substate.hs
+++ b/src/swarm-engine/Swarm/Game/State/Substate.hs
@@ -429,8 +429,9 @@ initDiscovery =
       -- since the master list of achievements is stored in UIState
       _gameAchievements = mempty
     , _structureRecognition =
-        StructureRecognizer (RecognizerAutomatons mempty mempty) $
-          RecognitionState emptyFoundStructures []
+        StructureRecognizer
+          (RecognizerAutomatons mempty mempty)
+          (RecognitionState emptyFoundStructures [])
     , _tagMembers = mempty
     }
 
--- a/src/swarm-engine/Swarm/Game/State/Substate.hs
+++ b/src/swarm-engine/Swarm/Game/State/Substate.hs
@@ -428,7 +428,9 @@ initDiscovery =
     , -- This does not need to be initialized with anything,
       -- since the master list of achievements is stored in UIState
       _gameAchievements = mempty
-    , _structureRecognition = StructureRecognizer (RecognizerAutomatons mempty mempty) emptyFoundStructures []
+    , _structureRecognition =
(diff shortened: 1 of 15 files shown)


src/swarm-engine/Swarm/Game/State/Substate.hs
```diff
@@ -428,7 +428,9 @@ initDiscovery =
     , -- This does not need to be initialized with anything,
       -- since the master list of achievements is stored in UIState
       _gameAchievements = mempty
-    , _structureRecognition = StructureRecognizer (RecognizerAutomatons mempty mempty) emptyFoundStructures []
+    , _structureRecognition =
+        StructureRecognizer (RecognizerAutomatons mempty mempty) $
+          RecognitionState emptyFoundStructures []
```
byorgey, line 433: Personally I find the `$` more confusing than helpful here.  It's much easier to see what's going on (IMHO) with more symmetric code like
```suggestion
        StructureRecognizer 
          (RecognizerAutomatons mempty mempty)
          (RecognitionState emptyFoundStructures [])
```
But this is just me nitpicking about which shade of blue to use on the bikeshed, so feel free to ignore.

src/swarm-topography/Swarm/Game/Scenario/Topography/Structure/Recognition/Registry.hs
```diff
@@ -72,8 +72,9 @@ removeStructure fs (FoundRegistry byName byLoc) =
   -- Swarm.Game.State.removeRobotFromLocationMap
   tidyDelete = NEM.nonEmptyMap . NEM.delete upperLeft
 
-addFound :: FoundStructure b a -> FoundRegistry b a -> FoundRegistry b a
-addFound fs@(FoundStructure swg loc) (FoundRegistry byName byLoc) =
+addFound :: Maybe (FoundStructure b a) -> FoundRegistry b a -> FoundRegistry b a
+addFound Nothing x = x
```
byorgey, line 76: Might be worth putting `addFound` back to the way it was (without the `Maybe`), then we can express the new `addFound`  as `maybe id addFound` (which could be inlined into the use site, or defined as a separate function).

byorgey (approved):

PR #2119, opened 2024-08-15
Do not show goal when ScenarioEnd is shown

### LINKED ISSUE #2117: Tutorial levels no longer prompt to start the next level upon goal completion
It used to be that the "Congratulations!" dialog popped up, with the "Next challenge!" button preselected, as soon as the player completes the tutorial goals.

Now, however, the "Goal" dialog (which the user has already seen) pops up instead.

I think it is important to provide a sense of continuity between tutorial stages by automatically prompting the user to continue to the next one.  As it is now, the user needs to somehow know to press <kbd>CTRL</kbd>+<kbd>q</kbd> to advance to the next tutorial stage.

xsebek:
* check for open quit game modal dialog and do not overwrite it with goal modal
* refactor the win condition path to make it easier to see
* closes #2117


byorgey (approved):
LGTM!

Issue #2109, opened 2024-08-11
Fix cabal check warnings


xsebek:
There are a few `cabal check` warnings shown in CI, that we should probably fix:
```
These warnings may cause trouble when distributing the package:
Warning: [doc-place] Please consider moving the file 'CHANGELOG.md' from the
'extra-source-files' section of the .cabal file to the section
'extra-doc-files'.
Warning: [missing-upper-bounds] On library 'swarm-engine', these packages miss
upper bounds:
- bytestring
- githash
- warp
Please add them. There is more information at https://pvp.haskell.org/
Warning: [missing-upper-bounds] On library 'swarm-lang', these packages miss
upper bounds:
- megaparsec
- parser-combinators
- prettyprinter
- split
- vty
Please add them. There is more information at https://pvp.haskell.org/
Warning: [missing-upper-bounds] On library 'swarm-topography', these packages
miss upper bounds:
- array
- unordered-containers
Please add them. There is more information at https://pvp.haskell.org/
Warning: [missing-upper-bounds] On library 'swarm-tournament', these packages
miss upper bounds:
- commonmark
- cookie
- exceptions
- servant-multipart
- utf8-string
- wai-extra
Please add them. There is more information at https://pvp.haskell.org/
Warning: [missing-upper-bounds] On library 'swarm-web', these packages miss
upper bounds:
- brick
- colour
- commonmark
- palette
Please add them. There is more information at https://pvp.haskell.org/
Warning: [missing-upper-bounds] On executable 'swarm', these packages miss
upper bounds:
- brick
- vty
Please add them. There is more information at https://pvp.haskell.org/
Warning: [missing-upper-bounds] On executable 'swarm-host-tournament', these
packages miss upper bounds:
- sqlite-simple
Please add them. There is more information at https://pvp.haskell.org/
```

byorgey:
Some of these might be spurious --- if we have upper bounds on a package in one place then we shouldn't have to add upper bounds every single place we require the same package.

xsebek:
@byorgey after a closer look, I do not think that is the case - that would have produced _a lot_ more warnings.

Rather it seems to be caused by other executables that do not depend on all the libraries.

xsebek:
Closed by #2132.

PR #2120, opened 2024-08-15
Remove Brick dependency from Web

xsebek:
* use the IO wrapper trick again for fun and profit :tm:
* part of #2109


kostmo (approved):
:heart_eyes:

PR #2126, opened 2024-08-18
exterminator scenario

kostmo:
```
scripts/play.sh --scenario data/scenarios/Challenges/Ranching/pied-piper.yaml --autoplay
```

In addition to the scenario, this PR:
1. Adds support for colorizing robot names
2. Improves diagnostic message for subworld portal spatial consistency enforcement

### DIFF
--- a/data/scenarios/Challenges/Ranching/00-ORDER.txt
+++ b/data/scenarios/Challenges/Ranching/00-ORDER.txt
@@ -3,3 +3,4 @@ capture.yaml
 powerset.yaml
 fishing.yaml
 gated-paddock.yaml
+pied-piper.yaml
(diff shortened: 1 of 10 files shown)


byorgey (approved):
This is great!

I love watching the "intended" solution which leads all the rats into the silo cellar and then kills them with mold.  However, I solved it in a much simpler way: after obtaining some mold, I grew a bunch in the silo and harvested it.  Then, I went to each door and placed a single mold right outside the door, and a single oats behind the mold.  Just as I hoped, the rats were attracted to the oats but died as soon as they exited the door and walked on the mold.  I don't particularly care --- I think the scenario is fun as it is --- but I'll leave it up to you whether you want to disallow such "unintended" solutions.

kostmo:
> I went to each door and placed a single mold right outside the door

Good find.  I believe I've closed this loophole now.

byorgey:
> I believe I've closed this loophole now.

Yes, it seems to work as intended now!

PR #2130, opened 2024-09-01
Refactor App using standalone functions

xsebek:
* Make `appMain` fit on the screen.
* Replace state `s1` and `s2` versions with `modifyIORef`.

### DIFF
--- a/app/game/Swarm/App.hs
+++ b/app/game/Swarm/App.hs
@@ -9,19 +9,27 @@
 -- Description: Application entry point
 --
 -- Main entry point for the Swarm application.
-module Swarm.App where
+module Swarm.App (
+  app,
+  appMain,
+  EventHandler,
+
+  -- * Demo web
+  demoWeb,
+) where
 
 import Brick
 import Brick.BChan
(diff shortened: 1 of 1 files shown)


app/game/Swarm/App.hs
```diff
@@ -47,7 +56,7 @@ app eventHandler =
     { appDraw = drawUI
     , appChooseCursor = chooseCursor
     , appHandleEvent = eventHandler
-    , appStartEvent = enablePasteMode
+    , appStartEvent = pure ()
```
byorgey, line 59: Why was `enablePasteMode` removed?

xsebek: I moved it to `buildVty` - I don't see why you would not want the paste mode on reinitialize, and we already have a similar mouse option set there.

When does the mode come into play? Currently, we only have a one-line REPL.

xsebek: Ah, it allows us to strip newlines and paste the whole definition into one line.

So it works with the refactor. 🙂

byorgey (approved):

Issue #1421, opened 2023-08-10
Update uses of `foldl'` to prepare for export from `Prelude`


byorgey:
See https://discourse.haskell.org/t/prelude-foldl-migration-guide/6950 .

byorgey:
I started on this a bit today but I think it will be better to just wait until we update to support GHC 9.10 (which currently requires building with `--allow-newer`).  At that time we will *have* to update in order to get rid of new warnings.

PR #2131, opened 2024-09-01
Allow GHC 9.10

xsebek:
* add GHC 9.10 to tested
* regenerate CI
* allow `base` 4.20

---

- blocked by `palette` package - diagrams/palette/pull/9

### DIFF
--- a/swarm.cabal
+++ b/swarm.cabal
@@ -38,7 +38,7 @@ maintainer: byorgey@gmail.com
 bug-reports: https://github.com/swarm-game/swarm/issues
 copyright: Brent Yorgey 2021
 category: Game
-tested-with: ghc ==9.2.8 || ==9.4.8 || ==9.6.5 || ==9.8.2
+tested-with: ghc ==9.2.8 || ==9.4.8 || ==9.6.5 || ==9.8.2 || ==9.10.1
 extra-doc-files:
   CHANGELOG.md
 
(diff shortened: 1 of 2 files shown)


swarm.cabal
```diff
@@ -38,7 +38,7 @@ maintainer: byorgey@gmail.com
 bug-reports: https://github.com/swarm-game/swarm/issues
 copyright: Brent Yorgey 2021
 category: Game
-tested-with: ghc ==9.2.8 || ==9.4.8 || ==9.6.5 || ==9.8.2
+tested-with: ghc ==9.2.8 || ==9.4.8 || ==9.6.5 || ==9.8.2 || ==9.10.1
```
nitinprakash96, line 41: Following up with https://github.com/swarm-game/swarm/pull/1996
We can probably drop GHC <= 9.4

xsebek: I think we should wait with GHC 9.4 since GHCup recommends it. But we could drop 9.2.

xsebek:
Note that more dependencies might need their bounds to be relaxed. It all builds and works with `--allow-newer`, but I would probably have to clean my cache to see if there are more issues.

byorgey:
> I am starting to like the relaxed `base <5` upper bound that avoids this frequent update.

In general, the problem with `base <5` is that if it *does* break, it does so silently, and anyone trying to build it will spend a half hour building all the dependencies before it finally fails, instead of having it fail immediately.  On the other hand, `base` updates rarely break anything.

I am not sure it matters so much here, since we still have to do the same amount of work updating of other dependencies regardless.

byorgey:
@xsebek can we revive this PR (or should we just close it and open a new one)?  `palette` is updated now, I did a local test build with GHC 9.10 and everything seems to work fine.  We may need to address #1421 as well.

xsebek:
Let's open a new one. 👍

byorgey:
See #2233 .

PR #2132, opened 2024-09-01
Make cabal check happy with dependency bounds

xsebek:
* fix ™️ the duplicate version bound problem once and for all using **common stanzas**
  * unify different repeated version bounds
  * add missing bounds (`swarm-tournament`)


kostmo (approved):

xsebek:
@kostmo @byorgey I moved the version bounds to common stanzas.

Please take a look if you are OK with this version.

xsebek:
To be clear, it is not my idea. 🙂  It was [suggested by people on Haskell Discourse](https://discourse.haskell.org/t/best-practices-for-public-cabal-sublibraries/10272/2) and is used by other packages:

https://github.com/haskell-works/cabal-cache/blob/main/cabal-cache.cabal

Issue #2138, opened 2024-09-05
Make robots selectable in F2 dialog and show details


kostmo:
Currently, the F2 robots dialog is non-interactive.

Since there is a lot of information we could show per robot, let's make the widget navigable and present full details for a single robot when selected.

This may subsume parts of #1341.

PR #2139, opened 2024-09-05
Group modal dialog states into a record

kostmo:
A refactoring towards #2138

A new field for the robots dialog will be added to the `UIDialogs` record next.

Also, move F2 robots dialog rendering code to a new module `Swarm.TUI.View.Robot`.


src/swarm-tui/Swarm/TUI/Controller.hs
```diff
@@ -96,12 +96,12 @@ import Swarm.TUI.Launch.Model
 import Swarm.TUI.Launch.Prep (prepareLaunchDialog)
 import Swarm.TUI.List
 import Swarm.TUI.Model
-import Swarm.TUI.Model.Goal
+import Swarm.TUI.Model.Dialog.Goal
+import Swarm.TUI.Model.Dialog.Popup (progressPopups)
+import Swarm.TUI.Model.Dialog.Structure
```
xsebek, line 101: It would be nice if these were re-exported from a common Dialog module.

xsebek (approved):
Thanks @kostmo! Every line moved from `View.hs` counts. 😄

Issue #2133, opened 2024-09-01
Add `debug` primitive providing current source location + enclosing function


kostmo:
While developing a scenario, it would be useful for debugging to be able to see which function a system robot is currently "in" as it executes its program.

Sometimes to achieve this I have placed a `say "function_name";` at the top of each function in a robot's program (i.e. "printf debugging") to accomplish this.  But it would be nice to have an automatic mechanism.

I imagine it is feasible:
* At program parse time, annotate each `Term` with the closest "function definition" that contains it.
* At runtime, inspect the top frame of the robot's `machine` and display the function name for the currently focused term in the F2 dialog.

kostmo:
> every time we look up a variable in order to evaluate it

Is that done on [this line](https://github.com/swarm-game/swarm/blob/589e2665b5005578a53f140fb27d0b5c991b22fe/src/swarm-engine/Swarm/Game/Step.hs#L579C7-L579C11)?

> we can set the "current function" to the name of the variable

I'd want to distinugish between variables that refer to "toplevel function definitions" vs. variables that refer to local definitions.  Is that distinction tracked in the `Term` type?

byorgey:
> Is that done on [this line](https://github.com/swarm-game/swarm/blob/589e2665b5005578a53f140fb27d0b5c991b22fe/src/swarm-engine/Swarm/Game/Step.hs#L579C7-L579C11)?

Hmm, yes, that's the line I had in mind, but now that I think about it, this won't work.  When we look up a variable in the environment, it is already evaluated; at some later point we may execute it (if it is a command) or reduce it (if it is a lambda that ends up getting applied to an argument) but in any case that happens at some later point unrelated to when the variable lookup happened.

Maybe annotating all terms would actually be the right way to accomplish this...

Regardless of how we accomplish it, I do worry that this feature might be difficult to maintain if we start doing more code optimization, see *e.g.* #1563 .

> I'd want to distinugish between variables that refer to "toplevel function definitions" vs. variables that refer to local definitions. Is that distinction tracked in the `Term` type?

It is not, but I don't think it would be too hard to enhance `Env` so it tracks additional information about each variable.

kostmo:
IIRC, the `Syntax'` type is a wrapper of the `Term` type, and is discarded once we are running the CESK `machine`?  If that were not the case, then at dialog display time we could infer the toplevel function name from the `SrcLoc` of the term.

byorgey:
[...]

PR #2140, opened 2024-09-06
navigable robots table

kostmo:
Towards #1341.  Should also be useful for #2133.

Screenshot from 2024-09-09 22-04-10

Uses the [`brick-tabular-list`](https://hackage.haskell.org/package/brick-tabular-list) package to render the <kbd>F2</kbd> robots dialog as a navigable list with tabular formatting.  Hitting <kbd>Tab</kbd> on a selected row shows details for that robot.

Also:
* Extracts some record definitions from `Swarm.TUI.Model.UI` into `Swarm.TUI.Model.UI.Gameplay`
* Removes re-export of the `Name` type from `Swarm.TUI.Model`
* Replace uses of `maximum` with the safer `maximum0`
* New `applyJust` combinator

### Showing a large robots list
```
scripts/play.sh -i data/scenarios/Challenges/Ranching/beekeeping.yaml --debug all_robots --speed 2 --autoplay
```
and with extra column:
```
scripts/play.sh -i data/scenarios/Challenges/Ranching/beekeeping.yaml --debug all_robots,robot_id --speed 2 --autoplay
```

### Showing a small robots list with details view and logs

```
scripts/play.sh -i data/scenarios/Testing/562-lodestone.yaml --debug all_robots,robot_id --speed 2 --autoplay
```

Log view:
Screenshot from 2024-09-13 17-25-23


src/swarm-topography/Swarm/Game/Universe.hs
```diff
@@ -82,3 +83,15 @@ defaultCosmicLocation = Cosmic DefaultRootSubworld origin
 
 offsetBy :: Cosmic Location -> V2 Int32 -> Cosmic Location
 offsetBy loc v = fmap (.+^ v) loc
+
```
kostmo, line 86: Add a "rendering" heading

src/swarm-tui/Swarm/TUI/Controller.hs
```diff
@@ -418,6 +420,11 @@ handleModalEvent = \case
                 refreshList $ uiState . uiGameplay . uiDialogs . uiStructure . structurePanelListWidget
               StructureSummary -> handleInfoPanelEvent modalScroll (VtyEvent ev)
             _ -> handleInfoPanelEvent modalScroll (VtyEvent ev)
+      Just RobotsModal -> case ev of
```
kostmo, line 423: Use a shared `zoom` into `uiState . uiGameplay . uiDialogs . uiRobot `

src/swarm-tui/Swarm/TUI/Controller/UpdateUI.hs
```diff
@@ -13,7 +13,9 @@ import Brick hiding (Direction, Location)
 import Brick.Focus
 
 -- See Note [liftA2 re-export from Prelude]
+
```
kostmo, line 16: rm

src/swarm-tui/Swarm/TUI/Controller/UpdateUI.hs
```diff
@@ -174,6 +180,28 @@ updateUI = do
           || newPopups
   pure redraw
 
+doRobotListUpdate :: GameState -> EventM Name AppState ()
+doRobotListUpdate g = do
+  gp <- use $ uiState . uiGameplay
+  dOps <- use $ uiState . uiDebugOptions
+
+  let rd =
```
kostmo, line 188: Combine `let` blocks

src/swarm-tui/Swarm/TUI/View.hs
```diff
@@ -607,14 +610,22 @@ drawDialog :: AppState -> Widget Name
...
   Nothing -> emptyWidget
 
 -- | Draw one of the various types of modal dialog.
 drawModal :: AppState -> ModalType -> Widget Name
 drawModal s = \case
   HelpModal -> helpWidget (s ^. gameState . randomness . seed) (s ^. runtimeState . webPort) (s ^. keyEventHandling)
-  RobotsModal -> robotsListWidget s
+  RobotsModal -> case s ^. uiState . uiGameplay . uiDialogs . uiRobot . robotsDisplayMode of
+    RobotList -> renderTheRobots $ s ^. uiState . uiGameplay . uiDialogs . uiRobot . robotListContent
+    SingleRobotDetails -> case maybeSelectedRID of
```
kostmo, line 623: TODO: Finish implementing details view

src/swarm-tui/Swarm/TUI/View/Robot.hs
```diff
@@ -28,17 +39,159 @@ import Swarm.Game.State.Substate
...
+  , _uiDbg :: Set DebugOption
+  }
+
+makeLenses ''RobotRenderingContext
+
+mkRobotDisplay :: RobotRenderingContext -> RobotListContent
+mkRobotDisplay c =
+  RobotListContent
+    { _libList = mixedTabularList RobotsList (mkLibraryEntries c) (LstItmH 1) (wprk uiDebug) wpr
+    , _libRenderers =
```
kostmo, line 69: Don't re-initialize the `_libRenderers` field, since it's constant with respect to all rendering context other than `uiDbg`

src/swarm-tui/Swarm/TUI/View/Robot.hs
```diff
@@ -28,17 +39,159 @@ import Swarm.Game.State.Substate
...
+  , _fStatus
+  , _fActns
+  , _fCmds
+  , _fCycles
+  , _fActivity
+  , _fLog
+  ]
+
+dc :: Set DebugOption -> ListFocused -> MixedCtxt -> RobotWidgetRow -> Widget Name
+dc uiDebug _ (MxdCtxt _ (MColC (Ix ci))) r =
```
kostmo, line 173: Improve naming of `dc`, `wpr`, wprk`, etc.

src/swarm-tui/Swarm/TUI/View/Robot.hs
```diff
@@ -28,17 +39,159 @@ import Swarm.Game.State.Substate
...
+  maybe emptyWidget (renderPlainCell . wWidget . ($ rPayload r)) (indexedAccessors V.!? ci)
+ where
+  indexedAccessors = V.fromList accessors
+  accessors = dropFirstColumn uiDebug accessorList
```
kostmo, line 177: Conslidate uses of the pattern `dropFirstColumn uiDebug accessorList`

src/swarm-tui/Swarm/TUI/View/Robot.hs
```diff
@@ -28,17 +39,159 @@ import Swarm.Game.State.Substate
...
+  renderPlainCell = padRight Max
+
+-- | For a single-constructor datatype like 'RobotWidgetRow',
+-- this implementation is trivial.
+wpr :: WidthsPerRow RobotWidgetRow Widths
+wpr = WsPerR $ \(Widths x) _ -> x
+
+wprk :: Set DebugOption -> WidthsPerRowKind RobotWidgetRow Widths
```
kostmo, line 185: Explain how this works: we take the maximum of all cell widths, **including the headers**, and add `1` for "padding".

src/swarm-tui/Swarm/TUI/View/Robot.hs
```diff
@@ -125,30 +270,38 @@ robotsListWidget s = hCenter table
...
-      Waiting {} -> txt "waiting"
+      Waiting {} ->
+        let tx = "waiting"
+         in WidthWidget (length tx) (str tx)
       _
-        | isActive robot -> withAttr notifAttr $ txt "busy"
-        | otherwise -> withAttr greenAttr $ txt "idle"
+        | isActive robot ->
+            let tx = "busy"
+             in WidthWidget (length tx) (withAttr notifAttr $ str tx)
```
kostmo, line 301: `WidthWidget` should be a `Functor` in the second field.  Then should re-use the local `strWidget` function here.

src/swarm-tui/Swarm/TUI/View/RobotDisplay.hs
```diff
@@ -0,0 +1,72 @@
...
+import Swarm.TUI.Model.Name
+
+data RobotRowPayload a = RobotRowPayload
+  { robID :: RID
+  , rPayload :: LibRobotRow a
+  }
+  deriving (Functor)
+
+data WidthWidget = WidthWidget
+  { wWidth :: Int
```
kostmo, line 22: Explain why we precompute the width

src/swarm-tui/Swarm/TUI/View/Robot.hs
```diff
@@ -28,17 +42,223 @@ import Swarm.Game.State.Substate
...
+                else withAttr rowHdrAttr
+         in attrFn $ padRight (Pad $ if wd > 0 then 0 else 1) $ padLeft Max (str $ show rh)
+    , width = \_ rh -> RowHdrW . (+ 2) . maximum0 $ map (length . show) rh
+    , toRH = \_ (Ix i) -> i + 1
+    }
+
+data ColumnExpansion
+  = Grow
+  | Minimal
+  deriving (Eq)
```
xsebek, line 138: Why not use [`Brick.Types.Size`](https://hackage.haskell.org/package/brick-2.4/docs/Brick-Types.html#t:Size)?

src/swarm-tui/Swarm/TUI/View/Robot.hs
```diff
@@ -28,17 +42,223 @@ import Swarm.Game.State.Substate
...
+         , _fActns
+         , _fCmds
+         , _fCycles
+         , _fActivity
+         , _fLog
+         ]
+
+drawCell :: Set DebugOption -> ListFocused -> MixedCtxt -> RobotWidgetRow -> Widget Name
+drawCell uiDebug _ (MxdCtxt _ (MColC (Ix ci))) r =
+  maybe emptyWidget (renderPlainCell . wWidget . ($ view rPayload r)) (indexedAccessors V.!? ci)
```
xsebek, line 192: What happens if the list is empty?

Does the tabular list require nonempty, or is that a self-imposed restriction?

xsebek (changes requested):
This is neat. 👍 

I have some nitpicks about the refactoring. It would have been nicer if that was a separate Pull Request, so that the changes here were easier to see.

The view seems to work well, but I was initially confused by the controls. It would be nice if there was a hint in UI about which keyboard shortcuts are used to control it.

src/swarm-tui/Swarm/TUI/View/Robot.hs
kostmo: The `brick-tabular-list` package does not require nonempty column list, but I establish the invariant from the definition of `getAccessorList`.  That said, it probably would make sense for `brick-tabular-list` to have a `NonEmpty` restriction, since a table with zero columns doesn't make sense.

src/swarm-tui/Swarm/TUI/View/Robot.hs
kostmo: Added a docstring.

byorgey (commented):
Cool!

I agree the controls seem a bit confusing.  To be consistent with the rest of the UI, I think Tab should only switch between focus within a single panel, not change which panel is displayed.  I suggest having Enter on a highlighted robot open the robot details panel (instead of Tab), then Tab within the robot details panel just switches between the different things to focus in that panel; Esc will close it and return to the list of all robots.

src/swarm-util/Swarm/Util.hs
```diff
@@ -273,6 +280,10 @@ applyWhen :: Bool -> (a -> a) -> a -> a
 applyWhen True f x = f x
 applyWhen False _ x = x
 
+applyJust :: Maybe (a -> a) -> a -> a
+applyJust Nothing x = x
+applyJust (Just f) x = f x
```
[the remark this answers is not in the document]
xsebek: Thanks, I was also intrigued by this helper function. 

Is any variant of this function available in our dependencies? 🤔 The only hoogle hit is [`thread`](https://hackage.haskell.org/package/composition-prelude-3.0.0.2/docs/Control-Composition.html#v:thread) from some alternative prelude.

src/swarm-tui/Swarm/TUI/View/Shared.hs
```diff
@@ -0,0 +1,15 @@
...
+-- UI view components shared across dialogs
+module Swarm.TUI.View.Shared where
+
+import Brick
+import Brick.Widgets.Center (hCenter)
+import Swarm.TUI.Model.Name
+import Swarm.TUI.View.Attribute.Attr (italicAttr)
+
+tabControlFooter :: Widget Name
+tabControlFooter = hCenter $ withAttr italicAttr $ txt "NOTE: [Tab] toggles focus between panes"
```
xsebek, line 15: If you make this polymorphic in the `Name` it could go to `Swarm.TUI.View.Util`, I think.

But maybe we will need a common set of shared view parts anyway.

src/swarm-tui/Swarm/TUI/Model/UI.hs
```diff
@@ -8,52 +8,14 @@
 -- SPDX-License-Identifier: BSD-3-Clause
 module Swarm.TUI.Model.UI (
```
xsebek, line 9: We don't follow this consistently (yet), but it might be nice to re-export the whole `Swarm.TUI.Model.UI.Gameplay` here, so that there are fewer changes in the imports.

It's a drop of water in the ocean though, so feel free to ignore this suggestion.

xsebek (commented):
Thanks @kostmo for going through my comments. 👍 

I like @byorgey's suggestion for the controls, I kept hitting Enter and closing the modal.

kostmo:
> Tab should only switch between focus within a single panel, not change which panel is displayed. I suggest having <kbd>Enter</kbd> on a highlighted robot open the robot details panel (instead of Tab), then <kbd>Tab</kbd> within the robot details panel just switches between the different things to focus in that panel; <kbd>Esc</kbd> will close it and return to the list of all robots.

I have now updated the controls to behave as described.  It feels a bit hacky though---perhaps at some point we would introduce the concept of a "stack" of dialogs so that things like the <kbd>Esc</kbd> would simply pop the top dialog from the stack, rather than special-casing logic for it.

xsebek (approved):
Works nicely, thanks @kostmo. 👍

kostmo:
This also closes #2138.

PR #2143, opened 2024-09-10
cache scenario preview widget

kostmo:
Fixes the lag issue noted in #1515.

# Demo

1. First, launch a toplevel scenario:
    ```
    scripts/play.sh -i data/scenarios/blank.yaml
    ```
2. Then use <kbd>Ctrl</kbd>+<kbd>q</kbd> to quit to the scenario selection menu.
3. Notice when using up-arrow to select the "Classic" and "Creative" scenarios, there is a delay as the preview is computed for the first time.
4. Finally, when subsequently selecting back and forth between the two again, notice that the lag is gone, as the cache is being used.

### DIFF
--- a/src/swarm-tui/Swarm/TUI/Model/Name.hs
+++ b/src/swarm-tui/Swarm/TUI/Model/Name.hs
@@ -92,6 +92,8 @@ data Name
     InventoryList
   | -- | The inventory item position in the InventoryList.
     InventoryListItem Int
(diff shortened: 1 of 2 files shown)


byorgey (approved):
That was... a lot easier than I thought it would be! :+1:

xsebek (approved):
That is a noticeable difference. 👍

PR #2147, opened 2024-09-15
Make Rolex show time

xsebek:
* show time when robot has `time` capability

Prompted by #2146, which used the fancy `rolex` entity for commands but surprisingly did not show time in UI.

You can test it with:
```sh
git cherry-pick ba8dae31  # see before and after adding rolex
cabal run swarm -O0 -- -i data/scenarios/Testing/562-lodestone.yaml --speed 2 --autoplay
```


kostmo (approved):

src/swarm-tui/Swarm/TUI/View.hs
```diff
@@ -538,14 +538,12 @@ drawClockDisplay lgTPS gs = hBox . intersperse (txt " ") $ catMaybes [clockWidge
...
--- | Check whether the currently focused robot (if any) has a clock
---   device equipped.
+-- | Check whether the currently focused robot (if any) has some kind
+-- of a clock device equipped.
 clockEquipped :: GameState -> Bool
 clockEquipped gs = case focusedRobot gs of
   Nothing -> False
-  Just r
-    | countByName "clock" (r ^. equippedDevices) > 0 -> True
-    | otherwise -> False
```
xsebek, line 548: This might be the only instance of `if p then True else False` in repo.

I am surprised HLint did not spot it.

PR #2153, opened 2024-09-23
use wall rendering

kostmo:
Makes use of #1285 to simplify wall rendering in several scenarios.

### DIFF
--- a/data/scenarios/Challenges/Sokoban/foresight.yaml
+++ b/data/scenarios/Challenges/Sokoban/foresight.yaml
@@ -92,7 +92,7 @@ entities:
       char: '#'
     description:
       - Unmovable barrier
-    properties: [known, unwalkable]
+    properties: [known, unwalkable, boundary]
 known: [mountain, water, 3D printer, flower]
 world:
   dsl: |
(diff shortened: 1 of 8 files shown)


xsebek (approved):
Thanks @kostmo, maybe this should be part of a tracking issue to get rid of old walls?

data/scenarios/Challenges/Sokoban/foresight.yaml
```diff
@@ -92,7 +92,7 @@ entities:
       char: '#'
     description:
       - Unmovable barrier
-    properties: [known, unwalkable]
+    properties: [known, unwalkable, boundary]
```
byorgey, line 95: I think this one will probably look strange with connected walls, won't it?

byorgey (approved):

data/scenarios/Challenges/Sokoban/foresight.yaml
kostmo: I don't mind the aesthetic, but it does look a bit like random ladders/staircases, which could be a bit misleading.
Maybe at some point I'll consider a more sophisticated renderer that eliminates interior boundaries.

image

byorgey: It's an interesting challenge to specify how such a more sophisticated renderer should work.  Like, I think I have an intuitive idea of what I think the result should look like (though it may not be the same as yours), but I think there is some interesting graph theory needed to compute the right drawing.  I will see if I can give it some thought.

PR #2160, opened 2024-09-29
Use the new wall in move tutorial

xsebek:
* add the new boundary wall to `entities.yaml`
* use the new boundary wall to simplify the scenario code
  * the door visual is achieved by using an invisible walkable boundary entity called "door"
* use instant wait
* directly check the base position in objective conditions 
* use one-line code snippets in the markdown description so that they are rendered on a single line


data/entities.yaml
```diff
@@ -635,6 +635,17 @@
...
 ############################################################
+- name: wall
+  display:
+    char: '+'
+    attr: entity
+  description:
+    - A wall
+  properties: [unwalkable, known, boundary]
+# following are "legacy" walls that do not connect automatically,
+# but require explicitly plaing each one - this can be used for
```
kostmo, line 646: typo
```suggestion
# but require explicitly placing each one - this can be used for
```

kostmo (approved):

byorgey (approved):

data/entities.yaml
xsebek: Fixed, thanks. 👍

data/scenarios/Tutorials/move.yaml
```diff
@@ -6,59 +6,68 @@ objectives:
...
         Previously you could move twice by chaining the move command:
-      - |
-        ```
-        move; move
-        ```
-      - To reuse that command without having to retype it press the upward arrow on your keyboard. This will allow you to select previous commands.
-      - Ahead of you is a six steps long corridor. Move to its end, i.e. the coordinates `(8,0)` marked with the second purple `flower`{=entity}.
-      - You can open this popup window at any time to remind yourself of the goal using **Ctrl+G**.
+
+        `move; move`{=snippet}
```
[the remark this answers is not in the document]
xsebek: Ah, I forgot how this works, thanks. 👍

PR #2166, opened 2024-10-11
Add github action for Restyled

kostmo:
Followed these instructions:
https://docs.restyled.io/docs/migrating-to-github-actions/


kostmo:
Verified that the new action works: #2167

xsebek (approved):
Thanks for the quick fix! 👍

PR #2170, opened 2024-10-12
Fix TyDef narrowToPosition

### LINKED ISSUE #2169: Type definition breaks LSP hover
**Describe the bug**

A type definition breaks LSP hover.

**To Reproduce**

Create a file `color.sw`:

```c
tydef Color = Bool end
def black = false end
def red = true end  // hover on red or true
```
any hover produces just:
```
tydef Color = Bool end

def black = false end

def red = true end

A type synonym definition.
```
But without the `tydef`:
```
red: Bool
```
```
true: Bool
A boolean literal.
```

**Expected behavior**

LSP should descend into `tydef`.

**Screenshots**

Screenshot 2024-10-12 at 1 54 07 PM

**Additional context**

Discovered in #2161.

xsebek:
`STyDef` is not actually an atom and we need to recurse further.

| Recursing | TyDef |
|--------|--------|
| false | tydef | 

Closes #2169.


byorgey (approved):
Aha, good catch.  It *used to be* an atom, but that changed in #1928 .

PR #2172, opened 2024-10-12
Improve type inference for record projection

byorgey:
When encountering a record projection `r.x`, we have to make sure that `r` has a record type.  Previously, if the type of `r` was not manifestly a record type, we simply emitted an error message like "Can't infer the type of a record projection: r.x".  This PR improves the situation in three related ways:

- First, we now expand the type of `r`, if it turns out to be a type alias or recursive type.  This was simply a bug: something like `tydef R = [x:Int] end; def f : R -> Int = \r. r.x` now works, as it should, but previously it was a type error.
- Second, if the type of `r` is provably *not* a record type, we generate an appropriate type mismatch error rather than an unhelpful "Can't infer" error.
- Finally, the "Can't infer" message itself is improved: it now says `In the record projection r.x, can't infer whether the LHS has a record type.  Try adding a type annotation.`.

### DIFF
--- a/src/swarm-lang/Swarm/Language/Typecheck.hs
+++ b/src/swarm-lang/Swarm/Language/Typecheck.hs
@@ -53,7 +53,7 @@ import Control.Effect.Catch (Catch, catchError)
 import Control.Effect.Error (Error)
 import Control.Effect.Reader
 import Control.Effect.Throw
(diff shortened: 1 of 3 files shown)


xsebek (approved):
Thanks for the quick fix! 👍

PR #2178, opened 2024-10-14
Scoped type variables

### LINKED ISSUE #2168: Make type variables scoped

When writing a complex function, it is helpful to use local type annotations. But this breaks when I use local variables:
```c
def id : forall x. x -> x = \x.
 let f : x = x in f  // From context, expected `x` to have type `s2`, but it actually has type `s1`
end
```

This is the same thing as Haskell without [`ScopedTypeVariables`](https://downloads.haskell.org/ghc/latest/docs/users_guide/exts/scoped_type_variables.html#):
```Haskell
idx :: forall x. x -> x
idx x = let y :: x
            y = x
        in y  -- Couldn't match expected type ‘x1’ with actual type ‘x’ ...
```

I would like to have scoped type variables by default.

It's possible that this would cause problems for later definitions if they were technically inside the scope. An alternative would be using a type annotation inside the body, like the Haskell example:
```Haskell
f :: [a] -> [a]
f (xs :: [aa]) = xs ++ ys
  where
    ys :: [aa]
    ys = reverse xs
```

**Additional context**

I discovered this issue while writing #2161. The only workaround is to _not_ specify the local definition signature or pass parameters with the local types.

byorgey:
Type variables in polymorphic types now scope over any nested local type signatures by default. Closes #2168.

- Any polymorphic type, whether written with an explicit `forall` or not, brings all its variables into scope in the body of its definition.
- Any type variables declared with an explicit `forall` are introduced as new variables, and in particular they will shadow any type variables already in scope with the same names.
- Any type variables not bound by an explicit `forall` *which are not already in scope* will be implicitly quantified.  Any such type variable names which are in scope will simply refer to the name in the enclosing scope.

So, for example, this will work:
```
def balance : RBNode k v -> RBTree k v = \t.
  let baseCase : RBTree k v = inr t in
  let balanceRR : RBNode k v -> RBTree k v = ...
  in undefined
end
```
The type signature on `balance` will bring `k` and `v` into scope in its body (even though there is not an explicit `forall` written); then the `k` and `v` in the type signatures on `baseCase` and `balanceRR` will both refer to the `k` and `v` bound in the type of `balance`.  In particular, this means that `inr t` is a valid definition for `baseCase` since their types match.  On the other hand, if one were to write
```
def balance : RBNode k v -> RBTree k v = \t.
  let baseCase : forall k v. RBTree k v = inr t in
  undefined
end
```
then the `k` and `v` in the type of `baseCase` would be *new* type variables which are different than the `k` and `v` in the type of `balance` (which would now be shadowed within the definition of `baseCase`).  This would now be a type error (which is what happens currently), since `baseCase` is required to be an `RBTree k v` for *any* types `k` and `v`, but `inr t` only has the specific type that was given as an input to `balance`.

Finally, if one wrote 
```
def balance : RBNode k v -> RBTree k v = \t.
  let baseCase : forall k. RBTree k v = inr t in
  undefined
end
```
then only `k` would be a new type variable, whereas `v` would refer to the `v` from the type of `balance`. (This would still be a type error, of course.)

In addition, what used to be `Polytype` is now two separate types: `RawPolytype` is what we get out of the parser, before implicit quantification has been carried out.  A `Polytype` is guaranteed to be properly quantified, i.e. variables bound by the forall are explicitly listed.  We carefully do not export the constructor so we can be sure that the typechecker is making sure that we never accidentally use a `RawPolytype` before we implicitly quantify over it.

### DIFF
--- a/src/swarm-lang/Swarm/Language/Types.hs
+++ b/src/swarm-lang/Swarm/Language/Types.hs
@@ -512,12 +512,20 @@ unPoly (Forall xs t) = (xs, t)
 ptVars :: Poly 'Quantified t -> [Var]
 ptVars (Forall xs _) = xs
 
+-- | Forget that a polytype has been properly quantified.
 forgetQ :: Poly 'Quantified t -> Poly 'Unquantified t
 forgetQ (Forall xs t) = Forall xs t
 
--- | A polytype without unification variables.
+-- | A regular polytype (without unification variables).  A @Polytype@
+--   (as opposed to a @RawPolytype@) is guaranteed to have implicit
+--   quantification properly applied, so that all type variables bound
+--   by the forall are explicitly listed.
 type Polytype = Poly 'Quantified Type
 
+-- | A raw polytype (without unification variables), which corresponds
+--   exactly to the way a polytype was written in source code.  In
+--   particular there may be type variables which are used in the type
+--   but not listed explicitly, which are to be implicitly quantified.
 type RawPolytype = Poly 'Unquantified Type
 
 instance PrettyPrec (Poly q Type) where
--- a/src/swarm-lang/Swarm/Language/Types.hs
+++ b/src/swarm-lang/Swarm/Language/Types.hs
@@ -742,13 +742,13 @@ quantify ::
   Poly 'Unquantified ty ->
   m (Poly 'Quantified ty)
 quantify (Forall xs ty) = do
-  -- Look at all variables which occur in the type but are not explicitly bound by the forall...
-  let unbound = tyVars ty `S.difference` S.fromList xs
-  -- ...and filter out those that are not bound in the context and
-  -- should therefore be implicitly quantified
   inScope <- ask @TVCtx
-  let implicit = S.filter (not . (`Ctx.member` inScope)) unbound
-
+  -- Look at all variables which occur in the type but are not
+  -- explicitly bound by the forall, and and are not bound in the
+  -- context.  Such variables must be implicitly quantified.
+  let implicit =
+        (tyVars ty `S.difference` S.fromList xs)
+          `S.difference` M.keysSet (Ctx.unCtx inScope)
   pure $ Forall (xs ++ S.toList implicit) ty
 
 -- | Absolute implicit quantification, i.e. assume there are no type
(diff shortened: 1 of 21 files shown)


byorgey:
@xsebek I think it works now.  I would appreciate any additional torture-testing you are able to do.

byorgey:
Actually, I think this is not ready to merge --- I'm afraid I've made it too easy to make mistakes and this will just lead to problems down the road.  Currently working on adding more type indexing so the Haskell type system will prevent us from using parsed polytypes that haven't been properly implicitly quanitified.  I'd still appreciate any feedback / testing anyone would like to do in the meantime.

byorgey:
@xsebek OK, I think this is ready to go now!

src/swarm-lang/Swarm/Language/Types.hs
```diff
@@ -650,6 +728,34 @@ type TCtx = Ctx Polytype
...
+  (Has (Reader TVCtx) sig m, Typical ty) =>
+  Poly 'Unquantified ty ->
+  m (Poly 'Quantified ty)
+quantify (Forall xs ty) = do
+  -- Look at all variables which occur in the type but are not explicitly bound by the forall...
+  let unbound = tyVars ty `S.difference` S.fromList xs
+  -- ...and filter out those that are not bound in the context and
+  -- should therefore be implicitly quantified
+  inScope <- ask @TVCtx
+  let implicit = S.filter (not . (`Ctx.member` inScope)) unbound
```
xsebek, line 750: The `unCtx` is published so you could use another `S.difference` with its `keysSet`. 
I am unsure if it would be efficient, but it could be a nice oneliner, and the comment would not be broken up.

src/swarm-lang/Swarm/Language/Types.hs
```diff
@@ -423,41 +432,110 @@ instance (UnchainableFun t, PrettyPrec t, SubstRec t) => PrettyPrec (TypeF t) wh
 ------------------------------------------------------------
 
 class Typical t where
-  foldT :: (TypeF t -> t) -> t -> t
+  -- Fold a type into a summary value of some type @a@, given an
+  -- "empty" value of type @a@ (for use at e.g. Pure nodes in a UType)
+  foldT :: a -> (TypeF a -> a) -> t -> a
+
+  -- Refold a type into another type
+  refoldT :: (TypeF t -> t) -> t -> t
```
xsebek, line 440: This looks truly magical. 🪄

xsebek (approved):
Works well and the documentation is neat. 👍

src/swarm-lang/Swarm/Language/Types.hs
byorgey: What seems magical about it?  Is there additional documentation that would be helpful?

PR #2180, opened 2024-10-15
use applyWhen

kostmo:
Continuing with the replacements started in this discussion: https://github.com/swarm-game/swarm/pull/2140#discussion_r1759787979


byorgey (approved):
Didn't realize we used `if ... then ... else id` so much!

xsebek:
@kostmo could we have a HLint rule to prevent adding it again?

PR #2187, opened 2024-10-19
Fix type inference for recursive let bindings

### LINKED ISSUE #2186: Bad application of execConst - `Fst [VInt 4]`
**Describe the bug**

While writing a benchmark for dictionaries/sets, I ran into a bug that caused Swarm to become unresponsive.
After trying to debug this and removing some code I got a runtime error instead:
```
Fatal error: Bad application of execConst:
Fst
[VInt 4]
```

<details>
  <summary>Click to see full exception</summary>
Screenshot 2024-10-19 at 4 15 07 PM
</details>

**To Reproduce**

Run the benchmark in d7ffc68:
```
run "example/recursive-containers.sw"
bench_insert 0
```

**Expected behavior**

I expect the benchmark not to cause Swarm to become unresponsive or report runtime type errors.

byorgey:
Fixes #2186.

When inferring the type of a recursive let binding with no provided type signature, we generate a unification variable to stand for its type, then infer the type of the body with the unification variable in the context.  Finally, we must unify the generated unification variable with the inferred type.  However, previously, we were doing this final unification with a direct call to `=:=` (which simply returns an `Either`) and discarding the result, which meant that we would simply ignore it when they failed to unify.

This PR replaces the bad call to `=:=` with a call to `unify`, and also makes the `=:=` import qualified to make it less likely that we will ever call `=:=` directly like this again.

This would only have ever caused a problem with (1) a recursive function definition (2) with no type annotation (3) with a type error (4) specifically in one of the recursive applications.  Apparently no one had ever done all of 1-4 at the same time before; @xsebek was the lucky winner. :smile:

### DIFF
--- a/src/swarm-lang/Swarm/Language/Typecheck.hs
+++ b/src/swarm-lang/Swarm/Language/Typecheck.hs
@@ -69,7 +69,7 @@ import Data.Set qualified as S
 import Data.Text (Text)
 import Data.Text qualified as T
 import Prettyprinter
-import Swarm.Effect.Unify (Unification, UnificationError, (=:=))
+import Swarm.Effect.Unify (Unification, UnificationError)
 import Swarm.Effect.Unify qualified as U
 import Swarm.Effect.Unify.Fast qualified as U
 import Swarm.Language.Context hiding (lookup)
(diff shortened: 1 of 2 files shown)


xsebek (approved):
Thanks for the quick fix. 👍

PR #2189, opened 2024-10-20
Move examples to data-files section

xsebek:
Otherwise, change to example would cause recompilation.

### DIFF
--- a/swarm.cabal
+++ b/swarm.cabal
@@ -47,7 +47,6 @@ extra-source-files:
   editors/vim/*.lua
   editors/vim/*.vim
   editors/vscode/syntaxes/*.yaml
-  example/*.sw
 
 extra-doc-files: docs/image/sublibrary-graph.svg
 data-dir: data/
@@ -61,6 +60,7 @@ data-files:
   test/standalone-topography/*.png
   test/standalone-topography/*.yaml
   worlds/*.world
+  example/*.sw
 
 source-repository head
   type: git


swarm.cabal
```diff
@@ -61,6 +60,7 @@ data-files:
   test/standalone-topography/*.png
   test/standalone-topography/*.yaml
   worlds/*.world
+  example/*.sw
```
xsebek, line 63: Ah, I guess we have it in `extra-source-files` because it is not in data directory?

Maybe we could workaround it by a symlink?

byorgey:
Can you explain in more detail?  I'm not sure I understand the problem we're trying to solve here.

xsebek:
@byorgey, if you want to iterate on a swarm file in the `example` directory, Cabal will check each library to see if it needs recompiling because the `example` is listed in extra sources. At least, that is my hypothesis based on observation. 🙂

byorgey (approved):
Seems fine to me.  I think `data-files` get installed in a special location when `cabal install` runs, which is irrelevant for `example`, but I don't think it matters that much.

byorgey (commented):
Hmm, actually, now I am not so sure.  I think the reason the tests are failing is that they are looking for the `example` directory---which isn't there because only things listed in `extra-source-files` get unpacked into the source tree when building + running tests.  That also means if someone does `cabal get` they will no longer get the `example` directory.  `extra-source-files` actually seems like the right place for it.

PR #2190, opened 2024-10-20
Add a progress bar vignette

xsebek:
* add a scenario showing an in-world progress bar

The motivation is that I eventually plan to use a similar system to report progress of tests.

| Recording |
|--------|
| progress |


byorgey (approved):
Looks nice!  It makes me wonder whether it would be worth adding a built-in, more streamlined way to include text entities, in cases like this when you want to label things on a world map.

xsebek:
Yeah, something like this would be nice:
```YAML
palette:
  - char: 'X'  # {name: "X", display: {char: "X"}, description: ["X"], properties: [known]}
```

PR #2191, opened 2024-10-20
Benchmark Dictionary implementation in Swarm

### LINKED ISSUE #2159: Dictionaries
It should be possible to program with dictionaries/maps in Swarm.  Some options include:

1. Now that we have recursive types, it's totally possible to just program dictionaries as a library, as is done in e.g. Haskell.  For example we could define a recursive type of binary trees and implement maps as balanced binary trees storing key/value pairs allowing for $O(\log n)$-time operations.  See #2161 .
2. That works well in Haskell since GHC is an amazingly good optimizing compiler; in Swarm-lang it would result in a rather slow dictionary library.  Taking a cue from e.g. Python, we could also have dictionaries built in as a language primitive, so under the hood they could use actual Haskell `Map`s.
3. One could also imagine trying to co-opt the record system to build dictionaries, but (1) records can't be polymorphic in the key type, (2) the type of a record must strictly list all its fields, and (3) records have heterogeneous value types, whereas dictionaries are homogeneous in the value type.

xsebek:
* add dictionary benchmarks _in Swarm language_
* add more tests
* fix discovered bugs

You can test it with:
```Haskell
run "example/recursive-containers.sw";
benchmark_tree
```

| Screenshot |
|--------|
| Screenshot 2024-10-20 at 5 10 06 PM | 

* Closes #2159


example/recursive-containers.sw
```diff
@@ -9,26 +9,94 @@
...
+  , pretty: d -> Text
+  ]
+end
+
+def undefined_dict : any -> IDict any k v =\empty.
+    [empty=empty, insert=\x.undefined, delete=\x.undefined, get=\x.undefined, contains=\x.undefined, pretty=\x.undefined]
+end
+
+tydef FlatSet a = List a end
+def flat_set : ISet (FlatSet a) a = undefined_set (inl ()) end
```
byorgey, line 51: What's the purpose of defining `flat_set` et al as records with lots of `undefined` at first, and then later redefining them?

example/recursive-containers.sw
```diff
@@ -510,4 +639,140 @@ def test_tree: Cmd Unit =
...
+end
+
+/*******************************************************************/
+/*                           BENCHMARKS                            */
+/*******************************************************************/
+
+def benchmark: Int -> s -> (s -> s) -> Cmd (Int * (Int * Int)) = \times.\s.\act.
+  let min = \x.\y. if (x > y) {y} {x} in
+  let max = \x.\y. if (x > y) {x} {y} in
+  let runM: (Int * Maybe (Int * Int)) -> s -> Int -> Cmd (Int * Maybe (Int * Int)) = \acc.\s.\n.
```
byorgey, line 651: Scoped type variables! :heart_eyes:

byorgey (approved):

example/recursive-containers.sw
xsebek: It's an overview, as it can be hard to spot between the Red-Black Tree implementation and tests.

It also works as a "header" of sorts, if you wanted to copy paste part of it in a scenario definition/other Swarm script and then run the file.

PR #2195, opened 2024-10-20
subworld-aware teleportation and location query commands

kostmo:
```
scripts/play.sh --scenario scenarios/Testing/144-subworlds/teleport-and-query.yaml --autoplay
```


byorgey (approved):
Looks good, and a very sensible generalization.

Issue #2106, opened 2024-08-10
Web API for base robot outputs 800MB per minute


xsebek:
**Describe the bug**

Trying to get base log after `run "example/list.sw"` does not terminate - at least not in any reasonable time.

Thats because the output is flooded with infinite (or exponential) stream of JSON `Syntax`:
```
[2902,2904],"_sTerm":{"contents":1,"tag":"TInt"},"_sType":[]}],"tag":"SApp"},"_sType":[]},{"_sComments":{"_afterComments":[],"_beforeComments":[]},"_sLoc":[2906,2909],"_sTerm":{"contents":"acc","tag":"
TVar"},"_sType":[]}],"tag":"SApp"},"_sType":[]}],"tag":"SApp"},"_sType":[]},{"_sComments":{"_afterComments":[],"_beforeComments":[]},"_sLoc":[2911,2918],"_sTerm":{"contents":[{"_sComments":{"_afterComm
ents":[],"_beforeComments":[]},"_sLoc":[2912,2914],"_sTerm":{"contents":[{"_sComments":{"_afterComments":[],"_beforeComments":[]},"_sLoc":"NoLoc","_sTerm":{"contents":"Div","tag":"TConst"},"_sType":[]}
,{"_sComments":{"_afterComments":[],"_beforeComments":[]},"_sLoc":[2912,2914],"_sTerm":{"contents":"i","tag":"TVar"},"_sType":[]}],"tag":"SApp"},"_sType":[]},{"_sComments":{"_afterComments":[],"_before
```

**To Reproduce**

Run swarm:
```sh
swarm --scenario creative --run "example/list.sw"
```

Get the output using cURL - I suggest piping to `wc` for 60 seconds.
Optionally observe the process memory usage with `ps`, `top`, or an alternative like `btop`.
```bash
ps -o uid,pid,ppid,ni,vsz,rss,stat,tty,time,command -p $(ps -o pid,comm | grep swarm | cut -f1 -d' ')
#  UID   PID  PPID NI      VSZ    RSS STAT TTY           TIME COMMAND
#  501 85641 44657  0 679915760 140512 S+   ttys003    0:02.74 swarm --scenario creative --run example/list.sw

curl --max-time 60 -s localhost:5357/robot/0 | wc
#       0    5760 874380819

ps -o uid,pid,ppid,ni,vsz,rss,stat,tty,time,command -p $(ps -o pid,comm | grep swarm | cut -f1 -d' ')
[...]

Issue #2107, opened 2024-08-10
Reuse environment definitions in `ToJSON` and `FromJSON`


xsebek:
**Is your feature request related to a problem? Please describe.**

**TLDR:** the problem is that `_envVars` inside `_envVars` cause exponential JSON size.
 
Take this simple example:
```
def m2 = m; m end
def m4 = m2; m2 end
def m8 = m4; m4 end
def m16 = m8; m8 end
def m32 = m16; m16 end
```

And observe the growth of JSON output:
| empty | m2 | m4 | m8 | m16 | m32 | m1024 |
|--------|--------|--------|--------|--------|--------|--------|
| 5 | 27 | 78 | 180 | 384 | 792 | 52200 |

This makes it impossible to get to other useful parts of robot JSON, like the log, if the program has enough definitions:
* #2106

**Describe the solution you'd like**

The definitions should be reused - the inner references should link (`{"link": "m8"}`) to outer definition.

If we can check that the definitions are the same, we could prune them from inner scope:

```diff
- deriving instance FromJSON Env
- deriving instance ToJSON Env
+ instance FromJSONE Env Env
+ instance ToJSON (Env, Env)  -- or some newtype WithEnv
```

**Describe alternatives you've considered**

The current derived instance is broken, so maybe we could only keep the top environment, or none at all.

byorgey:
This may actually be a good application for the (heretofore mythical) `ToJSONE`.  I just want to make sure we avoid doing $O(n^2)$ work to repeatedly check the same definitions for equality.

xsebek:
I think even $O(n^2)$ would be OK, since it would remove exponentially many inner references.

Outputting JSON is not done on main thread, so it would not freeze the game.

byorgey:
Good point.  Yes, we should not succumb to [premature optimization](https://wiki.c2.com/?PrematureOptimization) here.  Let's start with the simplest thing that works and we can optimize it later if necessary.

byorgey:
After some discussion with @xsebek on Discord, here's a sketch of an idea.  `Env` is just made up of a bunch of `Ctx` (which almost, but not quite always, have all the same variables).  So we can focus first on making this sort of sharing/recovery work for `Ctx`.

Currently `Ctx t` is just defined as a `newtype` for `Map Var t`.  The proposal is to keep this, but add some extra structure that allows us to remember the structure of how the `Ctx` was built, and quickly identify when two `Ctx` values are equal, without having to actually compare them:

```haskell
[...]

PR #2202, opened 2024-11-05
Contexts that can be serialized + deserialized while retaining and explicitly representing sharing

byorgey:
Towards #2107.  Instead of simply storing a `Map`, contexts now also store a structured tree representing how it was built, and cache homomorphic hash values so contexts can be compared + stored by hash.  This allows us to serialize contexts into a compact form that never duplicates any context, then deserialize them in a way that restores all the sharing.

The big thing that is left is to make `FromJSONE` instances for `Value` and `Env` (and anything which contains them).  I was working on these but got bogged down, so decided to not let the perfect be the enemy of the good.  You can see some commented-out code towards a `FromJSONE Value` instance.  But we don't use any such instances yet, so I think this can be merged as is, and the rest worked out later.


byorgey:
Not quite ready to merge --- still need to update more JSON instances --- but I would be happy to get any comments or questions at this point!

I've tried to put in a bunch of comments but it's quite possible, even likely, that it's still confusing.  Please let me know and I will happily add more comments or clarify anything that's confusing.

src/swarm-lang/Swarm/Language/Context.hs
```diff
@@ -87,18 +219,85 @@ vars = M.keys . unCtx
...
-addBinding :: Var -> t -> Ctx t -> Ctx t
-addBinding x t (Ctx c) = Ctx (M.insert x t c)
+addBinding :: Hashable t => Var -> t -> Ctx t -> Ctx t
+addBinding x t (Ctx m s@(CtxTree h _)) = Ctx (M.insert x t m) s'
+ where
+  s' = CtxTree h' (CtxUnion (CtxTree tHash (CtxSingle x t)) s)
+  tHash = singletonHash x t
+  h' = case M.lookup x m of
+    Nothing -> h + tHash
+    Just t' -> h - singletonHash x t' + tHash
```
xsebek, line 229: These hashes are spread out, could they be written in one function taking restructured tree to `Const Hash`? 🤔

test/unit/TestContext.hs
```diff
@@ -0,0 +1,63 @@
...
+ where
+  ctx1 = singleton "x" 3
+  ctx2 = singleton "x" 3 <> singleton "z" 6
+  ctx3 = singleton "x" 5 <> singleton "y" 7
+  ctx4 = singleton "x" 5 <> singleton "y" 7 <> singleton "z" 6
+  ctx5 = singleton "y" 10
+  bigCtx = fromMap . M.fromList $ zip (map (("x" <>) . showT) [1 :: Int ..]) [1 .. 10000]
+
+ctxsEqual :: Ctx Int -> Ctx Int -> Assertion
+ctxsEqual ctx1 ctx2 = do
```
xsebek, line 50: How about contexts not being equal?

```math
c_1 = c_2 \equiv hash c_1 = hash c_2
```

I think the equivalence tests do not prevent the trivial implementation `hash x = 0`.

src/swarm-lang/Swarm/Language/Context.hs
byorgey: Hmm, good question.  When I first saw your comment I thought "oh, of course, I will make a function like `hashTree :: Hashable t => CtxF CtxTree t -> CtxTree t` which rolls up one level of recursion while computing the appropriate top-level hash."  Unfortunately I quickly found that function cannot be written efficiently --- the problem is that in the delete and union cases, we also need the `Map Var t`'s in order to compute the hash (since we need to subtract off overlap when unioning, etc.).

Hmm, maybe I can write something like `Hashable t => CtxF Ctx t -> Ctx t`?  Let me try that...

byorgey: That worked great!  All hash computations are now in a single place, and all other definitions are one-liners.  Thanks for the suggestion!

test/unit/TestContext.hs
byorgey: I mean, the requirement to also have structural equality in addition to hash equality would not allow `hash x = 0` to pass.  But it's a good point, and we also want to make sure that hash collisions are as rare as I think they should be.  I will add some QuickCheck tests to make sure that randomly generated contexts have hashes that are equal iff the resulting maps are equal.

src/swarm-lang/Swarm/Language/Context.hs
```diff
@@ -10,72 +9,243 @@
...
+  deriving (Semigroup, Monoid) via Sum Int
+  deriving (Num) via Int
+
+instance Show CtxHash where
+  show (CtxHash h) = printf "%016x" h
+
+instance ToJSON CtxHash where
+  toJSON h = toJSON (show h)
+
+instance FromJSON CtxHash where
```
byorgey, line 67: Add test for round-tripping `CtxHash` through JSON

src/swarm-lang/Swarm/Language/Context.hs
```diff
@@ -10,72 +9,239 @@
...
+ctxFromTree :: CtxTree t -> Ctx t
+ctxFromTree tree = Ctx (varMap tree) tree
+ where
+  varMap (CtxTree _ s) = case s of
+    CtxEmpty -> M.empty
+    CtxSingle x t -> M.singleton x t
+    CtxDelete x _ s1 -> M.delete x (varMap s1)
+    CtxUnion s1 s2 -> varMap s2 `M.union` varMap s1
+
+-- | Roll up one level of context structure while building a new
```
kostmo, line 156: What does it mean to "roll up" here?

src/swarm-lang/Swarm/Language/Context.hs
```diff
@@ -87,18 +253,75 @@ vars = M.keys . unCtx
...
+-- | A 'CtxMap' maps context hashes to context structures.  Those
+--   structures could either be complete context trees, or just a
+--   single level of structure containing more hashes.
+type CtxMap f t = Map CtxHash (CtxF f t)
+
+-- | Reconstitute the context corresponding to a particular hash, by
+--   looking it up in a context map.
+getCtx :: CtxHash -> CtxMap CtxTree t -> Maybe (Ctx t)
+getCtx h m = case M.lookup h m of
+  Nothing -> Nothing
```
kostmo, line 284: `Nothing -> Nothing` indicates refactoring potential.  Replace with `fmap`?

src/swarm-lang/Swarm/Language/Context.hs
```diff
@@ -87,18 +253,75 @@ vars = M.keys . unCtx
...
+
+-- | Build a context map by keeping track of the incrementally built
+--   map in a state effect, and traverse the given context structure
+--   to add all subtrees to the map---but, of course, stopping without
+--   recursing further whenever we see a hash that is already in the
+--   map.
+buildCtxMap :: forall t m sig. Has (State (CtxMap CtxTree t)) sig m => Map Var t -> CtxTree t -> m ()
+buildCtxMap m (CtxTree h s) = do
+  cm <- get @(CtxMap CtxTree t)
+  case h `M.member` cm of
```
kostmo, line 300: Replace this `case` expression with `unless`?

src/swarm-lang/Swarm/Language/Context.hs
```diff
@@ -1,7 +1,6 @@
 {-# LANGUAGE DeriveDataTypeable #-}
+{-# LANGUAGE DerivingVia #-}
 {-# LANGUAGE OverloadedStrings #-}
-{-# LANGUAGE PatternSynonyms #-}
-{-# LANGUAGE ViewPatterns #-}
 
 -- |
```
kostmo, line 5: In addition to outlining this module's contents (e.g. "Contexts with a serialization implementation that preserves sharing") perhaps also briefly motivate the existence of Contexts (e.g. Robots need to store variables for...)

kostmo (approved):

src/swarm-lang/Swarm/Language/Context.hs
byorgey: I added more comments to clarify, hopefully it's clearer now?

The terms "roll" and "unroll" are commonly used to refer to things like `roll :: f (Fix f) -> Fix f` and `unroll :: Fix f -> f (Fix f)`, i.e. moving one layer of structure in and out of a tree represented as the fixpoint of a functor (which is essentially what we have here).

src/swarm-lang/Swarm/Language/JSON.hs
```diff
@@ -32,8 +32,38 @@ instance ToJSON Syntax
 instance ToJSON Value where
   toJSON = genericToJSON optionsMinimize
 
-instance FromJSON Value where
-  parseJSON = genericParseJSON optionsMinimize
+-- instance FromJSONE (CtxMap CtxTree t) Value where
+--   parseJSONE = withObjectE "Value" $ \v -> case Ae.toList v of
```
xsebek, line 36: Could you please add a TODO comment with a linked Issue?

xsebek (approved):
Thanks @byorgey, this enables debugging using the Web API again. 👍 

I considered removing the context in JSON output entirely, so having only hashes is better than what I would have aimed for. 😄

byorgey:
> Thanks @byorgey, this enables debugging using the Web API again. 👍
> 
> I considered removing the context in JSON output entirely, so having only hashes is better than what I would have aimed for. 😄

Great!  So can we close #2106 ?

src/swarm-lang/Swarm/Language/JSON.hs
byorgey: Good idea, done!

PR #2209, opened 2024-11-14
refactor in preparation for new structure recognizer

kostmo:
* Create a util function `adaptGameState` to simplify code in `updateEntityAt`
* Relocate and modify the `ensureStructureIntact` function to return the position of the first non-matching cell
* Supply width in the `StructureWithGrid` record


src/swarm-topography/Swarm/Game/Scenario/Topography/Structure/Recognition/Log.hs
```diff
@@ -70,7 +70,7 @@ data ParticipatingEntity e = ParticipatingEntity
   deriving (Functor, Generic, ToJSON)
 
 data IntactPlacementLog = IntactPlacementLog
-  { isIntact :: Bool
+  { intactnessFailure :: Maybe StructureIntactnessFailure
```
byorgey, line 73: :+1: https://existentialtype.wordpress.com/2011/03/15/boolean-blindness/

byorgey (approved):
Looks good to me!

PR #2211, opened 2024-11-17
Do requirements analysis for literal text argument to `use`

### LINKED ISSUE #1301: Perform requirements analysis on static argument to `use` command
From https://github.com/swarm-game/swarm/pull/1287#discussion_r1210858714:

If you write a program like `build {move; use "flerb" forward; move}` it will fail, because the requirements analysis will not figure out that we should automatically equip a `flerb` when building the child robot. You would have to explicitly write e.g. `build {move; require "flerb"; use "flerb" forward; move}`. We might consider adding something to the requirements analysis which will at least work when use is applied to a statically known device name, as in this example.

byorgey:
In the special case that we have something like `use "key"` we can statically know that it requires a `key` device.  For example, this means that now a program like
```
build { move; use "key"; move }
```
will automatically install a `key` device on the built robot.

Closes #1301.


xsebek (approved):
Nice! 👍

PR #2216, opened 2024-11-30
Support a generic "original" payload in structure recognizer

kostmo:
The recognizer algorithm operates on a `NonEmptyGrid` of `Maybe` elements.  However, there may exist some progenitor object with auxiliary metadata payload that is independent of the algorithm but should be presented along with the final recognized structure.

So long as an "extraction" function is provided that derives the nonempty grid of `a`s from the original object `b`, we can send `b` along for the ride through the algorithm without concern to `b`'s internals.

# Changes

Overall this refactoring makes the code more generic and simpler.

* Build the recognizer automata at scenario parse time
* Enforce grid no emptiness earlier in recognizer construction
* Subsume redundant `_structureDefs` member of `StaticStructureInfo` record with `_staticAutomatons` field
* Make some type signatures in the UI more generic
* Push the conversion from `Cell` to `Entity `further up the pipeline
* Remove some derived `Show` instances (`show`-ing a Scenario is not likely to be useful due to sheer volume of content)

### DIFF
--- a/src/swarm-scenario/Swarm/Game/Scenario.hs
+++ b/src/swarm-scenario/Swarm/Game/Scenario.hs
@@ -17,8 +17,8 @@ module Swarm.Game.Scenario (
   Scenario (..),
   ScenarioLandscape (..),
   ScenarioMetadata (ScenarioMetadata),
+  RecognizableStructureContent,
   staticPlacements,
-  structureDefs,
 
   -- ** Fields
   scenarioMetadata,
@@ -86,14 +86,16 @@ import Swarm.Game.Scenario.Objective.Validation
 import Swarm.Game.Scenario.RobotLookup
 import Swarm.Game.Scenario.Style
 import Swarm.Game.Scenario.Topography.Cell
+import Swarm.Game.Scenario.Topography.Grid
 import Swarm.Game.Scenario.Topography.Navigation.Portal
 import Swarm.Game.Scenario.Topography.Navigation.Waypoint (Parentage (..))
 import Swarm.Game.Scenario.Topography.Structure qualified as Structure
 import Swarm.Game.Scenario.Topography.Structure.Assembly qualified as Assembly
 import Swarm.Game.Scenario.Topography.Structure.Named qualified as Structure
 import Swarm.Game.Scenario.Topography.Structure.Overlay
-import Swarm.Game.Scenario.Topography.Structure.Recognition.Static
-import Swarm.Game.Scenario.Topography.Structure.Recognition.Symmetry
+import Swarm.Game.Scenario.Topography.Structure.Recognition.Precompute
+import Swarm.Game.Scenario.Topography.Structure.Recognition.Symmetry (renderRedundancy)
+import Swarm.Game.Scenario.Topography.Structure.Recognition.Type
 import Swarm.Game.Scenario.Topography.WorldDescription
 import Swarm.Game.Terrain
 import Swarm.Game.Universe
@@ -181,6 +183,8 @@ scenarioSolution :: Lens' ScenarioOperation (Maybe TSyntax)
 --   take during a single tick.
 scenarioStepsPerTick :: Lens' ScenarioOperation (Maybe Int)
 
+type RecognizableStructureContent = NonEmptyGrid (Maybe Cell)
+
 -- | All cosmetic and structural content of the scenario.
 data ScenarioLandscape = ScenarioLandscape
   { _scenarioSeed :: Maybe Int
@@ -190,10 +194,9 @@ data ScenarioLandscape = ScenarioLandscape
   , _scenarioKnown :: Set EntityName
   , _scenarioWorlds :: NonEmpty WorldDescription
   , _scenarioNavigation :: Navigation (M.Map SubworldName) Location
-  , _scenarioStructures :: StaticStructureInfo Cell
+  , _scenarioStructures :: StaticStructureInfo RecognizableStructureContent Entity
   , _scenarioRobots :: [TRobot]
   }
-  deriving (Show)
 
 makeLensesNoSigs ''ScenarioLandscape
 
@@ -220,7 +223,7 @@ scenarioKnown :: Lens' ScenarioLandscape (Set EntityName)
 scenarioWorlds :: Lens' ScenarioLandscape (NonEmpty WorldDescription)
 
 -- | Information required for structure recognition
-scenarioStructures :: Lens' ScenarioLandscape (StaticStructureInfo Cell)
+scenarioStructures :: Lens' ScenarioLandscape (StaticStructureInfo RecognizableStructureContent Entity)
 
 -- | Waypoints and inter-world portals
 scenarioNavigation :: Lens' ScenarioLandscape (Navigation (M.Map SubworldName) Location)
@@ -236,7 +239,6 @@ data Scenario = Scenario
   , _scenarioOperation :: ScenarioOperation
   , _scenarioLandscape :: ScenarioLandscape
   }
-  deriving (Show)
 
 makeLensesNoSigs ''Scenario
 
@@ -355,10 +357,15 @@ instance FromJSONE ScenarioInputs Scenario where
           namedGrids = map stuffGrid mergedStructures
           recognizableGrids = filter Structure.isRecognizable namedGrids
 
-      symmetryAnnotatedGrids <- mapM checkSymmetry recognizableGrids
+          -- We exclude empty grids from the recognition engine.
+          nonEmptyRecognizableGrids = mapMaybe (traverse getNonEmptyGrid) recognizableGrids
+
+      myAutomatons <-
+        either (fail . T.unpack . renderRedundancy) return $
+          mkAutomatons (fmap cellToEntity) nonEmptyRecognizableGrids
 
       let structureInfo =
-            StaticStructureInfo symmetryAnnotatedGrids
+            StaticStructureInfo myAutomatons
               . M.fromList
               . NE.toList
               $ NE.map (worldName &&& placedStructures) allWorlds
--- a/src/swarm-topography/Swarm/Game/Scenario/Topography/Structure/Recognition/Log.hs
+++ b/src/swarm-topography/Swarm/Game/Scenario/Topography/Structure/Recognition/Log.hs
@@ -6,12 +6,12 @@ module Swarm.Game.Scenario.Topography.Structure.Recognition.Log where
 
 import Data.Aeson
 import Data.List.NonEmpty (NonEmpty)
-import Data.List.NonEmpty qualified as NE
+import Data.List.NonEmpty.Extra qualified as NE
 import GHC.Generics (Generic)
 import Servant.Docs (ToSample)
 import Servant.Docs qualified as SD
 import Swarm.Game.Location (Location)
-import Swarm.Game.Scenario.Topography.Structure.Named (StructureName)
+import Swarm.Game.Scenario.Topography.Structure.Named (StructureName, name)
 import Swarm.Game.Scenario.Topography.Structure.Recognition.Static (
   OrientedStructure,
  )
@@ -20,7 +20,7 @@ import Swarm.Game.Universe (Cosmic)
 
 renderSharedNames :: ConsolidatedRowReferences b a -> NonEmpty StructureName
 renderSharedNames =
-  NE.nub . NE.map (getName . originalDefinition . wholeStructure) . referencingRows
+  NE.nubOrd . NE.map (name . originalItem . entityGrid . wholeStructure) . referencingRows
 
 data ParticipatingEntity e = ParticipatingEntity
   { entity :: e
--- a/src/swarm-topography/Swarm/Game/Scenario/Topography/Structure/Recognition/Precompute.hs
+++ b/src/swarm-topography/Swarm/Game/Scenario/Topography/Structure/Recognition/Precompute.hs
@@ -37,7 +37,6 @@ module Swarm.Game.Scenario.Topography.Structure.Recognition.Precompute (
 
   -- * Helper functions
   populateStaticFoundStructures,
-  getEntityGrid,
   lookupStaticPlacements,
   ensureStructureIntact,
 ) where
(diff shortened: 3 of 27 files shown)


src/swarm-topography/Swarm/Game/Scenario/Topography/Structure/Recognition/Precompute.hs
```diff
@@ -79,85 +82,74 @@ import Swarm.Util (histogram)
...
-  (Maybe b -> Maybe a) ->
-  [SymmetryAnnotatedGrid (Maybe b)] ->
-  RecognizerAutomatons (Maybe b) a
-mkAutomatons extractor xs =
-  RecognizerAutomatons
-    infos
-    (mkEntityLookup rotatedGrids)
+  (NamedArea b -> NonEmptyGrid (AtomicKeySymbol a)) ->
+  [NamedArea b] ->
+  Either T.Text (RecognizerAutomatons b a)
```
kostmo, line 93: Create a dedicated error type

src/swarm-topography/Swarm/Game/Scenario/Topography/Structure/Recognition/Static.hs
```diff
@@ -24,10 +18,10 @@ data RotationalSymmetry
   deriving (Show, Eq)
 
 data SymmetryAnnotatedGrid a = SymmetryAnnotatedGrid
-  { namedGrid :: NamedGrid a
-  , symmetry :: RotationalSymmetry
+  { symmetry :: RotationalSymmetry
+  , namedGrid :: a
```
kostmo, line 22: Rename this field

src/swarm-topography/Swarm/Game/Scenario/Topography/Structure/Recognition/Symmetry.hs
```diff
@@ -28,18 +29,18 @@ import Swarm.Util (commaList, failT, histogram, showT)
 --    2-fold symmetry.
 --    Warn if two opposite orientations were supplied.
 checkSymmetry ::
-  (MonadFail m, Eq a) => NamedGrid a -> m (SymmetryAnnotatedGrid a)
-checkSymmetry ng = do
+  Eq a =>
+  ExtractedArea b a ->
+  Either T.Text (SymmetryAnnotatedGrid (ExtractedArea b a))
```
kostmo, line 34: New error type

src/swarm-topography/Swarm/Game/Scenario/Topography/Structure/Recognition/Tracking.hs
```diff
@@ -54,48 +55,42 @@ entityModified ::
...
  where
-  entLookup = recognizer ^. automatons . automatonsByEntity
+  entLookup = autoRecognizer ^. automatonsByEntity
 
-  doAddition newEntity r = do
-    stateRevision <- case HM.lookup newEntity entLookup of
-      Nothing -> return oldRecognitionState
+  doAddition newEntity oldState =
+    case HM.lookup newEntity entLookup of
+      Nothing -> return oldState
```
kostmo, line 74: Use a combinator

byorgey:
@kostmo I am having trouble finding which modules contain the "main" changes (as opposed to other modules which "just" have to be updated to match the new types/definitions).  Could you mention which modules are the interesting ones to look at?

src/swarm-scenario/Swarm/Game/Scenario.hs
```diff
@@ -355,10 +357,15 @@ instance FromJSONE ScenarioInputs Scenario where
           namedGrids = map stuffGrid mergedStructures
           recognizableGrids = filter Structure.isRecognizable namedGrids
 
-      symmetryAnnotatedGrids <- mapM checkSymmetry recognizableGrids
+          -- We exclude empty grids from the recognition engine.
+          nonEmptyRecognizableGrids = mapMaybe (traverse getNonEmptyGrid) recognizableGrids
+
+      myAutomatons <-
+        either (fail . T.unpack . renderRedundancy) return $
+          mkAutomatons (fmap cellToEntity) nonEmptyRecognizableGrids
```
kostmo, line 365: @byorgey here is one interesting change.  Previously the orientations to be recognized were checked for redundancy (according to symmetry) in the scenario parser.  I have now moved that detail to within the recognizer API.

Related to this, we now instantiate the recognizer engine from within the scenario parsing code (instead of from within game state initialization), so that it can throw an error when a redundant recognition spec is provided.

src/swarm-topography/Swarm/Game/Scenario/Topography/Structure/Recognition/Precompute.hs
```diff
@@ -79,85 +81,74 @@ import Swarm.Util (histogram)
...
+      infos =
+        M.fromList $
+          map ((name . originalItem . grid . annotatedGrid &&& id) . process) onlyNonempties
+  return $
+    RecognizerAutomatons
+      infos
+      (mkEntityLookup rotatedGrids)
  where
-  rotatedGrids = concatMap (extractGrids extractor . namedGrid) xs
+  extractedItems = map (uncurry ExtractedArea . fmap (extractor . structure) . dupe) rawGrids
```
kostmo, line 104: Another major change: the "extraction" of `Entities` from `Cells` (which, from the perspective of the recognizer library, is the opaque type parameter `b`) is now only performed once in this single location, and the result is preserved and shared elsewhere it is needed.  The "extraction" function also used to operate piecewise on each cell within a structure.  Now, its contract is to take an entire "shape" of type `b` and produce a `NonEmptyGrid (Maybe a)`.

src/swarm-topography/Swarm/Game/Scenario/Topography/Structure/Recognition/Precompute.hs
```diff
@@ -79,85 +81,74 @@ import Swarm.Util (histogram)
...
-    Grid neGrid ->
-      let w = RowWidth . rectWidth . getNEGridDimensions $ neGrid
-       in Just $
-            StructureWithGrid wrapped d w $
-              applyOrientationTransformNE (Orientation d False) neGrid
+  StructureWithGrid b a
+extractOrientedGrid (ExtractedArea x neGrid) d =
+  StructureWithGrid d w $
+    ExtractedArea x $
+      applyOrientationTransformNE (Orientation d False) neGrid
```
kostmo, line 118: This got much simpler since I pushed the enforcement of `NonEmptyGrid`s further upstream.

src/swarm-topography/Swarm/Game/Scenario/Topography/Structure/Recognition/Precompute.hs
```diff
@@ -79,85 +81,74 @@ import Swarm.Util (histogram)
...
 
   f (subworldName, locatedList) = mapMaybe g locatedList
    where
     g (LocatedStructure (OrientedStructure theName d) loc) = do
       sGrid <- M.lookup theName definitionMap
-      x <- extractOrientedGrid extractor sGrid d
-      return $ PositionedStructure (Cosmic subworldName loc) x
+      return $
+        PositionedStructure (Cosmic subworldName loc) $
+          extractOrientedGrid (grid $ annotatedGrid sGrid) d
```
kostmo, line 152: This was the other place that "extraction" was recomputed.  Now we just re-use the result computed earlier.

src/swarm-topography/Swarm/Game/Scenario/Topography/Structure/Recognition/Precompute.hs
```diff
@@ -79,85 +81,74 @@ import Swarm.Util (histogram)
...
-  StaticStructureInfo b ->
-  [FoundStructure (Maybe b) a]
-lookupStaticPlacements extractor (StaticStructureInfo structDefs thePlacements) =
+  StaticStructureInfo b a ->
+  [FoundStructure b a]
+lookupStaticPlacements (StaticStructureInfo theAutomatons thePlacements) =
   concatMap f $ M.toList thePlacements
  where
-  definitionMap = M.fromList $ map ((name &&& id) . namedGrid) structDefs
+  definitionMap = theAutomatons ^. originalStructureDefinitions
```
kostmo, line 143: We also no longer need to recompute this `Map`.

src/swarm-topography/Swarm/Game/Scenario/Topography/Structure/Recognition/Static.hs
```diff
@@ -48,19 +42,3 @@ data LocatedStructure = LocatedStructure
 instance HasLocation LocatedStructure where
   modifyLoc f (LocatedStructure x originalLoc) =
     LocatedStructure x $ f originalLoc
-
-data StaticStructureInfo b = StaticStructureInfo
-  { _structureDefs :: [SymmetryAnnotatedGrid (Maybe b)]
-  , _staticPlacements :: Map SubworldName [LocatedStructure]
-  }
-  deriving (Show)
```
kostmo, line 56: This `StaticStructureInfo` record was relocated to the `Swarm.Game.Scenario.Topography.Structure.Recognition.Type` module.  And its `_structureDefs` member is made redundant.

src/swarm-topography/Swarm/Game/Scenario/Topography/Structure/Recognition/Tracking.hs
```diff
@@ -54,48 +55,41 @@ entityModified ::
...
-  doRemoval = do
+  doAddition newEntity =
+    maybe return logAndRegister $ HM.lookup newEntity entLookup
+   where
+    logAndRegister finder s = do
+      tell . pure . FoundParticipatingEntity $
+        ParticipatingEntity
+          newEntity
+          (finder ^. inspectionOffsets)
+      registerRowMatches entLoader cLoc finder s
```
kostmo, line 80: Here is just some independent refactoring to clean up this logic

src/swarm-topography/Swarm/Game/Scenario/Topography/Structure/Recognition/Type.hs
```diff
@@ -154,13 +155,9 @@ data ConsolidatedRowReferences b a = ConsolidatedRowReferences
...
--- | This wrapper facilitates naming the original structure
--- (i.e. the "payload" for recognition)
--- for the purpose of both UI display and internal uniqueness,
--- while remaining agnostic to its internals.
-data NamedOriginal b = NamedOriginal
-  { getName :: StructureName
-  , orig :: NamedGrid b
+data ExtractedArea b a = ExtractedArea
+  { originalItem :: NamedArea b
+  , extractedGrid :: NonEmptyGrid (AtomicKeySymbol a)
```
kostmo, line 160: This new `ExtractedArea` type is what propagates the original `b` through the algorithm.

src/swarm-topography/Swarm/Game/Scenario/Topography/Structure/Recognition/Type.hs
```diff
@@ -154,13 +155,9 @@ data ConsolidatedRowReferences b a = ConsolidatedRowReferences
   , theRowWidth :: RowWidth
   }
 
--- | This wrapper facilitates naming the original structure
--- (i.e. the "payload" for recognition)
--- for the purpose of both UI display and internal uniqueness,
--- while remaining agnostic to its internals.
-data NamedOriginal b = NamedOriginal
-  { getName :: StructureName
-  , orig :: NamedGrid b
```
kostmo, line 163: Don't need `NamedOriginal` anymore; we just access the `name` through the `NamedArea` record.

byorgey (approved):
Thanks, the additional comments were very helpful!  I think this looks good --- I'm always in favor of nice refactoring that makes things simpler + more modular!

src/swarm-topography/Swarm/Game/Scenario/Topography/Structure/Recognition/Log.hs
```diff
@@ -20,7 +20,7 @@ import Swarm.Game.Universe (Cosmic)
 
 renderSharedNames :: ConsolidatedRowReferences b a -> NonEmpty StructureName
 renderSharedNames =
-  NE.nub . NE.map (getName . originalDefinition . wholeStructure) . referencingRows
+  NE.nub . NE.map (name . originalItem . entityGrid . wholeStructure) . referencingRows
```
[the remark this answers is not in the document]
kostmo: Most likely the list is small (it's bounded by the number of user-defined structures/their sizes).  I was aware of this $O(N^2)$ issue but did not see an alternative in [`Data.List.NonEmpty`](https://hackage.haskell.org/package/base-4.20.0.1/docs/Data-List-NonEmpty.html#v:nub).   However, now that I look, I see [`Data.List.NonEmpty.Extra.nubOrd`](https://hackage.haskell.org/package/extra-1.8/docs/Data-List-NonEmpty-Extra.html#v:nubOrd), so I'll use that.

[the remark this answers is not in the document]
byorgey: Huh, `nubOrd` uses a custom red-black tree implementation.  I wonder why it doesn't just use `Set`.  Well, it doesn't matter.  Since the list we are nubbing is just a list of names, `nubOrd` should work well.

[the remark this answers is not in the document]
kostmo: Could the red-black implementation possibly be order-preserving?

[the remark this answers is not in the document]
kostmo: > I wonder why it doesn't just use `Set`.

I presume it is to minimize dependencies of the [`extra` package](https://hackage.haskell.org/package/extra).  Particularly, `extra` does not depend on [`containers`](https://hackage.haskell.org/package/containers-0.7/docs/Data-Set.html).

Incidentally, now that I've read the code I see that `nubOrd` is order-preserving whereas `nubSort` is not.  That fact would be good to note explicitly in the docstring (rather than indirectly by example).

Issue #1314, opened 2023-06-09
Add more typechecking stack frame types and pretty-print them in type errors


byorgey:
This is a follow-up to #1308.

* We need to add more constructors to the `TCFrame` type, and push them with appropriate calls to `withFrame` during type inference and checking.  See the existing three constructors as an example.
* Currently, pretty-printing of type error messages will simply print out every frame on the stack, but this might be overwhelming.  We probably need to do some work to select which frames from the stack to actually show.  At the very least we should probably show the first and last frames; beyond that maybe we can come up with some heuristics as to which additional frames are the most helpful to show.

byorgey:
I'm going to close this since especially after merging #2220 I think the status quo is fine, and I can't think of anything else in particular that would unambiguously improve error messages (as opposed to just adding extra clutter).  If we get reports of confusing messages or think of specific improvements then we can open separate issues in the future.

PR #2220, opened 2024-12-13
Add `TCAppL` and `TCAppR` typechecking stack frames, and remove `TCBind{L,R}`

byorgey:
Towards #1314.

- It seems like it could be helpful to report that we were checking the left- or right-hand side of a function application when reporting type errors, to help give context.
- On the other hand the "checking the LHS/RHS of a semicolon" never seemed very useful, since it would always come in a big chain and didn't really help localize the error, so I removed it.
- I'm open to suggestions for other context we could provide.  There is definitely a balance here between providing helpful context and providing too much.  For example, should we add something that says "while checking the LHS/RHS of a pair"?

### DIFF
--- a/src/swarm-lang/Swarm/Language/Typecheck.hs
+++ b/src/swarm-lang/Swarm/Language/Typecheck.hs
@@ -104,8 +104,8 @@ data TCFrame where
 instance PrettyPrec TCFrame where
   prettyPrec _ = \case
     TCLet x -> "While checking the definition of" <+> pretty x
-    TCAppL s -> "While checking the left-hand side of a function application: _" <+> prettyPrec 11 s
-    TCAppR s -> "While checking the right-hand side of a function application:" <+> prettyPrec 10 s <+> "_"
+    TCAppL s -> "While checking a function applied to an argument: _" <+> prettyPrec 11 s
+    TCAppR s -> "While checking the argument to a function:" <+> prettyPrec 10 s <+> "_"
 
 -- | A typechecking stack frame together with the relevant @SrcLoc@.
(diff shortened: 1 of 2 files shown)


src/swarm-lang/Swarm/Language/Typecheck.hs
```diff
@@ -892,11 +895,11 @@ infer s@(CSyntax l t cs) = addLocToTypeErr l $ case t of
   -- each time.
   SApp f x -> do
     -- Infer the type of the left-hand side and make sure it has a function type.
-    f' <- infer f
+    f' <- withFrame l (TCAppL x) $ infer f
     (argTy, resTy) <- decomposeFunTy f (Actual, f' ^. sType)
 
     -- Then check that the argument has the right type.
-    x' <- check x argTy
+    x' <- withFrame l (TCAppR f) $ check x argTy
```
xsebek, line 902: I like the comment here better then the LHS/RHS message, as "function argument" doesn't require remembering which side is which. 😅

xsebek (approved):
Nice, the error message for a function argument looks better. 👍 

What happens with `id 3 3`? Is that a RHS of LHS? 🤔

byorgey:
> What happens with `id 3 3`? Is that a RHS of LHS? 🤔

`id 3 3` reports that `id 3` was expected to have a function type but actually has type `Int`, while checking the LHS of the application `_ 3`.  

Maybe an example of what you were thinking of would be something like `"hi" + 2`, which now reports
```
1:1: Type mismatch:
  From context, expected `"hi"` to have type `Int`,
  but it actually has type `Text`

  - While checking the argument to a function: (+) _
  - While checking a function applied to an argument: _ 2
```

PR #2222, opened 2024-12-17
Add an extra `ghcup set` step to install instructions

byorgey:
As suggested on discord by ChucklesTheBeard, it may be necessary to do `ghcup set` after `ghcup install`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,6 +54,11 @@ unreleased features), read on.
 
            ghcup install ghc 9.8.2
 
+    1. Set the newly installed version of GHC as the current version
+       (if not already set):
+
+           ghcup set ghc 9.8.2
+
     1. Use `ghcup` to install `cabal`:
 
            ghcup install cabal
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -54,6 +54,11 @@ unreleased features), read on.
 
            ghcup install ghc 9.8.2
 
+    1. Set the newly installed version of GHC as the current version
+       (if not already set):
+
+           ghcup set ghc 9.8.2
+
```
xsebek, line 61:
```suggestion
           ghcup install ghc 9.8.2 --set
```

xsebek (changes requested):

xsebek (approved):
(I like the change, so the suggestion should be just a suggestion.)

README.md
byorgey: Oh, nice idea, I didn't know about the `--set` option!

Issue #2223, opened 2024-12-19
Handle `read` at user-defined types 


byorgey:
Currently, the `read` command (#2224) does not work with user-defined types.  It used to simply fail; then for a while it completely crashed the entire game; now after merging #2431 it always succeeds, even when it should not, allowing us to exploit the type system:
```
> read "3" : Bool -- this throws an exception, as it should
read: Could not read "3" at type Bool
> tydef Foo = Bool end
> read "3" : Foo   -- this should not succeed, but it does
it0 : Foo = 3
> if (read "3" : Foo) {} {}
Fatal error: Bad application of execConst ...
```
Fundamentally, the problem is that at runtime we no longer have easy access to tydefs that are in scope, so when we encounter a user-defined type like `Num` we don't know what it should expand to.  There are two ways we could potentially solve this: (1) we could keep track of in-scope tydefs at runtime (maybe we already do this somewhere? I forget, but even if so it would still need to be plumbed through to where we execute `read`), or (2) we could expand away any user-defined types in the type of `read` prior to runtime, *e.g.* at typechecking or elaboration time.  However, in that case we have to be careful to use the correct tydefs that are actually in scope.

byorgey:
Another idea: change the syntax of `read` to take a literal type argument, like `read Foo "3"`, rather than relying on a type annotation to provide a type argument implicitly.  Under the hood we actually turn it into something like `read Foo "3"` anyway.

The benefit of doing things this way is that when we see a type which is an argument to `read`, we can specifically substitute away all user-defined types it contains.  Currently, it is difficult to do this specifically for types which end up being annotations on a call to `read`, because the connection between the type annotation and the `read` call is less clear, and mediated by type inference.

One remaining question would be: do we just hard-code this into the parser as a special case, or do we generically allow parsing types as terms, and just check that `read` is applied to a type and a string?  Generically parsing types as terms might very well be ambiguous.  We could also consider adding a required sigil such as `@`, as in `read @Foo "3"`.

xsebek:
> Generically parsing types as terms might very well be ambiguous.

There are types like `Int * Int` or `{Unit}`. I guess `read` could expect the type as first argument similarly to `require` which would be unambiguous. But passing around types as terms sounds fun, so I am good with special syntax marker too. 👍

byorgey:
> I guess `read` could expect the type as first argument similarly to `require` which would be unambiguous. But passing around types as terms sounds fun, so I am good with special syntax marker too. 👍

[...]

Issue #116, opened 2021-09-29
Use `typewriter` device for reifying values as entities


byorgey:
A robot *has entities* in its inventory, and also *has values* referred to by its running program.  It's fun to contemplate ways to blur the lines between the two.  This ticket is really meant more to spur discussion than to represent a specific task that needs to be done (perhaps some specific tasks may spin out of it later).

* As mentioned in #115 , we could imagine simulating recursive data structures with nested boxes.  Wouldn't it be fun if a variable holding a list is actually represented as nested `cons` entities in your inventory?  Surely this isn't a great idea in general---it would be way less efficient---but it's fun to think about.
* On the other hand, it's not crazy to imagine wanting to reify, say, a number as a physical inventory entity, so that you can `give` it to another robot, who could reflect it back to an `int` value.  This would be like a poor man's version of #94 before you have developed the necessary tech to properly send values over channels.
* So perhaps we could have a generic mechanism for reifying and reflecting between values and entities.  Would this be interesting? Fun?  Worthwhile?

See also #455 which is probably a prerequisite.

Anrock:
Honestly, ints as physical entities sounds kinda weird. So far the game seems (at least to me) to be sort of realistic, i.e. robot have to have a specific device to do something, with various mechanics tied to devices like scanners. Summoning values into real world feels like cartoonish magic and out of place.

For poor man's versiof of #94 something like "data port" could be done, which is like antenna with 1 tile radius. Which also can be used for #22

byorgey:
> So far the game seems (at least to me) to be sort of realistic

It's not *that* realistic (and I don't intend it to be): there are lambdas and bits you can harvest! And you make branch predictors out of tree branches!  The world is already full of cartoonish, whimsical magic.

Anrock:
Oh, ok! Well then, pile of numbers and throwing ints around definitely fits this setting.

byorgey:
Hmm, how about a "paper" entity and a way to "print" a value onto a piece of paper and then "read" a value from a piece of paper?  Maybe reifying values into physical objects is not so magical after all... =)

byorgey:
So I'm going to make this concrete: how about designing a new `printer` device, which works as follows:

- It enables a new command `print : a -> cmd ()`.  If you have at least one `paper` entity in your inventory, then one is consumed, and you end up with a `printoutNNN` entity (the `NNN` should be chosen to make it unique, I guess?).
[...]

PR #2224, opened 2024-12-19
New `read` command to act as partial inverse to `format`

byorgey:
Towards #116.  This PR introduces a new built-in command `read`, which can (attempt to) convert text into values, *i.e.* a partial inverse to `format`.

It has type `read : Text -> a` and is a little bit magical, since its behavior depends on the inferred output type.  In particular, if reading at the expected output type succeeds, it will return the value read; otherwise it throws an exception.  In other words, in order to use `read` successfully you must already know what type you expect to get.

At first I gave it the type `Text -> (Unit + a)` and had it return `inl ()` whenever it failed, but then I realized that (1) you can make either version from the other (to convert `Text -> (Unit + a)` to `Text -> a`, `case` on the result and call `fail` if you get `inl`; to convert the other way, use `try`), but (2) it is simpler/easier to have the version that crashes to start and then use a `net` to turn it into the pure functional version if you wish, rather than the other way around (which requires a much harder-to-obtain `ADT calculator`).

From a technical point of view, the way this is achieved is by inserting the type inferred for `read` as an extra parameter during elaboration.  (Having types-as-special-values in the language might be useful for other things down the road as well.)  Note that this makes `read` strictly weaker than the `read` function in Haskell, which works via dictionary-passing.  For `read` in Swarm to work, its concrete type must be statically known at compile time.  For example here is a function that works in Haskell but not Swarm:
```
def testRead : a -> Bool = \x. read (format x) == x end
```
I think it should still be sufficiently usable, however, and the current implementation is much simpler than going to full-fledged type classes.

Currently `read` does not work on functions or delay types (even though such things are supported by `format`).  We may add them in the future, but it would be much more complex since it would require parsing arbitrary Swarm code and dealing properly with environments and closures etc.

`read` also does not currently work with user-defined types; see #2223 .


byorgey:
@kostmo, @xsebek, any thoughts on this?

data/entities.yaml
```diff
@@ -330,6 +330,32 @@
...
+    attr: plant
+    char: 'p'
+  description:
+    - A fast-growing plant with a pungent aroma, often found growing near rocks.
+    - |
+      When equipped as a device, parsley enables the `read`
+      command, which can be used to turn text into values:
+      ```
+      read : Text -> (Unit + a)
+      ```
```
xsebek, line 344: Elsewhere, I see `Read -> [tyQ| Text -> a |]` so this is probably outdated, right? :)

I think it would be fair to call out the limitations of `read`, like "unlike format, you should stick to specifying type explicitly and only using it for basic types, not debugging functions."

src/swarm-lang/Swarm/Language/Parser/Value.hs
```diff
@@ -0,0 +1,148 @@
...
+-- |
+-- SPDX-License-Identifier: BSD-3-Clause
+--
+-- Parser for values of the Swarm language, indexed by type.
+--
+-- Note that this is completely separate from the main Swarm language
+-- parser in the other Swarm.Language.Parser.* modules.  Parsing
+-- *values* only ever happens at runtime (e.g. to implement the 'read'
+-- command), and does not need to deal with comments, source position
+-- tracking, etc.
```
xsebek, line 13: Why is a separate parser necessary? I do not object to having a simpler one as well, but it would be useful to have the technical problem with the full parser documented here. 🤔

data/worlds/classic.world
```diff
@@ -75,6 +75,7 @@ overlay
     [ {stone}
     , mask (hash % 10 == 0) {stone, rock}
     , mask (hash % 100 == 0) {stone, lodestone}
+    , mask (hash % 300 == 0) {grass, parsley}
```
xsebek, line 78: A stray world test? 🙂

xsebek (approved):
I am all for having `format` and `read`. 👍 I think I had an `if (n == "0") ...` in some scenario at one point, so having a simple parser will be nice.

data/entities.yaml
byorgey: Good catch, I forgot to update the `parsley` description when I changed the type of `read`!  I also added an extra caveat about being unable to handle functions or delay types.

data/worlds/classic.world
byorgey: No, `parsley` is what provides the `read` capability, so I added some to grow in the world!

Fun English trivia: according to Wikipedia, parsley is native to Greece, and the English word "parsley" ultimately is derived from the Greek "petroselinon" meaning "rock-celery", so-called probably because it tends to grow on cliffs and among rocks.  That's why I added it to the biome with all the rocks.  :rock: :seedling: :rock: :smile:

src/swarm-lang/Swarm/Language/Parser/Value.hs
byorgey: Hmmm.... good question, let me think about this for a bit.  The alternative, I suppose, would be to use the full parser but then have a separate step afterwards to take an arbitrary term and check whether it actually looks like a value of the proper type...

byorgey: Thanks for bringing this up, I reimplemented it using the full swarm-lang parser followed by checking the type and then ensuring the parsed term is a value.  This turns out to be a lot simpler and gives a much clearer pathway forward for eventually being able to handle things like delay, function, and command types.

PR #2229, opened 2024-12-26
Fix shape recognition orientation edge case

kostmo:
This PR augments the inner map key (from `Cosmic Location` to `(Cosmic Location, AbsoluteDir)`) for the `foundByName` field in the recognition registry to account for a shape that may have two non-overlapping orientations placed at the same coordinate.

## Testing
```
scripts/test/run-tests.sh --test-options '--pattern "structure"'
```
## Other changes and refactoring

* Enhanced logging: add a new constructor for logging that indicates which shape (if any) was recognized
* Extract the inner logic from `entityModified` into a function `entityModifiedLoggable` that does not dictate the logging structure
* Introduce a new `RecognitionActiveStatus` parameter that allows API users to temporarily disable recognition of "contained" structures while constructing larger structures
* Inline the `registerBestStructureMatch` function


byorgey (approved):
LGTM!

PR #2233, opened 2024-12-26
Update to allow GHC 9.10 and drop GHC 9.2

### LINKED ISSUE #1421: Update uses of `foldl'` to prepare for export from `Prelude`

byorgey:
- Allow `base-4.21` and `template-haskell-2.22`
- Update CI + mergify config
- Update imports of `foldl'`: closes #1421 
- Update to testing with GHC-9.6.6 instead of 9.6.5, but stay with 9.8.2 instead of 9.8.4 since it's supported by HLS

### DIFF
--- a/.github/mergify.yml
+++ b/.github/mergify.yml
@@ -25,10 +25,10 @@ queue_rules:
           - or:
             - check-success=HLint
             - -files~=\.hs$
+          - check-success=Haskell-CI - Linux - ghc-9.10.1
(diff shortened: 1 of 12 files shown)


byorgey:
There are a bunch of other dependencies that could be updated too, but I decided to leave them for another PR.  See #2235 .

kostmo (approved):

PR #2237, opened 2024-12-27
Validate hex colors in `FromJSON` instance

### LINKED ISSUE #2236: Unparseable custom attribute colors cause a crash
**Describe the bug**
In a scenario description, if the `fg` or `bg` color specification in a custom attribute cannot be parsed, swarm crashes.

**To Reproduce**

Try loading the following scenario `.yaml` file:

```
version: 1
name: Custom attribute - invalid color
description: |
  Invalid color in custom attribute
creative: false
attrs:
  - name: bogus
    bg: "this is not a color"
terrains:
  - name: whatzit
    attr: bogus
    description: |
      A thingy
robots:
  - name: base
    dir: east
world:
  dsl: |
    {grass}
  palette:
    'B': [grass, null, base]
    '.': [whatzit]
  upperleft: [0, 0]
  map: |
    B.
```

This results in:

```
$ ./scripts/play.sh -i Testing/_Validation/unparseable-attr-color.yaml
swarm: Data.Colour.SRGB.sRGB24read: no parse
  error, called at ./Data/Colour/SRGB.hs:128:3 in colour-2.3.6-01a45f2577c459ed9632a48ba99a66e313619b5bf990fc54e630e96fb26912b8:Data.Colour.SRGB
```

Note that the scenario successfully validates and loads; the error is thrown sometime later, perhaps when it is being rendered.

**Expected behavior**
The scenario should fail to validate with a reasonable error message.

byorgey:
Fixes #2236.  Stores `Kolor` instead of `Text` inside a `HexColor`, and validates and parses it when reading from JSON.  This simplifies the code in multiple places where we were taking the `Text` out of a `HexColor` and parsing it.

### DIFF
--- a/src/swarm-scenario/Swarm/Game/Scenario/Style.hs
+++ b/src/swarm-scenario/Swarm/Game/Scenario/Style.hs
@@ -1,3 +1,5 @@
+{-# LANGUAGE ViewPatterns #-}
+
 -- |
 -- SPDX-License-Identifier: BSD-3-Clause
 --
@@ -7,9 +9,9 @@ module Swarm.Game.Scenario.Style where
 import Codec.Picture (PixelRGBA8 (..))
 import Data.Aeson
 import Data.Colour.Palette.Types (Kolor)
-import Data.Colour.SRGB (RGB (..), sRGB24read, toSRGB24)
+import Data.Colour.SRGB (RGB (..), sRGB24reads, toSRGB24, sRGB24show)
+import Data.Colour.SRGB.Linear (toRGB)
 import Data.Set (Set)
-import Data.Text (Text)
 import Data.Text qualified as T
 import GHC.Generics (Generic)
 import Swarm.Game.Entity.Cosmetic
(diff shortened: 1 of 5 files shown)


src/swarm-scenario/Swarm/Game/Scenario/Style.hs
```diff
@@ -38,17 +40,36 @@ instance FromJSON StyleFlag where
...
+  -- There is no Ord instance for Colour a, but we need one to use
+  -- with OccurrenceEncoder, so we make our own.
+  --
+  -- We use toRGB here since it does no conversions whatsoever, it
+  -- simply unpacks the raw color data into an RGB triple.  For the
+  -- purposes of an Ord instance, it doesn't matter: we just want a
+  -- consistent way to put a total ordering on colors as fast as
+  -- possible.
+  compare (HexColor (toRGB -> RGB r1 g1 b1)) (HexColor (toRGB -> RGB r2 g2 b2)) =
+    compare r1 r2 <> compare g1 g2 <> compare b1 b2
```
kostmo, line 58: would it be simpler to use [3-tuples](https://hackage.haskell.org/package/ghc-prim-0.13.0/docs/GHC-Tuple.html#t:Ord)?
```
compare (r1, g1, b1) (r2, g2, b2)
```

src/swarm-scenario/Swarm/Game/Scenario/Style.hs
```diff
@@ -38,17 +40,36 @@ instance FromJSON StyleFlag where
...
+
+instance ToJSON HexColor where
+  toJSON = toJSON . T.pack . sRGB24show . getHexColor
 
 instance ToPixel HexColor where
-  toPixel (HexColor colorText) = PixelRGBA8 r g b 255
+  toPixel (HexColor kolor) = PixelRGBA8 r g b 255
    where
-    temp :: Kolor
-    temp = sRGB24read $ T.unpack colorText
```
kostmo, line 50: Maybe want to ban `sRGB24read` in `.hlint.yaml` since it is partial

kostmo (approved):
Nice improvement.

src/swarm-scenario/Swarm/Game/Scenario/Style.hs
byorgey: Good idea, done.

Issue #2240, opened 2024-12-28
Overridden standard entities are still suggested as a means to provide a missing capability


byorgey:
**Describe the bug**
If a standard entity has been overridden by a custom entity with the same name, the standard entity will still be suggested if it could provide a missing capability, even if the custom entity does not provide that capability.

**To Reproduce**

Once #2241 is merged, load this scenario:

```
version: 1
name: Overridden entity capabilities
description: |
  Overridden standard entity should not still be suggested for its
  capabilities. The error message should suggest "- tank treads", not
  "- treads or tank treads".
creative: false
entities:
  - name: treads
    display:
      char: '#'
      attr: red
    description:
      - Broken treads
    properties: [known, pickable]
robots:
  - name: base
    dir: east
    devices:
      - logger
      - treads
solution: |
  move
world:
  dsl: |
    {grass}
  palette:
    'B': [grass, null, base]
  upperleft: [0, 0]
  map: |
    B
```

Then try to execute `move`.  The following error message is displayed:
```
You do not have the device required for:
  'move'
  Please equip:
  - treads or tank treads
```
Note that it is correct that we get an error message: even though `treads` are equipped, it is a custom entity that does not provide the `move` capability.  The problem is that `treads` are still suggested in the error message.

**Expected behavior**
The error message should mention only `tank treads` as a possibility for providing the `move` capability.

**Additional context**
Note that this depends on fixing #2239 first, i.e. #2241.

I believe the problem lies in the `Semigroup` instance for `EntityMap`.  If any entities are overridden, they must be appropriately removed from the map in the `entitiesByCap` field.

PR #2241, opened 2024-12-28
Make custom entities override built-in entities

### LINKED ISSUE #2239: Defining custom entities with the same name as an existing built-in entity has no effect
**Describe the bug**
If in a scenario description you define a custom entity whose name is the same as the name of a standard built-in entity, the custom entity definition has no effect.  Referring to that name will continue to reference the built-in entity.

**To Reproduce**

```
version: 1
name: Custom entity
description: |
  Custom entity overriding standard entity
creative: false
entities:
  - name: rock
    display:
      char: '$'
      attr: gold
    description:
      - Not your average rock!
robots:
  - name: base
    dir: east
    devices:
      - scanner
known: [rock]
world:
  dsl: |
    {grass}
  palette:
    'B': [grass, null, base]
    '$': [grass, rock]
  upperleft: [0, 0]
  map: |
    B$
```

Load the above scenario, `scan forward`, and then view the `rock` in your Compendium.  It is a standard, grey rock displayed with `o` and the standard description.

**Expected behavior**
The only sensible behaviors I can see would be to either
1. allow custom entities to override existing standard entities, OR
2. yield an error and refuse to validate the scenario.

Preferably (1).  Silently ignoring the custom entity seems like a bad default.

**Screenshots**

2024-12-27-195021_956x348_scrot

byorgey:
Fixes #2239.  The comments on the `Semigroup` instance for `EntityMap` *say* "it is possible for the latter 'EntityMap' to override members of the former with the same name.  This replacement happens automatically with 'Map'...", however, this was not correct.  The `Semigroup` instance for `Map` calls `Map.union`, which is *left*-biased (https://hackage.haskell.org/package/containers-0.7/docs/Data-Map-Lazy.html#t:Map).  So we want `n2 <> n1` here instead of `n1 <> n2`, so that the second `EntityMap` properly overrides the first.

Note that `c1 <> c2` is not correct either, for a more subtle reason.  In the case of overrides it needs to be manually adjusted, in a similar way to what is already done for `d1`; see #2240.  I will address that in a separate PR.

The attached scenario will succeed only if the `rock` entity is properly overridden so that it can provide the `move` capability.

Also had to apply a couple minor fixes to the `pied-piper` scenario, which it turns out tried to define custom entities named `key` and `wall` which were silently being ignored in preference to the standard entities of the same name; but then the solution inadvertently ended up depending on a property of the standard `key` (namely `pickable`) which the custom one did not have.

### DIFF
--- a/src/swarm-scenario/Swarm/Game/Entity.hs
+++ b/src/swarm-scenario/Swarm/Game/Entity.hs
@@ -424,15 +424,15 @@ data EntityMap = EntityMap
   }
   deriving (Eq, Show, Generic, ToJSON)
 
--- |
--- Note that duplicates in a single 'EntityMap' are precluded by the
--- 'buildEntityMap' function.
--- But it is possible for the latter 'EntityMap' to override
--- members of the former with the same name.
--- This replacement happens automatically with 'Map', but needs
--- to be explicitly handled for the list concatenation
--- of 'entityDefinitionOrder' (overridden entries are removed
--- from the former 'EntityMap').
+-- | Right-biased union of 'EntityMap's.
+--
+--   Note that duplicates in a single 'EntityMap' are precluded by the
+--   'buildEntityMap' function.  But it is possible for the right-hand
+--   'EntityMap' to override members of the left-hand with the same name.
+--   This replacement happens automatically with 'Map', but needs to
+--   be explicitly handled for the list concatenation of
+--   'entityDefinitionOrder' (overridden entries are removed from the
+--   former 'EntityMap').
 instance Semigroup EntityMap where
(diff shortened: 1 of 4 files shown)


src/swarm-scenario/Swarm/Game/Entity.hs
```diff
@@ -436,7 +436,7 @@ data EntityMap = EntityMap
 instance Semigroup EntityMap where
```
kostmo, line 436: Should we use the term "right-biased" in the docs?

kostmo (approved):

src/swarm-scenario/Swarm/Game/Entity.hs
byorgey: Good idea, done.

PR #2242, opened 2024-12-28
Fix Semigroup EntityMap instance to update Capabilities map properly when there are overridden devices

### LINKED ISSUE #2240: Overridden standard entities are still suggested as a means to provide a missing capability

byorgey:
Fixes #2240.  Fixes the `Semigroup EntityMap` instance so that devices which are overridden are removed from the lists of devices providing certain capabilities.

- Also adds a test scenario which generated an incorrect error message before this fix.
- Also adds the test scenario from #2241 to the test suite which was missed previously.

### DIFF
--- a/src/swarm-scenario/Swarm/Game/Entity.hs
+++ b/src/swarm-scenario/Swarm/Game/Entity.hs
@@ -104,7 +104,6 @@ import Data.Char (toLower)
 import Data.Either.Extra (maybeToEither)
 import Data.Foldable (Foldable (..))
 import Data.Function (on)
-import Data.Functor.Identity
 import Data.Hashable
 import Data.IntMap (IntMap)
 import Data.IntMap qualified as IM
@@ -448,17 +447,7 @@ instance Semigroup EntityMap where
     notOverridden = (`M.notMember` n2) . view entityName
     removeOverriddenDevices (Capabilities m) =
       Capabilities
-        . runIdentity
-        -- traverseMaybeWithKey allows us in one operation to filter
-        -- the NonEmpty list of devices for each capability, turn the
-        -- resulting lists into Maybe NonEmpty, and then remove from
-        -- the map any that became Nothing.  However, it is more
-        -- general than we need: it gives us access to the key which
-        -- we don't need (this explains the call to 'const'), and runs
-        -- in an arbitrary Applicative which we also don't need (this
-        -- explains the Identity).
-        . M.traverseMaybeWithKey
-          (const (Identity . NE.nonEmpty . NE.filter (notOverridden . device)))
+        . M.mapMaybe (NE.nonEmpty . NE.filter (notOverridden . device))
         $ m
 
 instance Monoid EntityMap where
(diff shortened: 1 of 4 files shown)


src/swarm-scenario/Swarm/Game/Entity.hs
```diff
@@ -429,16 +430,36 @@ data EntityMap = EntityMap
...
+        -- traverseMaybeWithKey allows us in one operation to filter
+        -- the NonEmpty list of devices for each capability, turn the
+        -- resulting lists into Maybe NonEmpty, and then remove from
+        -- the map any that became Nothing.  However, it is more
+        -- general than we need: it gives us access to the key which
+        -- we don't need (this explains the call to 'const'), and runs
+        -- in an arbitrary Applicative which we also don't need (this
+        -- explains the Identity).
+        . M.traverseMaybeWithKey
+          (const (Identity . NE.nonEmpty . NE.filter (notOverridden . device)))
```
kostmo, line 461: Instead of placing this verbose explanation inline here, shall we expose a simplified version of `traverseMaybeWithKey` in a util class?  Perhaps call the function `filterMaybes`?

kostmo (approved):

src/swarm-scenario/Swarm/Game/Entity.hs
byorgey: That's a good idea, I'll do that.

byorgey: Oh! It turns out it already exists in `Data.Map`, it's called `mapMaybeWithKey`.  Don't know why I didn't find that one before.

PR #2245, opened 2024-12-29
Add `print` + `erase` commands, and a challenge scenario showing off `print`

byorgey:
This PR is a complement/follow-up to #2224 .

- Add a new `print : Text -> Cmd Text` command which consumes a `paper` entity and produces an entity named `paper: <text>` where `<text>` is the given text to be printed on the paper
- Add an `erase : Text -> Cmd Unit` command which takes a printed-on paper entity and erases it back to a plain `paper`
- Add a challenge scenario necessitating the use of `print`

Especially eager to hear feedback from anyone who wants to try solving the new scenario!

```
./scripts/play.sh --scenario data/scenarios/Challenges/telephone.yaml
```


kostmo:
> Especially eager to hear feedback from anyone who wants to try solving the new scenario!

Am not sure how to obtain the `water` to make the `paper`, since I can't wade into the water to `grab` it.

byorgey:
> Am not sure how to obtain the water to make the paper

@kostmo maybe you need to refresh yourself with `Tutorials/require.yaml` :smile:

kostmo:
Again, I've abstained from peeking at the solution, but now I'm stuck on how to `give` the `paper` to the shuttle robot.  I lack `copper` or `iron` to craft a `grabber`.

byorgey:
> Again, I've abstained from peeking at the solution, but now I'm stuck on how to `give` the `paper` to the shuttle robot.  I lack `copper` or `iron` to craft a `grabber`.

Ah, you're not supposed to `give` it, just drop it at the terminal (in the center of the square at the end of the track) and it will pick it up.  I'll try to change the description to make this more clear.

src/swarm-engine/Swarm/Game/Step/Const.hs
```diff
@@ -1214,6 +1214,30 @@ execConst runChildProg c vs s k = do
...
+        robotInventory %= delete paper
+        robotInventory %= insert (paper & entityName .~ newEntityName)
+        return $ mkReturn newEntityName
+      _ -> badConst
+    Erase -> case vs of
+      [VText paperName] -> do
+        toErase <- ensureItem paperName "erase"
+        (paperName /= "paper")
+          `holdsOrFail` ["That is already blank!"]
+        ("paper: " `T.isPrefixOf` paperName)
```
kostmo, line 1230: It seems fragile in some way to use an "in-band" indicator for erasability and printability, in that it must be inferred by the entity's name.  Perhaps instead we want to define a "printable" property?  Then we could have different printable entities like "papyrus", "clay", "whiteboard".  Each of them can still be subject to the name transformation with the colon.  A property called "printable" would be inherent to the "unmarked" entity, whereas the property called "erasable" would be dynamically set once the entity undergoes a "print".

kostmo (approved):
Nice challenge.  [Here's my solution](https://gist.github.com/kostmo/3171b0cba23b7af258588e3ec0f32a88).

src/swarm-engine/Swarm/Game/Step/Const.hs
byorgey: Oh, that's a nice idea.  Let me play with that.

src/swarm-engine/Swarm/Game/Step/Const.hs
```diff
@@ -1215,27 +1215,30 @@ execConst runChildProg c vs s k = do
...
-        paper <-
-          lookupEntityName "paper" em
-            `isJustOrFail` ["Can't erase, I don't know what paper is."]
+        erased <-
+          lookupEntityName baseName em
+            `isJustOrFail` ["I've never heard of", indefiniteQ baseName <> "."]
+        (erased `hasProperty` Printable)
+          `holdsOrFail` ["You cannot erase", indefinite baseName <> "!"]
+        (not $ T.null printedMatter)
+          `holdsOrFail` ["That is already blank!"]
```
byorgey, line 1238: Should it be an error when you try to `erase` an already-blank, printable item?  Or should it just be a no-op?

src/swarm-engine/Swarm/Game/Step/Const.hs
byorgey: @kostmo What do you think?  I decided to just go with a single property `printable` that determines whether something can be both printed and erased.  Having a separate property `erasable` didn't seem to make sense to me since it's not something that would make sense to set on base entities, and the only reason to `erase` something is that you previously `print`ed on it.

byorgey: I guess this does mean that you can still `print` on something that has been previously `print`ed on, like this:
```
> print "paper" "3"
it0 : Text = "paper: 3"
> print "paper: 3" "7"
it1 : Text = "paper: 3: 7"
```
That's a bit strange, perhaps, but I don't think it's really a problem.

src/swarm-engine/Swarm/Game/Step/Const.hs
kostmo: I think a no-op makes sense.  Physically, what happens when you erase a blank paper?  It's still blank.

data/scenarios/Challenges/telephone.yaml
```diff
@@ -0,0 +1,223 @@
...
+description: |
+  Give another robot instructions to duplicate a pattern.
+creative: false
+objectives:
+  - id: 'paper'
+    teaser: Make some paper
+    goal:
+      - |
+        As part of a scheme to prove your intelligence to anyone
+        watching from space, the mystical design of X's and O's in the
```
kostmo, line 13:
```suggestion
        watching from space, the mystical geoglyph of X's and O's in the
```

data/scenarios/Challenges/telephone.yaml
```diff
@@ -0,0 +1,223 @@
...
+    bg: '#002f00'
+entities:
+  - name: mountain
+    display:
+      attr: snow
+      char: 'A'
+      priority: 9
+    description:
+      - An impassably tall mountain.
+    properties: [unwalkable, opaque]
```
kostmo, line 138: why did we need to redefine the `mountain` entity?

data/scenarios/Challenges/telephone.yaml
```diff
@@ -0,0 +1,223 @@
...
+    #.............................&P==========================AA====================S=======t&..............^.....TTTT.....#
+    #.............................&&&........................AAAA..........................&&&.....................TT......#
+    #..........................................................AAA.........................................................#
+    #...........................................................AAAA.........................................WWW...........#
+    #..........................................................AAAAAA........................................WW~W..........#
+    #...........................................................AAAA.......................................WW~WW...........#
+    #..........................................................AAAA..........................................WW............#
+    #.........................................................AAA..........................................................#
+    ########################################################################################################################
```
kostmo, line 223: While I was playing, I kind of wished that I could see the effect in on the left side without having to manually scroll the viewport.  What do you think about reducing the width of the map?

src/swarm-engine/Swarm/Game/Step/Const.hs
```diff
@@ -1214,6 +1214,33 @@ execConst runChildProg c vs s k = do
...
+          lookupEntityName baseName em
+            `isJustOrFail` ["I've never heard of", indefiniteQ baseName <> "."]
+        (erased `hasProperty` Printable)
+          `holdsOrFail` ["You cannot erase", indefinite baseName <> "!"]
+        not (T.null printedMatter)
+          `holdsOrFail` ["That is already blank!"]
+
+        robotInventory %= delete toErase
+        robotInventory %= insert erased
+        return $ mkReturn ()
```
kostmo, line 1242: Should this return the new entity name?

src/swarm-lang/Swarm/Language/Parser/Value.hs
```diff
@@ -20,7 +24,8 @@ import Text.Megaparsec qualified as MP
 
 readValue :: Type -> Text -> Maybe Value
 readValue ty txt = do
-  s <- eitherToMaybe $ readNonemptyTerm txt
+  txt' <- T.stripPrefix "paper:" txt <|> pure txt
```
kostmo, line 27: Can we avoid hard-coding this to `paper`?

src/swarm-lang/Swarm/Language/Syntax/Constants.hs
```diff
@@ -273,6 +273,10 @@ data Const
     Format
   | -- | Try to turn a string into a value
     Read
+  | -- | Print a string onto a piece of paper
```
kostmo, line 276:
```suggestion
  | -- | Print a string onto a printable surface
```

kostmo (commented):
Added a few suggestions

byorgey:
@kostmo thanks!  Also, I saw you asked why we had to redefine `mountain` but then removed it.  Presumably you figured out that it was to make the `mountain`s get drawn on top of the shuttle, so it looks like it's going through a tunnel.  However, I wondered about the `opaque` property.  Currently it only affects the `scout` command, but I wonder about making `opaque` also imply that the entity should be drawn with a higher priority than robots?  But perhaps having `opaque` mean several different things is just asking for trouble.

kostmo:
> but I wonder about making `opaque` also imply that the entity should be drawn with a higher priority than robots? But perhaps having `opaque` mean several different things is just asking for trouble.

I would be in favor of having the `opaque` property affect the display.  It is consistent with the physical interpretation w.r.t. the `scout` command.

src/swarm-lang/Swarm/Language/Parser/Value.hs
byorgey: Good catch, I missed this.  However, the right way to generalize it is not clear to me.  Since this lives in `swarm-lang` we can't get a list of entities and their properties.  A simple idea would be to just drop everything until the first colon, but colons can legitimately show up in `Text` values, for example `read "\"hi: there\"" : Text` should still work...

byorgey: OK, I hacked something together that strips up to the first colon, UNLESS the first colon is preceded by a double quote.  Ugly and possibly not quite correct but I think it will work.

PR #2247, opened 2024-12-29
generic cell modification

kostmo:
Makes the `classifyModification` function generic with type parameters, so that it can be moved "up" to the `swarm-topography` sublibrary.


byorgey (approved):
Looks good, though it seems even more general than that - it is really just about classifying the ways any value changes, not just cells.  I.e. it could live in swarm-util.

kostmo:
> Looks good, though it seems even more general than that - it is really just about classifying the ways any value changes, not just cells. I.e. it could live in swarm-util.

Good point.  I may consider merging `Swarm.Game.Scenario.Topography.Modify` and `Swarm.Game.Scenario.Topography.Terraform` in a follow up.

PR #2249, opened 2024-12-30
Counting flowers

byorgey:
Another challenge scenario to showcase the use of `print`/`read`.

### DIFF
--- a/data/scenarios/Challenges/_flower-count/judge.sw
+++ b/data/scenarios/Challenges/_flower-count/judge.sw
@@ -1,0 +1,35 @@
+def win =
+  try {
+    firestarter <- robotNamed "firestarter";
+    halt firestarter
+  } {};
+  try {
+    fire <- robotNamed "fire";
+    halt fire; wait 1;
+    reprogram fire { selfdestruct };
+  } {};
+  create "gold"
+end
+
+def judgeCount : Int -> Cmd Unit = \actual.
+  watch down;
+  wait 1024;
+  s <- scan down;
+  case s
+    (\_. return ())
+    (\p.
+      try {
+        let c = (read p : Int) in
+        if (c == actual) { win } {}
+      } {}
+    )
+end
+
+def forever = \c. c; forever c end
+
+def judge =
+  numFlowers <- resonate "flower" ((-59,-19),(60,20));
+  forever (judgeCount numFlowers);
+end;
+
+judge
--- a/data/scenarios/Challenges/flower-count.yaml
+++ b/data/scenarios/Challenges/flower-count.yaml
@@ -9,12 +9,12 @@ objectives:
     teaser: Count the flowers!
     goal:
       - |
-        Your evil nemesis, the sadistic supervillain Dr. Nemesis, is
-        at it again!  They have wired up some explosives to your True
-        Love and lit the fuse, but promised to stop it... IF you can
+        Your evil nemesis, the sadistic arithmomaniac supervillain Count Nemesis, is
+        at it again!  They have wired up some explosives to your
+        `True Love`{=robot} and lit the fuse, but promised to stop it... IF you can
         count all the flowers in time!
       - |
-        Specifically, Dr. Nemesis demands that you count the
+        Specifically, Count Nemesis demands that you count the
         total number of flowers in the 120x40 field bounded by walls,
         `print` the number on a piece of `paper`{=entity}, and place
         the paper at the origin, `(0,0)`.  If the number is correct,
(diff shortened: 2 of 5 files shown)


kostmo:
In order to provide useful playtesting feedback, I have not inspected the yaml file for the counting flowers scenario.  I haven't solved it yet, but here's what I've tried so far:
* Using the `typewriter` to `print` something (I get a `print: what is a paper` error)
* `meet`ing the True Love and `give`-ing them the `typewriter`
* `grab`-ing a `flower`
* `grab`-ing a `fuse`
* `unequip`ing each device in the inventory to see if I can `make` some other useful device, e.g. to facilitate the `resonate` command.  I inspected each equipped device to see their recipes.  The farthest I got on this front was to McGyver a `hyperloop` from the `strange loop` and `ADT calculator`.

My main idea would have been to `print` numbers sequentially, since the upper bound is <5000.  Visually the density looked around 10%, so perhaps it would have gone only 500 iterations.  But I am stuck at the `paper` error.

byorgey:
@kostmo Oh, sorry, this PR is not ready for review yet!  It is unsolvable since there is still more stuff I need to add.  I'll let you know when it's ready for you to give it a try.

byorgey:
> My main idea would have been to print numbers sequentially, since the upper bound is <5000. Visually the density looked around 10%, so perhaps it would have gone only 500 iterations.

Oh, hmm, I will have to think how to rule out this unintended solution. :smile:

byorgey:
@kostmo OK, this should be ready for test playing!  There's still one more thing I want to iron out but it's only cosmetic.

byorgey:
This is now all rebased and fixed up on top of the changes in #2245, should be ready for review.

data/scenarios/Challenges/flower-count.yaml
```diff
@@ -0,0 +1,177 @@
...
+        True Love a flower!
+    condition: |
+      truelove <- robotNamed "truelove";
+      as truelove { has "flower" }
+solution: |
+  run "scenarios/Challenges/_flower-count/solution.sw"
+robots:
+  - name: base
+    dir: north
+    devices:
```
kostmo, line 84: I wanted to programmatically signal the `base` robot when it was time to finally pick the `flower`.  I ended up just measuring the required `wait` time empirically, but if I possessed a `counter`, I could have used the built robots to each contribute something to my inventory when they were finished, and observe when it got to `12`.

data/scenarios/Challenges/_flower-count/judge.sw
```diff
@@ -0,0 +1,35 @@
...
+        let c = (read p : Int) in
+        if (c == actual) { win } {}
+      } {}
+    )
+end
+
+def forever = \c. c; forever c end
+
+def judge =
+  numFlowers <- resonate "flower" ((-59,-19),(60,20));
```
kostmo, line 31: I had a (off-by-one) bug in my code that I finally diagnosed by entering `creative` mode to use my own `resonate` command.  Perhaps that is the intended use of `creative` mode, but it may be fun to have an "in-scenario" method to double-check one's count.  Perhaps some oblique way to obtain a [`gradiator`](https://github.com/swarm-game/swarm/blob/83350513c7a4e16df591fac7da601be658baa0d5/data/scenarios/Challenges/Sliding%20Puzzles/3x3.yaml#L156) that is an instant-lose condition, but nonetheless can help a player debug.

data/scenarios/Challenges/flower-count.yaml
```diff
@@ -0,0 +1,177 @@
...
+      - scanner
+      - strange loop
+      - typewriter
+      - 3D printer
+      - branch predictor
+      - clock
+      - GPS receiver
+      - compass
+    inventory:
+      - [12, solar panel]
```
kostmo, line 105: These counts were a useful hint.

data/scenarios/Challenges/flower-count.yaml
```diff
@@ -0,0 +1,177 @@
...
+    teaser: Count the flowers!
+    goal:
+      - |
+        Your evil nemesis, the sadistic supervillain Dr. Nemesis, is
+        at it again!  They have wired up some explosives to your True
+        Love and lit the fuse, but promised to stop it... IF you can
+        count all the flowers in time!
+      - |
+        Specifically, Dr. Nemesis demands that you count the
+        total number of flowers in the 120x40 field bounded by walls,
```
kostmo, line 18: Let's motivate this objective with [arithmomania](https://en.wikipedia.org/wiki/Arithmomania#Folklore)?

kostmo (approved):
Here's [my solution](https://gist.github.com/kostmo/8b03a43dd2d441b57e6ebdf93d27832b).  I enjoyed having the fuse as time pressure.

PR #2257, opened 2025-01-03
atlas entity

kostmo:
* Created atlas entity
* Created recipe for it
* Replaced all usages of the index-based `waypoint` command with the list-based `waypoints`

### DIFF
--- a/data/entities.yaml
+++ b/data/entities.yaml
@@ -1183,11 +1183,10 @@
     char: 'a'
   description:
     - |
-      An atlas enables the `waypoint` command: `waypoint : Text -> Int -> Cmd (Int * (Int * Int))`.
-      Waypoints are keyed by name and index; the count of waypoints with the same name will be returned
-      along with the current index's location.
+      An atlas enables the `waypoints` command: `waypoints : Text -> (rec l. Unit + (Int * Int) * l)`.
+      Waypoints are keyed by name.
   properties: [pickable]
-  capabilities: [waypoint]
+  capabilities: [waypoints]
 - name: GPS receiver
   display:
     attr: device
--- a/data/entities.yaml
+++ b/data/entities.yaml
@@ -1177,6 +1177,17 @@
       all the nearby actors other than oneself.
   properties: [pickable]
   capabilities: [meet, meetAll]
+- name: atlas
+  display:
+    attr: device
+    char: 'a'
+  description:
+    - |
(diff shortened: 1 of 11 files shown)


data/entities.yaml
```diff
@@ -1223,6 +1223,17 @@
       all the nearby actors other than oneself.
   properties: [pickable]
   capabilities: [meet, meetAll]
+- name: atlas
+  display:
+    attr: device
+    char: 'a'
+  description:
+    - |
+      An atlas enables the `waypoint` command: `waypoint : Text -> Int -> Cmd (Int * (Int * Int))`.
```
byorgey, line 1232: Can I interest you in taking this opportunity to switch to something like `waypoints : Text -> rec l. Unit + (Int * Int) * l` instead?  The current type has always seemed to me like a strange workaround for not having lists.  I know that might be a bunch of work to convert everything, but I'm willing to help.

kostmo: I'm willing to switch to lists, if you can do the heavy lifting :)

byorgey: OK, let me work on a PR.

data/entities.yaml
```diff
@@ -1150,6 +1150,16 @@
...
   properties: [pickable]
   capabilities: [meet, meetAll]
+- name: atlas
+  display:
+    attr: device
+    char: 'a'
+  description:
+    - |
+      An atlas enables the `waypoints` command: `waypoints : Text -> (rec l. Unit + (Int * Int) * l)`.
+      This gets the x, y coordinates of the every waypoint with a given name.
```
byorgey, line 1160:
```suggestion
      This gets the x, y coordinates of every waypoint with a given name.
```

data/scenarios/Testing/836-pathfinding/_836-automatic-waypoint-navigation/rabbit.sw
```diff
@@ -1,16 +1,27 @@
...
       } {};
     } {};
     watch down;
-    let nextWaypointQuery = waypoint "wp" nextWpIdx in
-    teleport self $ nextWaypointQuery;
-    wait 1000;
-    visitNextWaypoint $ nextWpIdx + 1;
+
+    // Wrap around
+    let myList = case remainingList (\_. originalList) (\_. remainingList) in
```
byorgey, line 13: Ah, I like it, this is a clever way to use the list of waypoints incrementally without having to do an O(n) index operation every time.

byorgey (approved):

PR #2260, opened 2025-01-04
Make `structures`, `waypoints`, and `tagmembers` commands return lists

byorgey:
Change `structure`, `waypoint`, and `tagmembers` to return lists; also make `waypoint`, `tagmembers` and `hastag` pure functions, since their output within a given scenario will never change.

More specifically:

- Changes the type of `waypoint : Text -> Int -> (Int * Int)` to just return the waypoint with the given index (wrapping around), or throw an exception if there are no waypoints with the given name.
- Adds a new `waypoints : Text -> (rec l. Unit + Text * l)` function that returns the list of all waypoints with a given name.
- Change `structure` to `structures : Text -> Cmd (rec l. Unit + (Int * Int) * l)`, returning the list of all structure locations of the given name
- Change `tagmembers` to have type `tagmembers : Text -> (rec l. Unit + Text * l)`
- Change `hastag` to have type `hastag : Text -> Text -> Bool`

In the vast majority of cases, this greatly simplified the code for various scenario setups and solutions, though there is still a bit of code duplication for things like calculating the length of a list.  But even this can easily be improved once we have #495 and can make a standard library of list functions and import it within scenario code.

The reason I kept an indexing function for waypoints but not for structures or tags is that iterating through waypoints incrementally while doing other stuff in between seems like a common thing to want to do, and passing around a list of waypoints and manually indexing into it sounds tedious.  However, for structures and tags this is much less common; typically we just want to get a list and then do something with all of them right away.


byorgey:
@kostmo There's obviously still a bunch of work to do to fix up more failing tests, but I'm curious if you have any feedback at this point.  Does this seem like a reasonable direction?

byorgey:
However, I still don't really like the conceit of always returning a pair of the total number of waypoints and the waypoint with the requested index.  What about providing two functions: `waypoints : Text -> List (Int * Int)` and `waypoint : Text -> Int -> Int * Int` which does wrap-around indexing into the list for you (and crashes if the list is empty)?

Relatedly, @kostmo, do you know of a good reason why `tagmembers` and `waypoint` need to be in the `Cmd` monad?  `structures` I understand --- the structures you get can change depending on what entities get placed, so the sequencing matters.  However, aren't the results of `waypoint` and `tagmembers` fixed and deterministic within a given scenario?  It seems to me like they could just be pure functions.

kostmo:
> do you know of a good reason why `tagmembers` and `waypoint` need to be in the `Cmd` monad?

Looks like I didn't make any notes in #1635 or #1356, so I suppose the choice was arbitrary.  I'd be open to making them pure functions.

data/scenarios/Challenges/_dna/lab.sw
```diff
@@ -145,24 +144,19 @@ def makeDnaStrand = \receptacleLoc.
...
         teleport self receptacleLoc;
 
         waitUntilOccupied;
 
-        thingHere <- scan down;
-        case thingHere (\x. return $ inL x) (\item.
-            isOrganism <- hastag item "organism";
-            return $ inR item;
-        );
+        scan down;
```
byorgey, line 153: @kostmo the existing code called `hastag item "organism"` but then ignored the result.  Was that a mistake?

kostmo: Oh, it probably is a mistake.  It should return `inL` if it is not an organism.

data/scenarios/Challenges/_gallery/setup.sw
```diff
@@ -116,13 +124,10 @@ bust in the base's inventory increases monotonically.
 "idx" counts upwards.
 */
 def populateInventory = \baseCount. \idx.
-    
-    result <- tagmembers "bust" idx;
-    let total = fst result in
 
-    if (idx < total) {
+    if (idx < force bustCount) {
```
kostmo, line 128: What does `force` do here?

kostmo (approved):
Quite a feat!

data/scenarios/Challenges/_gallery/setup.sw
byorgey: When I first converted this scenario it was failing, and running it with `--autoplay` revealed that it was failing because the base started running its program *before* the level was set up.  The problem was that evaluating the definition of `bustCount` actually took multiple ticks, before the judge robot ever got around to executing its `instant setup`.  So I wrapped it in a delay so it wouldn't be evaluated at definition time.

I'm just realizing now, however, that this means it's going to be re-evaluated every time we `force` it (IIRC we don't have memoizing delay).  Perhaps a better idea would just be to wrap the *entire* program (including all the `def`s) in `instant`.

kostmo: For `word-search` (#999), I trapped the player behind a boulder until the playfield was set up :)  Though I think that may have been before the `instant` command was available.

https://github.com/swarm-game/swarm/blob/0c1d5f239ccdae13519108c37cade5eef3cccba5/data/scenarios/Challenges/_word-search/create-puzzle.sw#L244

byorgey: Strangely, wrapping the entire program in `instant` doesn't seem to work!  Not sure if this is a bug or what.

PR #2262, opened 2025-01-04
only meet interactive robots

kostmo:
As I was trying to solve #2249, I was rather confused after trying to `meet` a built robot and `give` them an item, but the item was not delivered.  It turns out that I had been `meet`-ing the `judge` robot instead.

We should distinguish between robots that are intended to be interactive or not, and only allow `meet`-ing an interactive robot.

### DIFF
--- a/src/swarm-scenario/Swarm/Game/Robot.hs
+++ b/src/swarm-scenario/Swarm/Game/Robot.hs
@@ -58,6 +58,7 @@ module Swarm.Game.Robot (
 
   -- ** Query
   robotKnows,
+  isInteractive,
 
   -- ** Constants
   hearingDistance,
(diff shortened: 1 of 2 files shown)


src/swarm-scenario/Swarm/Game/Robot.hs
```diff
@@ -276,6 +277,9 @@ inventoryHash = to (\r -> 17 `hashWithSalt` (r ^. (robotEntity . entityHash)) `h
 robotKnows :: Robot -> Entity -> Bool
 robotKnows r e = contains0plus e (r ^. robotInventory) || contains0plus e (r ^. equippedDevices)
 
+isInteractive :: Robot -> Bool
+isInteractive r = not $ r ^. robotDisplay . invisible && r ^. systemRobot
```
kostmo, line 281: We may want to improve this heuristic, or make it an explicit "interactive" property.

For example, I don't think we ever want to `meet` a "growing" `flower`.

byorgey: Yes, I agree, growing `tree`s, `flower`s, etc. are going to be a problem here.  We use robots internally to simulate growth --- but from the perspective of a player that is just an implementation detail, they should not be returned by `meet`.  This heuristic is better than nothing, but we will definitely want to improve it.  I am fine with either merging this as is and making an issue to improve it later, or coming up with a better heuristic in this PR.

An explicit boolean `interactive` property probably makes sense, and I think it makes sense for `interactive = true` to be the default, but invisible system robots can still be non-interactive by default.

byorgey (approved):
Good call, I think I also ran into this while developing the scenario, but I thought it was because I had creative mode turned on.  I didn't realize `meet` made no distinctions!

PR #2273, opened 2025-01-06
Use forM_/mapM_

kostmo:
Searched for instances of `return ()` and `pure ()`, especially `Nothing -> return ()`, which uncovered several opportunities to use `when/unless` and `mapM_/forM_`.

### DIFF
--- a/src/swarm-engine/Swarm/Game/Step/Util/Command.hs
+++ b/src/swarm-engine/Swarm/Game/Step/Util/Command.hs
@@ -99,19 +99,17 @@ ensureCanExecute ::
   Const ->
   m ()
 ensureCanExecute c =
-  gets @Robot (constCapsFor c) >>= \case
-    Nothing -> pure ()
-    Just cap -> do
-      isPrivileged <- isPrivilegedBot
-      -- Privileged robots can execute commands regardless
-      -- of equipped devices, and without expending
-      -- a capability's exercise cost.
-      unless isPrivileged $ do
-        robotCaps <- use robotCapabilities
-        let capProviders = M.lookup cap $ getMap robotCaps
-        case capProviders of
-          Nothing -> throwError $ Incapable FixByEquip (R.singletonCap cap) (TConst c)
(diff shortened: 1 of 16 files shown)


src/swarm-engine/Swarm/Game/Step/Util/Command.hs
```diff
@@ -99,19 +99,19 @@ ensureCanExecute ::
...
-      -- Privileged robots can execute commands regardless
-      -- of equipped devices, and without expending
-      -- a capability's exercise cost.
-      unless isPrivileged $ do
-        robotCaps <- use robotCapabilities
-        let capProviders = M.lookup cap $ getMap robotCaps
-        case capProviders of
-          Nothing -> throwError $ Incapable FixByEquip (R.singletonCap cap) (TConst c)
-          Just rawCosts -> payExerciseCost c rawCosts
+  gets @Robot (constCapsFor c) >>= mapM_ go
```
byorgey, line 102: Out of curiosity, does it work to write `gets @Robot (constCapsFor c) >>= mapM_ \cap -> do ...` ?  Either is fine though.

byorgey (approved):

src/swarm-tui/Swarm/TUI/Controller/SaveScenario.hs
```diff
@@ -90,40 +86,38 @@ unlessCheating a = do
...
-              . _NewGameMenu
-              . ix 0
-              . BL.listSelectedElementL
-              . _SISingle
-              . _2
-              .=
-          )
+    getNormalizedCurrentScenarioPath >>= mapM_ go
+ where
+  go p = do
```
kostmo, line 99: cannot inline the lambda here

src/swarm-engine/Swarm/Game/Step/Util/Command.hs
byorgey: Hmm, that's odd.  Oh well.

PR #2275, opened 2025-01-06
spiders demo

kostmo:
Intended to be a fun demo for inspiration.

```
scripts/play.sh --scenario data/scenarios/Fun/spider.yaml --hide-goal --debug creative,all_robots
```

### DIFF
--- a/data/scenarios/Fun/00-ORDER.txt
+++ b/data/scenarios/Fun/00-ORDER.txt
@@ -2,3 +2,4 @@ GoL.yaml
 logo-burst.yaml
 horton.yaml
 snake.yaml
+spider.yaml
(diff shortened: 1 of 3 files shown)


data/scenarios/Fun/_spider/monkey.sw
```diff
@@ -0,0 +1,55 @@
...
+        );
+        placeBoulder border;
+    } {};
+end
+
+def goDir = \f. \result.
+    let d = fst result in
+    if (d == down) {
+        spinWeb "mountain";
+        f;
```
byorgey, line 44: Could factor `f` out of both branches of the `if`.

byorgey (approved):
Fun!

data/scenarios/Fun/_spider/monkey.sw
```diff
@@ -0,0 +1,55 @@
...
+        spinWeb "mountain";
+    } {
+        turn d; move;
+    };
+    f;
+end;
+
+def followRoute = \item.
+    nextDir <- path (inL ()) (inR item);
+    case nextDir return $ goDir $ followRoute item;
```
byorgey, line 52: Would it be simpler to just say `case nextDir return goDir; followRoute item` and remove the `f` argument from `goDir`?

PR #2285, opened 2025-01-10
Rename `return` to `pure`

### LINKED ISSUE #700: Rename `return` to `pure`
People coming from an imperative programming background might be confused by `return`.  For example, they might be confused by the fact that `move; return (); move` does not ignore the second `move`;  they might be confused by the need for `return` in code like `if foo {grab; return ()} {}`; etc.  Some options for improving the situation include:
- Renaming `return` to something else (`pure`? `result`? `output`?)
- Adding some sort of warnings for uses of `return` that are likely to be confusing (like in the `move; return (); move` example above)

We should also give some thought to how we explain it in the tutorials and other documentation.

byorgey:
Closes #700.

The most interesting changes are in `swarm-lang/Swarm/Language/Parser/{Lex,Term}.hs`, where I remove the code for parsing swarm-0.5 and add an option to parse swarm-0.6 with `return` instead of `pure`, so `swarm format --v0.6` can be used to auto-convert code using `return` to code using `pure`.

Let me know how you feel about this change!  Another possible choice would be to leave `return` as a synonym for `pure`.  I could be convinced to do that as well.

### DIFF
--- a/app/game/Main.hs
+++ b/app/game/Main.hs
@@ -94,7 +94,7 @@ cliParser =
   widthOpt = option auto (long "width" <> metavar "COLUMNS" <> help "Use layout with maximum width")
 
   langVer :: Parser LanguageVersion
-  langVer = flag SwarmLangLatest SwarmLang0_5 (long "v0.5" <> help "Read (& convert) code from Swarm version 0.5")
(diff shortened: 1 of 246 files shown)


xsebek (approved):

kostmo (approved):
FWIW I'd be opposed to having synonyms.

PR #2293, opened 2025-01-15
Make FPS toggle and "recenter view" into global events

### LINKED ISSUE #1992: Global key shortcut (and better name) for `recenter`
The `recenter` UI feature is hard to discover (it is only available when the world map is focused), and poorly named (it does not recenter the map per se; what it really does is `view base`).  In the tutorial we teach players to run `view base` anyway.  I think we should either make a global key shortcut for this, or just get rid of it from the UI altogether.

Perhaps an even better solution is to wait until #1979 is merged, and then make it available as an event, but with no default keybinding.  Then players who really want it as a keyboard shortcut can bind it to whatever they want.

byorgey:
Remove FPS toggle and "recenter view" from the world panel and make them into globally available events with default keybindings (M-f and M-c, respectively).

Closes #1992.


xsebek (approved):

xsebek:
Neat, nice to see the keyboard event ADTs are useful. 🙂

PR #2295, opened 2025-01-20
Disallow `IntMap.!`

### LINKED ISSUE #2294: Crash with invalid IntMap key
**Describe the bug**
I was playing Swarm today when the entire app crashed with the error message:
```
swarm: IntMap.!: key 3971 is not an element of the map
  error, called at libraries/containers/containers/src/Data/IntMap/Internal.hs:601:17 in containers-0.6.8-d9b8:Data.IntMap.Internal
```

**To Reproduce**
I have no idea how to reproduce.  However, at the time I was viewing the robot (F2) panel.

byorgey:
`IntMap.!` is partial and should not be allowed.  Fixes #2294 .  Towards #1494.

### DIFF
--- a/.hlint.yaml
+++ b/.hlint.yaml
@@ -34,6 +34,7 @@
   - {name: undefined, within: [Swarm.Language.Key, TestUtil]}
   - {name: fromJust, within: []}
   - {name: Data.Colour.SRGB.sRGB24read, within: []}
+  - {name: Data.IntMap.!, within: []}
 # - {name: Data.Map.!, within: []} # TODO: #1494
 # - {name: error, within: []} # TODO: #1494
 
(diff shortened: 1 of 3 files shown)


xsebek (commented):
Is the plan to disallow `Map.!` later because it has more usage?

In these two cases, it seems it should still be an error when the collections do not contain expected items, but maybe one that includes more information.
I thought the maps used here should contain all active robots, but maybe I am reading the code wrong.

byorgey:
> Is the plan to disallow `Map.!` later because it has more usage?

I assume so.  See #1494.  `Map.!` was already commented out and I didn't look into it.

> In these two cases, it seems it should still be an error when the collections do not contain expected items, but maybe an error that includes more helpful information.

Ideally, yes, but since these are currently pure functions, it would be tedious to change their type to return some pure error type like a `Maybe` and then update all usage sites to propagate that error *all just in case there is a bug*.  And in the case of `Swarm.Game.State.robotsInArea` in particular, if there was a crash when using `IM.!` it means *that robot no longer exists*, and its ID has (perhaps temporarily) persisted in some other data structure.  I am pretty sure this scenario cannot cause much if any additional harm; the old robot ID will be purged since it does not exist any more.

xsebek (approved):

PR #2299, opened 2025-01-23
Minimize list of required device sets in error message

### LINKED ISSUE #2297: Devices can be reported multiple times in capability error message
**Describe the bug**
If there are multiple missing capabilities provided by the same device, that device is reported multiple times in the resulting error message.

**To Reproduce**
The easiest way I have found to reproduce is:
- Start the `blank` scenario with the `-x` option: `swarm -i blank -x`
- Now hit Ctrl-V to turn creative mode *off*
- Type `build {move; turn left}`
The resulting error message says something like 
```
build: You do not have required device, please obtain:
  - treads or tank treads
  - treads or tank treads
  - solar panel
```
Note that "treads or tank treads" is listed twice.

**Expected behavior**
There is no point in listing the exact same thing twice.  Equipping treads or tank treads will suffice to enable both capabilities.

byorgey:
Closes #2297.

### DIFF
--- a/src/swarm-engine/Swarm/Game/Exception.hs
+++ b/src/swarm-engine/Swarm/Game/Exception.hs
@@ -129,8 +129,8 @@ data IncapableFixWords = IncapableFixWords
 -- "obtain device", or "obtain consumables".
 formatIncapableFix :: IncapableFix -> IncapableFixWords
 formatIncapableFix = \case
-  FixByEquip -> IncapableFixWords "equip" "device"
-  FixByObtainDevice -> IncapableFixWords "obtain" "device"
+  FixByEquip -> IncapableFixWords "equip" "device(s)"
+  FixByObtainDevice -> IncapableFixWords "obtain" "device(s)"
   FixByObtainConsumables -> IncapableFixWords "obtain" "consumables"
 
 -- | Pretty print the incapable exception with an actionable suggestion
--- a/src/swarm-engine/Swarm/Game/Exception.hs
+++ b/src/swarm-engine/Swarm/Game/Exception.hs
@@ -153,7 +153,7 @@ formatIncapableFix = \case
 --   If God in troth thou wantest to play, try thou a Creative game.
 --
 -- >>> incapableError (R.singletonCap $ CExecute Appear) (TConst Appear)
--- You do not have the device required for:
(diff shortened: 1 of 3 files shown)


xsebek (approved):
I wonder, is this a full solution now? Or could we normalise this further?

For example a list of requirements $A \land (B \lor C) \land (A \lor B \lor C)$ could be just $A \land (B \lor C)$.

byorgey:
> I wonder, is this a full solution now? Or could we normalise this further?

Hmm, good point.  I went down a rabbit hole trying to figure out what prior work there is on this.  I think I finally figured out that the expressions we have are in "positive conjunctive normal form" (i.e. we have a conjunction of disjunctions, but variables can only occur positively, i.e. no negation), and for this constrained form finding a minimum equivalent normal form is easy: just repeatedly get rid of any clause which is a superset of another clause.

byorgey:
@xsebek I generalized it along the lines you suggested --- what do you think?

src/swarm-util/Swarm/Util/CNF.hs
```diff
@@ -0,0 +1,48 @@
...
+import Data.List (sortBy)
+import Data.Ord (comparing)
+import Data.Set (Set)
+import Data.Set qualified as S
+import Data.Set.NonEmpty (NESet)
+import Data.Set.NonEmpty qualified as NES
+
+type PCNF a = Set (NESet a)
+
+-- | Minimize an expression in positive conjunctive normal form.  That
```
kostmo, line 18: I have a hunch that the `boolexpr` package implements something in like this.

byorgey: Not really.  The `boolexpr` package does have a data type to represent expressions in conjunctive normal form, but with *signed* variables.  Minimizing CNF expressions *in general* is NP-hard (I'm pretty sure) and the `boolexpr` package does not have any such functions.

byorgey: But wait...!!  *We* already implement something like this... in `Swarm.Util.removeSupersets`. :facepalm:

byorgey:
OK, hopefully this is better!

xsebek (approved):

PR #2312, opened 2025-02-08
Fix completed scenario to show green status

xsebek:
* add lenses to a `Metric`
* decide completion status based on best record (either one will do)

Screenshot 2025-02-08 at 3 33 27 PM


byorgey (approved):
Looks great!  I like the new lenses too.

byorgey:
Hmm, actually, this doesn't seem to be working correctly for me.  After pulling this into `main` and building, it looks like I lost all my progress on the tutorials --- all are drawn with unfilled circles.  My progress is still saved, though, and if I switch back to a different branch the progress is shown correctly again.

Issue #2309, opened 2025-02-06
Don't call `error` when no tutorials are found


byorgey:
If no tutorials are found, we currently simply crash with a call to `error`.  I don't know how to reproduce, but one of my students somehow got Swarm installed and running but in such a way that it apparently could not find the data files, so it crashed with a message saying "No first tutorial found!".

We should instead somehow generate a nice error message explaining the problem.

See also #1494 .

xsebek:
> I don't know how to reproduce

Maybe they copied the data _files_ but not the inner folders? 🤔 Something like:
```sh
mkdir -p ~/.local/share/swarm/data/scenarios; cp data/scenarios/* ~/.local/share/swarm/data/scenarios/
# cp: data/scenarios/Tutorials is a directory (not copied).
```

> We should instead somehow generate a nice error message explaining the problem.

What if, in addition to adding the warning message to Messages, we made the Tutorials menu red and, when selected, opened a dialog popup explaining that the data files are missing?

It would be extra work, but we could reuse that for New Game when _all_ data files are missing.

byorgey:
> Maybe they copied the data _files_ but not the inner folders?

The student in question was definitely not that sophisticated in their use of the command line.  They just cloned the repo and ran `cabal run -O swarm:exe:swarm`.  This failed, e.g. because we had to install a few missing distro packages, but eventually it succeeded... but with no data files. We did various combinations of `cabal clean`, `cabal build`, `cabal install` but none of them fixed the issue.  Eventually we just abandoned the repo and did `cabal install swarm` to download and build it from Hackage, which succeeded.  Seems like ultimately some sort of bug in `cabal` but I wouldn't have the faintest idea how to go about reproducing, much less diagnosing, it.

Anyway, the ultimate point is that this *can* happen, so we should handle it gracefully.

> What if, in addition to adding the warning message to Messages, we made the Tutorials menu red and, when selected, opened a dialog popup explaining that the data files are missing?
> 
> It would be extra work, but we could reuse that for New Game when _all_ data files are missing.

That sounds like a nice idea!

PR #2314, opened 2025-02-08
Make `Tutorial` menu choice auto-start first *unsolved* tutorial scenario

### LINKED ISSUE #2308: Choosing `Tutorial` from the main menu should take you to the first unsolved tutorial scenario
If a player starts the tutorial but does not complete all the scenarios, they may be unsure how to continue later.  Selecting "Tutorial" from the main menu always starts at the very first tutorial scenario, and it is not obvious that one can select "New Game" and then choose a tutorial level.

Selecting `Tutorial` from the main menu should start the first *unsolved* tutorial scenario.  (If all tutorial scenarios have been completed, then it can just start the first tutorial scenario again.)

**Additional context**
I have at least one student who was confused about this and asked how to continue the tutorial without starting over from the beginning.  So I have concrete evidence that the status quo is not as clear as it could be.

byorgey:
Choosing `Tutorial` from the main menu now automatically starts playing the first *unsolved* tutorial scenario, rather than always starting with the first.  Closes #2308.

### DIFF
--- a/src/swarm-tui/Swarm/TUI/Controller.hs
+++ b/src/swarm-tui/Swarm/TUI/Controller.hs
@@ -62,6 +62,7 @@ import Swarm.Game.CESK (CESK (Out), Frame (FApp, FExec, FSuspend))
 import Swarm.Game.Entity hiding (empty)
 import Swarm.Game.Land
 import Swarm.Game.Robot.Concrete
+import Swarm.Game.Scenario (scenarioMetadata, scenarioName)
 import Swarm.Game.Scenario.Scoring.Best (scenarioBestByTime)
 import Swarm.Game.Scenario.Scoring.GenericMetrics
 import Swarm.Game.ScenarioInfo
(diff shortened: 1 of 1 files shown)


xsebek (approved):
I tested it, and it works well. 👍 

I was worried about what it would do after you had solved all the scenarios:
```sh
# --autoplay, but counts as solution:
for s in data/scenarios/Tutorials/*; do cabal run swarm -- -i $s --run <(yq '.solution' $s); done
```
Once all are solved, only the first tutorial will open, like the current state.

byorgey:
> Once all are solved, only the first tutorial will open, like the current state.

Yes, that was on purpose: that is accomplished by the code `firstUnsolved = (tutorials >>= find unsolved) <|> (tutorials >>= listToMaybe)`, which tries to find an unsolved tutorial, but if there isn't one (i.e. if `tutorials >>= find unsolved` results in `Nothing`) it falls back to `tutorials >>= listToMaybe` which takes the first element of the tutorials list.

Note that could still be `Nothing` if the tutorials list is empty --- see #2309.

PR #2316, opened 2025-02-08
Remove say command one message limit

### LINKED ISSUE #1592: log messages missing from "say" with multiple robots from same template
Two robots (created from the same template) that `say` at the same tick do not generate distinct log messages.

Demonstrated in #1593.

It may also be a problem when a single robot `say`s two different things, but both invocations of `say` are wrapped in the same `instant` block.

xsebek:
* remove the `say` command limitation that only keeps one last said message in hearing robot logs
* add test case from #1593 
  * add correct behavior check to scenario objective and integration test
* closes #1592
* closes #1593
* see the original merge request #565 for context

Co-authored-by: Karl Ostmo <kostmo@gmail.com>

### DIFF
--- a/src/swarm-engine/Swarm/Game/Step/Const.hs
+++ b/src/swarm-engine/Swarm/Game/Step/Const.hs
@@ -749,34 +749,21 @@ execConst runChildProg c vs s k = do
       [VText msg] -> do
         isPrivileged <- isPrivilegedBot
         loc <- use robotLocation
-
         -- current robot will be inserted into the robot set, so it needs the log
         m <- traceLog Said Info msg
         emitMessage m
         let measureToLog robLoc = \case
               RobotLog _ _ logLoc -> cosmoMeasure manhattan robLoc logLoc
               SystemLog -> Measurable 0
-            addLatestClosest rl = \case
-              Seq.Empty -> Seq.singleton m
-              es Seq.:|> e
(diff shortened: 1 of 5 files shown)


src/swarm-engine/Swarm/Game/Step/Const.hs
```diff
@@ -749,34 +749,18 @@ execConst runChildProg c vs s k = do
...
-                loc' <- use robotLocation
                 rid <- use robotID
-                return $ do
-                  guard $ hasLog && hasListen
-                  Just (rid, loc')
-              forM_ maybeRidLoc $ \(rid, loc') ->
-                robotInfo . robotMap . at rid . _Just . robotLog %= addLatestClosest loc'
+                return $ if hasLog && hasListen then Just rid else Nothing
+              forM_ maybeRidLoc $ \rid ->
+                robotInfo . robotMap . at rid . _Just . robotLog %= (|> m)
```
byorgey, line 763: Maybe this is simpler?  (It would have been simpler before, too...)
```
            addToRobotLog r = evalState r $ do
              hasLog <- hasCapability $ CExecute Log
              hasListen <- hasCapability $ CExecute Listen
              rid <- use robotID
              when (hasLog && hasListen) $
                robotInfo . robotMap . at rid . _Just . robotLog %= (|> m)
```

byorgey (approved):

PR #2318, opened 2025-02-09
Make `Metric` JSON format backward compatible

xsebek:
The change in #2312 mistakenly broke scenario "save" (best score) compatibility.
This customizes the JSON format and supports the original one.


byorgey (approved):
Looks good, and I can confirm that it successfully parses my existing save files.

PR #2321, opened 2025-02-09
Do not warn about hidden directories

### LINKED ISSUE #2320: Testing does not exist for 00-ORDER.txt
**Describe the bug**

When run without `--debug=testing`, a message about an unknown Testing shows up.

**To Reproduce**

Compare the following:

```
swarm
swarm --debug=testing
```

When you open Messages in the main menu.

**Expected behavior**

When the testing debug option is not selected, the order file should not complain about an unknown Testing directory.

**Screenshots**

Image

**Additional context**

The debug options split might have caused this warning message:
* #2094

xsebek:
* refactor `loadScenarioDir`
* do not warn about hidden directories not being present as they were deliberately not loaded
* closes #2320

### DIFF
--- a/src/swarm-engine/Swarm/Game/ScenarioInfo.hs
+++ b/src/swarm-engine/Swarm/Game/ScenarioInfo.hs
@@ -47,7 +47,7 @@ import Control.Monad.IO.Class (MonadIO (liftIO))
 import Data.Char (isSpace)
 import Data.Either (partitionEithers)
 import Data.Either.Extra (fromRight')
-import Data.List (intercalate, isPrefixOf, stripPrefix, (\\))
+import Data.List (intercalate, isPrefixOf, partition, stripPrefix, (\\))
 import Data.List.NonEmpty qualified as NE
 import Data.Map (Map)
 import Data.Map qualified as M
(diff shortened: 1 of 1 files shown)


src/swarm-engine/Swarm/Game/ScenarioInfo.hs
```diff
@@ -164,70 +166,73 @@ orderFileName = "00-ORDER.txt"
 testingDirectory :: FilePath
 testingDirectory = "Testing"
 
-readOrderFile :: (Has (Lift IO) sig m) => FilePath -> m [String]
-readOrderFile orderFile =
-  filter (not . null) . lines <$> sendIO (readFile orderFile)
+readOrderFile :: FilePath -> IO (Maybe [String])
```
byorgey, line 169: Why the change to a concrete `IO` type, instead of `Has (Lift IO)`?

xsebek: It's a simple IO function, so using sendIO is only noise.

byorgey: OK, fair enough.

xsebek:
@byorgey just to check, did you want to look at this later or did you intend to approve it? 🙂

byorgey (approved):
Sorry, slipped my mind, thanks for the nudge!  In addition to fixing the warning this looks like a very nice refactoring.

PR #2323, opened 2025-02-10
Move parse + typecheck failures from `LoadingFailure` to `SystemFailure`

byorgey:
Two constructors in `LoadingFailure` represented failures that are not necessarily limited to failures that can only happen when loading an asset, namely, parsing failures and typechecking failures.  For example, we could fail to parse or typecheck a term entered at the REPL, which has nothing to do with loading an asset.  This PR:

1. Moves those constructors into `SystemFailure`.
2. Adds a way to embed any `SystemFailure` into a `LoadingFailure` (via the new `General` constructor; name bikeshedding welcome)
3. Removes the `CustomMessage` constructor from `LoadingFailure` since now that can be achieved via `General . CustomFailure`.

This is a preliminary refactoring moving towards a PR for #495.

### DIFF
--- a/src/swarm-scenario/Swarm/Game/Entity.hs
+++ b/src/swarm-scenario/Swarm/Game/Entity.hs
@@ -490,7 +490,7 @@ validateEntityAttrRefs validAttrs es =
       AWorld n ->
         unless (Set.member (WorldAttr $ T.unpack n) validAttrs)
           . throwError
-          . General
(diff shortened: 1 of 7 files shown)


xsebek:
Regarding `General`, how about:
* Failure
* SystemFailure (the constructor name is free it seems?)
* OtherFailure

PR #2324, opened 2025-02-11
Clickable shortcuts in TUI

### LINKED ISSUE #2124: Clickable shortcuts in TUI

I would like to click the function shortcuts (help, robots, ...) and the shortcuts between World and REPL (unpause, step,...) with the mouse. This way I would not have to use a keyboard for every functionality.

Make the widgets `clickable` with name like `UIShortcut SwarmEvent`, so that the event can be handled on click.

**Additional context**
Screenshot 2024-08-16 at 12 16 42 AM

ussgarci:
closes #2124 

##### description:

This PR makes the function shortcuts (help, robots, ...) and the shortcuts between World and REPL (unpause, step,...) clickable with a mouse.


ussgarci:
Hello @xsebek @byorgey .  I have a couple questions.
1. Should I add some sort of hover indicator to show that some of the shortcuts are clickable?
2. For the shortcuts that are bound to more than one key i.e. `history` or `scroll` I didn't see an intuitive way to make them clickable so I left them unchanged.  Is that ok?

xsebek:
Hi, @ussgarci thanks for working on this! 🙂

1. If you want to, sure. It did not occur to me that it would be possible.
2. Could we split them to two buttons, like Speed [<]/[>]?

ussgarci:
Hello @xsebek @byorgey

I did manage to get the double button working for speed.  I think I'd like to leave the hover effect for a follow-up issue if that is ok.  I also left the REPL-specific shortcuts untouched (history, pagedown, etc.. ).  Thank you for any feedback!

src/swarm-tui/Swarm/TUI/Controller.hs
```diff
@@ -350,6 +350,19 @@ handleMainEvent forceRedraw ev = do
           when shouldUpdateCursor $
             uiGameplay . uiWorldCursor .= mouseCoordsM
         REPLInput -> handleREPLEvent ev
+        (UIShortcut "Help") -> playStateWithMenu $ toggleModal HelpModal
+        (UIShortcut "Robots") -> playStateWithMenu $ toggleModal RobotsModal
+        (UIShortcut "Commands") -> playStateWithMenu $ toggleDiscoveryNotificationModal CommandsModal availableCommands
```
byorgey, line 355: Is there a reason that the Recipes (F3) and Messages (F5) shortcuts are not included as clickable?

src/swarm-tui/Swarm/TUI/View.hs
```diff
@@ -146,6 +146,9 @@ import Text.Printf
 import Text.Wrap
 import Witch (into)
 
+-- | The 2nd case is for key commands that have more than one btn in the ui
+type KeyCmd = Either (KeyHighlight, Text, Text) (KeyHighlight, [(Text, Text)], Text)
```
byorgey, line 150: Rather than having a type synonym for `Either`, I'd rather have an actual `data` type with two constructors like `SingleButton` and `MultiButton` or something descriptive like that.  (Then we can also get rid of the tuples at the same time.)  When reading things like `Just (Left (...))` in the code it's very hard to know what is going on; something like `Just (SingleButton (...))` would be easier to read and easier to track down the purpose.

src/swarm-tui/Swarm/TUI/View.hs
```diff
@@ -932,34 +935,43 @@ drawKeyMenu s =
...
-          , keyM SE.ShowCESKDebugEvent
-          , "debug"
+          ( Left
+              ( if uig ^. uiShowDebug then Alert else NoHighlight
+              , keyM SE.ShowCESKDebugEvent
+              , "debug"
+              )
           )
-      , Just (NoHighlight, keyM SE.IncreaseTpsEvent <> "/" <> keyM SE.DecreaseTpsEvent, "speed")
+      , -- , Just (Left (NoHighlight, keyM SE.IncreaseTpsEvent <> "/" <> keyM SE.DecreaseTpsEvent, "speed"))
```
byorgey, line 955: Should be able to delete this commented-out code.

byorgey (commented):
This is looking great!  I tried it out and it seems to work well.  Just a few comments/suggestions below.

src/swarm-tui/Swarm/TUI/Controller.hs
ussgarci: Whoops.  I'll add those asap.

src/swarm-tui/Swarm/TUI/View.hs
ussgarci: Agreed!  I'll make that change.

ussgarci:
@byorgey I have addressed your suggestions.  Thank you for the feedback.  Please let me know if you would like any other changes.

src/swarm-tui/Swarm/TUI/View.hs
```diff
@@ -1320,7 +1335,8 @@ drawRecipe me inv (Recipe ins outs reqs time _weight) =
     hBox
       [ padRight (Pad 1) $ str (show n) -- how many?
       , fmtEntityName missing ingr -- name of the input
-      , padLeft (Pad 1) $ -- a connecting line:   ─────┬
+      , padLeft (Pad 1) $ -- a connecting line:   ─────┬ -- a connecting line:   ─────┬
+      -- a connecting line:   ─────┬
```
byorgey, line 1339: This comment got duplicated somehow?

byorgey (approved):
Looks great, thanks!  There is one comment that got strangely duplicated, but feel free to clean that up and then add the `merge me` label once it's ready!

PR #2326, opened 2025-02-12
Add test for save file compatibility

### LINKED ISSUE #2319: Add save file format compatibility test

The change in #2312 mistakenly broke scenario "save" (best score) compatibility.
Luckily, it was found (and soon fixed in #2318) promptly, but a change like this should not get past CI.

We should test that the best scores that a player has already saved can be parsed in CI.

Additive changes can extend the file, but a new file must be added once we have any required fields.

In the case of #2312:
* add one file format for the previous tuple-like version
* add a new file with the named field version, as it is required in the alternative parser

Unit tests for individual types would be nice, but they would not cover my use case. I want to make code refactorings fearlessly and not worry that I will break compatibility.

**Additional context**

I thought a bit about how we would maintain the files, and I think it would work like this:
* if a new (optional!) best metric is added, we could add it to the latest test file or be pedantic and add a whole new test file
* if a change is significant and has required fields, we could add a `version` tag to the top of the format and parse the rest accordingly
* if the amount of test files grows too large, we could delete the ones from unsupported versions - players could migrate by starting swarm with newer versions (this would require overwriting the files with the new format, which we do not automatically do)

This may seem like a lot of pointless work for one accidental change, which could be undone, but:
1. this change is good ™  and blocking format improvements would be annoying
2. we should be mindful of this if we want players to happily upgrade to the latest version and not be sad that they have to start over
3. if the test is fast, developers will be happy

xsebek:
* closes #2319

### DIFF
--- a/data/test/saves/backstory-0.6.0.0.yaml
+++ b/data/test/saves/backstory-0.6.0.0.yaml
@@ -1,0 +1,58 @@
+# obtained by:
+# git checkout 0.6.0.0
+# cabal run swarm -- --scenario Tutorials/Backstory --run <(echo 'say "Ready!"')
+path: Tutorials/backstory.yaml
+status:
+  contents:
+    - initialCode: /dev/fd/15
+      seedVal: null
+    - - Completed
+      - attemptmetrics:
+          codemetrics:
+            astSize: 3
+            sourceTextLength: 13
+          durationmetrics:
+            elapsed: 1.696629
+            elapsedticks: 5
+        started: 2025-02-12T16:05:31.987851+01:00
+    - bestbyastsize:
+        - Completed
+        - attemptmetrics:
+            codemetrics:
+              astSize: 3
+              sourceTextLength: 13
+            durationmetrics:
+              elapsed: 0.337978
+              elapsedticks: 5
+          started: 2025-02-12T16:05:31.987851+01:00
+      bestbycharcount:
+        - Completed
+        - attemptmetrics:
+            codemetrics:
+              astSize: 3
+              sourceTextLength: 13
+            durationmetrics:
+              elapsed: 0.337978
+              elapsedticks: 5
+          started: 2025-02-12T16:05:31.987851+01:00
+      bestbyticks:
+        - Completed
+        - attemptmetrics:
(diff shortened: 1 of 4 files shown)


test/integration/Main.hs
```diff
@@ -610,3 +613,19 @@ testEditorFiles =
...
+    , checkLoaded "backstory" "latest"
+    ]
+ where
+  checkLoaded scenario version = testCase ("save from version " <> version) $ do
+    ef <- decodeFileEither @ScenarioInfo $ "data/test/saves/" <> scenario <> "-" <> version <> ".yaml"
+    case ef of
+      Left e -> assertFailure $ prettyPrintParseException e
+      Right si -> case si ^. scenarioStatus of
+        Played _par (Metric Completed _) _best -> pure ()
+        other -> assertFailure $ "scenario save file loaded wrong - contains: " <> show other
```
xsebek, line 631: I am open to ideas for making the test better. The regression I caused would show this failure:
```
Tests
  Save files
    save from version 0.6.0.0: FAIL
      test/integration/Main.hs:621:
      Aeson exception:
      Error in $.status.contents[1]: parsing Swarm.Game.Scenario.Scoring.GenericMetrics.Metric(Metric) failed, expected Object, but encountered Array
      Use -p '/save/&&/save from version 0.6.0.0/' to rerun this test only.
```

data/test/saves/backstory-0.6.0.0.yaml
```diff
@@ -0,0 +1,58 @@
+# obtained by:
+# git checkout 0.6.0.0
+# cabal run swarm -- --scenario Tutorials/Backstory --run <(echo 'say "Ready!"')
+path: Tutorials/backstory.yaml
+status:
+  contents:
+    - initialCode: /dev/fd/15
```
byorgey, line 7: lol, I am not sure what the right thing to do is here, but surely it is not useful to store `initialCode: /dev/fd/15` :laughing: 

(This is not a comment about this PR, but about what we store in `initialCode` in general)

byorgey (approved):
Looks good to me.  Certainly having this is much better than nothing.

PR #2328, opened 2025-02-23
Remove old individual wall entities

### LINKED ISSUE #2154: Remove old walls from entities.yaml
I want to replace the default walls with the new connecting wall from #1285 in `entities.yaml`.

This would make the new wall a more easily discoverable good default.

> We can't quite get rid of all walls, as illustrated by the bucket brigade challenge map.

_Originally posted by @byorgey in https://github.com/swarm-game/swarm/pull/2153#issuecomment-2368111287_

In challenges that require custom walls the old walls would need to be included. This will make it explicit that the scenario needs custom wall rendering at the cost of being more verbose.

byorgey:
Closes #2154.

- Remove specific wall segment entities (like `upper left corner`) from `entities.yaml`
- Convert most scenarios to use the new `wall` entity with the `boundary` property
- Add custom wall segment entities to a few scenarios that need them


xsebek (approved):
Neat! 👍

PR #2329, opened 2025-02-24
Improvements to comment preservation in `swarm format`

### LINKED ISSUE #2105: Inline comment is moved inside several brackets
Consider this code:
```
if true
  {(1,2)} /* comment /*
  {(3,4)}
```
Running the formatter on this via `swarm format` produces
```
if a {(2, 1 /* comment */)} {(3, 4)}
```
where the comment has been moved all the way inside the tuple.  This is certainly not ideal.  I am not yet sure whether this is just a case our formatting algorithm does not handle, or whether this is actually a bug in the way it is implemented.

> Thinking about this more, I am actually surprised that it does not print `{(2 * i, 1)} /* last part */`.  I thought it was supposed to do that; I will have to think about whether this is a bug.

_Originally posted by @byorgey in https://github.com/swarm-game/swarm/pull/2103#discussion_r1712083008_

byorgey:
Fixes #2105.

This started out with me just trying to figure out what was going on with #2105, but turned into a bug-hunting fest.  This PR:

- Adds some integration tests in `data/test/language-snippets/format` --- we ensure that any files in that directory are unchanged by `swarm format`.
- Adds an `SParens` constructor to the `Term` data type, to explicitly represent parentheses from the parsed input.
    - This is necessary in order to assign a correct source span to parenthesized expressions.
    - Along the way I regularized the parsing of parenthesized things (unit value, normal parentheses, and tuples).
- Fixes the way we attach suffix comments to AST nodes to actually do what it said: attach each suffix comment to the shallowest node which ends before it.
    - We do this via a reverse postorder traversal, but the code we had before was not actually doing a reverse postorder traversal.
    - The other thing we had to fix to make this work was to assign tighter source spans to parsed AST nodes.
        - The source span assigned to a given node used to include any whitespace consumed after it.
        - But if that whitespace included any comments, we could not determine where they should go, since the correct AST node "ended after them".
        - Now we use some extra state to remember when we started consuming whitespace, so we can assign a source span that ends at that point.
- Does a better job preserving comments in a few places in the pretty-printer (*e.g.* chains of lambdas in definitions, infix operator applications).


test/integration/TestFormat.hs
```diff
@@ -0,0 +1,35 @@
...
+  defaultFormatCfg = FormatConfig Stdin Stdout Nothing SwarmLangLatest
+
+  checkFormat :: FilePath -> TestTree
+  checkFormat f = testCase f $ do
+    mcontent <- readFileMayT f
+    case mcontent of
+      Nothing -> assertFailure $ "Failed to load " <> f
+      Just content -> case formatSwarm defaultFormatCfg content of
+        Left err -> assertFailure (into @String err)
+        -- Make sure they are the same, but ignore leading + trailing whitespace
```
xsebek, line 33: Why ignore leading and trailing whitespace? Will the files not be the same?

data/test/language-snippets/format/list.sw
```diff
@@ -0,0 +1,306 @@
+
+/*******************************************************************/
+/*                       LIST ENCODING IN INT                      */
+/*                                                                 */
```
xsebek, line 4: Nice test case. 👍 😄

xsebek (approved):
Nice! It looks like `format` is pretty close to usable on all our code.

data/test/language-snippets/format/list.sw
byorgey: Indeed!  This was the original source for #2105 in the first place - the formatter almost worked on list.sw except for one line.

test/integration/TestFormat.hs
byorgey: I don't remember exactly.  I think I just didn't want to be bothered getting such things exactly right, e.g. whether there should be a trailing newline and so on.  But we could try to be more precise about it if you think it would be worth it.

byorgey:
@xsebek you said "pretty close to usable" - do you have any specific examples in mind where it's not ideal?

data/test/language-snippets/format/list.sw
```diff
@@ -0,0 +1,306 @@
...
+
+// Split length into 7-bit parts and prefix all but last with 1
+def to1numA: Int -> (Int * Int) 
+  = \i.
+  let nextPart = shiftL 7 i in
+  if (nextPart == 0) {
+
+    // last part
+    (2 * i, 1)
+  } {
```
xsebek, line 112: @byorgey, this stands out to me. It seems we are adding an extra line before comments, but that leads to weird cases like this one. Also, the space between definitions is either one or two lines, depending on whether there is a comment before it.

test/integration/TestFormat.hs
xsebek: I think it would be better to test precisely what we expect the formatter to do.

If this gets annoying on Windows, I would be okay with the tests ignoring the trailing newline, but I would like to know that the formatter does not add multiple lines.

byorgey: Oh, now I remember why, because my editor adds trailing newlines by default, but the formatter does not produce trailing newlines.  Creating files with no trailing newlines would just be annoying.  I will change it so it tests up to a single trailing newline but otherwise requires the whitespace to be exact.

data/test/language-snippets/format/list.sw
byorgey: Hmm, I don't remember why I originally did it that way, but I agree the extra line before comments does not look good.  Removed!

Issue #2300, opened 2025-01-25
Look into use of `MonoidMap`


byorgey:
[monoidmap](https://hackage.haskell.org/package/monoidmap) is a really nice package which defines total ` k -> v` mappings where `v` must be a `Monoid`.  Any `k -> v` mapping where `v` is `mempty` is not stored.  I only just discovered this package (thanks to [this toot](https://mathstodon.xyz/@pmidden@fosstodon.org/113888153842687719)) but it occurs to me that we might be able to use it to simplify some things; for example I'm pretty sure `Inventory` fits into this model very nicely.

jonathanknowles:
Hi @byorgey 

Thanks for the mention!

Out of curiosity, I had a look at your `Inventory` type.

Depending on the semantics you have in mind, I think it may be possible to use `MonoidMap` for the `byName` field:
```patch
-   byName ::       Map Text IntSet
+   byName :: MonoidMap Text IntSet
```

Full branch: https://github.com/swarm-game/swarm/compare/main...jonathanknowles:swarm:jonathanknowles/use-monoidmap-in-inventory

In the above branch, `lookupByName` and `insertCount` no longer need to consider the case where `byName` has no entry for a given name, as `MonoidMap` will map all such names to `IntSet.empty`.

Does do you think this approach is consistent with the existing semantics? If so, I'd be happy to turn it into a PR. Let me know!

---

As for the `counts` field, perhaps it's possible to replace this with something like `MonoidMap Int (MonoidMap Entity Count)`? (The inner map being a _bag_ of entities.)

Though this would mean not storing the following:
- any mappings from `Int` to empty bags of entities
- any mappings from `Entity` to `Count 0`

And it seems from elsewhere in the same module that you do want to retain the possibility of storing zero counts:
https://github.com/swarm-game/swarm/blob/d3ba8c5db8ac4496ffec40defb248dfca304043f/src/swarm-scenario/Swarm/Game/Entity.hs#L857-L858

Another issue is that `MonoidMap` is not (yet) optimised for `Int` keys (there is not yet an `IntMonoidMap`), so it might be less performant for your use case.

byorgey:
Hi @jonathanknowles , thanks!

> Do you think this approach is consistent with the existing semantics? If so, I'd be happy to turn it into a PR. Let me know!

I haven't studied the code in detail, but yes, this seems extremely plausible and is exactly the kind of thing I had in mind!  I would be very happy if you wanted to turn this into a PR.

[...]

PR #2334, opened 2025-03-07
Use `MonoidMap` for `robotsByLocation`.

jonathanknowles:
Related to https://github.com/swarm-game/swarm/issues/2300.

This PR redefines `Robots.robotsByLocation` to use a nested `MonoidMap` instead of a nested `Map`:

```patch
  data Robots = Robots
    { ...
-   , _robotsByLocation ::       Map SubworldName (      Map Location IntSet)
+   , _robotsByLocation :: MonoidMap SubworldName (MonoidMap Location IntSet)
    , ...
    }
```

Queries of the original nested `Map` did not distinguish between mappings to `Nothing` and mappings to `Just mempty`, whether in the outer map or the inner map. By using `MonoidMap`, we can simplify such queries, as well as functions that create or update the map.

For example, `robotsAtLocation` no longer needs to explicitly handle cases where the outer map does not contain a given `SubworldName`, or where the inner map does not contain a given `Location`:

```patch
  robotsAtLocation loc gs =
    mapMaybe (`IM.lookup` (gs ^. robotInfo . robotMap))
-     . maybe [] IS.toList
-     . M.lookup (loc ^. planar)
-     . M.findWithDefault mempty (loc ^. subworld)
+     . IS.toList
+     . MM.get (loc ^. planar)
+     . MM.get (loc ^. subworld)
      . view (robotInfo . robotsByLocation)
      $ gs
```

Insertion and removal can now be expressed in terms of simple nested adjustments, avoiding the need to have special handling code for mappings to null values. It's arguably now easier to see the inherent symmetry of these operations:

```hs
addRobotToLocation rid rLoc =
  robotsByLocation %=
  MM.adjust (MM.adjust (IS.insert rid) (rLoc ^. planar)) (rLoc ^. subworld)

removeRobotFromLocation rid rLoc =
  robotsByLocation %=
  MM.adjust (MM.adjust (IS.delete rid) (rLoc ^. planar)) (rLoc ^. subworld)
```

### DIFF
--- a/src/swarm-engine/Swarm/Game/State.hs
+++ b/src/swarm-engine/Swarm/Game/State.hs
@@ -90,6 +90,7 @@ import Data.IntMap qualified as IM
 import Data.IntSet qualified as IS
 import Data.Map qualified as M
 import Data.Maybe (fromMaybe, mapMaybe)
+import Data.MonoidMap qualified as MM
 import Data.Sequence (Seq ((:<|)))
 import Data.Sequence qualified as Seq
 import Data.Text (Text)
@@ -228,9 +229,9 @@ winSolution :: Lens' GameState (Maybe TSyntax)
 robotsAtLocation :: Cosmic Location -> GameState -> [Robot]
 robotsAtLocation loc gs =
   mapMaybe (`IM.lookup` (gs ^. robotInfo . robotMap))
-    . maybe [] IS.toList
-    . M.lookup (loc ^. planar)
-    . M.findWithDefault mempty (loc ^. subworld)
+    . IS.toList
+    . MM.get (loc ^. planar)
+    . MM.get (loc ^. subworld)
     . view (robotInfo . robotsByLocation)
     $ gs
 
@@ -245,9 +246,10 @@ robotsInArea (Cosmic subworldName o) d rs = mapMaybe (rm IM.!?) rids
   rm = rs ^. robotMap
   rl = rs ^. robotsByLocation
   rids =
-    concatMap IS.elems $
-      getElemsInArea o d $
-        M.findWithDefault mempty subworldName rl
+    concatMap IS.elems
+      . getElemsInArea o d
+      . MM.toMap
+      $ MM.get subworldName rl
 
 -- | The base robot, if it exists.
 baseRobot :: Traversal' GameState Robot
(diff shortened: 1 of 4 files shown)


src/swarm-engine/Swarm/Game/State.hs
```diff
@@ -245,9 +246,10 @@ robotsInArea (Cosmic subworldName o) d rs = mapMaybe (rm IM.!?) rids
   rm = rs ^. robotMap
   rl = rs ^. robotsByLocation
   rids =
-    concatMap IS.elems $
-      getElemsInArea o d $
-        M.findWithDefault mempty subworldName rl
+    concatMap IS.elems
+      . getElemsInArea o d
+      . MM.toMap
```
jonathanknowles, line 251: This is a zero-cost coercion. See:

https://github.com/jonathanknowles/monoidmap/blob/f54db7f51a5bed6b4213674be96a4e87475a78c6/components/monoidmap-internal/Data/MonoidMap/Internal.hs#L574-L575

jonathanknowles:
Regarding the `Restyled` check: I was able to manually check the style with `scripts/normalize/code-format.sh`.

byorgey (approved):
I wonder if we can get rid of `surfaceEmpty`, it seems to only be used in one other place.

jonathanknowles:
Hi @byorgey 

As suggested, I've pushed this branch to the `swarm` repository and created a new PR here: https://github.com/swarm-game/swarm/pull/2336

Thanks again for your review!

jonathanknowles:
> I wonder if we can get rid of `surfaceEmpty`, it seems to only be used in one other place.

@byorgey I did wonder about that. I'd be happy to take a look and see if there's a nice way to remove it. (Will create a subsequent PR if so.)

jonathanknowles:
> I wonder if we can get rid of `surfaceEmpty`, it seems to only be used in one other place.

It looks like there is already a similar function in `extra` (which is already a dependency of several sublibraries of `swarm`):

https://hackage.haskell.org/package/extra-1.8/docs/Control-Monad-Extra.html#v:guarded

I haven't yet tested it, but I suspect it's the same function (perhaps with the predicate inverted).

byorgey:
> It looks like there is already a similar function in `extra`

Oh, good find.  Yes, I'm quite sure it is the same function with the predicate inverted, although they are implemented very differently.

PR #2337, opened 2025-03-09
Use `MonoidMap` for `robotsWatching`.

jonathanknowles:
Related to https://github.com/swarm-game/swarm/issues/2300.

This PR redefines `Robots.robotsWatching` to use `MonoidMap` instead of `Map`:

```patch
  data Robots = Robots
    { ...
-   , _robotsWatching ::       Map (Cosmic Location) IntSet
+   , _robotsWatching :: MonoidMap (Cosmic Location) IntSet
    , ...
    }
```

Queries of the original `Map` did not distinguish between mappings to `Nothing` and mappings to `Just mempty`. By using `MonoidMap`, we can simplify such queries, as well as functions that create or update the map.

For example, within `wakeWatchingRobots`:

```patch
      botsWatchingThisLoc =
        mapMaybe (`IM.lookup` rMap) $
          IS.toList $
-           M.findWithDefault mempty loc watchingMap
+           MM.get                   loc watchingMap
```

Within `purgeFarAwayWatches`, we also no longer need to manually remove empty sets of robots, as `MonoidMap.mapWithkey` will remove these automatically, requiring only a single traversal of the map:

```patch
- robotInfo . robotsWatching %= M.filter (not . IS.null) . M.mapWithKey f
+ robotInfo . robotsWatching %=                           MM.mapWithKey f
```

### DIFF
--- a/src/swarm-engine/Swarm/Game/State/Robot.hs
+++ b/src/swarm-engine/Swarm/Game/State/Robot.hs
@@ -66,6 +66,8 @@ import Data.List.NonEmpty qualified as NE
 import Data.Map (Map)
 import Data.Map qualified as M
 import Data.Maybe (fromMaybe, mapMaybe)
+import Data.MonoidMap (MonoidMap)
+import Data.MonoidMap qualified as MM
 import Data.Set qualified as S
 import Data.Tuple (swap)
 import GHC.Generics (Generic)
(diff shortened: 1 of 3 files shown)


byorgey (approved):

byorgey:
@jonathanknowles our typical workflow is that once a PR is approved, we leave it up to whoever opened the PR to kick off the merge (by adding the `merge me` label) once they are totally happy with it.  In the case that a reviewer suggested some changes but approved the PR, that means trusting the person who opened the PR to apply the suggestions as they see fit before merging, without requiring another review---of course that doesn't really apply here since I didn't suggest any changes.  But in any case, feel free to add the `merge me` label!

jonathanknowles:
> @jonathanknowles our typical workflow is that once a PR is approved, we leave it up to whoever opened the PR to kick off the merge (by adding the `merge me` label) once they are totally happy with it. In the case that a reviewer suggested some changes but approved the PR, that means trusting the person who opened the PR to apply the suggestions as they see fit before merging, without requiring another review---of course that doesn't really apply here since I didn't suggest any changes. But in any case, feel free to add the `merge me` label!

Thanks for clarifying! I've added the `merge me` label. (I noticed that it's also mentioned in https://github.com/swarm-game/swarm/blob/main/CONTRIBUTING.md.)

PR #2338, opened 2025-03-09
Replace `surfaceEmpty` with `guarded`.

jonathanknowles:
This PR replaces `surfaceEmpty` with [`guarded`](https://hackage.haskell.org/package/extra-1.8/docs/Control-Monad-Extra.html#v:guarded) (from [`extra`](https://hackage.haskell.org/package/extra)) in `getBoundingBox`.

Since `guarded` was only made available in version `1.8` of `extra`, we need to bump the upper bound for `extra`, which in turn requires us to bump the upper bound for `lsp`.

Note that `guarded p == surfaceEmpty (not . p)`:
```hs
> guarded             even  (2 :: Int) == [2]
True
> surfaceEmpty (not . even) (2 :: Int) == [2]
True
```
```hs
> guarded             odd  (2 :: Int) == Nothing
True
> surfaceEmpty (not . odd) (2 :: Int) == Nothing
True
```
```hs
> guarded (not . null) "My Name" == Just "My Name"
True
> surfaceEmpty   null  "My Name" == Just "My Name"
True
```

### DIFF
--- a/src/swarm-scenario/Swarm/Game/World/Render.hs
+++ b/src/swarm-scenario/Swarm/Game/World/Render.hs
@@ -20,6 +20,7 @@ import Control.Carrier.Throw.Either (runThrow)
 import Control.Effect.Lift (Lift, sendIO)
 import Control.Effect.Throw
 import Control.Lens (view, (^.))
+import Control.Monad.Extra (guarded)
 import Control.Monad.Logger
 import Control.Monad.Trans (MonadIO)
 import Data.Aeson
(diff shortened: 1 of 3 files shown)


jonathanknowles:
@byorgey

Just curious if this solution (for eliminating `surfaceEmpty`) would work for you?

If so, after PR #2336 is merged, I can update this PR to actually remove the function.

byorgey:
Yes, this looks great to me.

jonathanknowles:
> Yes, this looks great to me.

@byorgey Thanks for confirming!

I've now updated the PR (removing the function).

byorgey (approved):

PR #2339, opened 2025-03-09
Use `MonoidMap` for `waitingRobots`.

jonathanknowles:
Related to #2300.

This PR redefines `Robots.waitingRobots` to use `MonoidMap` instead of `Map`:

```patch
  data Robots = Robots
    { ...
-   , _waitingRobots ::       Map TickNumber [RID]
+   , _waitingRobots :: MonoidMap TickNumber [RID]
    , ...
    }
```

This allows us to simplify various functions that query or update this map.

In addition, garbage collection of mappings to empty lists is now automatic.

### DIFF
--- a/src/swarm-engine/Swarm/Game/State/Robot.hs
+++ b/src/swarm-engine/Swarm/Game/State/Robot.hs
@@ -62,13 +62,10 @@ import Data.IntSet (IntSet)
 import Data.IntSet qualified as IS
 import Data.IntSet.Lens (setOf)
 import Data.List (partition)
-import Data.Map (Map)
-import Data.Map qualified as M
 import Data.Maybe (fromMaybe, mapMaybe)
 import Data.MonoidMap (MonoidMap)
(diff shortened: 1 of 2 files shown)


src/swarm-engine/Swarm/Game/State/Robot.hs
```diff
@@ -374,7 +365,7 @@ wakeWatchingRobots myID currentTick loc = do
         [ (currentTick, currTickWakeable)
         , (addTicks 1 currentTick, nextTickWakeable)
         ]
-      newInsertions = M.filter (not . null) $ M.fromList wakeTimeGroups
+      newInsertions = MM.fromList wakeTimeGroups
```
jonathanknowles, line 368: `MonoidMap.fromList` already arranges that the underlying support map has no mappings to `[]`, so there's no need to explicitly filter them out.

byorgey (approved):
I never realized how much code we have that is essentially just dealing with the possibility of things being empty!

PR #2340, opened 2025-03-09
Use original variable names in error messages about Skolem variables

### LINKED ISSUE #2102: Improve error messages involving skolem variables
There are two kinds of error messages we could get involving skolem variables.

* We can simply get a type mismatch when we try to unify a skolem variable with something else.  However, the current error is confusing since it references a made-up variable.  For example:
  ```
  > def f : a -> a = \x. 3 end
  1:22: Type mismatch:
    From context, expected `3` to have type `s1`,
    but it actually has type `Int`

    - While checking the definition of f
  ```
  It would be better if the error said something about polymorphism and mentioned the type variable `a`.  This might require keeping track of a mapping between generated skolem variables and type variables in the polymorphic type.

* When a skolem variable escapes, the error message is even worse since the error explicitly mentions the word "skolem":
  ```
  > \f. (f:forall a. a -> a) 3
  1:5: Skolem variable s3 would escape its scope
  ```
  This case is more obscure and probably won't happen very much (especially once we fix #2101), but it would still be nice to come up with a better error message that at least hints at what the problem might be.

byorgey:
Each type variable now carries two variable names: one, the original, user-supplied name of the type variable, and another, the (potentially different) current freshly generated name.  These can diverge specifically in the case that we skolemize a user-supplied polymorphic type.  We use the second name for all comparisons, but the first name when pretty-printing.

For example, `def f : a -> a = \x. 3 end` used to generate an error message saying something like "expected `3` to have type `s1` ; it now says "expected `3` to have type `a`".

Closes #2102.  That issue also talks about improving the error message that says "skolem variable would escape its scope", but at the moment I can't even find a way to get that error message to be generated at all, so I propose to not worry about it for now.

There is one specific situation where this can be confusing, when one type variable is shadowed by another with the same name.  For example:
```
> def f : a -> (a * Int) = \\x. let g : forall a. a * Int = (x, 3) in g end
1:59: Type mismatch:\n  From context, expected `x` to have type `a`,\n  but it actually has type `a`
```
Not sure whether it's worth doing anything about this; open to suggestions.

### DIFF
--- a/src/swarm-lang/Swarm/Effect/Unify/Fast.hs
+++ b/src/swarm-lang/Swarm/Effect/Unify/Fast.hs
@@ -312,7 +312,7 @@ unifyF t1 t2 = case (t1, t2) of
   (TyConF {}, _) -> unifyErr
   -- Note that *type variables* are not the same as *unification variables*.
   -- Type variables must match exactly.
-  (TyVarF v1, TyVarF v2) -> case v1 == v2 of
+  (TyVarF _ v1, TyVarF _ v2) -> case v1 == v2 of
     True -> pure t1
     False -> unifyErr
   (TyVarF {}, _) -> unifyErr
--- a/src/swarm-lang/Swarm/Effect/Unify/Naive.hs
+++ b/src/swarm-lang/Swarm/Effect/Unify/Naive.hs
@@ -155,7 +155,7 @@ unifyF t1 t2 = case (t1, t2) of
     True -> compose <$> zipWithM unify ts1 ts2
     False -> unifyErr
   (TyConF {}, _) -> unifyErr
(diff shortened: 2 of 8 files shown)


xsebek (approved):
@byorgey this is great! I remember that any error messages with skolemized type names were hard for me to understand, but this should fix that in most cases. 👍 

Regarding the same name, I take it that the type names do not keep the source span like values do?
If they did, the error messages could continue with "`a` declared at `source.sw`: line 12 char 42" and similarly for the other `a`.

byorgey:
> I take it that the type names do not keep the source span like values do?

Yeah, they don't, but they probably should!  That would be a bigger change though. Filed as #2343.

PR #2342, opened 2025-03-10
Use `MonoidMap.foldrWithKey` in `wakeWatchingRobots`.

jonathanknowles:
Tiny PR that changes `filteredWaiting` to use `MonoidMap.foldrWithKey` instead of `MonoidMap.toList` and `foldr`, thus avoiding the creation of an intermediate list.

### DIFF
--- a/src/swarm-engine/Swarm/Game/State/Robot.hs
+++ b/src/swarm-engine/Swarm/Game/State/Robot.hs
@@ -349,9 +349,10 @@ wakeWatchingRobots myID currentTick loc = do
       -- Step 3: Take these robots out of their time-indexed slot in "waitingRobots".
       -- To preserve performance, this should be done without iterating over the
       -- entire "waitingRobots" map.
-      filteredWaiting = foldr f waitingMap $ MM.toList wakeTimesToPurge
+      filteredWaiting :: MonoidMap TickNumber [RID]
+      filteredWaiting = MM.foldrWithKey f waitingMap wakeTimesToPurge
        where
-        f (k, botsToRemove) = MM.adjust (filter (`S.notMember` botsToRemove)) k
+        f k botsToRemove = MM.adjust (filter (`S.notMember` botsToRemove)) k
 
       -- Step 4: Re-add the watching bots to be awakened ASAP:
       wakeableBotIds = map fst wakeTimes


byorgey (approved):

src/swarm-engine/Swarm/Game/State/Robot.hs
```diff
@@ -349,9 +349,10 @@ wakeWatchingRobots myID currentTick loc = do
       -- Step 3: Take these robots out of their time-indexed slot in "waitingRobots".
       -- To preserve performance, this should be done without iterating over the
       -- entire "waitingRobots" map.
-      filteredWaiting = foldr f waitingMap $ MM.toList wakeTimesToPurge
+      filteredWaiting :: MonoidMap TickNumber [RID]
+      filteredWaiting = MM.foldrWithKey f waitingMap wakeTimesToPurge
        where
-        f (k, botsToRemove) = MM.adjust (filter (`S.notMember` botsToRemove)) k
+        f k botsToRemove = MM.adjust (filter (`S.notMember` botsToRemove)) k
```
jonathanknowles, line 355: If (hypothetically) these maps were to use sets instead of lists, we might be able to simplify this further:

```hs
filteredWaiting :: MonoidMap TickNumber (S.Set RID)
filteredWaiting = waitingMap `MM.monus` wakeTimesToPurge
```
The [`monus`](https://hackage.haskell.org/package/monoidmap-0.0.4.1/docs/Data-MonoidMap.html#v:monus) operation uses the [Monus](https://hackage.haskell.org/package/monoid-subclasses-1.2.6/docs/Data-Monoid-Monus.html#t:Monus) subtraction operator [`<\>`](https://hackage.haskell.org/package/monoid-subclasses-1.2.6/docs/Data-Monoid-Monus.html#v:-60--92--62-) to subtract each value in the second map from its matching value in the first map. In the case of `Set`, `<\>` corresponds to `Set.difference`.

(Though I can definitely appreciate there are performance considerations behind the choice to use lists here.)

byorgey: Hmm, yeah, I'm really not sure what the performance implications would be here.  Might be worth trying and benchmarking, though it might be tricky figuring out a good way to stress test this code specifically.

PR #2344, opened 2025-03-18
Get rid of "world offset" + add burned patch around base

### LINKED ISSUE #242: Surround base with "burned" terrain to make it easier to see
Currently every robot that looks `down` is shown as `Ω`, which is confusing and hard to see.

We could fix this by simply changing the default char for robots (say `x`), but I propose a bolder solution, making base **bigger**:

```
┌───┐
│ Ω │
└───┘
```
First of all, it is cute to have a little person inside, and secondly we can show base with treads:
```
╓───╖
║ Ω ║
╙───╜
```
This would clearly show why small robot treads are not enough to drive the base around!

---

EDIT: On second thought, I guess this is *actually reasonable* if paired with structures (#112).
- If the `base` was not a robot, but a structure, then it could take whatever weird shape we like.
- If we add #115, then base could have its own inventory.
- It would solve the #92 since `base` would only be used to check that we are still inside (outside is ~~blocked~~ toxic)
- Moving it around on giant treads would be *cool*.

---

The current idea (see discussion below) is to just make some special "burned" terrain spawn around the base.  This would involve creating a new kind of terrain with its own (dark brown?) color, and editing the world generation function to generate that terrain near the base.

### LINKED ISSUE #1889: World DSL feature to refer to location of base
The `base` does not always end up at `(0,0)` relative to the world DSL expression used, in particular if `offset: true` is set in the scenario configuration.  There ought to be a way within the world DSL to refer to wherever the base ends up. 

See original discussion at https://github.com/swarm-game/swarm/issues/242#issuecomment-1690555056

### LINKED ISSUE #2024: Add starting game entities using DSL

We shift the base to greener pastures, but that makes everything more complicated and the DSL less powerful. I want base to stay at the zero.

Use world DSL to add whatever we need somewhere in the vicinity of base.

* Observe that base does not shift then.
* Move the shifting code to unit tests.
* Make the world a simpler and happier place ™️.

We could get the base shift in DSL and then continue adding things.

**Additional context**

This should enable adding burned terrain around base using the DSL.

byorgey:
This PR:

- Gets rid of the "world offset" feature that used to search for a "good" offset for a world function based on the presence of certain resources - both the code and the `offset` field in the world description.
- Tweaks the classic world description so that
    - The starting biome for the base has a scattering of lambdas, trees, copper ore, and rocks
    - The starting biome is adjacent to biomes containing forest, rocks, copper, + a few other necessary things
    - Anecdotally, at least, I tried playing a few games and it seemed like I was always able to find the starting resources I needed.
- Adds a "burned patch" around the base (closes #242)
    - #242 contemplates making this configurable.  I chose to leave that to a separate PR if we want to do that.
- Closes #1889 since it is no longer necessary --- the base is always at (0,0)
- Closes #2024.


xsebek (approved):
@byorgey I didn't have time to check how the new world looks but the code changes look good to me. 👍 

One little thing: I expected the offset to also come up in JSON schema, but I don't see that in the changeset.

byorgey:
> One little thing: I expected the offset to also come up in JSON schema, but I don't see that in the changeset.

It's there, you must have just missed it: https://github.com/swarm-game/swarm/pull/2344/files#diff-d5b8130cb7ce3e4047d13573336dcaf88dc9a19c34dc9708cc1e7e2a738ab1c2

Issue #1907, opened 2024-06-06
Minimize size of generated JSON for `CESK`


byorgey:
As an experiment towards #50 (see also #347), I started up a creative scenario, loaded some code from a `.sw` file, ran a base program, and then serialized a snapshot of the base's CESK machine to a file.  Unfortunately it was almost 3MB in size!  However, most of it was unnecessary junk such as empty comments records.  If we want saving games to be practical we will need to work a lot harder to minimize the size of generated JSON.

As one simple example, if a `Comments` record is empty we should be able to omit it entirely from the output.  I'm sure there is lots more similar low-hanging fruit.

byorgey:
I suspect there is a lot of bloat from multiple closures containing `Env`s that contain duplicate bindings.  In memory, there is a lot of sharing going on, but we lose the sharing when we serialize.

byorgey:
Another bit of low-hanging fruit is to make custom instances for `Type` and `Term/Syntax` where `ToJSON` pretty-prints to a string and `FromJSON` runs the parser.

xsebek:
@byorgey, how about splitting low-hanging fruits from this one? 🙂

byorgey:
> @byorgey, how about splitting low-hanging fruits from this one? 🙂

Well, to be honest, I'm not sure what they are.  I already addressed the one example I mentioned.

xsebek:
@byorgey we could split off some improvements (numbers are for Blank scenario base):

* empty comments:
  ```yaml
  _sComments:
    _afterComments: []
    _beforeComments: []
  ```
* empty location:
  ```yaml
  _sLoc: NoLoc
  ```
* capabilities include the _complete description of the entity_ that provides it _repeatedly_ (8x calculators):
    ```yaml
    - - charat
      - - device:
            capabilities:
              getMap:
                - - charat
                  - ingredients: []
                - - tochar
                  - ingredients: []
            description: |-
              A handy lookup table for converting from characters to numeric codes and back, shaped for some reason into a ring.  When equipped, it enables two functions:

              `charat : Int -> Text -> Int` returns the numeric code of the character at a specific index in a (0-indexed) `Text` value.

              `tochar : Int -> Text` creates a singleton (length-1) `Text` value containing a character with the given numeric code.
            display:
              attr: silver
              char: O
              priority: 1
            name: decoder ring
            properties:
              - pickable
            tags: []
          useCost:
            ingredients: []
    ```
* empty ingredients (399x):
  ```yaml
[...]

PR #2346, opened 2025-03-23
Tweaks towards minimizing generated JSON

byorgey:
Towards #1907 .

- Get rid of empty comments
- Get rid of empty `sLoc`
- Get rid of empty `tags`
- Get rid of default `priority`

The major things still remaining from #1907 are the empty `ingredients` and duplicated device descriptions.


xsebek (approved):
Nice! 👍

PR #2349, opened 2025-03-26
Check movement when unequipping, and add more code comments

### LINKED ISSUE #1435: Not all achievements are achievable

byorgey:
Towards #1435.  In this PR:

- A bunch of extended/updated code comments I added while getting my head around the way we do movement checks
- When executing `unequip`, checks whether moving to the current cell would be allowed without the device equipped.
    - If it would not be allowed because the path is blocked, simply disallow unequipping the device and throw an exception instead.
        - I am not sure how to test this case.  It would require having some sort of custom device that grants e.g. a special capability to move on top of entities that would normally block movement.  I am not even sure if that is possible right now. @kostmo ?
    - If moving to the current cell would be disallowed because of liquid, then the robot drowns.  In addition, if it is `base`, the `DestroyedBase` achivement is granted.

Thus, in order to get the `DestroyedBase` achievement in classic mode, you must:
- Build some `tank treads` and `equip` them
- Equip a `boat`
- Drive into the middle of some water
- `unequip` the boat

This seems deliberate enough that no one would do it by accident, i.e. it's not going to frustrate beginning players.  (There are probably some scenarios where it is easier to achieve, e.g. because the `base` actually starts out with `tank treads`.)

### DIFF
--- a/src/swarm-engine/Swarm/Game/Step/Const.hs
+++ b/src/swarm-engine/Swarm/Game/Step/Const.hs
@@ -107,7 +107,8 @@ import Text.Megaparsec (runParser)
 import Witch (From (from), into)
 import Prelude hiding (Applicative (..), lookup)
 
--- | How to handle failure, for example when moving to blocked location
+-- | How to handle failure, for example when moving into liquid or
+--   attempting to move to a blocked location
 data RobotFailure = ThrowExn | Destroy | IgnoreFail
 
 -- | How to handle different types of failure when moving/teleporting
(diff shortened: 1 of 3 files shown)


xsebek (approved):
The comments are neat! 👍 Would it be simpler if the base self-destructed in both cases?

byorgey:
> Would it be simpler if the base self-destructed in both cases?

Yes, I suppose it would be simpler, I'm just not sure that would make in-game sense.  The principle I was going for is that unequipping a device on a cell should result in an effect which is similar to what happens if you try to move onto the cell without the device equipped in the first place.
- Moving onto a liquid entity without a floatation device equipped means the robot drowns (well, except base just gives you an error about not wanting to self destruct).  Therefore unequipping a flotation device on a liquid entity does the same (with the exception that the base drowns too).
- On the other hand, if you attempt to move onto a cell with a blocking entity without a device that would let you do that, nothing happens (an exception is thrown).  Therefore unequipping on the cell does something similar.

PR #2358, opened 2025-03-29
Water disappears when placed, via new `evanescent` property

byorgey:
Based on the [discussion here](https://github.com/swarm-game/swarm/issues/1435#issuecomment-2425192480).  The idea is that you can `grab` a bucket full of water, but if you try to `place` it somewhere you just dump it and it disappears --- it doesn't create a permanent lake.  This also solves the awkward issue that technically if you created a lake underneath yourself without a `boat` equipped you should drown.

### DIFF
--- a/data/entities.yaml
+++ b/data/entities.yaml
@@ -478,14 +478,14 @@
     char: ' '
   description:
     - Liquid dihydrogen monoxide, which seems to be plentiful on this planet.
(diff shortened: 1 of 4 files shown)


xsebek (approved):
I know [what](https://en.wikipedia.org/wiki/Evanescence) I am listening to tonight. 😄

PR #2282, opened 2025-01-09
[WIP] toplevel record reorganization

kostmo:
Towards #2276.

Issue #2276, opened 2025-01-06
Move `GameState` and `UIGameplay` into new subrecord in `AppState`


kostmo:
Currently, `AppState` has a dedicated field `RuntimeState` for scenario-independent state, and a field `GameState` for scenario-dependent state.  However, `UIState` contains an awkward mix of active-play-dependent widget states and play-independent states (see #2265).

https://github.com/swarm-game/swarm/blob/e7b4d657dee7516117eb1ae6a5b43b7bb0ef66d1/src/swarm-tui/Swarm/TUI/Model.hs#L176-L181

When embarking on #2265 I noticed that `GameState` tends to be accessed often alongside `UIGameplay`.  I propose a new toplevel member `PlayingState` of the `AppState` that shall contain both `GameState` and `UIGameplay`.  Eventually, `PlayingState` should be made nullable.

This refactoring may be accomplished in phases:
1. Move `GameState` underneath a new toplevel record `PlayingState`.
2. Move `UIGameplay` out of `UIState` into a second field of `PlayingState`.
3. Wrap the `PlayingState` field in a `Maybe`
4. Make `scenarioRef` mandatory in `PlayingState` (#2265)

byorgey:
Sounds good to me.  When we first created the `AppState` record it seemed like a sensible organization to put all the gameplay stuff in one record and the UI stuff in another.  However, I think you're right that the better top-level organizational principle should be "stuff that persists across scenarios" and "stuff that only exists while playing a scenario".

PR #2362, opened 2025-03-30
toplevel records zoom

kostmo:
Continues toward #2276.

This PR incrementally pushes the state scope to the narrower `PlayState` rather than `AppState` for many utility functions, re-simplifying the code in many cases where an extra `playState` lens layer was added in #2282.  Counting the occurrences of the `playState` lens has been a good metric for the progress on this refactoring.

For functions where this is not yet feasible, common reasons are:
* need access to the `uiState . uiMenu`
    * In particular, `toggleModal`/`openModal` need access to the `Menu` for use by `generateModal`.
        * And more specifically, for the "win", "lose", and "quit" modals, either to determine if it is `NoMenu` or to determine which scenario comes next.
* need access to `uiState . uiDebugOptions`
* need access to the `ScenarioCollection` (by way of `normalizeScenarioPath`) for:
    * `saveScenarioInfoOnFinishNocheat`
    * `saveScenarioInfoOnQuit`
* Need access to `uiState . uiAchievements` for:
    * `runGameTick`
 
This exercise is informative as to which record members we may want to re-organize.

### DIFF
--- a/src/swarm-tui/Swarm/TUI/Controller.hs
+++ b/src/swarm-tui/Swarm/TUI/Controller.hs
@@ -328,51 +328,53 @@ handleMainEvent forceRedraw ev = do
     VtyEvent vev
       | isJust (s ^. playState . uiGameplay . uiDialogs . uiModal) -> handleModalEvent vev
     MouseDown (TerrainListItem pos) V.BLeft _ _ ->
-      playState . uiGameplay . uiWorldEditor . terrainList %= BL.listMoveTo pos
+      Brick.zoom playState $ uiGameplay . uiWorldEditor . terrainList %= BL.listMoveTo pos
     MouseDown (EntityPaintListItem pos) V.BLeft _ _ ->
-      playState . uiGameplay . uiWorldEditor . entityPaintList %= BL.listMoveTo pos
-    MouseDown WorldPositionIndicator _ _ _ -> playState . uiGameplay . uiWorldCursor .= Nothing
+      Brick.zoom playState $ uiGameplay . uiWorldEditor . entityPaintList %= BL.listMoveTo pos
+    MouseDown WorldPositionIndicator _ _ _ ->
+      Brick.zoom playState $ uiGameplay . uiWorldCursor .= Nothing
     MouseDown (FocusablePanel WorldPanel) V.BMiddle _ mouseLoc ->
       -- Eye Dropper tool
-      EC.handleMiddleClick mouseLoc
+      Brick.zoom playState $ EC.handleMiddleClick mouseLoc
     MouseDown (FocusablePanel WorldPanel) V.BRight _ mouseLoc ->
       -- Eraser tool
-      EC.handleRightClick mouseLoc
+      Brick.zoom playState $ EC.handleRightClick mouseLoc
     MouseDown (FocusablePanel WorldPanel) V.BLeft [V.MCtrl] mouseLoc ->
       -- Paint with the World Editor
-      EC.handleCtrlLeftClick mouseLoc
+      Brick.zoom playState $ EC.handleCtrlLeftClick mouseLoc
     MouseDown n _ _ mouseLoc ->
       case n of
-        FocusablePanel WorldPanel -> do
-          mouseCoordsM <- Brick.zoom (playState . gameState) $ mouseLocToWorldCoords mouseLoc
+        FocusablePanel WorldPanel -> Brick.zoom playState $ do
+          mouseCoordsM <- Brick.zoom gameState $ mouseLocToWorldCoords mouseLoc
           shouldUpdateCursor <- EC.updateAreaBounds mouseCoordsM
           when shouldUpdateCursor $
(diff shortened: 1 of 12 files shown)


src/swarm-tui/Swarm/TUI/Controller.hs
```diff
@@ -327,50 +328,53 @@ handleMainEvent forceRedraw ev = do
     VtyEvent vev
       | isJust (s ^. playState . uiGameplay . uiDialogs . uiModal) -> handleModalEvent vev
     MouseDown (TerrainListItem pos) V.BLeft _ _ ->
-      playState . uiGameplay . uiWorldEditor . terrainList %= BL.listMoveTo pos
+      Brick.zoom playState $ uiGameplay . uiWorldEditor . terrainList %= BL.listMoveTo pos
```
byorgey, line 331: What's the benefit of using `Brick.zoom playState` for a single line of code like this?  It seems unnecessarily complicated to me.

src/swarm-tui/Swarm/TUI/Controller.hs
```diff
@@ -327,50 +328,53 @@ handleMainEvent forceRedraw ev = do
...
     MouseDown (EntityPaintListItem pos) V.BLeft _ _ ->
-      playState . uiGameplay . uiWorldEditor . entityPaintList %= BL.listMoveTo pos
-    MouseDown WorldPositionIndicator _ _ _ -> playState . uiGameplay . uiWorldCursor .= Nothing
+      Brick.zoom playState $ uiGameplay . uiWorldEditor . entityPaintList %= BL.listMoveTo pos
+    MouseDown WorldPositionIndicator _ _ _ ->
+      Brick.zoom playState $ uiGameplay . uiWorldCursor .= Nothing
     MouseDown (FocusablePanel WorldPanel) V.BMiddle _ mouseLoc ->
       -- Eye Dropper tool
-      EC.handleMiddleClick mouseLoc
+      Brick.zoom playState $ EC.handleMiddleClick mouseLoc
```
byorgey, line 338: Are these calls to `Brick.zoom` necessary?

**Edited to add**: Oh, wait, I see, the type of e.g. `EC.handleMiddleClick` changed from `... -> EventM Name AppState ()` to `... -> EventM Name PlayState ()`.  Makes sense.

src/swarm-tui/Swarm/TUI/Controller.hs
```diff
@@ -327,50 +328,53 @@ handleMainEvent forceRedraw ev = do
...
-        InventoryListItem _ -> Just RobotPanel
-        InfoViewport -> Just InfoPanel
-        REPLViewport -> Just REPLPanel
-        REPLInput -> Just REPLPanel
-        WorldEditorPanelControl _ -> Just WorldEditorPanel
-        _ -> Nothing
-      case n of
-        FocusablePanel x -> setFocus x
-        _ -> return ()
+      m <- use $ uiState . uiMenu
```
byorgey, line 355: Use `playStateWithMenu` here?

src/swarm-tui/Swarm/TUI/Controller.hs
```diff
@@ -383,19 +387,21 @@ handleMainEvent forceRedraw ev = do
             wh <- use $ keyEventHandling . keyDispatchers . to worldDispatcher
             void $ B.handleKey wh k m
           WorldPanel | otherwise -> continueWithoutRedraw
-          WorldEditorPanel -> EC.handleWorldEditorPanelEvent ev
+          WorldEditorPanel -> do
+            m <- use $ uiState . uiMenu
+            Brick.zoom playState $ EC.handleWorldEditorPanelEvent m ev
```
byorgey, line 392: `playStateWithMenu`?

src/swarm-tui/Swarm/TUI/Controller.hs
kostmo: I suppose I was hoping to factor them out, but I'm not sure that will be possible.  Reverting them for now.

byorgey (approved):

src/swarm-tui/Swarm/TUI/Controller/Util.hs
```diff
@@ -159,63 +162,84 @@ mouseLocToWorldCoords (Brick.Location mouseLoc) = do
...
+  m a
 zoomGameState f = do
   gs <- use $ playState . gameState
   (gs', a) <- liftIO (Fused.runM (runTimeIO (Fused.runState gs f)))
   playState . gameState .= gs'
   return a
 
-onlyCreative :: (MonadState AppState m) => m () -> m ()
+-- | Modifies the game state using a fused-effect state action.
+zoomGameState' ::
```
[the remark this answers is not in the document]
kostmo: For now I've created `zoomGameStateFromAppState` and `zoomGameStateFromPlayState`.  The larger problem IMO is that either of these functions are necessary, since `Brick.zoom` doesn't work.

PR #2365, opened 2025-03-30
Don't re-validate REPL input when moving the cursor left or right

### LINKED ISSUE #2364: Don't re-process REPL input when moving the cursor
There is a noticeable lag when trying to move the cursor quickly through the REPL input.  *e.g.* if you type a moderately large input and then try to hold down an arrow key to move to another part of the input, the cursor moves noticeably slowly.  I think this is probably due to the fact that we re-process the REPL input (i.e. parse and typecheck it) after every REPL event.  However, there is no reason to re-process when moving the cursor, since cursor movement cannot possibly affect whether the input parses or typechecks.

byorgey:
Moving the cursor L/R within the REPL input now feels much snappier.  There's no need to re-validate the entire input when only moving the cursor.  Closes #2364.

### DIFF
--- a/src/swarm-tui/Swarm/TUI/Controller.hs
+++ b/src/swarm-tui/Swarm/TUI/Controller.hs
@@ -693,7 +693,12 @@ handleREPLEventTyping = \case
         uiState . uiGameplay . uiREPL . replPromptType %= \case
           CmdPrompt _ -> CmdPrompt [] -- reset completions on any event passed to editor
           SearchPrompt a -> SearchPrompt a
-        modify validateREPLForm
+
+        -- Now re-validate the input, unless only the cursor moved.
+        case ev of
+          Key V.KLeft -> pure ()
+          Key V.KRight -> pure ()
+          _ -> modify validateREPLForm
 
 insertMatchingPair :: Char -> EventM Name (Editor Text Name) ()
 insertMatchingPair c = modify . applyEdit $ TZ.insertChar c >>> TZ.insertChar (close c) >>> TZ.moveLeft


src/swarm-tui/Swarm/TUI/Controller.hs
```diff
@@ -693,7 +693,12 @@ handleREPLEventTyping = \case
         uiState . uiGameplay . uiREPL . replPromptType %= \case
           CmdPrompt _ -> CmdPrompt [] -- reset completions on any event passed to editor
           SearchPrompt a -> SearchPrompt a
-        modify validateREPLForm
+
+        -- Now re-validate the input, unless only the cursor moved.
+        case ev of
+          Key V.KLeft -> pure ()
+          Key V.KRight -> pure ()
+          _ -> modify validateREPLForm
```
byorgey, line 701: Note, it would be shorter to write something like 
```
unless (ev `elem` [Key V.KLeft, Key V.KRight]) $ modify validateREPLForm
```
However, that doesn't work, since it requires an `Eq` instance for `Event`, which in turn requires an `Eq` instance for `AppEvent`, but `AppEvent` can contain a function so that's a no-go.

kostmo: Not exactly "better", but:
```
unless (null [() | Key x <- [ev], x `notElem` [V.KLeft, V.KRight]]) $
    modify validateREPLForm
```

Incidentally, I also tried using a refutable pattern with `do` notation and the `Maybe` instance of `MonadFail`, but I couldn't get it to work:

```
let maybeExempt = do
       Key x <- ev
       guard $ x `notElem` [V.KLeft, V.KRight]
       return ()
unless (null maybeExempt) $
    modify validateREPLForm
```
It yields the error:
```
    • Couldn't match expected type ‘AppEvent’
                  with actual type ‘BrickEvent n0 e0’
    • In the pattern: Key x
```

kostmo (approved):

src/swarm-tui/Swarm/TUI/Controller.hs
byorgey: Hmm, yes, I hadn't considered using a pattern match in a list comprehension.  Maybe if we add more keys to the list in the future we could consider shortening it like this.  But in this case I think there's virtue in keeping it simple and obvious and not trying to be too clever, even if it means the code is a bit longer.

> I also tried using a refutable pattern with `do` notation and the `Maybe` instance of `MonadFail`, but I couldn't get it to work:

It looks like the types `n0` and `e0` are ambiguous, so it can't reduce `BrickEvent n0 e0` to `AppEvent`.  It could probably be solved by adding some type signatures or explicit type applications, but I'm not motivated to figure out exactly how!

PR #2366, opened 2025-03-30
simplify unequip check

kostmo:
Simplify #2349 with the beloved `forM_`.


byorgey (approved):
Nice!  Eventually I'll remember to use this myself. :smile:

PR #2368, opened 2025-04-01
Disentangle more state parameters

kostmo:
Toward #2276.

# In this PR
* Decompose `updateUI` into smaller functions
* add documentation for `UIGameplay` and `UIState` records
* As described in #2362, the `uiState . uiMenu` is used in a number of places, often just to check whether it is set to `NoMenu`.  We instead precompute this condition and pass it as a parameter.
* Instead of retrieving from the implicit `AppState`, we pass the `ScenarioCollection` ( `runtimeState . scenarios` ) explicitly to `getNormalizedCurrentScenarioPath`.

# Long-term goal

The "main layers" of the UI can either comprise the active playfield (`drawGameUI`) or the out-of-game menu (`drawMenuUI`):

https://github.com/swarm-game/swarm/blob/68e57fc040283b0d2659a6336ed6521b20c684f5/src/swarm-tui/Swarm/TUI/View.hs#L150-L154

Since these are mutually exclusive, we should extricate the state needed for either one, rather than passing the complete `AppState` to both.

In fact there are a couple of places in the code that describe "unreachable combinations" of state:

https://github.com/swarm-game/swarm/blob/68e57fc040283b0d2659a6336ed6521b20c684f5/src/swarm-tui/Swarm/TUI/View.hs#L158-L160

and

https://github.com/swarm-game/swarm/blob/68e57fc040283b0d2659a6336ed6521b20c684f5/src/swarm-tui/Swarm/TUI/Controller.hs#L155-L158

As part of the deconflation of state, we should also be able to solve this: https://github.com/swarm-game/swarm/issues/1064#issuecomment-1410915316

### DIFF
--- a/src/swarm-tui/Swarm/TUI/Controller.hs
+++ b/src/swarm-tui/Swarm/TUI/Controller.hs
@@ -432,14 +432,20 @@ handleModalEvent = \case
       _ -> do
         mdialog <- preuse $ playState . uiGameplay . uiDialogs . uiModal . _Just . modalDialog
         playStateWithMenu $ toggleModal QuitModal
+
+        menu <- use $ uiState . uiMenu
+        let isNoMenu = case menu of
+              NoMenu -> True
+              _ -> False
+
         case dialogSelection =<< mdialog of
-          Just (Button QuitButton, _) -> quitGame
+          Just (Button QuitButton, _) -> quitGame isNoMenu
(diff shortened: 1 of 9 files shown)


kostmo:
This is ready for review @byorgey

byorgey (approved):

PR #2373, opened 2025-04-03
Custom error message for missing `end` (#1141)

### LINKED ISSUE #1141: Custom error message for missing `end`
I've found a common mistake in my own swarm-lang code to be a missing `end` keyword at end of a `def`.  One reason for the frequency of this error is probably that the languages I, and I suspect most others, are most familiar with (e.g. Python and Haskell) do not require this keyword at the end of a function.

The error manifests itself far away (at the end of the file) and as the message `"unexpected end of input"`.
Since it is a common error, could we special-case this error to have a more user-friendly message?  Perhaps in the parser we can count to make sure that the number of `end` keywords matches the number of `def` keywords.

ussgarci:
Closes #1141 

When you omit a closing `end` for function definition you will see the following.

"4:1:\n |\n4 | \n | ^\nunexpected end of input\nexpecting "!=", "&&", "()", "++", "<=", "==", ">=", "def", "false", "let", "require", "requirements", "true", "tydef", "||", '"', '$', '(', '*', '+', '-', '.', '/', ':', ';', '<', '>', '[', '\', '^', **'end' keyword for definition of 'your-function'**, '{', built-in user function, direction constant, integer literal, or variable name\n"


src/swarm-lang/Swarm/Language/Parser/Term.hs
```diff
@@ -92,11 +93,20 @@ parseTermAtom2 =
...
+        --   <*> optional (symbol ":" *> parsePolytype)
+        --   <*> (symbol "=" *> parseTerm <* reserved "end")
+        --   <*> (optional (symbol ";") *> (parseTerm <|> (eof $> sNoop)))
+        <|> do
+          reserved "def"
+          locVar@(LV _srcLoc nameText) <- locTmVar
+          mTy <- optional (symbol ":" *> parsePolytype)
+          _ <- symbol "="
+          body <- parseTerm
+          reserved "end" <?> ("missing 'end' keyword for definition of '" <> T.unpack nameText <> "'")
```
byorgey, line 107: I had not thought about doing this in the parser!  Does it work?  I worry that the `parseTerm` right before `reserved "end"` will be quite happy to keep going, including the subsequent `def`.  I think it won't realize anything is wrong until it gets all the way to the end and realizes the original `def` was never closed by `end`...

ussgarci: Kinda? The message ends up as one of the possible expected alternatives so you see:

"4:1:\n  |\n4 | <empty line>\n  | ^\nunexpected end of input\nexpecting \"!=\", \"&&\", \"()\", \"++\", \"<=\", \"==\", \">=\", \"def\", \"false\", \"let\", \"require\", \"requirements\", \"true\", \"tydef\", \"||\", '\"', '$', '(', '*', '+', '-', '.', '/', ':', ';', '<', '>', '[', '\\', '^', **'end' keyword for definition of 'a'**, '{', built-in user function, direction constant, integer literal, or variable name\n"

Which isn't a very clear message...  Maybe I'm taking the wrong approach.  Do you have any suggestions about where would be a better place to implement the check?

byorgey: Oh, I see!  That makes sense.  Given that it *is* now possible to nest definitions, it might be the best we can do.

byorgey (approved):
Cool, I like this approach.  It's easy to imagine a better error message, but this is a relatively simple change and way better than the status quo (also, I don't even know how to make a better error message).

PR #2375, opened 2025-04-04
Use off-the-shelf ordered map

kostmo:
Towards #1494.

Previously, `ScenarioCollection` used a custom implementation of an ordered map.  Furthermore, it used a partial function `M.!` when converting to a list.

We now use `Data.Map.Ordered` instead.

### DIFF
--- a/src/swarm-engine/Swarm/Game/ScenarioInfo.hs
+++ b/src/swarm-engine/Swarm/Game/ScenarioInfo.hs
@@ -1,4 +1,6 @@
 {-# LANGUAGE TemplateHaskell #-}
+{-# LANGUAGE TypeFamilies #-}
+{-# OPTIONS_GHC -fno-warn-orphans #-}
 
 -- |
 -- SPDX-License-Identifier: BSD-3-Clause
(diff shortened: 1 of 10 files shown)


src/swarm-engine/Swarm/Game/ScenarioInfo.hs
```diff
@@ -137,8 +152,7 @@ normalizeScenarioPath col p =
 
 -- | Convert a scenario collection to a list of scenario items.
 scenarioCollectionToList :: ScenarioCollection -> [ScenarioItem]
-scenarioCollectionToList (SC Nothing m) = M.elems m
-scenarioCollectionToList (SC (Just order) m) = (m M.!) <$> order
```
kostmo, line 141: Here is the partial function that was eliminated.

src/swarm-engine/Swarm/Game/ScenarioInfo.hs
```diff
@@ -128,17 +143,16 @@ normalizeScenarioPath col p =
...
           let n =
                 stripPrefix (d </> "scenarios") canonPath
                   & maybe canonPath (dropWhile (== pathSeparator))
           return n
 
 -- | Convert a scenario collection to a list of scenario items.
 scenarioCollectionToList :: ScenarioCollection -> [ScenarioItem]
-scenarioCollectionToList (SC Nothing m) = M.elems m
-scenarioCollectionToList (SC (Just order) m) = (m M.!) <$> order
+scenarioCollectionToList (SC xs) = map snd $ OM.assocs xs
```
byorgey, line 155: Is there not an `OM.elems` function or something similar?

src/swarm-engine/Swarm/Game/ScenarioInfo.hs
```diff
@@ -219,19 +233,20 @@ loadScenarioDir scenarioInputs loadTestScenarios dir = do
...
     when (dirName /= testingDirectory) (warn $ OrderFileWarning orderFileShortPath NoOrderFile)
-    pure $ SC Nothing scenarioMap
+    pure . SC . OM.fromList $ M.toList scenarioMap
 
   -- warn if the ORDER file does not match directory contents
   loadOrderedScenarioDir :: [String] -> Map FilePath ScenarioItem -> m ScenarioCollection
   loadOrderedScenarioDir order scenarioMap = do
     let missing = M.keys scenarioMap \\ order
-        (loaded, notPresent) = partition (`M.member` scenarioMap) order
+        loadedEithers = map (traverse (`lookupEither` scenarioMap) . dupe) order
```
byorgey, line 242: I find this line somewhat inscrutable; I had to type it into ghci bit by bit and stare at the types before I was able to understand it.  Maybe we could split it out into a utility function?  Even just being able to see the type signature would help a lot.

byorgey (approved):
Nice, I didn't know about `ordered-containers`!

src/swarm-engine/Swarm/Game/ScenarioInfo.hs
kostmo: Weirdly not.  Defined a local util function for it.

src/swarm-engine/Swarm/Game/ScenarioInfo.hs
kostmo: I've re-written this with more explicit tuples.  I admit that exploiting the Functor instance of a tuple tends to obscure things, especially deep within some point-free chain.

PR #2379, opened 2025-04-06
Fix robot modal memory leak by only rendering modal when it is displayed

### LINKED ISSUE #2370: Memory leak from updates to robots modal dialog
**Describe the bug**
I had started Swarm with `play.sh -i classic` and accidentally left it running all day.  I might have called `run` on a file of definitions; I'm not sure.  But in any case I didn't do anything beyond that.  Just now I noticed my computer was slow and realized that Swarm was using 20% of my memory. =(

byorgey:
Fixes #2370 .  Only rerender the robots modal dialog when it is actively being displayed.  This does indeed seem to get rid of the memory leak; here's a heap profile with this fix included:

robotDisplay-heap-fixed

The only downside is that it now takes a noticeable fraction of a second for the robots dialog to be updated each time it is displayed.  However, I consider this a much better alternative to a memory leak.  I suppose we could try to update the robots modal immediately when it is toggled on, but personally I'd rather leave that as a future improvement.

### DIFF
--- a/src/swarm-tui/Swarm/TUI/Controller/UpdateUI.hs
+++ b/src/swarm-tui/Swarm/TUI/Controller/UpdateUI.hs
@@ -185,8 +185,11 @@ updateUI = do
 
   newPopups <- generateNotificationPopups
 
-  dOps <- use $ uiState . uiDebugOptions
-  Brick.zoom playState $ doRobotListUpdate dOps g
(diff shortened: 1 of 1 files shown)


xsebek (approved):
Well done, @byorgey, how did you find it?

PR #2382, opened 2025-04-07
Narrow the scope of more functions

kostmo:
More incremental refactoring to disentangle state.

Relocates `ProgressionState` to `Swarm.TUI.Model`, which had been added to `Swarm.Game.State.Runtime` in #2381.  We also introduce new record layer `ScenarioState` as a direct child of `PlayState` to contain `GameState` and `UIGameplay`, while keeping `ProgressionState` as a direct child of `PlayState`.

```
data PlayState = PlayState
  { _scenarioState :: ScenarioState
  , _progression :: ProgressionState
  }
```
and
```
data ScenarioState = ScenarioState
  { _gameState :: GameState
  , _uiGameplay :: UIGameplay
  }
```
and
```
data ProgressionState = ProgressionState
  { _scenarios :: ScenarioCollection ScenarioInfo
  , _attainedAchievements :: Map CategorizedAchievement Attainment
  , _uiPopups :: PopupState
  }
```

Now, the following functions have state dependency narrowed to only `PlayState`:
* `runGameTick`
* `saveScenarioInfoOnFinish`
* `saveScenarioInfoOnQuit`
* `saveScenarioInfoOnFinishNocheat`
* `doGoalUpdates`

Many other functions now have an even narrower dependency, on `ScenarioState`.

## Future work

Eventually, we shall wrap the `ScenarioState` member in a `Maybe`, so that the `scenarioRef` in `UIGameplay` can be made non-optional.

Another simplification will be to extract the dependency upon `Menu` from `generateModal`, by implementing `nextScenario` in terms of a linked list of scenarios (#2408)


byorgey (approved):

src/swarm-tui/Swarm/TUI/Controller/UpdateUI.hs
```diff
@@ -160,37 +162,34 @@ checkLogUpdated fr = do
...
 
-  newPopups <- generateNotificationPopups
+  newPopups <- Brick.zoom playState generateNotificationPopups
 
-  -- Update the robots modal only when it is enabled.  See #2370.
-  curModal <- use $ playState . uiGameplay . uiDialogs . uiModal
-  when ((view modalType <$> curModal) == Just RobotsModal) $ do
-    dOps <- use $ uiState . uiDebugOptions
-    Brick.zoom playState $ doRobotListUpdate dOps g
+  Brick.zoom (playState . scenarioState) $ doRobotListUpdate dOps g
```
xsebek, line 192: @kostmo @byorgey This seems to have caused a regression of #2370.

PR #2385, opened 2025-04-09
Single authoritative source for ScenarioInfo

kostmo:
Currently, the `ScenarioInfo` records for each `Scenario` are stored in two places:
* The list inside `RuntimeInfo`
* The `Menu`

This has been the source of multiple bugs: see https://github.com/swarm-game/swarm/pull/1243#pullrequestreview-1411539743, https://github.com/swarm-game/swarm/issues/1064#issuecomment-1410915316

In this PR we consolidate on the `RuntimeState` as the single authoritative source for this information and retrieve it from there at all sites that used to reference the `Menu` (in the UI).

# Implementation details

We use the `ScenarioCollection` type to represent the hierarchical structure of scenarios.  Until now this has included `ScenarioInfo` for each leaf node.  We want to re-use this type for the UI menu, but without duplicating each `ScenarioInfo`.  Therefore, we parameterize `ScenarioCollection` (and likewise `ScenarioItem`/`ScenarioInfoPair`) with a payload, which can be either a `ScenarioInfo` (for the authoritative information source) or a `ScenarioPath` (for referencing the authoritative source).

## Conversion

We can "downgrade" a `ScenarioCollection ScenarioInfo` to a `ScenarioCollection ScenarioPath` via its functor instance:
```
fmap (ScenarioPath . view scenarioPath)
```

From a `ScenarioCollection ScenarioPath`, we can recover the `ScenarioInfo` using the `loadScenarioInfoFromPath` function.

# Impact

Although the refactoring touches all type signatures involving `ScenarioInfo`, much complexity was eliminated, including:
* Deletion of the `mkNewGameMenu` function
* Removed most of the code in the `saveScenarioInfoOnQuit` function
* Eliminates a reload from disk in `constructAppState`

# Test procedure

1. Remove tutorial progress:
    ```rm ~/.local/share/swarm/saves/Tutorials_*```
2. Start the game:
    ```./scripts/play.sh```
4. Select "New game".  Observe that no tutorials show indication of progress.
5. Start the first tutorial
6. Use <kbd>Ctrl</kbd>+<kbd>q</kbd> to exit the tutorial back to the Tutorials menu
7. Observe that the tutorial entry shows an indication of progress.

### DIFF
--- a/src/swarm-tui/Swarm/TUI/Model/StateUpdate.hs
+++ b/src/swarm-tui/Swarm/TUI/Model/StateUpdate.hs
@@ -236,7 +236,7 @@ constructAppState rs ui key opts@(AppOpts {..}) = do
       }
 
 -- | Load a 'Scenario' and start playing the game.
-startGame :: (MonadIO m, MonadState AppState m) => ScenarioInfoPair -> Maybe CodeToRun -> m ()
+startGame :: (MonadIO m, MonadState AppState m) => ScenarioInfoPair ScenarioInfo -> Maybe CodeToRun -> m ()
 startGame siPair = startGameWithSeed siPair . LaunchParams (pure Nothing) . pure
 
 -- | Re-initialize the game from the stored reference to the current scenario.
@@ -248,14 +248,19 @@ startGame siPair = startGameWithSeed siPair . LaunchParams (pure Nothing) . pure
 --
 -- Since scenarios are stored as a Maybe in the UI state, we handle the Nothing
 -- case upstream so that the Scenario passed to this function definitely exists.
-restartGame :: (MonadIO m, MonadState AppState m) => Seed -> ScenarioInfoPair -> m ()
-restartGame currentSeed siPair = startGameWithSeed siPair $ LaunchParams (pure (Just currentSeed)) (pure Nothing)
+restartGame ::
+  (MonadIO m, MonadState AppState m) =>
+  Seed ->
+  ScenarioInfoPair ScenarioInfo ->
+  m ()
+restartGame currentSeed siPair =
+  startGameWithSeed siPair $ LaunchParams (pure (Just currentSeed)) (pure Nothing)
 
 -- | Load a 'Scenario' and start playing the game, with the
 --   possibility for the user to override the seed.
 startGameWithSeed ::
   (MonadIO m, MonadState AppState m) =>
-  ScenarioInfoPair ->
+  ScenarioInfoPair ScenarioInfo ->
   ValidatedLaunchParams ->
   m ()
 startGameWithSeed siPair@(_scene, si) lp = do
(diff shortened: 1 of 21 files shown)


byorgey:
Haven't looked through the changes yet, but just wanted to report quickly that I tried it out, and some scenarios in the New Game menu no longer have any indicator next to them (*e.g.* see `Classic` in the screenshot below):

new-menu

src/swarm-tui/Swarm/TUI/View.hs
```diff
@@ -232,14 +235,19 @@ drawNewGameMenuUI (l :| ls) launchOptions = case displayedFor of
...
+  drawStatusInfo s sPath = case currentScenarioInfo of
+    Nothing -> emptyWidget
+    Just si -> case si ^. scenarioStatus of
+      NotStarted -> txt " ○ "
+      Played _script _latestMetric best | isCompleted best -> withAttr greenAttr $ txt " ● "
+      Played {} -> case s ^. scenarioOperation . scenarioObjectives of
+        [] -> withAttr cyanAttr $ txt " ◉ "
+        _ -> withAttr yellowAttr $ txt " ◎ "
+   where
+    currentScenarioInfo = appState ^? runtimeState . scenarios . scenarioItemByPath sPath . _SISingle . _2
```
kostmo, line 250: The status indicator will not be rendered if the path cannot be looked up in the toplevel `scenarios` collection.

kostmo:
> some scenarios in the New Game menu no longer have any indicator next to them

@byorgey can you check out the commit in #2388 and see what the paths for the broken scenarios are?  For me, "First steps" has path `Tutorials/world101.yaml`.  All other tutorial paths start with `Tutorials/`.
Screenshot from 2025-04-09 17-14-10

byorgey:
> can you check out the commit in #2388 and see what the paths for the broken scenarios are?

Ah, indeed, the path for `Classic game` is `/home/brent/projects/swarm/./data/scenarios/classic.yaml`, and the paths for the other broken scenarios are similar.  Weird, I wonder how they got like that.  All the scenarios that work have a path that starts with `Tutorials/`.   I looked in the save files in `~/.local/share/swarm/saves/` and verified that the paths stored there are exactly the same --- the save file called `Tutorials_world101.yaml` has `path: /home/brent/...` but e.g. `Tutorials_bind2.yaml` has `path: Tutorials/bind2.yaml`.

kostmo:
> I wonder how they got like that

I think the issue is #2390.  I have pushed one more commit to #2385; @byorgey could you please check if that fixes your issue?

byorgey:
> I think the issue is #2390. I have pushed one more commit to #2388; @byorgey could you please check if that fixes your issue?

@kostmo Yes, that fixes it.

src/swarm-tui/Swarm/TUI/Model/StateUpdate.hs
```diff
@@ -146,6 +146,15 @@ initPersistentState opts@(AppOpts {..}) = do
   let initRS' = addWarnings initRS (F.toList warnings)
   return (initRS', initUI, initKs)
 
+loadScenarioInfoFromPath ::
```
byorgey, line 149: To me the word `load` implies reading something from disk.  Can we name this something like `scenarioInfoFromPath`, `scenarioInfoByPath`, `getScenarioInfoFromPath`, etc?

byorgey (approved):
What a fantastic refactoring!  Removes a bunch of code and gets rid of potential bugs at the same time, what more could one ask for?

PR #2389, opened 2025-04-10
Assign scenario path deeper in game state construction

kostmo:
There was a comment that started with:

> Beware: ...

I've refactored that logic to avoid the forewarned mistake.

### DIFF
--- a/src/swarm-tui/Swarm/TUI/Model/StateUpdate.hs
+++ b/src/swarm-tui/Swarm/TUI/Model/StateUpdate.hs
@@ -296,9 +296,6 @@ startGameWithSeed siPair@(ScenarioWith _scene si) lp = do
       (prevBest t)
 
   scenarioToAppState (pathifyCollection siPair) lp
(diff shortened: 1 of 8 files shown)


src/swarm-tui/Swarm/TUI/Model/StateUpdate.hs
```diff
@@ -296,9 +296,6 @@ startGameWithSeed siPair@(_scene, si) lp = do
       (prevBest t)
 
   scenarioToAppState (pathifyCollection siPair) lp
-  -- Beware: currentScenarioPath must be set so that progress/achievements can be saved.
-  -- It has just been cleared in scenarioToAppState.
-  playState . gameState . currentScenarioPath .= Just p
```
kostmo, line 301: Getting rid of this warning is the main benefit of this refactor.

byorgey (approved):

byorgey:
Why didn't the usual CI checks run on this PR?

Issue #2303, opened 2025-01-28
Devices/entities with no recipe


byorgey:
This is a new umbrella issue for entities or devices without a recipe. Eventually we should create recipes so the entities are craftable in classic mode.

- [ ] `Elmer's glue`
- [ ] `ash`
- [ ] `binoculars`
- [ ] `blueprint`
- [ ] `caliper`
- [ ] `decoder ring`
- [ ] `linotype`
- [ ] `tape drive`
- [ ] `wedge`

Related: #2030.

Older versions of this issue: #26 , #1268.

PR #2391, opened 2025-04-11
Recipes for binoculars, caliper, glue

kostmo:
Towards #2303

Introduces recipes for:
* binoculars
* caliper
* Elmer's glue

For glue, additionally introduced new entities:
* wheat
* millstone
* flour

Also sprinkled some wheat in the world DSL as per requirements.

### DIFF
--- a/data/worlds/classic.world
+++ b/data/worlds/classic.world
@@ -89,6 +89,7 @@ overlay
   ( overlay
     [ {grass}
     , mask (hash % 20 == 10) {grass, cotton}
+    , mask (hash % 20 == 10) {grass, wheat}
     , mask (hash % 20 == 0)  {grass, flower}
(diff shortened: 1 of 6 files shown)


data/worlds/classic.world
```diff
@@ -89,6 +89,7 @@ overlay
   ( overlay
     [ {grass}
     , mask (hash % 20 == 10) {grass, cotton}
+    , mask (hash % 20 == 10) {grass, wheat}
```
byorgey, line 92: Since both `cotton` and `wheat` use the same mask, one will completely override the other (I forget which).  Should use something like e.g. `hash % 20 == 11` for `wheat`.

byorgey (approved):
I like this!

Issue #2392, opened 2025-04-11
Tutorial play test Mega Issue


xsebek:
I am slowly working through the tutorials. Since I last played through them, some have changed, and new ones have been added.

I have found some bugs and things to improve, which I plan to track here:

* [ ] Robot detail modal is too tiny to read messages
* [ ] Robot detail modal time is not formatted in clock format
* [ ] The Debug tutorial message in the crashed robots log has line breaks which do not show well in the wrapped log
* [ ] The common "[Tab] toggles focus between panes" is confusing, especially in the tutorial that teaches how to navigate between panes
* [x] The scan tutorial sounds like you need to scan all entities - "scan them" - but scanning one is enough.
* [x] The give tutorial should remind players that they can use bind notation to get the result of grab.
* [x] The give tutorial requires players to count how distant entities are. Maybe add a chess board or at least move base to 0,0.
* [x] The give tutorial is easy to fail, because there is only one piece of each item, and you can not `salvage` robots.
* [x] The lambda tutorial move pattern is hard for me to see, which makes it intimidating. Also, it has old non-connected walls.
* [ ] The First steps tutorial should remind players that unknown entities are `?` and maybe hint which one is lambda.
* [ ] The First steps lambda is pretty far (-26, 12) and there are two other purple entities closer.
* [ ] The Farming tutorial requires water, but does not call it out - robot has to go with boat to water to not drown.

PR #2395, opened 2025-04-12
Draft: Update debug scenario and robot detail modal

### LINKED ISSUE #2228: Improve layout/use of space in Robots (F2) dialog
image

Some ideas discussed on Discord:

> xsebek: We could use the horizontal space better here for sure 😅
> xsebek: Maybe we could cycle between a full message view and statistics tab for a single robot 🤔 otherwise we will always have space that could have been used for messages or would have to put messages below statistics, possibly out of view
> ChucklesTheBeard: I was thinking maybe "hit enter on a log entry to show an expanded view of that entry, but even a horizontal scroll would be better than standard system messages getting cut off

xsebek:
* rename to debug.yaml to match scenario name
* tweak the map so that it is easier to see where the robot should crash
* change the crashed robot log hint so that it does not have line breaks inside sentences
* change hint robot code so that it does not spam logs
* use more available space in the robot detail pane to make the log readable - closes #2228 
* use hexadecimal time format in `Tick` pretty instance
* part of #2392


data/scenarios/Tutorials/debug-hint.sw
```diff
@@ -0,0 +1,111 @@
...
+  }
+end;
+
+// An infinite while cycle that carries a state.
+def iterate = \state.\com.
+  n <- com state;
+  iterate n com;
+end;
+
+tydef RobotState = [gave_win: Bool, said_log_missing: Bool, said_loc: (Int * Int)] end;
```
byorgey, line 19: oooh records :cd:

byorgey (approved):
I played through the improved tutorial and it works well!

byorgey:
Looks like the CI is failing because `test/integration/Main.hs` still refers to `Tutorials/crash` instead of `Tutorials/debug`.

xsebek:
I will add a few more improvements, stay tuned! 📀

src/swarm-engine/Swarm/Game/Step/Util/Command.hs
```diff
@@ -183,9 +183,12 @@ updateRobotLocation oldLoc newLoc
   | otherwise = do
       newlocWithPortal <- applyPortal newLoc
       rid <- use robotID
+      t <- use $ temporal . ticks
+      invis <- use $ robotDisplay . invisible
       zoomRobots $ do
         removeRobotFromLocationMap oldLoc rid
         addRobotToLocation rid newlocWithPortal
+        unless invis $ wakeWatchingRobots rid t newLoc
```
xsebek, line 191: @kostmo I would also like to change the `watch` to notify robots. I want to avoid system robots having to busy-wait, and this would help a lot.

But to do this radical change, I would need help. Four challenges that use `watch` fail with this change, and I am unable to decipher how they work:
* Challenges/gopher
* Challenges/Sliding Puzzles/3x3
* Challenges/Ranching/capture
* Challenges/Sokoban/foresight.yaml

Could you please check if the solutions could work with waking by (visible) robots?

---

<sup>I thought about making a new command for compatibility, but that would need a new watch map. If possible, I would like to avoid the mental overload of too many commands both for players and developers.</sup>

data/scenarios/Tutorials/debug.yaml
```diff
@@ -4,23 +4,26 @@ description: |
...
 objectives:
   - goal:
-      - Before you send your robots far away from the `base` you need to learn how to figure out what went wrong with them if they crash.
       - |
-        In this challenge, you should start by
-        sending a robot to walk four steps straight east into the mountain,
-        crashing deliberately.  However, you must make sure it has a `logger`{=entity},
-        so we can see what command failed.  The simplest way to ensure
+        Before you send your robots far away from the `base` you need to learn how
+        to figure out what went wrong with them if they crash.
```
xsebek, line 9: @byorgey I am considering adding an objective to salvage a built robot in the same space first.
I worry that this scenario teaches too much at once.

What do you think?

byorgey: Sure, I think that makes sense.

PR #2403, opened 2025-04-13
initial all-contributors data

byorgey:
Made a quick first pass.  Feel free to add or remove contribution types for yourself.  We should also try to go through and add e.g. people who filed bug reports or ideas but didn't make a PR.


.all-contributorsrc
```diff
@@ -1,4 +1,289 @@
 {
   "projectName": "swarm",
-  "projectOwner": "swarm-game"
+  "projectOwner": "swarm-game",
+  "repoType": "github",
+  "commitConvention": "angular",
+  "contributors": [
```
xsebek, line 6: Is there an order to the list?
I feel a bit uncomfortable on the first position. 😄

xsebek (approved):

.all-contributorsrc
byorgey: Good question, it looks like it just preserves the arbitrary order in which I added people (I think I added you first, since you were in my mind after you reviewed the previous PR, then I went back and added everyone from the list of github contributors more systematically).  We could sort them alphabetically or something, though that order would not be preserved as we add more in the future.  Maybe the better idea is to just order them by the date of their first contribution, since that's what order it would be in if we had added them incrementally along the way.

byorgey: OK, I've updated the order to be roughly when people made their first contribution.

PR #2404, opened 2025-04-18
Improve Give tutorial

xsebek:
* split the `give` tutorial into two parts
* replace grabbed items in give tutorial
* part of #2392

### DIFF
--- a/data/scenarios/Tutorials/give.yaml
+++ b/data/scenarios/Tutorials/give.yaml
@@ -18,7 +18,7 @@ objectives:
       - |
         This is a training simulation, so the items will reappear after a while.
       - |
-        Start with `build`ing a robot to `grab` the `bit (0)`{=entity} east of the base.
+        Start with `build`ing a robot to `grab` the `bit (0)`{=entity} east of the base, then return to `give` it to the `base`.
     condition: |
       as base { has "bit (0)" }
   - id: give_rest
@@ -30,13 +30,13 @@ objectives:
         Where will you find the item names, so you can `give base "item name"`?
         That's the neat part, you won't!
       - |
-        You will get the name from the `grab` command result:
+        You will get the name from the `grab` command result, for example:
         ```
         build { move; w <- grab; log w; turn back; move; give base w }
         ```
       - |
         **TIP:** Instead of trial and error, you can get the distance to items
-        by clinking on them in the world map, which will show the coordinates.
+        by clicking on them in the world map, which will show the coordinates.
     prerequisite: give_bit
     condition: |
       as base {
--- a/data/scenarios/Tutorials/give.yaml
+++ b/data/scenarios/Tutorials/give.yaml
@@ -28,7 +28,7 @@ objectives:
         Good job. Now send robots to grab the rest!
       - |
         Where will you find the item names, so you can `give base "item name"`?
-        That's the best part, you won't!
+        That's the neat part, you won't!
       - |
         You will get the name from the `grab` command result:
         ```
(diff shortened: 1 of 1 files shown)


data/scenarios/Tutorials/give.yaml
```diff
@@ -3,32 +3,59 @@ name: Give
...
+        Where will you find the item names, so you can `give base "item name"`?
+        That's the neat part, you won't!
+      - |
+        You will get the name from the `grab` command result:
+        ```
+        build { move; w <- grab; log w; turn back; move; give base w }
+        ```
+      - |
+        **TIP:** Instead of trial and error, you can get the distance to items
+        by clinking on them in the world map, which will show the coordinates.
```
byorgey, line 39:
```suggestion
        by clicking on them in the world map, which will show the coordinates.
```

data/scenarios/Tutorials/give.yaml
```diff
@@ -3,32 +3,59 @@ name: Give
...
         is often the special `base` robot, but could be any robot.
-      - In this scenario, you will need to collect the items scattered around your `base`, by `build`ing one or more robots to collect them and then `give` them to you.
+      - |
+        In this scenario, you will need to collect the items scattered
+        around your `base`, by `build`ing one or more robots to collect
+        them and then `give` them to you.
+      - |
+        This is a training simulation, so the items will reappear after a while.
+      - |
+        Start with `build`ing a robot to `grab` the `bit (0)`{=entity} east of the base.
```
byorgey, line 21:
```suggestion
        Start with `build`ing a robot to `grab` the `bit (0)`{=entity} east of the base, then return to `give` it to the `base`.
```
I know we have just been explicitly talking about `give`, but I think it's worth repeating the instruction here.  The first time I tried this scenario I read "start by building a robot to grab the bit" and I figured turning around and giving it to the base would come later.  So I did literally what it said, I built a robot to go and `grab` the bit and... nothing happened.

data/scenarios/Tutorials/give.yaml
```diff
@@ -3,32 +3,59 @@ name: Give
...
+  - id: give_rest
+    teaser: Fetch the rest
+    goal:
+      - |
+        Good job. Now send robots to grab the rest!
+      - |
+        Where will you find the item names, so you can `give base "item name"`?
+        That's the neat part, you won't!
+      - |
+        You will get the name from the `grab` command result:
```
byorgey, line 33:
```suggestion
        You will get the name from the `grab` command result, for example:
```
Just to make it clear that you are not necessarily supposed to run this exact code.

data/scenarios/Tutorials/give.yaml
```diff
@@ -47,17 +74,27 @@ robots:
       - [10, scanner]
       - [10, treads]
       - [10, solar panel]
+  - name: replacer
+    system: true
+    program: |
+      def replace: Cmd Unit =
+          i <- scan down;
+          watch down; wait 2024; wait 32;
```
byorgey, line 82: This is an example where I think it's helpful to have a version of `watch` which *only* pays attention to entity changes, not robots entering the cell --- because we have nothing useful to do when a robot shows up, only when the entity disappears.

byorgey (approved):

data/scenarios/Tutorials/give.yaml
xsebek: I agree that this use only cares about entity changes, but it has to handle errors anyway.

It is generally wrong to assume that if watching robot was woken up, then the cell now (doesn't) contain an entity. Other robots could have gone before it.

That "something changed in the cell at least once" is all the robot knows.

But it could have also been woken up at the end of wait time, as I did not implement infinity yet.

byorgey: True, all good points.

xsebek:
@mergifyio requeue

PR #2410, opened 2025-04-21
Disentangle modal types

### LINKED ISSUE #2408: Implement `nextScenario` in terms of state not dependent on the `Menu`
Currently, the `generateModal` function depends on `Menu` (from `uiState . uiMenu`) because of `nextScenario`.  However, it's weird to store that state in the menu itself; we could instead carry around a linked list of "next scenarios" somewhere in `PlayState`.

kostmo:
Closes #2408.

There are two distinct categories of modals which both go through the same function for toggling/generating/opening.  However, one of these categories requires access to more state than the other.

By introducing an additional layer of sum types, we can simplify the code path for one of these categories which requires less state access.

# Test procedure

## Test case 1

1. `rm ~/.local/share/swarm/saves/Tutorials_*`
1. Launch game
2. Select "Tutorials" from toplevel menu
3. Complete first tutorial, select "Next challenge!" in popup dialog
4. Complete second tutorial, select "Next challenge!" in popup dialog
5. Observe that we are now playing the "Crafting" tutorial.

## Test case 2

1. Launch game
2. Select "New game"
3. Select "Tutorials" category
4. Select an arbitrary tutorial
5. Complete two in sequence, using "Next challenge!" each time
6. Observe correct sequence was followed

### DIFF
--- a/src/swarm-tui/Swarm/TUI/Controller.hs
+++ b/src/swarm-tui/Swarm/TUI/Controller.hs
@@ -33,7 +33,7 @@ import Brick.Focus
 import Brick.Keybindings qualified as B
 import Brick.Widgets.Dialog
 import Brick.Widgets.Edit (Editor, applyEdit, editContentsL, handleEditorEvent)
-import Brick.Widgets.List (handleListEvent)
+import Brick.Widgets.List (handleListEvent, listElements)
 import Brick.Widgets.List qualified as BL
 import Brick.Widgets.TabularList.Mixed
 import Control.Applicative (pure, (<|>))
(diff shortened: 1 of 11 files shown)


src/swarm-tui/Swarm/TUI/Controller.hs
```diff
@@ -692,46 +701,47 @@ handleREPLEventTyping m = \case
...
         case formSt of
           CmdPrompt {} -> continueWithoutRedraw
           SearchPrompt _ -> resetREPL "" (CmdPrompt [])
       ControlChar 'd' -> do
-        text <- use $ uiGameplay . uiREPL . replPromptText
+        text <- use $ scenarioState . uiGameplay . uiREPL . replPromptText
         if text == T.empty
-          then toggleModal QuitModal m
+          then toggleEndScenarioModal QuitModal m
           else continueWithoutRedraw
```
kostmo, line 725: This is the only keyboard input that requires the wider `PlayState`, rather than the narrow `ScenarioState`.

kostmo:
This is ready for review, @byorgey.

byorgey (approved):
Another nice refactoring!  Sorry it took me so long to get around to reviewing!

PR #2415, opened 2025-04-27
Add quasiquoter for raw, untyped ASTs

byorgey:
We already have quasiquoters for types and for fully typechecked ASTs. This PR adds a quasiquoter for raw, untyped ASTs, with the ability to splice existing `Syntax` nodes into quasiquotes.  I needed this for #2414 and decided to split it out into a separate PR.


src/swarm-lang/Swarm/Language/Syntax/AST.hs
```diff
@@ -71,6 +71,8 @@ data Term' ty
     TAntiText Text
   | -- | A Boolean literal.
     TBool Bool
+  | -- | An antiquoted Haskell variable name of type Syntax.
+    TAntiSyn Text
```
xsebek, line 75: Why not `TAntiVarName`?

src/swarm-lang/Swarm/Language/Parser/QQ.hs
```diff
@@ -40,3 +43,27 @@ quoteTypeExp s = do
...
+
+quoteASTExp :: String -> TH.ExpQ
+quoteASTExp s = do
+  loc <- TH.location
+  let pos =
+        ( TH.loc_filename loc
+        , fst (TH.loc_start loc)
+        , snd (TH.loc_start loc)
+        )
+  parsed <- runParserTH pos (fully sc parseTerm) s
```
xsebek, line 64: Why does `runParserTH` not take TH location directly?

src/swarm-lang/Swarm/Language/Syntax/Pretty.hs
```diff
@@ -70,6 +70,7 @@ instance PrettyPrec (Term' ty) where
     TAntiInt v -> "$int:" <> pretty v
     TText s -> fromString (ushow s)
     TAntiText v -> "$str:" <> pretty v
+    TAntiSyn v -> "`" <> pretty v
```
xsebek, line 73: Why the tick? I would have expected `$var:`.

Wouldn't this be problematic if we added infix call like Haskell has with ticks on both sides of the identifier?

xsebek (approved):
Cool, I wonder how you plan to use it. 🙂

src/swarm-lang/Swarm/Language/Syntax/AST.hs
byorgey: All antiquotations are variable names; the interesting part is the *type* of the antiquoted variable.

src/swarm-lang/Swarm/Language/Syntax/Pretty.hs
byorgey: I don't know, I just thought it looked cool.  I can change it to `$syn:` to be consistent with the other ones.

src/swarm-lang/Swarm/Language/Parser/QQ.hs
byorgey: That... is a good question!

byorgey:
> I wonder how you plan to use it.

You'll see shortly, but it's to implement your suggestion of having `format` add definitions for `fst` and `snd` when converting from version 0.6 code that uses them.  It was going to be super annoying to write out the actual syntax trees of type `Syntax` corresponding to the definitions.  This way, we can just write actual inline Swarm code.

PR #2424, opened 2025-05-03
Add Swiss Army Robot achievement for equipping all craftable devices in classic game

### LINKED ISSUE #1030: Achievement for equipping all canonical devices
The achievement could be titled "Swiss Army Robot".

See [comment](https://github.com/swarm-game/swarm/pull/1029#issuecomment-1399961935).

byorgey:
Closes #1030.

- Adds new `EquippedAllDevices` achievement
- Moves the recipe graph analysis code from `swarm-doc` to a new module `Swarm.Game.Recipe.Graph` in `swarm-scenario`.
- Adds a new field `craftableDevices` to the `discovery` subrecord in `gameState`, containing a set of entity names which are available in the world or recursively craftable from what is available.
- Grant the achievement when the player's equipped devices become a superset of the craftable devices, and the current scenario is `classic.yaml`.

This achievement is extremely difficult!  I have not actually been able to test it directly.  Here is a way to test that it seems to be working:

- Temporarily remove the check whether the current scenario is `classic.yaml`
- Load `Tutorials/equip`
- `equip "solar panel"`
- Grab the 3D printer and equip it
- You should be granted the achievement.

For now I have restricted the achievement only to classic mode, but am open to arguments that we should somehow open it up more.

### DIFF
--- a/src/swarm-scenario/Swarm/Game/Recipe/Graph.hs
+++ b/src/swarm-scenario/Swarm/Game/Recipe/Graph.hs
@@ -16,7 +16,7 @@ import Data.Set (Set)
 import Data.Set qualified as Set
 import Data.Tuple (swap)
 import Swarm.Failure (simpleErrorHandle)
-import Swarm.Game.Entity (Entity, EntityMap (entitiesByName), entityProperties, entityYields, EntityProperty(Pickable))
+import Swarm.Game.Entity (Entity, EntityMap (entitiesByName), EntityProperty (Pickable), entityProperties, entityYields)
 import Swarm.Game.Entity qualified as E
 import Swarm.Game.Land
 import Swarm.Game.Recipe (Recipe, recipeCatalysts, recipeInputs, recipeOutputs)
--- a/src/swarm-scenario/Swarm/Game/Recipe/Graph.hs
+++ b/src/swarm-scenario/Swarm/Game/Recipe/Graph.hs
@@ -1,0 +1,160 @@
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE RecordWildCards #-}
+{-# LANGUAGE TemplateHaskell #-}
+
+-- |
+-- SPDX-License-Identifier: BSD-3-Clause
+-- Description: calculate the graph of craftable entities
+-- and their dependencies via recipes.
+module Swarm.Game.Recipe.Graph (
+  RecipeGraph (..),
+  classicScenarioRecipeGraph,
+  scenarioRecipeGraph,
+) where
+
+import Control.Lens (view, (^.))
+import Data.Map.Lazy (Map)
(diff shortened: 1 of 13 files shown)


xsebek (approved):
Cool challenge. 👍

src/swarm-scenario/Swarm/Game/Recipe/Graph.hs
```diff
@@ -0,0 +1,146 @@
...
+-- This is almost a BFS, but the requirement is that the set of entities
+-- required for recipe is subset of the entities known in Level N.
+--
+-- If we ever depend on some graph library, this could be rewritten
+-- as some BFS-like algorithm with added recipe nodes, but you would
+-- need to enforce the condition that recipes need ALL incoming edges.
+recipeLevels :: EntityMap -> [Recipe Entity] -> Set Entity -> [Set Entity]
+recipeLevels emap recipeList start = levs
+ where
+  recipeParts r = ((r ^. recipeInputs) <> (r ^. recipeCatalysts), r ^. recipeOutputs)
```
xsebek, line 122: IIRC I wrote this naively. It should check the catalysts separately. 🤔

So if some starting device was not craftable and at the same time needed in recipe, the achievement would be impossible.

It's probably fine though. 🤞

byorgey: Hmm, I'm not 100% sure I understand.  Why would we need to treat catalysts differently than recipe inputs in this calculation?

xsebek: Because at the start you have devices equipped that you can not craft yet. The algorithm wrongly thinks you also have them ready as ingredients.

I don't have a real example on mind, but imagine if you had equipped lambda and lambdas were created from "alphabet soup", then the graph would show curry as available immediately from lambda at start. It should instead show curry only after you have obtained lambda _as ingredient_.

Hopefully it's only a rare edge case and does not matter for the achievement. I just wanted to point it out since you said it's difficult.

byorgey: Ah, I see.  So it's not really about catalysts specifically, but about whether you have enough of uncraftable entities.

- If you have a device equipped that is needed as a catalyst, great.
- If you have a device equipped that is needed as an ingredient, well, you can `unequip` it, but then it will be gone and you don't have it equipped any more.
- There could also be some uncraftable entity which is available in the world, but is non-renewable and available only in a limited amount.  Or maybe you start with some in your inventory but it's not obtainable from the world.  In either case there might not be enough in order to make all the things you are supposed to make.

Having the recipe graph analysis take all this into account sounds extremely complicated (maybe even undecidable). I think we can just leave that up to careful design and not worry about it in the analysis.

PR #2431, opened 2025-05-09
Disallow calls to `error`

byorgey:
Towards #1494.  This PR eliminates a number of calls to `error`, explicitly marks the rest as safe, and disallows future calls to `error` via the `hlint` configuration.


xsebek (approved):

src/swarm-scenario/Swarm/Game/Scenario/Topography/WorldPalette.hs
```diff
@@ -71,11 +69,13 @@ constructWorldMap mappedPairs maskChar =
...
 
+  renderMapCell :: Maybe CellPaintDisplay -> Char
   renderMapCell maybeC = case maybeC of
     Nothing -> maskChar
     Just c ->
-      -- NOTE: This lookup should never fail
-      M.findWithDefault (error "Palette lookup failed!") k $
+      -- NOTE: This lookup should never fail; if it does for some
+      -- reason, silently return a space rather than crash the game
+      M.findWithDefault ' ' k $
```
kostmo, line 78: Perhaps make this a visible character like 'Z', just in case we we actually do encounter it at runtime and need to search for it in the source code.

kostmo (approved):

src/swarm-scenario/Swarm/Game/Scenario/Topography/WorldPalette.hs
byorgey: Good idea, done.

PR #2436, opened 2025-05-11
Add some optional goals to the classic scenario

byorgey:
Adds some optional goals to the classic scenario (thanks to @xsebek for the idea), to help nudge players in the direction of some useful goals.  This should be considered just a start --- there are many more goals we could add after these (e.g. crafting a metal drill; mining quartz; creating a deep mine + mining gold, silver, and mithril; crafting many other devices, etc.) as well as more whimsical goals we could add just for fun.

Also removes the `known` property from several entities, which I discovered while testing this PR.

### DIFF
--- a/data/scenarios/classic.yaml
+++ b/data/scenarios/classic.yaml
@@ -32,3 +32,132 @@ world:
   scrollable: false
   dsl: |
     "classic"
+objectives:
+  - id: trees
+    teaser: Trees
+    goal:
+      - |
+        Some goals will be shown here, to gently nudge you towards
+        some useful accomplishments if you'd like some guidance.
+        However, these goals are optional!  Feel free to ignore them
+        and choose your own path.
+      - |
+        One of the first things you will probably want is a few `tree`{=entity}s.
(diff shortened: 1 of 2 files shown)


data/scenarios/classic.yaml
```diff
@@ -32,3 +32,132 @@ world:
...
+        Establish some mines and collect both `copper ore`{=entity}
+        and `iron ore`{=entity}.
+    optional: true
+    prerequisite: drill
+    condition: |
+      as base { i <- has "iron ore"; c <- has "copper ore"; pure (i && c) }
+  - id: unwinnable
+    goal:
+      - |
+        The classic scenario is not actually winnable (yet).
```
xsebek, line 160: Do we have an issue for that? 🙂

xsebek (approved):

data/scenarios/classic.yaml
byorgey: I don't think so, but personally I think we need to add a bunch more features before we start thinking about win conditions!

Issue #2343, opened 2025-03-10
Save source location information in types


byorgey:
We save source spans in the AST representing terms, but we don't save any source location information for types; but we should!

One place this recently came up is #2340 , where it's possible to have an error message referring to two different type variables with the same name, where one shadows the other.  It could be nice to have source location info in order to disambiguate them in the error message.

I'm not yet sure whether this means saving source location info in every single type and threading it through everywhere, or whether we should have two separate types: one representing an abstract type, and one representing concrete parsed syntax for a type with embedded source info.

byorgey:
See also #1764 .  We need source locations in types to be able to figure out what type name the user is hovering on, to implement jump-to-definition for type definitions.

PR #2438, opened 2025-05-13
Expand all tydefs in `infer` and `check`

### LINKED ISSUE #2223: Handle `read` at user-defined types

### LINKED ISSUE #2437: Redefining user types can break type soundness
**Describe the bug**
If a user-defined type is redefined, it can break type soundness, because the inferred type of a function can mention user-defined type names which are redefined in between declaring and using the function.

**To Reproduce**
```
> tydef Foo = Int end
> def f : Int -> Foo = \x. x + 1 end
> tydef Foo = Bool end
> if (f 3) {} {}
```
The above code typechecks but causes a CESK machine fatal error, *i.e.* it breaks type soundness.  The problem is that the definition of `f` was checked in the context of the definition `Foo = Int`, but later when it is used, `f`'s type `Int -> Foo` is incorrectly interpreted in the context of the definition `Foo = Bool`.

**Expected behavior**
Ordered roughly from easiest to hardest, here are three things I can think of that we could do instead:
1. Disallow `tydef`s that redefine a previously defined type.
    - This would be very simple to implement but probably annoying in practice.  I have definitely defined a type alias at the REPL and then realized that it was wrong and wanted to change it.
3. Greedily expand all `tydef`s so that they only ever show up in source code, never in the recorded inferred type of a definition.  In the above example, `f`'s type would then be recorded as `Int -> Int`, so `if (f 3)` would not type check.
    - I think this is relatively straightforward; the downside would be that the inferred types shown for things at the REPL would no longer mention user-defined types.
    - See #2438 .
4. Store a context of relevant type definitions along with each type which were in scope at the time it was defined, so we know the correct definition to use to interpret it.
    - This seems like the most technically correct solution, but would probably require extensive refactoring.
5. **Updated to add**: here is a new idea, inspired by what `ghci` seems to do in this situation.  Just introduce one extra layer of indirection: when we process `tydef T = blah`, do not directly store a mapping from `T` to `blah`.  Instead, generate a new globally unique name, let's call it `T746`.  Then store two mappings `T -> T746` and `T746 -> blah`.
    - When processing a type like `f : T -> T`, substitute once to get `T746 -> T746`. This ensures that if the definition of `T` changes later, the type of `f` will not change.
    - However, as long as the mapping `T -> T746` is still in scope, we can display the type of `f` as `T -> T` instead of `T476 -> T476`.  If `T` is ever redefined, then the type of `f` will have to be displayed differently.
    - Note, however, this does not solve #2223.

**Additional context**
Related: #2223; options (2) or (3) above would also fix that issue.

Fundamentally, the issue here is that we want each subsequent `tydef` of the same name to *shadow* the previous definition, not *mutate* it.  In other words, any uses of a defined type name after the second `tydef` should refer to it, but prior uses should continue to refer to the previous definition.

Here is what `ghci` does in a similar situation:
```
λ> type Foo = Int
λ> f :: Foo -> Foo; f x = x + 1
λ> type Foo = Bool
λ> :type f
f :: Ghci10.Foo -> Ghci10.Foo
```

byorgey:
This PR does two related things:

1. Immediately expand all user tydefs in `infer` and `check`, so inferred types are always fully expanded and will never contain user tydefs.  This is a bit annoying, especially in the case of tydefs created to abbreviate long, complicated types, but seems necessary to solve soundness issues like #2437.
    - This closes #2223, since now any type annotation on `read` will be immediately expanded, so no user-defined types will be encountered when evaluating the `read` at runtime.
    - This also closes #2437, since the inferred type of a function stored in the context can no longer contain a user-defined type, and so we cannot confuse the type system by redefining one.
2. Throw a proper error in case an undefined type constructor is encountered in `expandTydef`.
    - In #2431 we got rid of a call to `error`.
    - However, instead we just returned the same, unchanged and unexpanded type, which caused its own issues: when doing unification, if we see the same pair of types a second time, we assume by coinduction that the types unify.  This doesn't work if a type can expand to itself, and was the cause of the fatal CESK error in #2223.
    - This PR handles things properly, by throwing an exception which will be propagated appropriately in case anything like this ever happens again.

### DIFF
--- a/src/swarm-lang/Swarm/Effect/Unify/Fast.hs
+++ b/src/swarm-lang/Swarm/Effect/Unify/Fast.hs
@@ -39,6 +39,7 @@ import Data.Map.Merge.Lazy qualified as M
 import Data.Monoid (First (..))
 import Data.Set (Set)
 import Data.Set qualified as S
+import Swarm.Util.Effect (withThrow)
 import Swarm.Effect.Unify
 import Swarm.Effect.Unify.Common
 import Swarm.Language.Types hiding (Type)
@@ -253,7 +254,9 @@ unify ty1 ty2 = do
         (UTyRec x ty, _) -> unify (unfoldRec x ty) ty2
         (_, UTyRec x ty) -> unify ty1 (unfoldRec x ty)
         (UTyUser x1 tys, _) -> do
-          ty1' <- expandTydef x1 tys
+          ty1' <- withThrow
+            (\(UnexpandedUserType _) -> UndefinedUserType (UTyUser x1 tys))
+            (expandTydef x1 tys)
           unify ty1' ty2
         (_, UTyUser {}) -> unify ty2 ty1
         (Free t1, Free t2) -> Free <$> unifyF t1 t2
--- a/src/swarm-lang/Swarm/Language/Kindcheck.hs
+++ b/src/swarm-lang/Swarm/Language/Kindcheck.hs
@@ -20,6 +20,7 @@ import Prettyprinter (hsep, nest, pretty, vsep, (<+>))
 import Swarm.Language.Types
 import Swarm.Pretty (PrettyPrec (..), ppr)
 import Swarm.Util (number)
+import Swarm.Util.Effect (withThrow)
 
 -- | Kind checking errors that can occur.
 data KindError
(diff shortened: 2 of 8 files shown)


xsebek:
Doesn't this go against the direction of https://github.com/swarm-game/swarm/issues/2343? 🤔 
I worry that once we expand the type synonyms we won't be able to get back.

I thought we had an issue to create type context for LSP/error messages/something but I can't find it right now.

byorgey:
> Doesn't this go against the direction of #2343? 🤔

I guess substituting away user-defined types does make it a little harder to maintain a good correspondence between types and source locations; but that's already true: we are already substituting definitions of user-defined types, the question is just whether we do it lazily or eagerly.  Other than that, I see this and #2343 as orthogonal.  Unless I'm missing something specific you have in mind, I don't see how this makes #2343 any harder.

> I worry that once we expand the type synonyms we won't be able to get back.

Well, that's a good worry, because we definitely can't.  And that is definitely the downside of this PR.  For example if you have
```
tydef T = [x: Int, y: Int, longFieldName: Int -> (rec l. Unit + Int * l)] end;
def f : T -> T = \t. ... end
```
then the type of `f` is sadly going to be displayed as `[x: Int, y: Int, longFieldName: Int -> (rec l. Unit + Int * l)]  ->  [x: Int, y: Int, longFieldName: Int -> (rec l. Unit + Int * l)] `, and there's no easy way to display it as `T -> T`.

> I thought we had an issue to create type context for LSP/error messages/something but I can't find it right now.

That doesn't ring a bell, I'm not sure what you're referring to.  But if you remember/find it, let me know.

I have had a new thought on how we could solve these issues without giving up on the convenience of lazily expanded type definitions.  I will update #2437 with my idea.

byorgey:
Closing in favor of #2450.

PR #2446, opened 2025-05-25
Add Replay game script

xsebek:
* add Web API `repl/history/current`
* add ticks to REPL history data
* add script to convert JSON history to `.sw` with inserted `wait`


scripts/gen/replay.py
```diff
@@ -0,0 +1,61 @@
...
+    elif isinstance(input, str):  # JSON string
+        return json.loads(input)
+    else:
+        raise Exception(f"Can not parse '{type(input)}'")
+
+
+if __name__ == "__main__":
+    import argparse
+
+    parser = argparse.ArgumentParser(description="Process command JSON with wait times.")
```
xsebek, line 47: I have started to work on a REPL mode that will replay the input at proper tick, but it's going to take time and this might be useful to get some save-like functionality until then.

byorgey (approved):

PR #2448, opened 2025-05-25
Fix two space leaks

patrickestrada:
This PR adds strictness in the `insert` function from the `Swarm.Util.WindowedCounter `module to fix a space leak caused by the laziness of the `Just` constructor. This PR also adds strictness to the `_commandsHistogram` field in the `ActivityCounts` record in `Swarm.Game.Robot.Activity` by using `Data.Map.Strict` instead of `Data.Map`. Doing so fixes another space leak.

### DIFF
--- a/src/swarm-engine/Swarm/Game/Robot/Activity.hs
+++ b/src/swarm-engine/Swarm/Game/Robot/Activity.hs
@@ -16,7 +16,7 @@ module Swarm.Game.Robot.Activity (
 
 import Control.Lens hiding (Const, contains)
 import Data.Aeson qualified as Ae (FromJSON (..), ToJSON (..))
-import Data.Map (Map)
+import Data.Map.Strict (Map)
 import GHC.Generics (Generic)
 import Swarm.Game.Tick
 import Swarm.Language.Syntax (Const)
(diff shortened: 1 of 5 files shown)


xsebek (approved):
Thanks for fixing those leaks. 👍 

Out of curiosity, did you find them by browsing code or was this an issue when playing the game?

patrickestrada:
I found them when I was trying to see if I could reduce the performance degradation caused by using Brick.Animation to implement popups. I built the program using ghc 9.12.2 with -O1 and the following ghc-options: -rtsopts -finfo-table-map -fdistinct-constructor-tables. Then I ran the program with +RTS -hi -l -i0.5, ran eventlog2html on the resulting eventlog, and opened the resulting html file. 

Looking at the details tab I noticed some big thunks being built up. I switched back to the main branch to see if it was my changes causing the problem and noticed that the issue was still there. Tracking down the thunks eventually led me to the space leaks.

As an aside, do you know why the restyled check is failing?

xsebek:
It seems restyled did not handle your fork properly and failed to pull it with git.

This used to work even with forks, but since most of the PRs happen inside swarm repo, it's possible it broke and we did not notice.

If you did not, could you please run fourmolu locally (see scripts) and push? This would run the check again and in the worst case we would merge it manually, then deal with it separately.

byorgey:
@patrickestrada, congrats on having your first PR accepted to Swarm!   We appreciate the contribution and we're really glad to welcome you as part of the [community](https://github.com/swarm-game/swarm/blob/main/COMMUNITY.md).  In a minute we will send you an invite to the repo (see https://github.com/swarm-game/swarm/blob/main/CONTRIBUTING.md#i-have-push-access-to-the-swarm-repository-now-what ).  If you use Discord, feel free to also join the Swarm server at https://discord.gg/kp8MuSgkPw .

PR #2449, opened 2025-05-25
Update restyled.yml for forks

xsebek:
This should allow #2448 to pass.

See https://github.com/restyled-io/actions?tab=readme-ov-file#handling-forks.


byorgey (approved):
Nice!  I knew restyled was broken for forks but I hadn't realized we could fix it via our config.

PR #2450, opened 2025-05-27
Add version numbers to variables, to track shadowing for user types

### LINKED ISSUE #2437: Redefining user types can break type soundness

byorgey:
Closes #2437.

- Create a new type `TDVar` specifically to represent names of type definitions.  In addition to a name, `TDVar` also contains a version number, which is incremented every time the same name is redefined.  This means that redefining a type name just *shadows* the old one, and any types which refer to the old one continue to do so, because they refer to a variable with a different version number.
- Generalize `Ctx` over the variable type, so we can use both `Ctx Var` and `Ctx TDVar`.
- Remove `Monoid TDCtx` instance.  It was only being used when looking up a `Maybe TDCtx` via a traversal-used-as-a-lens, and with the version numbering stuff I'm not even sure what the combining operation should do any more.
- Replace kind checking with `process{Poly}Type` functions that do *both* name resolution and kind checking.
- Add a bunch of comments.

The pretty-printing is not perfect since you can't tell the difference between the different versions of a shadowed tydef name.  However, I've left that to #2452.


src/swarm-lang/Swarm/Language/Types.hs
```diff
@@ -816,8 +824,50 @@ data TydefInfo = TydefInfo
 makeLenses ''TydefInfo
 
 -- | A @TDCtx@ is a mapping from user-defined type constructor names
---   to their definitions and arities/kinds.
-type TDCtx = Ctx TydefInfo
+--   to their definitions and arities/kinds.  It also stores the
+--   latest version of each name (for any names with more than one
+--   version), so we can tell when a type definition has been
+--   shadowed.
+data TDCtx = TDCtx {getTDCtx :: Ctx TydefInfo, getTDVersions :: Map Text Int}
```
xsebek, line 831: Monoid Map for Sum might be nice to eliminate the fromMaybe.

src/swarm-lang/Swarm/Language/Var.hs
```diff
@@ -0,0 +1,98 @@
...
+-- | Pretty-print a variable, given an extra argument representing the
+--   latest version of any variable with this name.  If this variable
+--   is the latest version, just print its name.  If it is not the
+--   latest version, print @name%version@.
+prettyVar :: Int -> Var -> Doc ann
+prettyVar latest (Var x n)
+  | latest > n = pretty (x <> "%" <> showT n)
+  | otherwise = pretty x
+
+-- TODO(#2452): overhaul pretty-printing to take extra
```
xsebek, line 86: So Pretty (WithE Ctx) for syntax?

src/swarm-lang/Swarm/Language/Types.hs
byorgey: Hmm, that sounds like a nice idea in theory, but I just tried it and ran into an issue with missing `Hashable` and `ToJSON` instances for `MonoidMap`.  I suppose we could add some orphan instances that work via conversion to `Map`...

byorgey:
> If not, wouldn't it be better to have separate Var and TyVar (data/new)types? 🤔

Yes, maybe you're right!

src/swarm-lang/Swarm/Language/Types.hs
xsebek: It's not an orphan if you declare it for our TDCtx type :)

byorgey: Good point!

src/swarm-lang/Swarm/Language/Context.hs
```diff
@@ -325,21 +320,24 @@ buildCtxMap m (CtxTree h s) = do
       CtxDelete x t s1 -> buildCtxMap (M.insert x t m) s1
       CtxUnion s1 s2 -> buildCtxMap m s1 *> buildCtxMap m s2
 
+newtype ConstHash v t = ConstHash CtxHash
+  deriving (Eq, Show, Generic, ToJSON, FromJSON)
```
xsebek, line 324: So this is a double const at the type level?

xsebek (approved):
Thanks for the detailed comments. 👍 

I hope this works well, it would be neat if we could do something similar with normal variables and make the serialization possible and pretty printing nicer. 🙂

byorgey:
> neat if we could do something similar with normal variables

Earlier you told me I should make `Var` and `TDVar` into different types.  Make up your mind. :smile: 

More seriously, I don't know of any reason to do something similar with normal variables.

src/swarm-lang/Swarm/Language/Context.hs
byorgey: Yes, it is.  At first I tried just using `Const (Const CtxHash)` but that didn't work for some reason.  I don't remember why.

xsebek:
@mergifyio requeue

byorgey:
@mergifyio requeue

PR #2455, opened 2025-05-30
Add stock command

### LINKED ISSUE #2206: Consider giving `require X` and `require n X` different names to avoid confusion
> I am not surprised that `require 0 "treads"` did not work.  `require X` and `require n X`, although they look similar, are quite distinct: the first requires a device to *be equipped*, and the second requires a certain number of copies of a device *in the inventory*.  (I realize this is potentially confusing; open to suggestions for making the distinction more clear / easier to understand.)

_Originally posted by @byorgey in https://github.com/swarm-game/swarm/issues/921#issuecomment-1356402374_

Currently, `require "boat"` means you require a boat to be equipped, and `require 5 "boat"` means you require 5 boats in your inventory.  Although superficially similar these are quite distinct commands, and in order to reduce confusion / help players remember what they mean, it might be worth renaming one or both of them.

oliverpauffley:
Add stock command to replace the `require n "entity"` syntax. Also
made some changes to change the language around "RequireDevice" since
it's now clearer that this doesn't clash with "RequireInventory".

fixes #2206

### DIFF
--- a/data/scenarios/Testing/201-require/00-ORDER.txt
+++ b/data/scenarios/Testing/201-require/00-ORDER.txt
@@ -1,8 +1,8 @@
 201-require-device.yaml
 201-require-device-creative.yaml
 201-require-device-creative1.yaml
-201-require-entities.yaml
-201-require-entities-def.yaml
+201-stock-entities.yaml
+201-stock-entities-def.yaml
 533-reprogram-simple.yaml
 533-reprogram.yaml
 1664-require-system-robot-children.yaml
--- a/data/scenarios/Tutorials/stock.yaml
+++ b/data/scenarios/Tutorials/stock.yaml
@@ -1,18 +1,18 @@
 version: 1
-name: Require inventory
+name: Stock
 description: |
-  Learn how to require inventory when building robots.
+  Learn how to stock inventory when building robots.
 objectives:
   - goal:
       - |
         In the previous tutorial challenge, you learned how to use `require`{=snippet}
         to require specific devices to be equipped.
-        Sometimes, instead of requiring equipped devices, you require supplies in your inventory.
+        Sometimes, instead of requiring equipped devices, you stock supplies in your inventory.
       - |
-        In this case, you can write `require <int> <name>`{=snippet} to require a certain number
+        In this case, you can write `stock <int> <name>`{=snippet} to stock a certain number
         of copies of a certain entity to be placed in your inventory. For example:
         ```
-        build {require 2 "flower"; move; place "flower"; move; place "flower"}
+        build {stock 2 "flower"; move; place "flower"; move; place "flower"}
         ```
         This would build a robot with two `flower`{=entity}s in its inventory which the robot can then place.
       - Your goal in this challenge is to cover the entire 4x4 gray area with `rock`{=entity}s!
@@ -38,7 +38,7 @@ solution: |
   def mp = move; place "rock" end;
   def rr = turn right; move; turn right; x4 mp; turn back; x4 move end;
   build {
-    require 16 "rock";
+    stock 16 "rock";
     x4 rr
   };
 robots:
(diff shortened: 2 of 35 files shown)


data/scenarios/Testing/201-require/00-ORDER.txt
```diff
@@ -1,8 +1,8 @@
 201-require-device.yaml
 201-require-device-creative.yaml
 201-require-device-creative1.yaml
-201-require-entities.yaml
-201-require-entities-def.yaml
+201-stock-entities.yaml
+201-stock-entities-def.yaml
```
oliverpauffley, line 5: It's possible that this should now be moved out from this set of tests/scenarios?

data/scenarios/Tutorials/stock.yaml
```diff
@@ -1,18 +1,18 @@
...
 description: |
-  Learn how to require inventory when building robots.
+  Learn how to stock inventory when building robots.
 objectives:
   - goal:
       - |
         In the previous tutorial challenge, you learned how to use `require`{=snippet}
         to require specific devices to be equipped.
-        Sometimes, instead of requiring equipped devices, you require supplies in your inventory.
+        Sometimes, instead of requiring equipped devices, you stock supplies in your inventory.
```
oliverpauffley, line 10: Happy to try and reword this if needed. It used to follow nicely that you just learnt about `requiring` devices and now you are `requiring` inventory but it now doesn't follow as well.

data/scenarios/Tutorials/stock.yaml
```diff
@@ -63,10 +63,10 @@ robots:
...
 world:
   palette:
-    'Ω': [grass, null, base]
-    '.': [grass]
-    '_': [stone]
-    '+': [stone, wall]
+    "Ω": [grass, null, base]
+    ".": [grass]
+    "_": [stone]
+    "+": [stone, wall]
```
oliverpauffley, line 69: My editor does this automatically so it's a little annoying!

Yaml spec says single or double quotes is fine there is just some differences to how quoting is handled. 

Happy to change back if this is too unrelated.

data/scenarios/Tutorials/stock.yaml
byorgey: The connection still makes sense since both are about providing entities to a built robot.  How about something like "Sometimes, instead of requiring entities to be equipped as devices, you need them to be stocked in the inventory"?

src/swarm-engine/Swarm/Game/Step/Const.hs
```diff
@@ -16,26 +16,70 @@ import Control.Carrier.State.Lazy
...
+  parts,
+  use,
+  uses,
+  view,
+  (%=),
+  (+=),
+  (.=),
+  (<+=),
+  (<>=),
+ )
```
byorgey, line 32: I assume splitting out all the import lists onto separate lines like this is something else your editor did automatically?  I do find it pretty annoying, it uses up a ton of vertical space for no good reason, especially when the list of imports is short.  If there's an easy way for you to revert these reformatted imports, that would be nice --- but ultimately it's not a big deal, don't worry about it if it would be a pain.

byorgey (approved):
Looks great!

src/swarm-engine/Swarm/Game/Step/Const.hs
oliverpauffley: I can revert no worries. I was hoping the auto format would fix it up.

byorgey: Feel free to add the `merge me` label once you're happy with it to kick off the merge bot.  Though I think the Windows build may be broken at the moment (https://github.com/chhackett/vty-windows/issues/19).

byorgey:
@all-contributors please add @oliverpauffley for code

PR #2460, opened 2025-06-01
Bump Fourmolu formatter to 0.18.0.0

### LINKED ISSUE #2459: Update Fourmolu formatter to version 0.18
We should update our restyled configuration to use latest Fourmolu 0.18.

Currently we use 0.13 from 2023-06-09.

This would make it easier to setup for newcomers - just download the latest Fourmolu.

It also formats better and allows latest GHC extensions, so we will inevitably have to migrate at some point.

#### Notes

If this produces a larger diff, we could add a `.git-blame-ignore-revs` file.

xsebek:
* closes #2459

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,0 +1,27 @@
+# git blame ignore list.
+#
+# This file contains a list of git hashes of revisions to be ignored by git blame.
+# These revisions are considered "unimportant" in that they are unlikely to be what
+# you are interested in when blaming.
+#
+# Instructions:
+# - Only large (generally automated) reformatting or renaming CLs should be
+#   added to this list. Do not put things here just because you feel they are
+#   trivial or unimportant. If in doubt, do not put it on this list.
+# - Precede each revision with a comment containing the first line of its log.
+#   For bulk work over many commits, place all commits in a block with a single
+#   comment at the top describing the work done in those commits.
+# - Only put full 40-character hashes on this list (not short hashes or any
+#   other revision reference).
+# - Append to the bottom of the file (revisions should be in chronological order
+#   from oldest to newest).
+# - Because you must use a hash, you need to append to this list in a follow-up
+#   CL to the actual reformatting CL that you are trying to ignore.
+#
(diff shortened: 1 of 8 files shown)


byorgey (approved):

.github/restyled.yaml
```diff
@@ -1,6 +1,6 @@
 restylers_version: stable
 restylers:
   - fourmolu:
-      image: 'restyled/restyler-fourmolu:v0.13.0.0'
+      image: 'restyled-io/restyler-fourmolu:v0.18.0.0'
```
xsebek: Not sure why the CI does not pick up the new version. 🤔 
```
Error response from daemon: manifest for restyled/restyler-fourmolu:v0.18.0.0 not found: manifest unknown: manifest unknown
 WARN: Retrying docker-pull
       {
         "attempt": 1,
         "exitCode": 1,
         "limit": 5
       }
```

Based on the restyled repo docs, it was updated in January:
https://github.com/restyled-io/restylers/blob/main/_docs/restylers.md#fourmolu
and there is an image with this name and version:
https://gallery.ecr.aws/restyled-io/restyler-fourmolu

.git-blame-ignore-revs
```diff
@@ -0,0 +1,27 @@
...
+# - Because you must use a hash, you need to append to this list in a follow-up
+#   CL to the actual reformatting CL that you are trying to ignore.
+#
+# This note was taken from the chromium project file and should scare off
+# anyone that would like to put their manual "refactoring" commit here.
+# This magic should be used sparingly.
+# https://chromium.googlesource.com/chromium/src/+/HEAD/.git-blame-ignore-revs
+
+# Bump Fourmolu formatter to 0.18.0.0
+7c1247f80e2de34bbea52eb07ecb12954ba4d7a5
```
xsebek, line 27: This will need to be merged with rebase (actually fast forward) otherwise the hash will not match.

byorgey: I think you can do that manually.  Let me know if you need additional admin privileges.

PR #2461, opened 2025-06-01
Explicit term-level type argument to `read` command

### LINKED ISSUE #2223: Handle `read` at user-defined types

byorgey:
- Add `@` syntax for writing types as terms.
- Add a type of types, `Type`.
- Give the `read` command the type `Type -> Text -> a`, and a special typechecking rule such that the type argument (which must be a type literal) becomes the output type.  In other words it really should have a dependent type like `(tau : Type) -> Text -> tau` but I'm not about to add actual dependent types to Swarm. :smile:
- Eagerly expand any user-defined types in an argument to `read`.  Fixes #2223.
- Add some tests.

### DIFF
--- a/data/scenarios/Challenges/_flower-count/judge.sw
+++ b/data/scenarios/Challenges/_flower-count/judge.sw
@@ -19,7 +19,7 @@ def judgeCount : Int -> Cmd Unit = \actual.
     (\_. pure ())
     (\p.
       try {
-        let c = (read p : Int) in
+        let c = read @Int p in
         if (c == actual) { win } {}
       } {}
     )
--- a/src/swarm-lang/Swarm/Language/Parser/Term.hs
+++ b/src/swarm-lang/Swarm/Language/Parser/Term.hs
@@ -107,6 +107,7 @@ parseTermAtom2 =
           <*> pure Nothing
           <*> (optional (symbol ";") *> (parseTerm <|> (eof $> sNoop)))
         <|> SRcd <$> brackets (parseRecord (optional (symbol "=" *> parseTerm)))
+        <|> TType <$> (symbol "@" *> parseTypeAtom)
     )
     <|> parseLoc (mkTuple <$> parens (parseTerm `sepBy` symbol ","))
     <|> parseLoc (TDelay (TConst Noop) <$ try (symbol "{" *> symbol "}"))
--- a/test/unit/TestLanguagePipeline.hs
+++ b/test/unit/TestLanguagePipeline.hs
@@ -755,6 +755,12 @@ testLanguagePipeline =
                 "(read @Int \"3\") 4"
                 "1:1: Type mismatch:\n  From context, expected `(read @Int \"3\")` to be a function,\n  but it actually has type `Int`"
             )
+        , testCase
+            "read at type stored in a variable"
+            ( process
+                "def T : Type = @(Bool * Int) end; read T \"(True, 3)\""
+                "1:35: The `read` command must be given a literal type as its first argument (Swarm does not have dependent types); found `T` instead."
+            )
         ]
     ]
  where
--- a/test/unit/TestLanguagePipeline.hs
+++ b/test/unit/TestLanguagePipeline.hs
@@ -747,6 +747,15 @@ testLanguagePipeline =
                 "3:15:\n  |\n3 |  def z = 3 end\n  |               ^\nunexpected end of input\nexpecting \"!=\", \"&&\", \"()\", \"++\", \"<=\", \"==\", \">=\", \"def\", \"false\", \"let\", \"require\", \"requirements\", \"stock\", \"true\", \"tydef\", \"||\", '\"', '$', '(', '*', '+', '-', '.', '/', ':', ';', '<', '>', '@', '[', '\\', '^', 'end' keyword for definition of 'x', '{', built-in user function, direction constant, integer literal, or variable name\n"
             )
         ]
+    , testGroup
+        "Typechecking for read #2461"
+        [ testCase
+            "read type argument propagates"
+            ( process
+                "(read @Int \"3\") 4"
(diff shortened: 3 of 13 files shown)


test/unit/TestLanguagePipeline.hs
```diff
@@ -744,7 +744,16 @@ testLanguagePipeline =
...
-                "3:15:\n  |\n3 |  def z = 3 end\n  |               ^\nunexpected end of input\nexpecting \"!=\", \"&&\", \"()\", \"++\", \"<=\", \"==\", \">=\", \"def\", \"false\", \"let\", \"require\", \"requirements\", \"stock\", \"true\", \"tydef\", \"||\", '\"', '$', '(', '*', '+', '-', '.', '/', ':', ';', '<', '>', '[', '\\', '^', 'end' keyword for definition of 'x', '{', built-in user function, direction constant, integer literal, or variable name\n"
+                "3:15:\n  |\n3 |  def z = 3 end\n  |               ^\nunexpected end of input\nexpecting \"!=\", \"&&\", \"()\", \"++\", \"<=\", \"==\", \">=\", \"def\", \"false\", \"let\", \"require\", \"requirements\", \"stock\", \"true\", \"tydef\", \"||\", '\"', '$', '(', '*', '+', '-', '.', '/', ':', ';', '<', '>', '@', '[', '\\', '^', 'end' keyword for definition of 'x', '{', built-in user function, direction constant, integer literal, or variable name\n"
+            )
+        ]
+    , testGroup
+        "Typechecking for read #2461"
+        [ testCase
+            "read type argument propagates"
+            ( process
+                "(read @Int \"3\") 4"
```
xsebek, line 752: Could we have one test that shows what happens when you store a `Type` in a variable and use that as first argument?

data/scenarios/Challenges/_flower-count/judge.sw
```diff
@@ -19,7 +19,7 @@ def judgeCount : Int -> Cmd Unit = \actual.
     (\_. pure ())
     (\p.
       try {
-        let c = (read p : Int) in
+        let c = read @Int p in
         if (c == actual) { win } {}
```
xsebek, line 23: I suppose we can't auto convert it, because the type might not have been specified? 🤔

src/swarm-lang/Swarm/Language/Parser/Term.hs
```diff
@@ -107,6 +107,7 @@ parseTermAtom2 =
           <*> pure Nothing
           <*> (optional (symbol ";") *> (parseTerm <|> (eof $> sNoop)))
         <|> SRcd <$> brackets (parseRecord (optional (symbol "=" *> parseTerm)))
+        <|> TType <$> (symbol "@" *> parseTypeAtom)
```
xsebek, line 110: 🤩

xsebek (approved):

data/scenarios/Challenges/_flower-count/judge.sw
byorgey: I mean, technically, we could go through and look for syntax specifically of the form `read <expr> : <type>` and convert it to `read <type> <expr>`.  But you're right that might not catch all uses, since the type might have just been inferred.  However, the bigger reason I don't think it's worth trying to do auto-conversion is that `read` itself was added since the 0.6 release.  So someone upgrading directly from 0.6 to 0.7 would have no existing uses of `read` to convert.

src/swarm-lang/Swarm/Language/Parser/Term.hs
byorgey: Yeah, doing it this more general way turned out to be way nicer than making `read` a special syntax form.  It actually wasn't even that hard since we already had types-as-values for the implementation of `read`.

test/unit/TestLanguagePipeline.hs
byorgey: Oh, drat.  That's a really good suggestion, and it reveals a problem.  The typechecking for `read` only works when it has a literal type argument.  Without full-blown dependent types, we can't typecheck uses of `read` where its output type depends on a *value* passed in at runtime.  With this PR as it stands, you *can* write
```
def T : Type = @Int end; read T "3"
```
and it works --- except that the type inferred for `read T "3"` is `forall a. a`, because the special case in the typechecker for `read` applied to a literal type argument is not triggered.  Of course you can use this to crash the CESK machine, e.g.
```
def T : Type = @Int end; (read T "3") 4
```
I suppose the solution is to add a check in the typechecker to *ensure* that the first argument to `read` is a literal type and not any other kind of expression (with a suitably apologetic error message).

byorgey: Fixed by making it an error to give anything other than a type literal as the first argument to `read`.

So technically, right now, you can define variables storing types, like `def mytype : Type = @(Int * Bool) end`, but there's nothing you can do with them!

PR #2470, opened 2025-06-05
Convert `require n` -> `stock n` with `format --v0.6`

byorgey:
Update the parser so it accepts `require n` instead of `stock` when parsing in 0.6 mode, so `format --v0.6` can be used to convert.

### DIFF
--- a/src/swarm-lang/Swarm/Language/Parser/Term.hs
+++ b/src/swarm-lang/Swarm/Language/Parser/Term.hs
@@ -75,13 +75,8 @@ parseTermAtom2 =
         <|> TInt <$> integer
         <|> TText <$> textLiteral
         <|> TBool <$> ((True <$ reserved "true") <|> (False <$ reserved "false"))
-        <|> reserved "require"
-          *> ( TRequire
-                 <$> (textLiteral <?> "device name in double quotes")
-             )
-        <|> reserved "stock"
-          *> (TStock . fromIntegral <$> integer)
-          <*> (textLiteral <?> "entity name in double quotes")
+        <|> reserved "require" *> parseRequire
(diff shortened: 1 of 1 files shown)


oliverpauffley (approved):
Ah this is so you can parse both `require` and `stock` depending on the version?

Do you ever get rid of the old version?

And question just to check I understand. The `view languageVersion` is using the `ReaderT` monad to get the current version?

byorgey:
> Ah this is so you can parse both `require` and `stock` depending on the version?

Right.  Specifically if someone runs `swarm format --v0.6` then it parses code from Swarm 0.6, and pretty-prints it in 0.7 syntax.
 
> Do you ever get rid of the old version?

Yes, for now we only keep one old version around at a time.  Version 0.6 had a flag called --v0.5 which could read and convert code from version 0.5 to version 0.6, but we deleted that code a while ago.

> Ok I was nearly correct the `view` is using the lens to access the ReaderT values?

Yes, `Parser` is defined as `ReaderT ParserConfig (...)`: https://github.com/swarm-game/swarm/blob/5827eb2f17cd28aea91143c2b1031194d8ae73e3/src/swarm-lang/Swarm/Language/Parser/Core.hs#L103 .  The `view` is accessing the `ParserConfig` in the `ReaderT` and projecting out the `languageVersion` field.  Actually just now I had to go look up the type of `view` to remember how this works.  Normally you would think of `view` has having a type like `Getter a b -> a -> b`, but it turns out that it has a more general type like `MonadReader m a => Getter a b -> m b`.  Since `((->) a` is an instance of `MonadReader` it can still be used at type `Getter a b -> a -> b`, but it can also be used with e.g. `ReaderT`.

PR #2471, opened 2025-06-06
Remove data step from release

xsebek:
There is another step creating the ZIP file afterwards, so I think this additional step was a mistake.

See #2046.


byorgey (approved):
:shrug:

PR #2473, opened 2025-06-08
refactor narrowToPosition to use pathToPosition

oliverpauffley:
When getting the lsp response for hover we need to narrow the syntax tree to by a given position. This change refactors that function to do so by building a list of syntax terms and then selecting the last. This isn't useful for Hover but will be used when we need to jump to position within the tree.


src/swarm-lang/Swarm/Language/LSP/Hover.hs
```diff
@@ -88,60 +88,86 @@ posToRange myRope foundSloc = do
...
+    TAntiSyn {} -> mempty
+    SSuspend {} -> mempty
+
+  bindPath :: Maybe LocVar -> Syntax' ty -> Syntax' ty -> [Syntax' ty]
+  bindPath mlv s1@(Syntax' _ _ _ lty) s2 = mlvToSyntax mlv <|> d s1 <|> d s2
+   where
+    mlvToSyntax :: Maybe LocVar -> [Syntax' ty]
+    mlvToSyntax = \case
+      Just lv -> d $ locVarToSyntax' lv (getInnerType lty)
+      Nothing -> []
```
oliverpauffley: This was a single line before since we were dealing only with the Maybe monad. Now we have Maybe and List I couldn't figure out a one liner. Maybe this is clearer anyway but happy to change if you have something.

src/swarm-lang/Swarm/Language/LSP/Hover.hs
```diff
@@ -88,60 +88,86 @@ posToRange myRope foundSloc = do
...
-  TRobot {} -> Nothing
-  TAntiInt {} -> Nothing
-  TAntiText {} -> Nothing
-  TAntiSyn {} -> Nothing
-  SSuspend {} -> Nothing
+  NonEmpty (Syntax' ty)
+pathToPosition s0 pos = s0 :| innerPath s0
  where
+  innerPath :: Syntax' ty -> [Syntax' ty]
+  innerPath (Syntax' _ t _ ty) = case t of
```
byorgey, line 119: I don't think this is doing the same thing anymore.  The `<|>` operators used to choose the first valid result, using `Maybe`.  So e.g. something like `d (locVarToSyntax' lv lty) <|> d s1 <|> d s2` meant "first try `locVarToSyntax' lv lty`, then try descending into `s1`, then `s2`, and pick the first that gives us a valid result".  Now the `<|>` operator is working over the list of results, so it means to concatenate.  But we do not want to concatenate all the results from all the subterms.

Now, in theory, only one of the subterms will actually contain the given position, and the other ones will return the empty list.  But I still don't think this works correctly: `innerPath` will always end up returning just a singleton list with the deepest syntax node, rather than returning a sequence of syntax nodes representing the path down through the tree.

Stepping back a bit, I think the issue is that simply replacing `Maybe` (which represented success or failure) with `[]` means we are thinking of the list as a "list of successes"; but that's not actually what we want here.  I think `innerPath` should actually have a type like `Syntax' ty -> Maybe [Syntax' ty]` --- that is, it could succeed or fail to return a path of syntax nodes.  Then the `<|>` operators will be working over the `Maybe` which is exactly what we want.

src/swarm-lang/Swarm/Language/LSP/Hover.hs
```diff
@@ -88,60 +88,86 @@ posToRange myRope foundSloc = do
...
+  descend ::
+    ExplainableType ty =>
+    -- \| position
+    Int ->
+    -- \| next element to inspect
+    Syntax' ty ->
+    [Syntax' ty]
+  descend p s1@(Syntax' l1 _ _ _) = do
+    guard $ withinBound p l1
+    return $ narrowToPosition s1 p
```
byorgey: I think `descend` must now call `innerPath` rather than `narrowToPosition`, and should have a type like `Syntax' ty -> Maybe [Syntax' ty]`, just like `innerPath`.  This is probably the place to add the root node onto the beginning of the path, so something like
```
  guard $ withinBound p l1
  return $ (s1 :) <$> innerPath s1 p
```
so that the `s1` node gets consed onto the front of the path returned by `pathToPosition`.  That way, as `innerPath` keeps recursively calling `descend` on its way down the tree, it will build up a path of all the syntax nodes encountered along the way.

src/swarm-lang/Swarm/Language/LSP/Hover.hs
```diff
@@ -88,59 +89,79 @@ posToRange myRope foundSloc = do
...
-  TRobot {} -> Nothing
-  TAntiInt {} -> Nothing
-  TAntiText {} -> Nothing
-  TAntiSyn {} -> Nothing
-  SSuspend {} -> Nothing
+  NonEmpty (Syntax' ty)
+pathToPosition s0 pos =
+  case innerPath s0 of
+    Nothing -> s0 :| []
+    Just ss -> s0 :| ss
```
byorgey: This is probably better written `s0 :| fromMaybe [] (innerPath s0)`.

src/swarm-lang/Swarm/Language/LSP/Hover.hs
```diff
@@ -88,59 +89,79 @@ posToRange myRope foundSloc = do
...
+  descend ::
+    ExplainableType ty =>
+    -- \| position
+    Int ->
+    -- \| next element to inspect
+    Syntax' ty ->
+    Maybe [Syntax' ty]
+  descend p s1@(Syntax' l1 _ _ _) = do
+    guard $ withinBound p l1
+    liftA2 (:) (Just s1) (innerPath s1)
```
byorgey: I prefer `(s1 :) <$> innerPath s1` although it's a matter of taste.  Pick whichever you think is more readable.

src/swarm-lang/Swarm/Language/LSP/Hover.hs
```diff
@@ -88,59 +89,79 @@ posToRange myRope foundSloc = do
...
+    TAntiInt {} -> mempty
+    TAntiText {} -> mempty
+    TAntiSyn {} -> mempty
+    SSuspend {} -> mempty
+
   d = descend pos
+  -- try and decend into the syntax element if it is contained with position
+  descend ::
+    ExplainableType ty =>
+    -- \| position
```
byorgey: These pipes should not have backslashes in front of them.

byorgey (commented):
Looking much better!

src/swarm-lang/Swarm/Language/LSP/Hover.hs
```diff
@@ -88,59 +91,76 @@ posToRange myRope foundSloc = do
...
   d = descend pos
+  -- try and decend into the syntax element if it is contained with position
+  descend ::
+    ExplainableType ty =>
+    Int ->
+    Syntax' ty ->
+    Maybe [Syntax' ty]
+  descend p s1@(Syntax' l1 _ _ _) = do
+    guard $ withinBound p l1
+    (s1 :) <$> innerPath s1
```
xsebek:
```suggestion
  descend p s1@(Syntax' l1 _ _ _) = do
    guard $ withinBound p l1
    pure $ case innerPath s1 of
      Nothing -> [s1]
      Just iss -> s1 : iss
```

xsebek:
I changed the unit test function so that it does not spew the Syntax Show instance:
```Haskell
  assertEqualTerms (Syntax' _ expected _ _) (Syntax' _ actual _ _) = 
    unless (actual == expected) $ assertFailure
      ("expected: " ++ prettyString expected
      ++ "\n but got: " ++ prettyString actual
      )
```

src/swarm-lang/Swarm/Language/LSP/Hover.hs
xsebek: I could be wrong, but I think that there used to be `Nothing` which would now mean `Just []`. But what you have with `mempty` is `Nothing :: Maybe [Syntax]` which means failure.

So there needs to be a base case which stops the recursion when you can't go into _inner_, but it should not let the failure bubble up to the top.

I choose to put it into `descend`, because its shorter to write, but you can write it differently if you want. I could also be missing some case, but at least it makes the tests pass.

byorgey (approved):
Looks good to me!

PR #2482, opened 2025-06-08
Add proper Swarm history replay

xsebek:
* adds `--replay history.json` CLI option
  * this stores the parsed JSON and sets a `Replaying` REPL mode
* add Web API runtime log
* clean up log entry JSON instances
* use custom keybindings when showing other REPL modes (Handling/Replaying)


scripts/gen/replay.sh
```diff
@@ -0,0 +1,55 @@
+#!/usr/bin/env bash
+
+PORT="5357"
+HOST="localhost"
+
+function help {
+  echo "Save Swarm REPL history for replay"
+  echo
+  echo "This is a shortcut for saving Swarm REPL via cURL as JSON:"
+  echo "curl --silent $HOST:$PORT/repl/history/current"
```
xsebek, line 10: I would like to add the in-game option to save history separately.

@byorgey could you please test this (using script to save) and see if the replay works well? 🙂

byorgey: I played the classic scenario for a little while, saved the history, and tried replaying it... but it used a different seed!  I think the seed needs to be saved as part of the history file, perhaps?

xsebek: Ah, I used it for challenges which don't have a seed or classic with `--seed 0`.

Adding it to the history would complicate things a bit.

The full replay format should probably include more information, like the scenario name and if it was creative.

byorgey: Ah, OK, so you see this as just a step along the path to a full replay format?  Makes sense, I can try it with `--seed`.

byorgey: Seems to work well!

xsebek: > Ah, OK, so you see this as just a step along the path to a full replay format?

Yes, this is a manual stage, replay would have a simple format and then full game save would also have the current state. I just have to add a save file dialog to the game, then its incremental steps. 😅

PR #2483, opened 2025-06-09
Split out `Texel` from `Display` 

### LINKED ISSUE #570: Refactor `Display` to make it simpler/more robust
There are several things about `Display` that seem a bit awkward.  It definitely feels like there is a better design hiding in there but we haven't been able to figure out what it is yet.

> I wonder if that is a good responsibility division - perhaps there is a better abstraction/wrapper for displayable things that have orientation, and their drawing depends on it. :thinking: 
I am hoping things would get simpler if robots (or later other entities) set the drawn character from their direction map when they are themselves changing their orientation, keeping the source of truth in the robot (or entity).
Caching seems like overengineering to me - I am worried it would just cause bugs.

_Originally posted by @xsebek in https://github.com/swarm-game/swarm/pull/556#discussion_r922486446_

### LINKED ISSUE #1662: Robots and entities should take on underlying cell background unless overridden

byorgey:
In this PR:

- Keep `Display` purely as a record of configuration options that specify how an entity should be drawn.
    - Remove fields (such as `boundaryOverride` and `curOrientation`) that were just caching more concrete rendering info.
    - Remove `Monoid` instance.
- Split out a new `Texel` type which represents the concrete way a single cell should be drawn.
    - `Texel`s are compositional: each `Texel` optionally has a character + foreground color, and optionally has a background color, each with a priority.  Combining `Texel`s combines the fg + bg separately according to the priorities.
    - This way, to decide how to draw a cell we can just render each thing (terrain, entity, robots) in that cell to a `Texel` and then combine all the `Texel`s.
    - This therefore closes #1662 , and closes #2264, and closes #1672.
    - I've also adjusted some of the colors so that (IMO) it looks a bit better/higher contrast, now that background colors show through behind entities
- Get rid of the `WorldAttr` type and just use `Attribute`
- A bit of reorganization, under the new `Swarm.Game.Cosmetic.*` namespace:
    - Rename `Swarm.Game.Display` to `Swarm.Game.Cosmetic.Display`
    - Create `Swarm.Game.Cosmetic.Texel`
    - Move `Swarm.Game.Entity.Cosmetic` to `Swarm.Game.Cosmetic.Color` (it was not really specific to entities)
    - Move `Swarm.Game.Entity.Cosmetic.Assignment` to `Swarm.Game.Cosmetic.Assignment`
    - Split out a new `Swarm.Game.Cosmetic.Attribute` module.
- Consolidate a whole bunch of code into the `renderDisplay` function, which turns a `Display` into a `Texel`
- Split out new module `Swarm.TUI.View.Static`

Closes #570.


src/swarm-scenario/Swarm/Game/Cosmetic/Attribute.hs
```diff
@@ -0,0 +1,35 @@
...
+-- 'Swarm.TUI.View.Attribute.Attr'.)
+module Swarm.Game.Cosmetic.Attribute where
+
+import Data.Hashable (Hashable)
+import Data.Text (Text)
+import Data.Yaml
+import GHC.Generics (Generic)
+
+-- | An internal attribute name.
+data Attribute = ABlank | ARobot | AEntity | AWorld Text
```
xsebek, line 17: Nice, I like how this is split off. 👍

src/swarm-scenario/Swarm/Game/Cosmetic/Display.hs
```diff
@@ -10,105 +10,67 @@
 -- Description: TUI rendering of entities
 --
 -- Utilities for describing how to display in-game entities in the TUI.
-module Swarm.Game.Display (
+module Swarm.Game.Cosmetic.Display (
```
xsebek, line 13: Could you please add a `Cosmetic.hs` module? It would probably just re-export all inner modules, unless there are some functions that you want to keep inside Cosmetic.

src/swarm-tui/Swarm/TUI/View.hs
```diff
@@ -534,36 +540,52 @@ drawGameUI s =
...
+  -- We want "with" before an entity (if any), "on" before the
+  -- terrain, and "at" at the end (before the coordinates)... but any
+  -- given preposition should only be used if there is something
+  -- before it.  Since the terrain will always exist, the
+  -- possibilities are as follows:
+
+  -- R with E on T at
+  -- R on T at
+  -- E on T at
+  -- T at
```
xsebek: Maybe we could write those possibilities out by hand? 🤔 Something like:
```Haskell
let nonEmptyTexel t = if texelIsEmpty t then Nothing else Just t
    mEntity = renderTexel <$> nonEmptyTexel (renderEntityCell worldEditor ri cCoords)
    mRobot = renderTexel <$> nonEmptyTexel (renderRobotCell aMap g cCoords)
    terrain = renderTexel $ renderTerrainCell worldEditor ri cCoords
in case (mEntity, mRobot) of
  (Just entity, Just robot) -> robot <> " with " <> entity <> " on " <> terrain
  (Just entity, Nothing) -> entity <> " on " <> terrain
  (Nothing, Just robot) -> robot <> " on " <> terrain
  (Nothing, Nothing) -> terrain
```

xsebek (approved):
This change looks great! 👍  I can't believe terrain had an orientation map. 😄

src/swarm-scenario/Swarm/Game/Cosmetic/Display.hs
byorgey: Good idea, done.

src/swarm-tui/Swarm/TUI/View.hs
byorgey: Yeah, good idea, done.  I was trying too hard to be clever.

PR #2487, opened 2025-06-10
Optimize structure composition

### LINKED ISSUE #1611: Optimize structure composition
Structure definitions, as implemented in #1332, are recursively assembled.

The current method is inefficient, in that a given sub-structure may be referenced as part of two different parent structures, and is itself re-assembled each time.

Instead, we should form a DAG of structure references and visit deepest first, caching in a map as we go.  Then, if a given sub-structure is referenced more than once, we don't have to re-assemble it.

patrickestrada:
This PR changes the structure assembly logic to build a graph of structure references, topologically sort them while checking for cycles, then assembles the structure, caching the already-merged structures in a HashMap so that we don't have to re-assemble structures.

Because two different structures in the tree of structure definitions may share the same name, we use the path of structure names from a named structure to the root as a unique identifier.

This PR resolves #1611


src/swarm-topography/Swarm/Game/Scenario/Topography/Structure/Assembly.hs
```diff
@@ -1,126 +1,249 @@
...
+
+cycleError :: PathToRoot -> [PathToRoot] -> Text
+cycleError path dfsPath = T.unwords ["Structure graph contains a cycle:", cycleText]
  where
-  Structure origArea subStructures subPlacements subWaypoints = baseStructure
+  cyc = (path :) . reverse . (path :) . takeWhile (/= path) $ dfsPath
+  cycleText = brackets . T.intercalate " -> " . fmap showPath $ cyc
 
-  originatedWaypoints = map (Originated parentPlacement) subWaypoints
+-- | Given a graph constructed via 'mkGraph', this function does a dfs on the graph to find
```
byorgey: We already have some code to do cycle detection in `Swarm.Util.Graph`, though it does not currently do a topological sort.  It would be nice if we could reuse that code.

Do we actually need the topological sort?  In Haskell one can use the very nice trick of constructing a recursively defined map which is lazy in the values, and the runtime system will implicitly figure out the right order in which to compute them (e.g. see https://byorgey.github.io/blog/posts/2023/04/11/competitive-programming-in-haskell-topsort-via-laziness.html and https://byorgey.github.io/blog/posts/2025/06/10/comprog-hs-intro.html).

patrickestrada: I don't think the topological sort is strictly necessary, but I figure we may as well do it as long as we have to do a dfs to check for cycles. If we cut the topological sort, I don't think we'd even need to construct a recursively defined map, as we could just cache structures in a Map or HashMap after we assemble them.

I'm using HashMap largely because I'm guessing that the Ord instance for `type PathToRoot = [StructureName]` is much slower on average than the Hashable instance. That's also why I'm reluctant to use the cycle detection from Swarm.Util.Graph, as the cycle detection from `Swarm.Util.Graph` is reliant on the Ord instance for the key.

byorgey: > I'm guessing that the Ord instance for `type PathToRoot = [StructureName]` is much slower on average than the Hashable instance. That's also why I'm reluctant to use the cycle detection from Swarm.Util.Graph, as the cycle detection from `Swarm.Util.Graph` is reliant on the Ord instance for the key.

I understand the impulse, but this sounds very much like premature optimization to me.  I don't know whether the `Ord` instance for `PathToRoot` is actually slower than the `Hashable` instance --- in fact, it could be the opposite, since the `Ord` instance only has to examine two lists far enough to find a place where they are different, whereas the `Hashable` instance always has to examine an entire list to compute a hash.  It would require benchmarking to know for sure.  But even if using `Hashable` were significantly faster, we're talking about a few dozen structures at most, so it really won't make much difference.  And even if someone were to make a scenario with hundreds of thousands of structures, the cost of composing them only has to be paid *once* while processing the scenario, so it's not likely to make much noticeable difference to players.  We should be optimizing for code simplicity, not squeezing out marginal performance gains.

patrickestrada: That's a fair point. Just changed the implementation.

src/swarm-topography/Swarm/Game/Scenario/Topography/Structure/Assembly.hs
```diff
@@ -1,126 +1,213 @@
...
+overlaySingleStructure (MergedStructure inputArea inputPlacements inputWaypoints) (MergedStructure overlayArea overlayPlacements overlayWaypoints, pose@(Pose loc orientation)) = MergedStructure mergedArea mergedPlacements mergedWaypoints
+ where
+  mergedWaypoints = inputWaypoints <> map (fmap $ placeOnArea overlayArea) overlayWaypoints
+  mergedPlacements = inputPlacements <> map (placeOnArea overlayArea) overlayPlacements
+  mergedArea = overlayGridExpanded inputArea pose overlayArea
+  placeOnArea (PositionedGrid _ overArea) =
+    offsetLoc (coerce loc)
+      . modifyLoc (reorientLandmark orientation $ getGridDimensions overArea)
+
+-- | Given the structure graph, ensure that the graph is not acyclic and that all placements are valid
```
kostmo: "not cyclic"

byorgey:
@patrickestrada is there still more you are planning to do with this PR?  Or is it ready for review + merge?  (I ask just because you still have it marked as a "Draft", not sure if that was intentional.)

src/swarm-topography/Swarm/Game/Scenario/Topography/Structure/Assembly.hs
```diff
@@ -1,126 +1,213 @@
 {-# LANGUAGE OverloadedStrings #-}
+{-# OPTIONS_GHC -Wno-unused-imports #-}
```
byorgey: Can we get rid of this option (and remove any unused imports)?

src/swarm-topography/Swarm/Game/Scenario/Topography/Structure/Assembly.hs
```diff
@@ -1,126 +1,213 @@
...
+{-# OPTIONS_GHC -Wno-unused-top-binds #-}
```
byorgey: What is the purpose of this option?

byorgey (approved):
Looks good to me!

src/swarm-topography/Swarm/Game/Scenario/Topography/Structure/Assembly.hs
patrickestrada: Corrected

src/swarm-topography/Swarm/Game/Scenario/Topography/Structure/Assembly.hs
patrickestrada: I got rid of the option.

src/swarm-topography/Swarm/Game/Scenario/Topography/Structure/Assembly.hs
patrickestrada: Got rid of the option.

PR #2488, opened 2025-06-15
Add Web API endpoint for runtime and robot logs

xsebek:
* Add Web API endpoint to stream runtime and robot logs
* Improve logs JSON instances
* Split off from #2482

You can test this with:
```bash
curl -N -s 'localhost:5357/runtime/log'
curl -N -s 'localhost:5357/robot/0/log'
```

### DIFF
--- a/src/swarm-doc/Swarm/Doc/Wiki/Cheatsheet.hs
+++ b/src/swarm-doc/Swarm/Doc/Wiki/Cheatsheet.hs
@@ -14,7 +14,6 @@ module Swarm.Doc.Wiki.Cheatsheet (
 
 import Control.Effect.Lift
 import Control.Lens (view, (^.))
-import Control.Lens.Combinators (to)
 import Data.Foldable (find, toList)
 import Data.List (transpose)
 import Data.List.Extra (enumerate)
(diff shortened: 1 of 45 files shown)


src/swarm-engine/Swarm/Log.hs
```diff
@@ -22,19 +23,34 @@ module Swarm.Log (
...
 import GHC.Generics (Generic)
+import Servant.Docs qualified as SD
 import Swarm.Game.Location (Location)
-import Swarm.Game.Tick (TickNumber)
+import Swarm.Game.Tick (TickNumber (..))
 import Swarm.Game.Universe (Cosmic)
+import Swarm.Util.JSON (optionsMinimize)
+
+enumJsonOptions :: Options
+enumJsonOptions =
```
byorgey: I think this already exists as `Swarm.Util.JSON.optionsUntagged`.

byorgey (approved):

src/swarm-engine/Swarm/Log.hs
xsebek: Good point, fixed. 👍

Issue #2193, opened 2024-10-20
Mechanism to render text within the landscape


kostmo:
See https://github.com/swarm-game/swarm/pull/2190#pullrequestreview-2380652838:
>  would be worth adding a built-in, more streamlined way to include text entities, in cases like this when you want to label things on a world map

and
> Yeah, something like this would be nice:
> ```YAML
> palette:
>   - char: 'X'  # {name: "X", display: {char: "X"}, description: ["X"], properties: [known]}
> ```

byorgey:
Here's an even more streamlined strawman proposal: if any character is used in a map but that character is not specified in the palette, then a default entity is generated with a display identical to the character that was used.

Well, I'm not sure I like that, since then you no longer get as much help catching mistakes where you accidentally used the wrong character in a map, or forgot to include something in the palette, etc.

The thing I'm trying to avoid is even having to write
```
- char: 'a'
- char: 'b'
- char: 'c'
```
and so on, it's super tedious to have to list all the characters you want to use.  What about allowing something like 
```
palette:
  - chars: "A-Za-z"
```
?

PR #2492, opened 2025-06-17
Simple photocopier challenge

byorgey:
This PR adds a simple photocopier challenge, where you have to scan for some invisible entities and reproduce a pattern.  It's similar to the existing `telephone` challenge but much stripped down, and may help give players an idea of how to find quartz mines.


data/scenarios/Challenges/photocopier.yaml
```diff
@@ -0,0 +1,98 @@
...
+    'J': [grass, null, judge]
+  map: |
+    ++++++++++++.....++++++++++++
+    +BXX.......+.....+..........+
+    +..XX......+.....+..........+
+    +...XX.....+.....+..........+
+    +..XXXX....+.....+..........+
+    +.XX..XXX.X+.....+..........+
+    +XX.....XX.+.....+J.........+
+    ++++++++++++.....++++++++++++
```
xsebek: Could this please have sizes that are a power of two?

Or at least add some visual cue. For example using letter entities:
```
       SOURCE            COPY
    ++++++++++++.....++++++++++++
```
This would also help those that confuse left and right to tell where they should scan and where to place. 😅

data/scenarios/Challenges/photocopier.yaml
```diff
@@ -0,0 +1,98 @@
...
+author: Brent Yorgey
+description: |
+  Copy an invisible pattern
+creative: false
+objectives:
+  - id: photocopy
+    goal:
+      - |
+        Inside the left-hand rectangular frame is a pattern drawn with
+        invisible pixels.  Your job is to copy the invisible pattern
```
xsebek:
```suggestion
        `invisible pixel`{=entity}s.  Your job is to copy the invisible pattern
```

data/scenarios/Challenges/photocopier.yaml
byorgey: Oh, good idea.  I might just do #2193 first.

byorgey: @xsebek how does it look now?

PR #2494, opened 2025-06-18
Optional `chars` palette field to specify auto-generated character entities

### LINKED ISSUE #2193: Mechanism to render text within the landscape

byorgey:
Adds an optional `chars` field to the `palette` specification of a world, which expects a string like `A-Za-f0-9!?` specifying a set of characters that can be used to stand for auto-generated default entities that just look like themselves.  Can be used to write text or create strictly cosmetic designs in a world.

Closes #2193 .  In the immediate future, I plan to use it for https://github.com/swarm-game/swarm/pull/2492#discussion_r2153283183 .

### DIFF
--- a/src/swarm-topography/Swarm/Game/Scenario/Topography/Structure.hs
+++ b/src/swarm-topography/Swarm/Game/Scenario/Topography/Structure.hs
@@ -21,7 +21,7 @@ import Swarm.Game.Location
 import Swarm.Game.Scenario.Topography.Grid
 import Swarm.Game.Scenario.Topography.Navigation.Waypoint
 import Swarm.Game.Scenario.Topography.Placement
-import Swarm.Game.Scenario.Topography.ProtoCell
+import Swarm.Game.Scenario.Topography.Palette
 import Swarm.Game.Scenario.Topography.Structure.Named
 import Swarm.Game.Scenario.Topography.Structure.Overlay
 import Swarm.Game.Scenario.Topography.Structure.Recognition.Static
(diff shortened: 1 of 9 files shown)


src/swarm-topography/Swarm/Game/Scenario/Topography/Structure.hs
```diff
@@ -138,7 +138,7 @@ paintMap maskChar pal g = do
 
   usedChars = Set.fromList $ allMembers g
   paletteKeys = M.keysSet $ unPalette pal
-  unusedPaletteChars = Set.difference paletteKeys usedChars
+  unusedPaletteChars = Set.difference paletteKeys (usedChars <> paletteChars pal)
```
xsebek: Do chars silently win over rest of palette?

byorgey: I'm not sure I understand your question.

- `chars` has *lower* priority than the rest of the palette.  So *e.g.* if you have `chars: 'A-Z'` but then define `'F': [grass, whatever]`, then `F` will be defined as `[grass, whatever]` and all the other letters will be default character entities.
- However, this code has nothing to do with that.  This code is trying to warn the user about unused palette entries.  At first, when I used something like `chars: 'A-Za-z'` and wrote a message, it warned me about every single letter I didn't use!  This seemed more tedious than helpful, so I consider everything in a `chars` declaration as "used".  I suppose I could try to be a bit more careful about this: if you say `chars: A-Z` and don't use `F`, it shouldn't warn you, of course.  But if you say `chars: A-Z` and also define `'F': [grass]` then it should warn you if you don't use F in the map.

PR #2497, opened 2025-06-20
Add tests for narrowToPosition

oliverpauffley:
I want to refactor narrowToPosition and I was getting confused if I had broken it or not. So I thought the best plan would be to add the tests first then refactor. Will follow up with another PR for the changes.

### DIFF
--- a/swarm.cabal
+++ b/swarm.cabal
@@ -130,7 +130,7 @@ common JuicyPixels
   build-depends: JuicyPixels >=3.3 && <3.4
 
 common QuickCheck
-  build-depends: QuickCheck >=2.14 && <2.16
+  build-depends: QuickCheck >=2.14 && <2.17
 
 common SHA
   build-depends: SHA >=1.6.4 && <1.6.5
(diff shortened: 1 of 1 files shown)


oliverpauffley:
Yes I did make the test aim for small terms so that they aren't so long!

oliverpauffley:
Also I have no idea how to resolve the `LOC` being different between versions? I guess this is down to parsing differences?

oliverpauffley:
> > Also I have no idea how to resolve the LOC being different between versions? I guess this is down to parsing differences?
> 
> What version do you mean?

Yeah windows vs linux versions. Sorry not very clear.

I think changing the assert makes sense here I will look into that

test/unit/TestLSP.hs
```diff
@@ -27,67 +31,126 @@ simplifyWarning (VU.VarUsage (S.LV _ v) scope) = UnusedVar v scope
...
+        , testCase "record with unused var" $
+            checkFile
+              "lambda-with-record-unused.sw"
+              [UnusedVar "y" VU.Lambda]
+        ]
+    , testGroup
+        "narrowToPosition"
+        [ testCase "narrow to TVar" $
+            assertEqualTerms
+              (S.Syntax' (S.SrcLoc 57 59) (S.TVar "m2") (S.Comments mempty mempty) ())
```
xsebek: I would suggest putting `NoLoc` instead of `SrcLoc` so that it's clear the location is not tested.

PR #2505, opened 2025-06-26
Destroy base if it has `life support system` unequipped for 256 ticks

### LINKED ISSUE #1388: Consequences for unequipping `life support system`
I feel like unequipping the `Life Support System` device should not be allowed, but you can do it without any problems and can continue the game as normal.

byorgey:
When the `base` unequips the `life support system`:
- Log a dire warning
- Spawn a system robot that counts down from 256
    - If the base re-equips the `life support system` at any point, cancel the countdown
    - Otherwise, destroy the base at the end of the countdown (and grant `DestroyedBase` achievement)

This PR also adds a new command `destroy :: Actor -> Cmd Unit` which destroys another robot.  For obvious reasons, this requires `God` capability.

Closes #1388.

### DIFF
--- a/src/swarm-engine/Swarm/Game/Step/Const.hs
+++ b/src/swarm-engine/Swarm/Game/Step/Const.hs
@@ -432,6 +432,13 @@ execConst runChildProg c vs s k = do
         equippedDevices %= delete item
         robotInventory %= insert item
 
+        when (myID == 0 && itemName == "life support system") $ do
+          -- If the base unequips the life support system, you die!
+          -- This is one of only two (known) ways to get the
+          -- `DestroyedBase` achievement.
+          selfDestruct .= True
+          when (myID == 0) $ grantAchievementForRobot DestroyedBase
+
         -- Now check whether being on the current cell would still be
         -- allowed.
         loc <- use robotLocation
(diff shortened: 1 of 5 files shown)


src/swarm-engine/Swarm/Game/Step/Const.hs
```diff
@@ -158,6 +158,17 @@ execConst runChildProg c vs s k = do
       destroyIfNotBase $ \case False -> Just AttemptSelfDestructBase; _ -> Nothing
       flagRedraw
       return $ mkReturn ()
+    Destroy -> case vs of
+      [VRobot rid] -> do
+        myID <- use robotID
+        if (myID == rid)
```
xsebek: HLint does not like this. 😄

src/swarm-engine/Swarm/Game/Step/Const.hs
```diff
@@ -448,7 +478,7 @@ execConst runChildProg c vs s k = do
           PathLiquid _ -> do
             -- Unequipping a device that gives the Float capability in
             -- the middle of liquid results in drowning, EVEN for
-            -- base!  This is currently the only (known) way to get
+            -- base!  This is currently the other known way to get
             -- the `DestroyedBase` achievement.
```
xsebek:
```suggestion
            -- base!  This gives the `DestroyedBase` achievement,
            -- the other way is unequipping life support.
```

xsebek (approved):

PR #2512, opened 2025-06-29
Add `ekg` metrics

### LINKED ISSUE #909: Performance monitoring
The goal is to measure the performance of the different parts of the game. Here are some available options:

- [prometheus.io](https://prometheus.io), haskell packages [prometheus-client](https://hackage.haskell.org/package/prometheus-client) or [prometheus](https://hackage.haskell.org/package/prometheus). Prometheus works by polling a text http endpoint, it supports counter and gauge. it can be used to query the metric directly, or it can be configured as a [grafana](https://grafana.com/) data source.
- [opentelemetry.io](https://opentelemetry.io/), haskell packages [opentelemetry](https://hackage.haskell.org/package/opentelemetry) or [tracing](https://hackage.haskell.org/package/tracing). Telemetry supports "spans" and events. Logs can be processed offline, for example with https://profiler.firefox.com/ , or sent to a live service like https://zipkin.io/ or https://www.jaegertracing.io/ .
- [ekg](https://hackage.haskell.org/package/ekg-core-0.1.1.7/docs/System-Metrics.html), a standalone solution in pure haskell.

I think ekg would be the most user friendly as it should work straight from the web interface. Though we might get better results using an extra service.

xsebek:
* add `ekg-core` metrics for game tick
* add fused-effect for measuring CPU time
* add command line `--metrics PORT` option to configure or disable `ekg-wai` API and website
* Closes #909

### DIFF
--- a/src/swarm-engine/Swarm/Game/State/Runtime.hs
+++ b/src/swarm-engine/Swarm/Game/State/Runtime.hs
@@ -13,6 +13,7 @@ module Swarm.Game.State.Runtime (
 
   -- ** Lenses
   webPort,
+  metricsPort,
   upstreamRelease,
   eventLog,
   appData,
@@ -45,6 +46,7 @@ import System.Metrics qualified as Metrics
 
 data RuntimeState = RuntimeState
   { _webPort :: Maybe Int
+  , _metricsPort :: Maybe Int
   , _upstreamRelease :: Either (Severity, Text) String
   , _eventLog :: Notifications LogEntry
   , _stdGameConfigInputs :: GameStateConfig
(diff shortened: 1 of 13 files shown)


xsebek:
You can test it by opening the webpage `http://localhost:6543` in browser:

|Screenshot 2025-06-29 at 5 10 32 PM|
|---|

Or access the metrics directly:
```bash
curl -s -H "Accept: application/json" localhost:6543/game.tick_time | jq
```
```json
{
  "count": 2442,
  "max": 0,
  "mean": 0.014012881244881249,
  "min": 0,
  "sum": 34.21945600000001,
  "type": "d",
  "variance": 0.002380784686224228
}
```

src/swarm-engine/Swarm/Game/State/Runtime.hs
```diff
@@ -131,3 +143,4 @@ stdGameConfigInputs :: Lens' RuntimeState GameStateConfig
 -- | Free-form data loaded from the @data@ directory, for things like
 --   the logo, about page, tutorial story, etc.
 appData :: Lens' RuntimeState (Map Text Text)
+metrics :: Lens' RuntimeState Metrics.Store
```
byorgey, line 151: Missing Haddock comment?

Issue #2510, opened 2025-06-28
Only redraw changed world cells


byorgey:
The idea
-------------

I just had an epiphany this morning... one of things that is extremely obvious once you think of it, but somehow I had never thought of it before.  I am quite certain that one of the biggest low-hanging fruits in terms of optimizing the game's speed is to find a way to **recompute + redraw only the specific world cells that have changed every tick** rather than redrawing the entire world every time anything changes.

How do we know this is worth pursuing?
-----------------------------------------------------------

- Start a new creative game.  Make the terminal as large as possible.
- Hit Alt-F to open the FPS view.  On my computer, I get something close to 30 FPS (which is the cap).
- Now, program a robot to move around, e.g. `def forever = \c. c ; forever c end` and then `build {forever (move; turn right)}`.
- Watch the FPS plummet.  On my laptop, with a 63x239 terminal, I get about 4-5 FPS, which means at the default speed of 16 ticks/second it is only drawing a frame about every 4th tick.
- Now, with the robot still running, make the terminal much smaller.
- Watch the FPS skyrocket again.  On my computer, if I shrink the terminal to 20x80 I get basically 30 FPS again.

The obvious conclusion is that actually computing each tick is not the bottleneck; the bottleneck is drawing the world on the screen.  Currently, we are careful to cache + not redraw the world if nothing has changed, which is why the game runs at 30 FPS before any robots have been built; but if anything changes at all, we set a dirty flag and redraw the entire world view.  But this is incredibly wasteful, since in most cases only a few cells have actually changed their appearance.

I don't know which takes more time: *recomputing* the appearance of a cell (which involves evaluating the world DSL, doing lookups up in the robot map, and so on), or actually *drawing* it on the screen (which involves various `brick` and `vty` internal details), but it doesn't matter.  Only recomputing + redrawing the cells that have actually changed should be a big win.

Technical approach
------------------------------

As for the technical details of how this could work, I am not 100% sure.  I suppose the first thing would be to replace the single dirty bit with a dirty *set* of coordinates.  Practically speaking, we could then change the type of `flagRedraw` to take some coordinates; then the typechecker would force us to find every use of `flagRedraw` and think carefully about which coordinates need to be marked dirty.

I am less sure about the rest.  It may involve refactoring the way we draw the world view, and/or asking for some help from the `brick` folks to figure out the best way to achieve our goals.  But we could probably just start and figure it out along the way.

[...]

PR #2513, opened 2025-06-29
Render world view directly as vty Image

byorgey:
We used to create the world view by rendering each cell as a 1x1 brick `Widget`, then using `brick` combinators like `hBox` and `vBox` to put them together.  In this PR, based on advice from https://github.com/jtdaugherty/brick/discussions/531 , I directly create a vty `Image` representing the world view and then use brick's `raw` function to turn it into a `Widget`.

This is extremely low-hanging fruit and does seem to save some rendering time, presumably because brick `Widget`s have a bunch of extra overhead (for tracking extents, dealing with clicks, etc.) that we now avoid.  For example, on my machine, with a 61x238 terminal running `def forever = \c. c ; forever c end; build {forever (move; turn right)}` in a fresh Creative world:

- Prior to this PR: ~4.8 FPS
- With this PR: ~7.5 FPS

This is a small step towards #2510, but we are still redrawing the entire world every frame.  I plan to follow this up with a more sophisticated approach that should save more time.

### DIFF
--- a/data/scenarios/Tutorials/bind2.yaml
+++ b/data/scenarios/Tutorials/bind2.yaml
@@ -93,6 +93,7 @@ robots:
       - [1, treads]
       - [1, grabber]
       - [1, compass]
+known: [Hastur, pedestal]
 world:
   palette:
     'Ω': [stone, null, base]


xsebek (approved):
Cool! 👍

PR #2515, opened 2025-07-04
View chunk caching for fast world drawing

### LINKED ISSUE #2387: Cached world view for scenarios is not regenerated when terminal size changes
**Describe the bug**
If the terminal size is changed while in the New Game menu, the world preview for any scenarios that were visited while the terminal was smaller retain their smaller size 

**To Reproduce**
1. Start the game
2. Choose `New Game`
3. Make the terminal bigger
4. Move down to `Creative game`
5. Move back and forth between `Classic game` and `Creative game` and observe that the world preview for `Creative game` is an appropriate size for the new terminal size, but the world preview for `Classic game` is still smaller

**Expected behavior**
The world preview size should always match the current terminal size.

**Screenshots**

Image

Image

Image

**Additional context**
I assume this has something to do with caching.  We should probably just invalidate the cached world previews when the terminal size changes.

### LINKED ISSUE #2510: Only redraw changed world cells

byorgey:
Following up on the discussion at https://github.com/jtdaugherty/brick/discussions/531, and building on #2513, this PR splits up the world view into chunks and individually caches each chunk, leading to *much* faster render times.  Closes #2510.

Details
----------

- The chunks are 4 cells x 4 cells (empirically, this seems to give the best performance; see below), and aligned to multiples of 4.  That is, there is always a chunk from (0,0) to (3,3), and a chunk from (0,4) to (3,7), and so on; to calculate which chunk contains a given cell we just need to do some right bit shifts.  This is similar to the way we load world tiles (though world tiles are much larger).  See `Swarm.TUI.Model.ViewChunk`.
- To draw the world view, we generate a grid of chunks which completely cover the desired viewing region, render them all, then crop to the desired region.  See `Swarm.TUI.View.worldWidget`.
- Each view chunk is cached, so it does not need to be re-rendered if nothing in that chunk has changed.
- Note that another nice side benefit of basing the view chunks on world rather than viewing window coordinates is that the game is still very fast when scrolling the world (e.g. when `view`ing a robot which is moving around), because we can still reuse all the rendered view chunks that have not changed; they just end up at a different place on the screen.
- During each tick, we accumulate a set of "dirty" cells which have changed since the last frame.  When rendering a frame, we first invalidate the cache entries for all the view chunks containing dirty cells.
- It is also possible to set a flag to invalidate the entire cache and redraw the world view from scratch, which is necessary in certain situations, e.g. when the currently viewed robot learns about a new entity, so the display of that entity may need to change from a question mark everywhere it occurs.  See #2527.
- Along the way I also fix #2387 by invalidating the cache on terminal resize events at the top level, instead of only when running a scenario.

Performance analysis
--------------------------------

I did some quick & dirty performance analysis by opening a new Creative game in a 61 x 238 terminal, hitting Alt-F to show the FPS meter, and executing
```
def forever = \c. c ; forever c end
build {move; forever (move; turn right)}
```

- Prior to this PR, it maxed out at about 7-8 FPS.
- With this PR, it now runs at 30 FPS - but only because we cap the game at 30 FPS.  When I comment out the line of code that caps the FPS, it runs at up to about 75 FPS.

For a bit more robust (but still pretty quick & dirty) analysis, and to find the best value for `viewChunkBits` (i.e. the value of $k$ such that each view chunk is $2^k \times 2^k$ cells), I opened a Blank game in a 61 x 328 terminal and executed
```
def forever = \c. c ; forever c end
def x = \n. \c. if (n == 0) {} {c ; x (n-1) c} end
x 20 (build {forever (x 20 move; turn left)}; turn left; wait 13)
```
$k = 2$ was the clear winner, with the game running at about 20 FPS when set to 64 ticks per second.  It became slower with both smaller values of $k$ (presumably because of all the extra work brick has to do composing lots of tiny view chunk widgets into a big rectangle) and with larger (presumably because the chunks are too large so we spend too much time unnecessarily re-rendering cells that have not changed).

Admittedly my tests were not very robust, and the optimal value of $k$ might depend a lot on one's hardware and how big you would like your world view to be.  I suppose a nice follow-up could be to make the value of $k$ configurable via a command-line parameter.

Follow-up issues
-------------------------

- #2527
- #2528 
- #2529

### DIFF
--- a/src/swarm-engine/Swarm/Game/State/Redraw.hs
+++ b/src/swarm-engine/Swarm/Game/State/Redraw.hs
@@ -64,8 +64,4 @@ resetRedraw = const initRedraw
 -- | Does the world pane need redrawing at all?
 needsRedraw :: Redraw -> Bool
 needsRedraw r =
-  or
-    [ r ^. redrawWorld
-    , r ^. drawFrame
-    , not (S.null $ r ^. dirtyCells)
-    ]
+  r ^. redrawWorld || r ^. drawFrame || not (S.null $ r ^. dirtyCells)
--- a/src/swarm-engine/Swarm/Game/State/Redraw.hs
+++ b/src/swarm-engine/Swarm/Game/State/Redraw.hs
@@ -1,0 +1,74 @@
+{-# LANGUAGE TemplateHaskell #-}
+
+-- |
+-- SPDX-License-Identifier: BSD-3-Clause
+--
+-- A subrecord containing state specific to deciding when and which
+-- parts of the world view to redraw.
+module Swarm.Game.State.Redraw (
+
+  Redraw,
+
+  -- * Lenses
+
+  dirtyCells,
+  redrawWorld,
+  drawFrame,
+
+  -- * Utility
+
+  initRedraw,
+  resetRedraw,
+  needsRedraw
+
+) where
+
+import Control.Lens ((^.), Lens')
+import Data.Set (Set)
+import Data.Set qualified as S
+import Swarm.Game.Location (Location)
+import Swarm.Game.Universe (Cosmic)
+import Swarm.Util.Lens (makeLensesNoSigs)
+
+-- | Subrecord of state relating to whether (and which parts of) the
+--   world view needs to redrawn.
+data Redraw = Redraw
+  { _dirtyCells :: Set (Cosmic Location)
+  , _redrawWorld :: Bool
+  , _drawFrame :: Bool
+  }
+
+makeLensesNoSigs ''Redraw
+
+-- | Which locations have changed since the previous frame?  We will
+--   need to be sure to redraw the view chunks containing these
+--   locations.
+dirtyCells :: Lens' Redraw (Set (Cosmic Location))
+
+-- | Should we redraw the entire world (i.e. invalidate the entire
+--   view chunk cache) next frame?
+redrawWorld :: Lens' Redraw Bool
+
+-- | Should we make sure that we redraw the world pane, even if the
+--   world view itself has not changed?  This might be relevant
+--   e.g. for things displayed around the border of the world pane.
+drawFrame :: Lens' Redraw Bool
+
+-- | Initial 'Redraw' state, indicating that nothing needs to be
+--   redrawn.
+initRedraw :: Redraw
+initRedraw = Redraw S.empty False False
+
+-- | Reset the redraw state, by setting it to 'initRedraw'.  Typically
+--   this should be called right after drawing each frame, to be ready
+--   to collect redrawing information for the next frame.
+resetRedraw :: Redraw -> Redraw
+resetRedraw = const initRedraw
+
+-- | Does the world pane need redrawing at all?
+needsRedraw :: Redraw -> Bool
+needsRedraw r = or
+  [ r ^. redrawWorld
+  , r ^. drawFrame
+  , not (S.null $ r ^. dirtyCells)
+  ]
--- a/src/swarm-engine/Swarm/Game/State/Redraw.hs
+++ b/src/swarm-engine/Swarm/Game/State/Redraw.hs
@@ -21,7 +21,7 @@ module Swarm.Game.State.Redraw (
   redrawWorld,
 ) where
 
-import Control.Lens (Lens', Getter)
+import Control.Lens (Getter, Lens')
 import Data.Set (Set)
 import Data.Set qualified as S
 import Swarm.Game.Location (Location)
--- a/src/swarm-engine/Swarm/Game/State/Redraw.hs
+++ b/src/swarm-engine/Swarm/Game/State/Redraw.hs
@@ -10,16 +10,18 @@ module Swarm.Game.State.Redraw (
 
   -- * Lenses
   dirtyCells,
-  redrawWorld,
+  redrawWorldFlag,
   drawFrame,
 
   -- * Utility
   initRedraw,
   resetRedraw,
   needsRedraw,
+  markDirtyCell,
+  redrawWorld,
 ) where
 
-import Control.Lens (Lens', (^.))
+import Control.Lens (Lens', Getter)
 import Data.Set (Set)
 import Data.Set qualified as S
 import Swarm.Game.Location (Location)
@@ -30,7 +32,7 @@ import Swarm.Util.Lens (makeLensesNoSigs)
 --   world view needs to redrawn.
 data Redraw = Redraw
   { _dirtyCells :: Set (Cosmic Location)
-  , _redrawWorld :: Bool
+  , _redrawWorldFlag :: Bool
   , _drawFrame :: Bool
   }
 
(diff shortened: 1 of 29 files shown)


src/swarm-engine/Swarm/Game/Step/Const.hs
```diff
@@ -396,24 +396,29 @@ execConst runChildProg c vs s k = do
         -- robotMap, overwriting any changes to this robot made
         -- directly in the robotMap during the tick.
         myID <- use robotID
-        focusedID <- use $ robotInfo . focusedRobotID
         if otherID /= myID
           then do
+            -- Check whether the receiving robot knew about this item before
+            inv <- gets (preview (robotInfo . robotMap . at otherID . _Just . robotInventory))
+            -- No idea why 'preuse' doesn't typecheck above, but (gets . preview) works
```
xsebek: Maybe its the transformer preuse and not the fused effect one?

I think I recently wrote `use . pre`.

src/swarm-tui/Swarm/TUI/Controller.hs
```diff
@@ -541,6 +541,10 @@ handleModalEvent = \case
 -- * returns to the previous menu
 quitGame :: Bool -> EventM Name AppState ()
 quitGame isNoMenu = do
+  -- Invalidate drawing cache so the world preview in the menu will be
+  -- refreshed
+  -- invalidateCache
```
xsebek: commented code?

src/swarm-engine/Swarm/Game/Step/Const.hs
```diff
@@ -432,9 +437,6 @@ execConst runChildProg c vs s k = do
             equippable <- use $ discovery . craftableDevices
             when (equippable `S.isSubsetOf` equipped) $ grantAchievementForRobot EquippedAllDevices
 
-          -- Flag the UI for a redraw if we are currently showing our inventory
-          when (focusedID == myID) flagRedraw
-
         return $ mkReturn ()
```
byorgey, line 439: Check if this means the inventory updates slowly now.

src/swarm-engine/Swarm/Game/Step/Const.hs
```diff
@@ -482,8 +483,6 @@ execConst runChildProg c vs s k = do
             selfDestruct .= True
             when (myID == 0) $ grantAchievementForRobot DestroyedBase
 
-        -- Flag the UI for a redraw if we are currently showing our inventory
-        when (focusedID == myID) flagRedraw
```
byorgey, line 486: Check this too.

src/swarm-tui/Swarm/TUI/Controller.hs
```diff
@@ -541,6 +541,10 @@ handleModalEvent = \case
...
+
```
byorgey: This was a failed attempt to fix #2387, should remove this.

src/swarm-tui/Swarm/TUI/Model/ViewChunk.hs
```diff
@@ -0,0 +1,82 @@
...
+ where
+  ul = mapCoords (`shiftL` viewChunkBits) <$> cc
+  lr = addTuple (ul ^. planar) (viewChunkSize - 1, viewChunkSize - 1)
+
+-- | @viewChunkCover r@ generates the smallest possible set of
+--   'ViewChunk's necessary to entirely cover the given rectangle @r@.
+--   The resulting 'ViewChunk's are returned in row-major order, that
+--   is, the first list represents the topmost row from left to right,
+--   the second list represents the second row, and so on.
+viewChunkCover :: Cosmic BoundsRectangle -> NonEmpty (NonEmpty ViewChunk)
```
xsebek, line 66: Could you please add a test for these properties?

src/swarm-engine/Swarm/Game/State/Redraw.hs
```diff
@@ -0,0 +1,67 @@
...
+makeLensesNoSigs ''Redraw
+
+-- | Which locations have changed since the previous frame?  We will
+--   need to be sure to redraw the view chunks containing these
+--   locations.
+dirtyCells :: Lens' Redraw (Set (Cosmic Location))
+
+-- | Should we redraw the entire world (i.e. invalidate the entire
+--   view chunk cache) next frame?
+redrawWorld :: Lens' Redraw Bool
```
xsebek: If these were setters and not lenses, I imagine redraw world would throw away the set and stop adding new elements.

But it's probably just a small optimisation. So feel free to ignore.

xsebek (approved):
Looks good, I will test it later. 👍 

I forgot we have FPS view.

src/swarm-tui/Swarm/TUI/Model/ViewChunk.hs
byorgey: Done!  Caught a few (minor) bugs in the process so I appreciate the nudge. :sweat_smile:

xsebek:
I tested it locally on my mac and at default zoom it is both 30fps. But once I made the font smaller to fit more world, the current version slowed down to 10fps, while this one stays at 30fps. 👍  (So pretty much same results.)

I assume this happens on the main thread, so saving time there will make the game feel more responsive. 🙂

src/swarm-engine/Swarm/Game/Step/Const.hs
byorgey: It doesn't.  The inventory gets redrawn immediately because its hash changes.

PR #2517, opened 2025-07-04
Add total and active robots metrics

xsebek:
* add `game.robots_active` and `game. robots_total` gauges

You can test it with:
```bash
cabal run swarm -O0 -- +RTS -T -RTS --scenario Creative --run <( echo 'def x=\\i.\\c.if(i<=0){}{_ <-c; x (i-1) c} end; x 100 (build {wait 100; x 20 move})')
```
and then open http://localhost:6543 in browser or with cURL:
```bash
curl -s -H "Accept: application/json" localhost:6543 | jq '.game'
```
```json
{
  "robots_active": {
    "type": "g",
    "val": 16
  },
  "robots_total": {
    "type": "g",
    "val": 46
  },
  ...
}
```

### DIFF
--- a/src/swarm-engine/Swarm/Game/Step.hs
+++ b/src/swarm-engine/Swarm/Game/Step.hs
@@ -108,8 +108,12 @@ gameTick = measureCpuTimeInSec runTick >>= updateMetrics
   updateMetrics (t, res) =
     use gameMetrics >>= \case
       Just metrics -> do
-        sendIO $ Counter.inc metrics.tickCounter
-        sendIO $ Distribution.add metrics.tickDistribution t
+        (fTotal, fActive) <- countRobots
+        sendIO $ do
+          Counter.inc metrics.tickCounter
+          Distribution.add metrics.tickDistribution t
+          Gauge.set metrics.robotsGauge $ fromIntegral fTotal
+          Gauge.set metrics.activeRobotsGauge $ fromIntegral fActive
         pure res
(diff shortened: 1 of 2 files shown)


src/swarm-engine/Swarm/Game/Step.hs
```diff
@@ -108,50 +109,64 @@ gameTick = measureCpuTimeInSec runTick >>= updateMetrics
...
     time <- use $ temporal . ticks
     zoomRobots $ wakeUpRobotsDoneSleeping time
+    ticked <- runActiveRobots
+    updateBaseReplState
+    -- Possibly update the view center.
+    modify recalcViewCenterAndRedraw
+    -- On new tick see if the winning condition for the current objective is met.
+    when ticked checkWinCondition
+    return ticked
+  runActiveRobots :: m Bool
```
byorgey: In my mind, there are two use cases for making local definitions in a `where` clause: (1) if you want to have some parameter in scope throughout all of them, or (2) if the local definitions are short and it's just a way to organize/split up the code for a single function.  In a case like this, with multiple longer functions, it seems to me it would be better to just promote them all to top-level functions.  That helps with testability, potential reuse, and readability.

xsebek: I agree, but the downside is that you can learn that `hypotheticalWinCheck` signature does not make sense and spend some time refactoring. 😄

PR #2522, opened 2025-07-06
Watch for robots

xsebek:


src/swarm-engine/Swarm/Game/Step/Util/Command.hs
```diff
@@ -182,8 +182,11 @@ updateRobotLocation oldLoc newLoc
   | otherwise = do
       newlocWithPortal <- applyPortal newLoc
       rid <- use robotID
+      t <- use $ temporal . ticks
+      invis <- use $ robotDisplay . invisible
       zoomRobots $ do
-        -- wakeWatchingRobots?
+        unless invis $ wakeRobotsWatchingForRobots rid t newLoc
+        unless invis $ wakeRobotsWatchingForRobots rid t oldLoc
```
xsebek, line 189: Commenting out these two lines fixes two tests, but not `Challenges/Sliding Puzzles/3x3`.

I suspect one of the side effects in `wakeWatchingRobots` is wrong because it does not expect to be called more than once in a robot step.

Why the `3x3` puzzle still fails is a mystery to me. 😕

xsebek: Oh, `3x3` fails for me locally in `main`. So it is an unexpected, but separate issue.

EDIT: It only fails in `-O0`. 😐

byorgey:
@xsebek As you recall I was initially resistant to the idea of making `watch` also watch for robots in addition to entities.  I didn't want it to become too complicated, and I thought it wouldn't give enough control: what if you *only* wanted to be woken on entity changes, or only on robot changes?  But in reflecting on it more, I think it is not too hard to recover this control: if you only care about, say, entity changes, then all you have to do is an extra `scan` when you are woken, and go back to sleep if something changed besides what you care about.  It actually seems simpler to me now to just have `watch` + `scan` + `meet` and use them in combination to specify what you care about.  So I also lean towards just having a single `watch` and a single `surveil` command.

data/scenarios/Tutorials/debug-hint.sw
```diff
@@ -13,96 +13,118 @@ def foreachF = \s.\e.\com.\state.
...
 // An infinite while cycle that carries a state.
-def iterate = \state.\com.
-  n <- com state;
-  iterate n com;
-end;
+def iterate = \state. \com. n <- com state; iterate n com end
 
-tydef RobotState = [gave_win: Bool, said_log_missing: Bool, said_loc: (Int * Int)] end;
+// -------------------------------------------------------------------------
+// Robot State - this allows us to skip some commands, so that this hont robot
```
byorgey: "hont"?

xsebek:
@byorgey thats what I tried in the branch with `reWatchEntities`. 🙂 

System robots can easily check it all in one tick with `instant`, other robots would have to use atomic and react in next tick or only watch one location.

data/scenarios/Tutorials/debug-hint.sw
xsebek:
```suggestion
// Robot State - this allows us to skip some commands, so that this hint robot
```

PR #2535, opened 2025-07-14
Redraw world when static changes

### LINKED ISSUE #2534: Static is only drawn in blocks where robots are moving
**Describe the bug**
When the entire view should be static, only some view chunks are drawn with static where something has been updated.

**To Reproduce**
Start a Classic game, and execute
```
r <- build {}; view r; salvage
```

**Expected behavior**
The entire world view should show shifting static.  Instead, just a single view chunk at the origin shows static.  If you then e.g. `build {move}` an additional view chunk of static will show up, as an additional view chunk is marked dirty.

**Screenshots**

Image

**Additional context**
This was definitely introduced in #2515.

byorgey:
Fixes #2534.

### DIFF
--- a/src/swarm-engine/Swarm/Game/State/Redraw.hs
+++ b/src/swarm-engine/Swarm/Game/State/Redraw.hs
@@ -10,7 +10,6 @@ module Swarm.Game.State.Redraw (
 
   -- * Lenses
   dirtyCells,
-  prevFocusRange,
   redrawWorldFlag,
   drawFrame,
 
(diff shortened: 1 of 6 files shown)


src/swarm-engine/Swarm/Game/State/Redraw.hs
```diff
@@ -55,24 +64,35 @@ drawFrame :: Lens' Redraw Bool
...
 -- | Does the world pane need redrawing at all?
 needsRedraw :: Redraw -> Bool
-needsRedraw (Redraw d w f) = w || f || not (S.null d)
+needsRedraw (Redraw d _ w f) = w || f || not (S.null d)
+
+-- | Update the current focus range.  If it has changed, mark the
+--   world for redrawing.
+updateFocusRange :: Maybe RobotRange -> Redraw -> Redraw
+updateFocusRange newR r
+  | _prevFocusRange r /= newR = redrawWorld $ r {_prevFocusRange = newR}
```
xsebek, line 86: I wish we could use the lens here and export the getter, but since it's only in this small module it's not important. 🤷

byorgey: Hmm, though perhaps we can use a lens and not export it at all.  I'll try that.

Issue #2526, opened 2025-07-06
Add Haskell logging infrastructure


xsebek:
**Is your feature request related to a problem? Please describe.**

Logging is a fundamental tool for debugging and making sense of program lifetime.

**Describe the solution you'd like**

Add structured logging from [`log-base`](https://hackage.haskell.org/package/log-base) to Runtime state.

The logs should be kept in memory (up to some buffer limit) and written to file.

Publishing the logs on the Web API or on separate port could be nice if it integrates to some industry standard solution like elastic. But if it has difficult setup or brings too many dependencies we can leave that for later.

**Describe alternatives you've considered**

Rolling our own logger is probably a bad idea, at least we should implement the structured logging interface.

PR #2537, opened 2025-07-20
Add developer logs using log-base library

xsebek:


src/swarm-engine/Swarm/Game/Step.hs
```diff
@@ -120,14 +127,17 @@ gameTick = measureCpuTimeInSec runTick >>= updateMetrics
...
-    zoomRobots $ wakeUpRobotsDoneSleeping time
-    ticked <- runActiveRobots
-    updateBaseReplState
-    -- Possibly update the view center.
-    modify (robotInfo %~ recalcViewCenter)
-    -- On new tick see if the winning condition for the current objective is met
-    when ticked hypotheticalWinCheck'
-    return ticked
+    Effect.localData ["tick" Effect..= time] $ do
+      Effect.logTrace_ "Running tick"
```
xsebek, line 131: @byorgey I realized I have not set a goal in #2526, what should we start logging for the issue to be _done_.

Do you have any suggestions? I am thinking:
* loading each part of state
* scenario start
* scenario end

Maybe also something in TUI to see how it would work in `EventM` monad.

byorgey: That sounds like a good start.

src/swarm-tui/Swarm/TUI/Model/StateUpdate.hs
```diff
@@ -144,6 +145,7 @@ mkRuntimeOptions AppOpts {..} =
     { startPaused = pausedAtStart
     , pauseOnObjectiveCompletion = autoShowObjectives
     , loadTestScenarios = Set.member LoadTestingScenarios debugOptions
+    , startLogging = True
```
byorgey, line 148: Does `startLogging` ever get set to `False`?  Will the game just be always writing out a log to disk by default?

xsebek: Oh, I should add a command line option or maybe finally add a config file.

PR #2538, opened 2025-07-20
Watch for robots as well

### LINKED ISSUE #2518: Add `watch` command for robots

The `watch` command only watches for entity changes, but especially the system robots want to watch for robots to enter and leave locations.

Add new command:
```Haskell
watchForRobots: (Int, Int) -> Cmd Unit
```
that will wake the robot if another robot enters or leaves the watched cell.

Implement it by adding new `MonoidMap` to `Robots` state and wake up when robots move in/out of the location.

Prove this works by updating the `Debug` tutorial.

Adding this behavior to current `watch` breaks a few challenge scenarios and they are quite complex for me to debug.
This would not allow scenario authors to _limit_ players to only watch for robots or entities.

As another thought, I am not sure watching for robots leaving the cell is necessary.
It does feel well defined to me, `deltaCell := None | RobotLeave | RobotEnter` and could be useful for writing scenarios.

It might be nice to rename `watch` to `watchForEntities` and once #495 is implemented add `watch` in Prelude.

**Additional context**

* This will allow optimizing the Debug scenario robot to not busy wait, see my failed attempt at the alternative approach
  in #2395

xsebek:
* fix `wakeWatchingRobots` to allow calling it more than once
* make Debug tutorial system robot more efficient using `watchRobots`
* change ranching scenario solution to work with watching for robots (do not step on the watched space too soon)
* move robot naming state to separate module
* move robots state and lenses to an internal module
* make gopher move silently
* allow empty appear parameter making robot invisible
* closes #2522 - **this PR is a rebased version with the new map and commands reverted**
* closes #2518
* closes #2395


src/swarm-engine/Swarm/Game/Step/Util/Command.hs
```diff
@@ -182,8 +182,11 @@ updateRobotLocation oldLoc newLoc
   | otherwise = do
       newlocWithPortal <- applyPortal newLoc
       rid <- use robotID
+      t <- use $ temporal . ticks
+      invis <- use $ robotDisplay . invisible
       zoomRobots $ do
-        -- wakeWatchingRobots?
+        unless invis $ wakeWatchingRobots rid t newLoc
+        unless invis $ wakeWatchingRobots rid t oldLoc
```
xsebek, line 189: It seems that as a side effect, the gopher scenario takes longer. Everything else takes roughly the same time, but somehow, this makes the gopher more (cca 3x) elusive. 🤔

byorgey: Longer, as in, more ticks?  Or more wall clock time?

byorgey (approved):

src/swarm-engine/Swarm/Game/Step/Util/Command.hs
xsebek: More ticks - the robots eventually catch the gopher, but it takes longer.

byorgey: Strange.

src/swarm-engine/Swarm/Game/Step/Const.hs
```diff
@@ -880,10 +880,14 @@ execConst runChildProg c vs s k = do
       [VText app, VInj hasAttr mattr] -> do
         -- Set the robot's display character(s)
         case into @String app of
+          [] -> do
+            robotDisplay . invisible .= True
```
xsebek, line 884: @byorgey what do you think? I could limit it to system robots if it is too powerful, but since it requires `beaglepuss` it sounds plausible to me. 🥸

byorgey: Sure, I like it, why not?

We may need to reconsider if we ever get some kind of adversarial multiplayer mode (i.e. playing against an opponent in a shared world over the network).  A `beaglepuss` is not that hard to make (`rubber` + `glass` + `string`) so this would be pretty OP if you could do stuff to your opponent with invisible robots.  But for now it's fun and doesn't really matter.

data/scenarios/Challenges/_gopher/gopher.sw
```diff
@@ -20,7 +20,7 @@ def goDir = \dist. \d1. \d2.
         turn d2;
     } {};
 
-    doN (abs dist) move;
+    doN (abs dist) (instant (appear "" (inl ()); move; appear "o" (inl ())));
```
xsebek, line 23: I forgot to commit this, but this should make the gopher visible yet move silently.

byorgey: I don't understand this --- how is this any different than `appear "o" (inl ()); doN (abs dist) move`?

Edited to add: oh, wait, I see, is it because `move`ing while invisible doesn't trigger any `watch`ing robots?

byorgey:
Looks like the `flower-count` test failed on GHC 9.8.2 --- probably just spurious?

PR #2540, opened 2025-07-28
Import

byorgey:
This still does not work at all, but it's inching closer and I figured it was worth opening a draft PR, so I can start working on a comprehensive PR description in parallel with finishing it up.


xsebek:
I tried to test it on the move tutorial:
```diff
   - name: 1P wall
     system: true
     program: |
-      def main = \a. pure noop end
       instant {
-        run "scenarios/Tutorials/move_surveil.sw";
+        import "scenarios/Tutorials/move_surveil.sw" in
         main [entity="wall", room=1]
       }
```
But the robots fail to run:
```
0:00:00.0: Fatal error: Import not found:
/Users/xsebek/Git/swarm-import/scenarios/Tutorials/move_surveil.sw
Please report this as a bug at <https://github.com/swarm-game/swarm/issues/new>.
```

Also, could you please fix the `Swarm/Doc/Pedagogy.hs`? Since it fails to build I cannot run integration tests or `cabal install`. So I run this:
```
cabal run swarm:exe:swarm -- --scenario Tutorials/Move --run <(echo 'move;move') --debug=all
```

byorgey:
> But the robots fail to run:

@xsebek what did you do to get that crash?  When I try it, the system robots in the move tutorial do not work properly, but at least it doesn't crash.  But I'm not sure if that's because I'm doing something differently than you, or if I incidentally fixed the crash in the meantime.

xsebek:
@byorgey I used import in Move tutorial thus:

<details>
  <summary>Click to expand diff</summary>

```diff
diff --git a/data/scenarios/Tutorials/move.yaml b/data/scenarios/Tutorials/move.yaml
index 5a20eb6e..ea7f35cc 100644
--- a/data/scenarios/Tutorials/move.yaml
+++ b/data/scenarios/Tutorials/move.yaml
@@ -149,54 +149,44 @@ robots:
   - name: 1P wall
     system: true
     program: |
-      def main = \a. pure noop end
       instant {
-        run "scenarios/Tutorials/move_surveil.sw";
+        import "scenarios/Tutorials/move_surveil.sw" in
         main [entity="wall", room=1]
       }
   - name: 2P wall
     system: true
     program: |
-      def main = \a. noop end
       instant {
-        run "scenarios/Tutorials/move_surveil.sw";
+        import "scenarios/Tutorials/move_surveil.sw" in
         main [entity="wall", room=2]
       }
   - name: 3P wall
     system: true
     program: |
-      def main = \a. noop end
-      instant {
-        run "scenarios/Tutorials/move_surveil.sw";
-        main [entity="wall", room=3]
-      }
+      import "scenarios/Tutorials/move_surveil.sw" in
+      main [entity="wall", room=3]
   #################
   ## GARDENERS   ##
   #################
   - name: 1P flower
     system: true
     program: |
-      def main = \a. noop end
       instant {
-        run "scenarios/Tutorials/move_surveil.sw";
+        import "scenarios/Tutorials/move_surveil.sw" in
         main [entity="flower", room=1]
       }
   - name: 2P flower
     system: true
     program: |
-      def main = \a. noop end
       instant {
-        run "scenarios/Tutorials/move_surveil.sw";
+        import "scenarios/Tutorials/move_surveil.sw" in
         main [entity="flower", room=2]
       }
   - name: 3P flower
     system: true
     program: |
-      def main = \a. noop end
-      instant {
-        run "scenarios/Tutorials/move_surveil.sw";
-        main [entity="flower", room=3]
-      }
+      import "scenarios/Tutorials/move_surveil.sw" in
+      main [entity="flower", room=3]
 entities:
   - name: door
     display:
```

</details>

Then run:
```bash
cabal run swarm:exe:swarm -- --scenario Tutorials/Move --run <(echo 'move;move') --debug=all --paused
```
After one step, you should see the robots all turn idle in the robot panel. Select any robot to view their log and you should see the "crash".

byorgey:
This is a weird one... the tests pass on Linux and Mac OS; on Windows, testing all the scenario solutions (including `Tutorials/move`) work fine, but we get a failure when testing for "no errors at loading time" saying that the `move` tutorial is missing.  Presumably this has something to do with the fact that the `move` tutorial is the only one that includes an `import`.  It says `import "blah.sw"`, and it is supposed to look for the `.sw` file in the same directory where the `.yaml` file was located, but maybe that fails for some reason...

byorgey:
Oops, caching environments does not work, since they can contain references to memory cells in the store, which are not cached (indeed, it doesn't really make sense to cache them).  I guess I should just take out the environment caching for now...

byorgey:
I added a few more scenarios that I switched to use `import`, which uncovered a few bugs (now fixed):

- The `farming` scenario used `import` in the `solution`, which wasn't being parsed in a way that it could correctly find relative imports.
- The `roadway` scenario revealed that caching environments would not work, as explained above.

I also created the very beginnings of a standard library in `data/lib/`, at least a standard library we can use in scenarios - not sure it makes sense to advertise or provide it to users.

byorgey:
I don't understand why the MacOS build is suddenly failing: https://github.com/swarm-game/swarm/actions/runs/20659588011/job/59319080169  I reran it twice with the same result.

data/scenarios/Vignettes/_roadway/coordinator.sw
```diff
@@ -1,4 +1,4 @@
-def forever : Cmd Unit -> Cmd Unit = \c. c ; forever c end
+import "../../../lib/control"
```
byorgey, line 1: Note, the `../../../` is annoying; I have already coded up a feature to allow writing things like `import "~swarm/lib/control"` to import something relative to Swarm's data directory, which I will include in a follow-up PR once this is merged.

src/swarm-doc/Swarm/Doc/Pedagogy.hs
```diff
@@ -100,19 +101,27 @@ extractCommandUsages :: Int -> ScenarioWith ScenarioPath -> TutorialInfo
...
-getDescCommands :: Scenario -> Set Const
+getDescCommands ::
+  forall phase.
+  ( Data (Anchor (ImportPhaseFor phase))
+  , Data (SwarmType phase)
+  , Typeable phase
+  , Typeable (ImportPhaseFor phase)
+  ) =>
+  Scenario phase ->
+  Set Const
```
xsebek: Does this need to be generic? I would assume swarm-doc only works with one phase.

src/swarm-engine/Swarm/Game/Step/Const.hs
```diff
@@ -339,17 +340,17 @@ execConst runChildProg c vs s k = do
       [VDir d] -> do
         when (isCardinal d) $ hasCapabilityFor COrient (TDir d)
         robotOrientation . _Just %= applyTurn d
-        use robotLocation >>= markDirty
+        use (robotLocation @Instantiated) >>= markDirty
```
xsebek: How about exporting a constrained type from another module? Something like List/Foldable in base:
```Haskell
module Robot.Generic where
robotLocation :: Lens' (Robot phase) Location
-- ------------------------------------------------
module Robot where
import Robot.Generic as G
type Robot = G.Robot Instantiated
robotLocation :: Lens' (Robot phase) Location
robotLocation = G. robotLocation
```

src/swarm-scenario/Swarm/Game/Robot.hs
```diff
@@ -182,74 +197,75 @@ type Robot = RobotR 'ConcreteRobot
...
 
 -- | The robot's current location, represented as @(x,y)@.  This is only
 --   a getter, since when changing a robot's location we must remember
 --   to update the 'Swarm.Game.State.robotsByLocation' map as well.  You can use the
 --   'Swarm.Game.Step.updateRobotLocation' function for this purpose.
-robotLocation :: Getter Robot (Cosmic Location)
+--
+--   This lens is actually used at multiple types; see Note [Robot
+--   lens types].
+robotLocation :: Getter (Robot phase) (RobotLocation phase)
```
xsebek: How about `grobot` prefix for generic? The location and inventory are used a lot with Instantiated robots.

I said in previous comment that I would prefer a separate module with the constrained types, but I can understand it is a chore to keep them synced.

xsebek (commented):
Great* work @byorgey!

I have gone through the first 100 files out of 168, so about 60% done. 😄 

The import process we have already gone over in the caching PR and talked about it on a call, so no surprises there. 

Most of the modules are just adding the phase parameter. A few robot lenses seem to cause most of the noise in Step/Const, so maybe those would be worth specializing.

I will finish the rest when I get home, but I don't expect I will find much to nitpick. 😄

<sub>*) So great in fact, that it froze the new review UI. I tried writing a comment and it took 10 seconds for each character. Switching to the old one worked, though it collapsed almost every file. 😄 </sub>

src/swarm-topography/Swarm/Game/Scenario/Topography/Structure/Assembly.hs
```diff
@@ -148,12 +151,24 @@ mergeStructures graph = mergedMap
...
     toPlace = pathPlacements ann
-    toPlaceMerged = map (\(PathPlacement p pose) -> (mergedMap ML.! p, pose)) toPlace
+    toPlaceMerged = map (\(PathPlacement p pose) -> (mergedMap ML.!? p, pose)) toPlace
+
+    -- In order to avoid using a partial map lookup function in toPlaceMerged, we use
+    -- this variant of overlaySingleStructure which does nothing if the first component
+    -- of the pair to be merged is Nothing.  However, if we ever actually get Nothing
+    -- here, it is a bug!
+    overlaySingleStructure' ms (Nothing, _) = ms
+    overlaySingleStructure' ms (Just ms', p) = overlaySingleStructure ms (ms', p)
```
xsebek: How would we find about such bug?

I think we can be relatively sure that this function works _iff we can detect errors._ So this pattern might be acceptable:
```Haskell
mergedMap ML.!? p
  ?? error ("Could not find p " <> show p <> " in graph while merging structures!")
```
My reasoning being, that players will not run into failure because we already tested all scenarios. If they create their own scenario and manage to trigger this path then at least they can report it instead of tearing their hair out with silent errors.

test/unit/TestLanguagePipeline.hs
```diff
@@ -741,7 +744,7 @@ testLanguagePipeline =
             "missing end"
             ( process
                 "def x = 3;\n def y = 3 end;\n def z = 3 end"
-                "3:15:\n  |\n3 |  def z = 3 end\n  |               ^\nunexpected end of input\nexpecting \"!=\", \"&&\", \"()\", \"++\", \"<=\", \"==\", \">=\", \"def\", \"false\", \"let\", \"require\", \"requirements\", \"stock\", \"true\", \"tydef\", \"||\", '\"', '$', '(', '*', '+', '-', '.', '/', ':', ';', '<', '>', '@', '[', '\\', '^', 'end' keyword for definition of 'x', '{', built-in user function, direction constant, integer literal, or variable name\n"
+                "3:15:\n  |\n3 |  def z = 3 end\n  |               ^\nunexpected end of input\nexpecting \"!=\", \"&&\", \"()\", \"++\", \"<=\", \"==\", \">=\", \"def\", \"false\", \"import\", \"let\", \"require\", \"requirements\", \"stock\", \"true\", \"tydef\", \"||\", '\"', '$', '(', '*', '+', '-', '.', '/', ':', ';', '<', '>', '@', '[', '\\', '^', 'end' keyword for definition of 'x', '{', built-in user function, direction constant, integer literal, or variable name\n"
```
xsebek, line 747: As a side note, that _"'end' keyword for definition of 'x'"_ is so far back I never noticed it before. 😄

src/swarm-util/Swarm/Language/Syntax/Import.hs
```diff
@@ -0,0 +1,402 @@
...
+{-# LANGUAGE PatternSynonyms #-}
+{-# LANGUAGE TypeData #-}
+{-# LANGUAGE TypeFamilyDependencies #-}
+{-# LANGUAGE UndecidableInstances #-}
+
+-- |
+-- SPDX-License-Identifier: BSD-3-Clause
+--
+-- Data types to represent Swarm-lang import locations.
+module Swarm.Language.Syntax.Import (
```
xsebek, line 15: As we discussed, I think it is a bit unfortunate to have Language module in utils.

Perhaps it would be better to have an equivalent code like:
```Haskell
-- UTIL
type SomeImportFailure = Text
-- or
data SomeImportFailure = InternalSomeImportFailure Text
```
```Haskell
-- LANGUAGE
someImportFailure :: ImportLoc Raw -> SomeImportFailure
fromImportFailure :: SomeImportFailure -> ImportLoc Raw
```

xsebek (approved):
OK, all checked! 👀 

I have a few comments, but nothing that would preclude merging. 🙂 

I look forward to helping with the follow up issues. 👍

data/test/import/f.sw
```diff
@@ -0,0 +1,3 @@
+import "g"
+
+def f : Int = g + 7 end
```
xsebek, line 3: What is the idea behind these tests? 🤔 
```mermaid
flowchart TD
    A
    B --> C
    C
    D --> E
    E
    F --> G
    G
```
I wanted to see if there is a test for #2659.

src/swarm-doc/Swarm/Doc/Pedagogy.hs
byorgey: Good point, I've monomorphized it now.

test/unit/TestLanguagePipeline.hs
byorgey: Yeah, we can probably do something to cut down on the length of these parse errors.  e.g. give names to more things so we get helpful names for categories of things instead of just a giant list of possible tokens.

byorgey: https://github.com/swarm-game/swarm/issues/2676

data/test/import/f.sw
byorgey: They are for testing recursive imports: https://github.com/swarm-game/swarm/blob/b903def261c9c76bb0c811d345bd721ed5676cad/test/unit/TestLanguagePipeline.hs#L765-L785

byorgey: It does test #2659, for example in making sure that `import f; pure (f + g)` is an error, since `g` is in scope in `f.sw` but not re-exported.

src/swarm-topography/Swarm/Game/Scenario/Topography/Structure/Assembly.hs
byorgey: Good point, I hadn't fully realized that this code will only be run while validating scenarios, not during gameplay.

src/swarm-util/Swarm/Language/Syntax/Import.hs
byorgey: So `fromImportFailure` would have to parse an import location from `Text`?  That sounds error-prone and fragile...

I think I am going to leave it as it is for now, but create an issue for it: #2677 .

PR #2543, opened 2025-07-29
Update tab completion to work at the cursor, not just at the end of REPL input

### LINKED ISSUE #2542: Tab completion should work in the middle of the input
Right now it seems like tab completion only works when the thing being completed is at the very end of the current REPL input.  This is particularly annoying now that we auto-insert closing parens and curly braces.  e.g. if you type an open paren, then start typing a variable name or entity name etc. and try to tab-complete it, it does not work.

Tab completion should operate on the string prior to the cursor, not on the end of the REPL input.

byorgey:
Update tab completion so it uses the REPL editor zipper to complete text just before the cursor, rather than text at the end.  Closes #2542.

### DIFF
--- a/src/swarm-tui/Swarm/TUI/Model/Repl.hs
+++ b/src/swarm-tui/Swarm/TUI/Model/Repl.hs
@@ -40,6 +40,8 @@ module Swarm.TUI.Model.Repl (
   replPromptType,
   replPromptEditor,
   replPromptText,
+  replPromptZipper,
+  replPromptTextBeforeCursor,
   replValid,
   replLast,
   replType,
@@ -53,7 +55,7 @@ module Swarm.TUI.Model.Repl (
   lastEntry,
 ) where
 
-import Brick.Widgets.Edit (Editor, applyEdit, editorText, getEditContents)
+import Brick.Widgets.Edit (Editor, applyEdit, editContentsL, editorText, getEditContents)
 import Control.Applicative ((<|>))
 import Control.Lens hiding (from, (.=), (<.>))
 import Data.Aeson (FromJSON, ToJSON, Value (..), object, parseJSON, toJSON, (.:), (.=))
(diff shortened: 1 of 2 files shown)


src/swarm-util/Swarm/Util.hs
```diff
@@ -325,6 +326,18 @@ unsnocNE (x :| xs) = go x xs
...
+--
+-- >>> iterateN 4 succ 'a'
+-- 'e'
+-- >>> iterateN 0 succ 'a'
+-- 'a'
+-- >>> iterateN 5 (*2) 1
+-- 32
+iterateN :: Int -> (a -> a) -> a -> a
+iterateN 0 _ = id
+iterateN n f = iterateN (n - 1) f . f
```
xsebek: This could cause confusion - containers use this name to construct a Seq.

How about applyN or nTimes?
https://hoogle.haskell.org/?hoogle=Int%20-%3E%20(a%20-%3E%20a)%20-%3E%20a%20-%3E%20a

src/swarm-tui/Swarm/TUI/Model/Repl.hs
```diff
@@ -379,9 +381,18 @@ replPromptEditor :: Lens' REPLState (Editor Text Name)
 replPromptText :: Lens' REPLState Text
 replPromptText = lens g s
  where
-  g r = r ^. replPromptEditor . to getEditContents . to T.concat
+  g r = r ^. replPromptEditor . to getEditContents . to T.unlines
```
xsebek, line 384: Is the unlines unused currently since the REPL is always one line?

xsebek (approved):

src/swarm-tui/Swarm/TUI/Model/Repl.hs
byorgey: Yes, that's true.  But since the zipper returns a list of lines, `T.unlines` is really the correct thing to do (I noticed that it was `T.concat`; I think previously I misunderstood what the return type of `getEditContents` represented).

PR #2547, opened 2025-07-30
Auto-insert matching close quotes

### LINKED ISSUE #2545: Auto-insert matching double quotes just like parens
Currently, when the user types a left delimeter (paren, curly brace, or square bracket), we automatically insert a matching right delimeter.  It would be really nice to do something similar for double quotes.  Of course, this is slightly trickier since double quotes do not come in "left" and "right" varieties, so it seems that we might have to work harder to decide when to do this.  However, I'm not sure that we do.  A proposed approach is as follows.  When the user types a double quote:

1. If he cursor is positioned just before a double quote, do not insert any quote character and simply move the cursor to the right one position.
2. Otherwise, insert two double quote characters and position the cursor in between them.

The only potentially annoying detail of this approach is that if the user deletes the ending double quote from a string, adds more characters, then tries to close the string with a double quote, two double quote characters will be entered, requiring the user to backspace over one of them.  If we wanted to be slightly more sophisticated, we could add an additional rule between rules (1) and (2) above:

- If there are currently an odd number of double quote characters before the cursor, then insert only a single double quote character.  (This is similar to what we currently do to determine how to do tab-completion.)

byorgey:
Closes #2545. Implements the three rules described there, i.e. when a double quote character is typed:

1. If the cursor is positioned just before a double quote, do not insert a new quote character; simply move the cursor to the right one position.
2. If there are currently an odd number of double quote characters before the cursor, then insert only a single double quote character.
3. Otherwise, insert two double quote characters and position the cursor in between them.

### DIFF
--- a/src/swarm-tui/Swarm/TUI/Controller.hs
+++ b/src/swarm-tui/Swarm/TUI/Controller.hs
@@ -738,6 +738,18 @@ handleREPLEventTyping m = \case
           CharKey c
             | c `elem` ("([{" :: String) -> insertMatchingPair c
             | c `elem` (")]}" :: String) -> insertOrMovePast c
+            | c == '"' -> do
+                tz <- use editContentsL
+                case TZ.currentChar tz of
+                  -- If currently on a quote, just move past it
+                  Just '"' -> modify . applyEdit $ TZ.moveRight
+                  _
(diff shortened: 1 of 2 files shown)


xsebek (approved):
So that's what the previous PR was building infrastructure for, neat!

Issue #2270, opened 2025-01-05
`instant` seems to have no effect when wrapped around definitions


byorgey:
**Describe the bug**
`instant`, when wrapped around a term that includes some expensive `def`s, seems to have no effect.

**To Reproduce**
In the below scenario, the judge does an expensive computation (naive `fib 10`), then places a `rock`. The base robot just executes `move; grab`.  Since the entire judge program is wrapped in `instant`, this ought to work: the entire judge program including the expensive computation and the `place` command should all execute in a single tick, placing the `rock` in time for the `base` to `grab` it.  However, the base crashes with a "nothing here to grab" error, and the rock shows up later.

```
version: 1
name: Instant wrapped defs
description: |
  `instant` should work when wrapped around definitions
creative: false
objectives:
  - goal:
      - Grab the rock
    condition: |
      as base { has "rock" }
robots:
  - name: base
    dir: east
    devices:
      - logger
      - treads
      - grabber
  - name: judge
    dir: east
    system: true
    program: |
      instant (
      def fib : Int -> Int = \n.
        if (n <= 1) {n} {fib (n-1) + fib (n-2)}
      end

      def x = fib 10 end

      create "rock"; place "rock"
      )
solution: |
  move; grab
known: [rock]
world:
  dsl: |
    {grass}
  palette:
    'B': [grass, null, base]
    'j': [grass, null, judge]
  upperleft: [0, 0]
  map: |
    Bj
```

**Additional context**
Ran into this while preparing #2260.  See the discussion here: https://github.com/swarm-game/swarm/pull/2260#discussion_r1903359926

byorgey:
Ohhhh, I understand why this happens: it's because Swarm is a strict language, so the argument to `instant` is evaluated first, before passing it to `instant`!  In this case, evaluating the argument includes evaluating the values of the definitions.

So this is not exactly a bug, but it is confusing, and we might want to consider special-casing it somehow...

PR #2552, opened 2025-07-30
Give `instant` and `atomic` types `{Cmd a} -> Cmd a`

### LINKED ISSUE #2521: Change type of `instant/atomic` to `{Cmd a} -> Cmd a`
In #2270 we had an issue because the argument to `instant` was evaluated strictly, which led to unexpected results --- in particular, the CESK machine spent a while first evaluating the argument non-atomically before finally executing it atomically.  In #2271 we solved this by adding a special case for `atomic/instant` to also evaluate the argument atomically.

If we just make the type of `atomic`/`instant` to `{Cmd a} -> Cmd a` instead of `Cmd a -> Cmd a`, it should solve the problem without needing a special case.

byorgey:
This is an alternative, better solution to #2270, which allows us to delete the special-case code added in #2271.  By delaying the argument to `instant` and `atomic`, it will be evaluated at execution time (and thus happen atomically) rather than before.  Closes #2521.

### DIFF
--- a/data/entities.yaml
+++ b/data/entities.yaml
@@ -546,8 +546,12 @@
     attr: entity
     char: 'O'
   description:
-    - A polished, rounded piece of glass, capable of focusing rays of light.
+    - |
+      A polished, rounded piece of glass, capable of focusing rays of light.
+    - |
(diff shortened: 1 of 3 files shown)


xsebek (approved):
Nice 👍

PR #2555, opened 2025-07-30
Load world tiles in parallel

xsebek:
* Make tile size 2**5=32
* Load world tiles in parallel using [Control.Parallel.Strategies](https://hackage-content.haskell.org/package/parallel-3.2.2.0/docs/Control-Parallel-Strategies.html#v:parMap)
* [ ] TODO: Add metrics
* [ ] TODO: Add `-N[x]` to cabal file 

Tested with
```sh
cabal run swarm -- --debug=all --scenario Creative +RTS -s -N10  # try setting number of threads
```

### DIFF
--- a/swarm.cabal
+++ b/swarm.cabal
@@ -294,6 +294,9 @@ common pandoc
 common pandoc-types
   build-depends: pandoc-types >=1.23 && <1.24
 
+common parallel
+  build-depends: parallel >=3.2 && <3.3
+
 common parser-combinators
   build-depends: parser-combinators >=1.2 && <1.4
 
@@ -608,6 +611,7 @@ library swarm-scenario
     mtl,
     murmur3,
     palette,
+    parallel,
     parser-combinators,
     prettyprinter,
     random,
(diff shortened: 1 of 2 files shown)


xsebek:
> - [ ] Add metrics

I tried to add metrics and actually measure how long does it take us to load tiles. There were a couple of blockers, which made this task bigger than I initially expected:

* Some operations used the pure functions and were not caching tiles. Adding effects was not trivial and once I gave up and used `IO` for one function (`mtlEntityAt`), suddenly half of step functions needed `IO` transitively.
  * I also wrote "mock effects" for tests, so the tests and benchmarks don't need `IO` if they are not using metrics.
* I was getting lost at which pure world functions were missing stateful caching counterparts, so I decided to split the module to `World` folder and re-export. But there it got mixed up with world DSL. So I moved that into its own folder as well.
* The world tiles cache is _lazy_. The world state also seemed lazy?! It took me a while to force the computation and get believable metrics.

@byorgey I will split out smaller PRs from this one and get it back to one file change, don't worry. 😄

swarm.cabal
```diff
@@ -1198,8 +1202,9 @@ executable swarm
     -hiedir=.hie/app/game
 
   default-language: Haskell2010
-  ghc-options: -threaded
-  default-extensions: ImportQualifiedPost
```
byorgey, line 1202: Why is `default-extensions: ImportQualifiedPost` removed?

xsebek: It's already in `common`.

PR #2564, opened 2025-08-23
Grant achievement for unequipping a welder

### LINKED ISSUE #2288: Achievement for `unequip "welder"` (or make it impossible)
It is possible to `unequip "welder"`, but after that it is not possible to equip it again.  We should either make this impossible, or perhaps give an achievement for doing it.

byorgey:
Closes #2288.

### DIFF
--- a/src/swarm-engine/Swarm/Game/Achievement/Description.hs
+++ b/src/swarm-engine/Swarm/Game/Achievement/Description.hs
@@ -110,6 +110,13 @@ describe = \case
       "`equip` all craftable devices simultaneously."
       Gruelling
       True
+  GameplayAchievement UnequippedWelder ->
+    AchievementInfo
+      "Put That Back!"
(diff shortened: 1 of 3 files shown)


xsebek (approved):
Nice. 😆 How did you find out about it? Is it ever useful to install it on a robot? 🤔

byorgey:
I don't remember when or how I first realized this.  I mean, I filed #2288 back in January but I don't remember what I was thinking about at the time.  I just remember realizing that it was possible to get "stuck" if you used your `welder` to `unequip` itself and then you couldn't re-`equip` it (or anything).

It is possible to install a `welder` on a robot, and I could imagine it being useful, if *e.g.* you programmed a robot to bootstrap itself by going to fetch some device from wherever it is being produced and then `equip` the device on itself.  Or if you have only a single device of some sort and need to pass it around from robot to robot for some reason.  But I imagine most that get this achievement will do it while playing around with `equip` and `unequip` on the base robot.

PR #2568, opened 2025-08-25
Tear the World asunder

xsebek:
* split World into smaller modules and re-export
* also re-export `Coords` for convenience
* split from #2555 
* depends on #2567 

The idea is to have Tile and Function modules separate from World, instead of having to scroll past them.

Then World functions get split into a Pure interface and a monadic/stateful one that should match (some are missing) and be preferred as it will implicitly use cache (it is currently not very used). 

In the next PR, the monadic interface will be improved with metrics...


src/swarm-scenario/Swarm/Game/World/Pure.hs
```diff
@@ -0,0 +1,145 @@
...
+
+-- | Update the entity (or absence thereof) at a certain location,
+--   returning an updated 'World' and a Boolean indicating whether
+--   the update changed the entity here.
+--   See also 'updateM'.
+update ::
+  Coords ->
+  (Maybe Entity -> Maybe Entity) ->
+  World t Entity ->
+  (World t Entity, CellUpdate Entity)
```
xsebek, line 108: One thing that I noticed with the split is that while most functions are polymorphic in their arguments, `update` does not bother.

Now if `hash` returned the cached hash, we could make it any hashable.

But I would even consider monomorphising all the functions to use entity. The type checker needs to be nudged on every step and it makes the type errors harder to decipher.

Also Tile ID should be a (new)type, when I read `MultiWorld Int e` I had no idea what it was supposed to do.

But that is another diversion from adding metrics. 😅

xsebek:
I realize now I may have used too ominous tone when writing the PR description. 😄 But in reality this should be a small harmless refactoring. I copied the files first to preserve git history then trimmed them down.

byorgey (approved):

src/swarm-scenario/Swarm/Game/World/Pure.hs
byorgey: Agreed!

Issue #2158, opened 2024-09-28
Make sure Haddocks are up-to-date


byorgey:
> I look forward to seeing the haddock in the next release.

That reminds me, I haven't looked at the Haddocks in a long time, probably worth a pass at some point to make sure things are still well-explained, and putting more prominent links to the built Haddock documentation in our materials about "how to contribute".

_Originally posted by @byorgey in https://github.com/swarm-game/swarm/issues/2155#issuecomment-2380889866_

PR #2571, opened 2025-08-26
Add links to module guide + Haddocks in CONTRIBUTING

byorgey:
Towards #2158.


xsebek (approved):
We should publish the "latest main" documentation from build artifacts. 😅

Then we could more easily see where it needs to be improved.

byorgey:
> We should publish the "latest main" documentation from build artifacts. 😅
> 
> Then we could more easily see where it needs to be improved.

Yes, we should: #281 :smile:

PR #2574, opened 2025-09-04
Handle metrics port bind in second swarm

### LINKED ISSUE #2539: Metrics throw exception in second swarm
**Describe the bug**
Running second swarm throws an exception because of used metrics port.
**To Reproduce**
Running swarm twice.

**Expected behavior**
Handle the exception, turn off metrics.

xsebek:
* Closes #2539

Compare Messages in first swarm:
```
[Metrics API] started on :6543
```
In second:
```
[Metrics API] Can not start on port '6543': Network.Socket.bind: resource busy (Address already in use)
```


app/game/Swarm/App.hs
```diff
@@ -135,9 +138,18 @@ defaultMetrics = 6543
...
 startMetricsThread (Just 0) _ = pure $ Left "Metrics API disabled."
 startMetricsThread mPort store = do
-  let p = fromMaybe 6543 mPort
-  _ <- WaiMetrics.forkServerWith store "localhost" p
-  pure $ Right p
+  let port = fromMaybe 6543 mPort
+  portCheck <- checkPortFree port
+  print portCheck
+  case portCheck of
+    Right () -> Right port <$ WaiMetrics.forkServerWith store "localhost" port
```
xsebek, line 144: I wanted to `catch` exception from the metrics server. But that does not work because it does `forkFinally` and then if the port binding fails in another thread, it uses `throwTo` to throw an _asynchronous exception_ to the main thread. The main thread has meanwhile exited the catch block.

byorgey: Sure.  Technically, of course, there is a race condition here but I don't think it matters.

xsebek:
@mergifyio requeue

PR #2581, opened 2025-09-13
Track world metrics in all stateful functions

xsebek:
* add metric effect to lookup and update functions
* replace MTL style entity lookup with fused-effects version
  * add fake time effect for tests
* refactor the robot metrics to use the new metric fused-effect
* refactor copy-paste zoom* functions to call one function that accepts a lens
* split from #2555

This is the last part that I wanted to split off from #2555. I rewrote most of it and I think I arrived at a cleaner solution for the MTL function and the signatures of Step functions.

### DIFF
--- a/src/swarm-engine/Swarm/Game/Step/Path/Finding.hs
+++ b/src/swarm-engine/Swarm/Game/Step/Path/Finding.hs
@@ -25,7 +25,6 @@
 -- distance to prevent programming errors from irrecoverably freezing the game.
 module Swarm.Game.Step.Path.Finding where
 
-import Control.Carrier.State.Lazy
 import Control.Effect.Lens
 import Control.Lens ((^.))
 import Control.Monad (filterM, guard)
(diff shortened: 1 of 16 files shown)


src/swarm-engine/Swarm/Game/Step/Path/Finding.hs
```diff
@@ -145,7 +147,8 @@ pathCommand parms = do
   goalReachedFunc loc = case target of
     LocationTarget gLoc -> return $ loc == gLoc
     EntityTarget eName -> do
-      me <- entityAt $ Cosmic currentSubworld loc
+      worlds <- Fused.use $ landscape . multiWorld
+      let me = lookupCosmicEntity (Cosmic currentSubworld $ locToCoords loc) worlds
```
xsebek, line 151: I gave up and used pure functions for path finding, which seems to have fixed the issue.

PR #2582, opened 2025-09-21
Have inventory search look at entity description in addition to names

### LINKED ISSUE #1879: Search within entity descriptions as well as titles
In the inventory panel, we can enter fuzzy search mode by hitting forward slash.  This is cool, but sometimes you can't remember the name of the entity you are looking for.  Maybe you only remember something about its description.  We should consider searching within item descriptions as well as titles.  I don't know if this is a good idea or not but it seems worth considering.

byorgey:
Closes #1879 .

At first I tried doing fuzzy search within entity descriptions, just like we currently do for searching in entity names, but subjectively that yielded way too many false positives.  The letters of any word you might type in could be found somewhere in almost all entity descriptions.  Doing fuzzy search within names but literal substring search within descriptions subjectively seems to yield useful results.

### DIFF
--- a/src/swarm-scenario/Swarm/Game/Entity.hs
+++ b/src/swarm-scenario/Swarm/Game/Entity.hs
@@ -671,8 +671,7 @@ entityNameFor _ = to $ \e ->
     Just pl -> pl
     Nothing -> plural (e ^. entityName)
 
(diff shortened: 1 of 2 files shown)


xsebek (approved):
Neat, thanks for updating the docs as well. (I think I might have changed it to use markdown.)

Issue #1764, opened 2024-02-12
Support jump-to-definition


byorgey:
**Is your feature request related to a problem? Please describe.**
I have emacs configured so that `Alt-.` jumps to the definition of the name under the cursor.  It would be cool if this worked in Disco too.

**Describe the solution you'd like**
Specifically, I'd like the LSP server to support the `textDocument/definition` capability, see https://microsoft.github.io/language-server-protocol/specifications/specification-3-14/#textDocument_definition .  Since we already keep track of position information while parsing, I imagine this should not be all that hard, although there very likely may be difficulties I don't know about.

oliverpauffley:
Im interested in picking this up and maybe some of the other LSP related ones in #979. Will have a go this week and at Zurihac

PR #2583, opened 2025-09-28
Add support for jump to definition

oliverpauffley:
Partially addresses #1764. We are still missing the ability to jump to type level definitions since we don't have locations for these yet.

- Adds an lsp handler to support jump to value level definitions.
- Moves some shared lsp code from `Hover` to `Position` since it seemed to be shared across more than hover support.

### DIFF
--- a/src/swarm-lang/Swarm/Language/LSP/Definition.hs
+++ b/src/swarm-lang/Swarm/Language/LSP/Definition.hs
@@ -66,7 +66,7 @@ boundVars (Syntax' _ t _ _) = case t of
   STydef {} -> mempty
   SApp {} -> mempty
   SPair {} -> mempty
-  SDelay s -> mempty
+  SDelay {} -> mempty
   SRcd {} -> mempty
   SProj {} -> mempty
   SAnnotate {} -> mempty


src/swarm-lang/Swarm/Language/LSP/Definition.hs
```diff
@@ -0,0 +1,117 @@
...
+import System.IO (stderr)
+
+data DefinitionResult = TError ContextualTypeErr | PError ParserError | Unsupported | NotFound | Found [J.Range]
+
+findDefinition ::
+  J.NormalizedUri ->
+  J.Position ->
+  VirtualFile ->
+  DefinitionResult
+findDefinition _ p vf@(VirtualFile _ _ myRope) =
```
oliverpauffley, line 29: I'm not super happy with this function. It seems to be doing a lot and the multiple wheres don't seem that right to me. But it works so maybe ok?

src/swarm-lang/Swarm/Language/LSP/Definition.hs
```diff
@@ -0,0 +1,105 @@
...
+usageName :: Syntax' a -> Maybe Var
+usageName (Syntax' _ t _ _) = case t of
+  (TVar n) -> Just n
+  _ -> Nothing
+
+syntaxVars :: Syntax' a -> [(SrcLoc, Var)]
+syntaxVars (Syntax' _ t _ _) = case t of
+  (SLet _ _ lv _ _ _ _ _) -> [lvToLoc lv]
+  (STydef lv _ _ _) -> [(lvSrcLoc lv, tdVarName $ locVal lv)]
+  (SApp s1 _) -> syntaxVars s1
```
byorgey: I may be misunderstanding how this works, but it seems to me that `syntaxVars` should not call itself recursively on subterms in the `SApp`, `SPair`, and `SDelay` cases.  Indeed, it seems to cause weird behavior if I write something like
```
def f : Int -> (Int * (Int -> Int)) = \x. (x, \y. y) end
```
and then try to jump to the definition of the `y`: it shows *two* definition sites, but it's just the same one twice.

The idea is that when processing syntax like `def f = \x. (x, \y.y) end` for the position of the final `y`, we first generate a path of subterms to that position: `[y,  \y.y,   (x, \y.y),   \x. (x, \y.y),    def f = ... end]`  By the time we get to a pair like `(x, \y.y)`, we have already seen the relevant subterm when going up the path, so there's no need to process it again.  All we are doing is going up the path and looking for places where variables are bound/introduced: lambdas, `def`/`let`, bind.

So for `SApp`, `SPair`, `SDelay` I think `syntaxVars` should simply return an empty list.

I actually think it should also return an empty list in the `SRcd` case, because record literals do not actually bind variables: in `def f = \x. [x = x]`, the last `x` actually refers to the lambda, not to the `x =` in the record.

byorgey: A better name for `syntaxVars` might therefore be `boundVars`, since it is not extracting all variables but just the places where they are defined/bound.

src/swarm-lang/Swarm/Language/LSP/Definition.hs
```diff
@@ -0,0 +1,105 @@
...
+    | name == name' = P.posToRange myRope pos
+    | otherwise = Nothing
+
+-- | find the name of the syntax element if it is a value level variable
+usageName :: Syntax' a -> Maybe Var
+usageName (Syntax' _ t _ _) = case t of
+  (TVar n) -> Just n
+  _ -> Nothing
+
+syntaxVars :: Syntax' a -> [(SrcLoc, Var)]
```
byorgey: I think maybe `syntaxVars` ought to have a return type of `Maybe (SrcLoc, Var)` rather than a list.  Any syntax node binds at most one variable.

src/swarm-lang/Swarm/Language/LSP/Definition.hs
```diff
@@ -0,0 +1,105 @@
...
+      Left e -> TError e
+      Right pt -> do
+        let path = P.pathToPosition pt $ fromIntegral absolutePos
+        let usage = usageName $ NE.last path
+
+        case usage of
+          Nothing -> Unsupported
+          Just u -> do
+            let pathTerms = concatMap syntaxVars (NE.drop 1 . NE.reverse $ path)
+            case mapMaybe (maybeDefPosition u) pathTerms of
```
byorgey: I think we should take the *first* binding site with the correct name we find along the path, rather than all of them.  For example, if I write
```
def f : Int -> Int = \x.
  x + ((\x. x + 2) 3)
end
```
and then do jump-to-def on the `x` in `x+2`, it shows me *two* potential definition sites, one for each lambda.  But this is incorrect, since the second lambda shadows the first.  The `x` in `x+2` is unambiguously defined by the nested lambda.

src/swarm-lang/Swarm/Language/LSP/Definition.hs
byorgey: It's probably possible to refactor it to look nicer, but nothing obvious comes to me off the top of my head, and it's not really that big of a deal.  We can refactor it later if we want.

src/swarm-lang/Swarm/Language/LSP/Definition.hs
oliverpauffley: Yeah that makes sense

src/swarm-lang/Swarm/Language/LSP/Definition.hs
oliverpauffley: Yeah that makes a lot of sense. I took a little too much inspiration from the usage code but the case is different here. Makes it much easier too

src/swarm-lang/Swarm/Language/LSP/Definition.hs
```diff
@@ -0,0 +1,103 @@
...
+      Left e -> TError e
+      Right pt -> do
+        let path = P.pathToPosition pt $ fromIntegral absolutePos
+        let usage = usageName $ NE.last path
+
+        case usage of
+          Nothing -> Unsupported
+          Just u -> do
+            let pathTerms = mapMaybe boundVars (NE.drop 1 . NE.reverse $ path)
+            case mapMaybe (maybeDefPosition u) pathTerms of
```
byorgey: Instead of `mapMaybe` followed by `head`, you should be able to `find (maybeDefPosition u)`.

oliverpauffley: Yeah I changed it up a bit.

Interesting haskell thing of having a line that is doing a *lot*. I understand what it does but must take some brain power to read

byorgey (approved):
Looks great!

PR #2584, opened 2025-10-04
Change `SRcd` to contain a list of tuples instead of a `Map`

byorgey:
It makes sense for, say, a record type to consist of a `Map` from field names to types.  However, we were also storing record literal AST nodes using a `Map`, but after a question from @oliverpauffley I realized this is probably not the right thing to do.  For one thing, the AST should store the fields in the same order as they were written (otherwise pretty-printing would always sort record fields alphabetically).  For another, the field names should be stored as `LocVar`s, not just `Var`s, so we know where they were in the source.


byorgey:
@oliverpauffley what do you think?  Will this help?

PR #2588, opened 2025-10-10
Cache imports

xsebek:
Adds a global cache for checked + elaborated modules, keyed by canonical import location, so we only do the work of loading + checking a given import once (until it changes).

### DIFF
--- a/src/swarm-engine/Swarm/Game/CESK.hs
+++ b/src/swarm-engine/Swarm/Game/CESK.hs
@@ -87,7 +87,6 @@ import Control.Lens (Lens', Traversal', lens, traversal, (^.))
 import Data.Aeson (FromJSON (..), ToJSON (..), genericParseJSON, genericToJSON)
 import Data.IntMap.Strict (IntMap)
 import Data.IntMap.Strict qualified as IM
-import Data.Set (Set)
 import Data.Strict.Tuple (Pair)
 import GHC.Generics (Generic)
 import Prettyprinter (Doc, Pretty (..), encloseSep, hsep, (<+>))
@@ -101,7 +100,6 @@ import Swarm.Language.Module (Module, moduleTerm)
 import Swarm.Language.Requirements.Type (Requirements)
 import Swarm.Language.Syntax
 import Swarm.Language.Syntax.Import qualified as Import
-import Swarm.Language.TDVar (TDVar)
 import Swarm.Language.Types
 import Swarm.Language.Value as V
 import Swarm.Pretty (PrettyPrec (..), pparens, ppr)


src/swarm-util/Swarm/Language/InternCache.hs
```diff
@@ -0,0 +1,80 @@
+{-# LANGUAGE BlockArguments #-}
+module Swarm.Language.InternCache
```
byorgey: Should add a comment with a link to the source, https://chrispenner.ca/posts/intern-cache

src/swarm-lang/Swarm/Language/Load.hs
```diff
@@ -217,33 +229,43 @@ validateImport loc = maybe (pure ()) validate . moduleTerm
     TConst Noop -> pure ()
     t -> throwError $ ImpureImport loc (prettyText t)
 
+-- | Cache file contents for import location.
+fileCache :: (MonadIO m) => InternCache m (ImportLoc Import.Resolved) Text
```
byorgey: Why do we need to cache file contents separately from caching processed modules?

src/swarm-lang/Swarm/Language/Load.hs
```diff
@@ -170,31 +180,33 @@ resolveImport parent loc = do
   -- Compute the canonicalized location for the import, and record it
   canonicalLoc <- resolveImportLoc (unresolveImportDir parent <//> loc)
   add $ S.singleton canonicalLoc
+  
+  sendIO (IC.lookupCached moduleCache canonicalLoc) >>= \case
```
byorgey: The big problem that I can see is that this will never consider reloading a file from disk.  In other words, consider the following steps:

1. Start Swarm.
2. Load a scenario that imports `foo/bar.sw`.
3. Edit `foo/bar.sw`.
4. Reload the scenario.

You would like to get the new version of `foo/bar.sw`, of course, but you will get the old, cached version.  To get the new version you would have to completely restart Swarm.

One solution to this I have considered is to add extra data to import locations that allow us to uniquely identify them in time as well as space: a "last modified" timestamp in the case of local files, and a content hash in the case of remote files.

However, there also needs to be a mechanism to identify when we are willing to update.  For example, if we are loading a scenario and encounter `import https://foo` seven times while recursively loading imports, we should only fetch it over the network the first time --- even if (*especially* if!) the remote file changes while we are still in the process of loading.  However, if we then restart the scenario, we have to fetch it over the network again to see whether it changed.

src/swarm-lang/Swarm/Language/Load.hs
xsebek: Well, parsing the file could fail, so in that case we could reuse a little bit.

But mainly it was easier for me to write as a first test. 😁

src/swarm-lang/Swarm/Language/Load.hs
xsebek: Yes, I was thinking that "last modified time" would need to be tracked as well, but I was not sure how exactly.

Here is my proposal:
* add the timestamp info _only to the cache_, so the import location data stay the same
* when loading file use `getModificationTime` and compare it with the time in cache
  * if they match, reuse the value in cache
  * if not, overwrite the value in cache - the previous value will never be useful again

I have not given network files much thought, do they have a timestamp we could compare?
Personally, I would not mind if they required a Swarm restart to update.

byorgey: > I have not given network files much thought, do they have a timestamp we could compare?
> Personally, I would not mind if they required a Swarm restart to update.

Oh, good point!  I like this, it's much simpler.  It seems reasonable that files fetched over the network are much less likely to change (and if they do, I might not even be expecting it).  Hence it makes sense to never invalidate them from the cache.

As for files on disk, I like your proposal.  I still think to be 100% correct we need to account for the fact that while, say, recursively loading all the code for a given scenario, we do not want to update a cached module even if it changes on disk in the middle of loading --- because all the imports of a given module should be consistent across a given scenario.  However, I think we can make an issue for this / put it off until later.  It won't happen very frequently.

src/swarm-util/Swarm/Util/InternCache.hs
```diff
@@ -0,0 +1,56 @@
...
+)
+where
+
+import Data.HashMap.Strict (HashMap)
+import Data.HashMap.Strict qualified as HashMap
+import Data.Hashable (Hashable)
+import UnliftIO.STM
+
+-- | An 'InternCache' is a key-value map, parameterized by the monad
+--   in which it operates, the key type, and the value type.
```
xsebek:
```suggestion
-- | An 'InternCache' is a key-value map that lives in the IO monad.
```

src/swarm-util/Control/Carrier/Accum/ActuallyStrict.hs
xsebek, line 1: With https://github.com/fused-effects/fused-effects/pull/468 merged (good job!) can this be replaced by using latest `fused-effects:v1.1.2.6`?

LICENSE
```diff
@@ -29,3 +29,37 @@ DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
 THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
 (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
 OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+------------------------------------------------------------
+
+Swarm.Util.InternCache copyright Chris Penner 2025
+SPDX-License-Identifier: BSD-3-Clause
```
xsebek: If you want to use the simple cache, does it even make sense to attribute it? The simple version looks very different to me. Maybe even rename it to "global cache".

src/swarm-lang/Swarm/Language/Cache.hs
```diff
@@ -0,0 +1,69 @@
...
+import Swarm.Language.Value (Env)
+import Swarm.Util.InternCache (InternCache)
+import Swarm.Util.InternCache qualified as IC
+import System.Directory (getModificationTime)
+import System.IO.Unsafe (unsafePerformIO)
+
+-- | A global cache for fully parsed, typechecked, + elaborated
+--   modules, to avoid reloading the same module from disk repeatedly.
+--   However, note that modules which are no longer referenced from
+--   outside the cache can be garbage collected.
```
xsebek: This will not happen with the simple cache right?

src/swarm-lang/Swarm/Language/Load.hs
```diff
@@ -115,42 +58,62 @@ resolve ::
...
--- | Convert a 'SourceMap' into a suitable form for 'findCycle'.
-toImportGraph :: SourceMap Resolved -> [(ResLoc, ResLoc, [ResLoc])]
-toImportGraph = map processNode . M.assocs
- where
-  processNode (imp, Module _ _ imps) = (imp, imp, S.toList imps)
+-- -- | Convert a 'SourceMap' into a suitable form for 'findCycle'.
+-- toImportGraph :: SourceMap Resolved -> [(ResLoc, [ResLoc])]
+-- toImportGraph = map processNode . M.assocs
+--  where
+--   processNode (imp, m) = (imp, S.toList (moduleImports m))
```
xsebek: I guess this commented out code can be deleted, because you changed the graph functions instead?

src/swarm-lang/Swarm/Language/Load.hs
```diff
@@ -184,34 +148,61 @@ resolveImport ::
 resolveImport parent loc = do
   -- Compute the canonicalized location for the import, and record it
   canonicalLoc <- resolveImportLoc (unresolveImportDir parent <//> loc)
+
+  -- Note, the purpose of this set is not to track which imports have
+  -- been loaded yet; rather, the purpose is simply to record the
+  -- complete set of imports for a given module.
   add $ S.singleton canonicalLoc
```
xsebek: This seems a bit tricky, I find myself wishing for an explanatory comment like GHC uses. Then the comments could link to it. 

We don't have that in the CI (yet), but if such a comment would be help here, then I would be happy to set it up.

src/swarm-lang/Swarm/Language/Module.hs
```diff
@@ -0,0 +1,72 @@
+{-# LANGUAGE TypeFamilies #-}
+{-# LANGUAGE UndecidableInstances #-}
+
+-- |
+-- SPDX-License-Identifier: BSD-3-Clause
+--
+-- A module is an AST along with
```
xsebek: With what? 🤔

src/swarm-lang/Swarm/Language/Module.hs
```diff
@@ -0,0 +1,72 @@
...
+
+-- | The context for a module, containing names and types of things
+--   defined in the module (once typechecking has run).
+type family ModuleCtx (phase :: Phase) where
+  ModuleCtx Raw = ()
+  ModuleCtx Resolved = ()
+  ModuleCtx Inferred = (UCtx, TDCtx)
+  ModuleCtx Typed = (TCtx, TDCtx)
+  ModuleCtx Elaborated = (TCtx, TDCtx)
+  ModuleCtx Instantiated = (TCtx, TDCtx)
```
xsebek: I am a bit paranoid about the lazy pair. Maybe we could have a strict struct for these two things? I would actually appreciate having a name for the two components, as `UCtx` and `TDCtx` are not as familiar to me.

src/swarm-lang/Swarm/Language/Module.hs
```diff
@@ -0,0 +1,72 @@
...
+
+-- | With a newly parsed, raw module we don't know any import
+--   information.  After that, we cache the set imports for a module
+--   to enable easy import cycle checks.
+type family ModuleImports (phase :: Phase) where
+  ModuleImports Raw = ()
+  ModuleImports _ = Set (ImportLoc Import.Resolved)
+
+data ModuleProvenance where
```
xsebek, line 61: What does this do? 🙂 The field has a note "Where the module came from" but I not sure what is this knowledge useful for.
Also I do not know what NoProvenance means, is that a root module? When I first saw that parameter, I thought it would mean that provenance should not be tracked.

EDIT: I found this note in the other module:
```Haskell
-- | Provenance of the source code was obtained, relative to
  --   which imports should be interpreted.  If Nothing, use
  --   the current working directory.
  Maybe FilePath ->
```
So maybe `NoProvenance` could be called `FromCurrentDirectory`?

src/swarm-lang/Swarm/Language/Pipeline.hs
```diff
@@ -79,25 +91,57 @@ processTerm ::
...
+
+  -- Elaborate term + collected imports
+  let tmElab = elaborate tmTy
+      srcMapElab = fmap elaborateModule srcMapTy
+
+  -- Insert all newly checked + elaborated modules into the module
+  -- cache, and delete them from the environment cache in case they
+  -- are replacing a previously cached + evaluated version
+  let newModule loc m = insertCached moduleCache loc m >> deleteCached envCache loc
+  void . sendIO $ M.traverseWithKey newModule srcMapElab
```
xsebek: It might be useful to have a union/difference operation on the global map, so that you only need to take the `TVar` once.

src/swarm-util/Control/Carrier/Accum/ActuallyStrict.hs
byorgey: Oh, good point, I didn't realize it had been released.

LICENSE
byorgey: Good point, it started out as a copy but then changed a lot over time as I figured out what we really needed.

src/swarm-lang/Swarm/Language/Cache.hs
byorgey: Ah, right, fixed.

src/swarm-lang/Swarm/Language/Pipeline.hs
byorgey: Would that really be an improvement?  In theory, when using a `TVar` it's better to split up big operations into smaller atomic operations, so they can succeed or retry individually, rather than requiring the entire thing to succeed as one big transaction.  Again, it's moot at the moment since everything is single-threaded, but eventually it could matter if we add some concurrency.

src/swarm-lang/Swarm/Language/Module.hs
byorgey: Added a bunch of comments on `ModuleProvenance`, thanks for pointing out it was missing!

src/swarm-lang/Swarm/Language/Module.hs
byorgey: Making a strict struct is awkward because of the need to have `UCtx` in the `Inferred` phase and `TCtx` everywhere else.  Do we make two different structs?  A struct parameterized by the `UCtx`/`TCtx`?  In the end I decided to just use a generic strict pair type, and added a bunch of comments to help explain what the components are.

src/swarm-lang/Swarm/Language/Load.hs
byorgey: I added a bunch of comments + a Note, see what you think now!

src/swarm-lang/Swarm/Language/Pipeline.hs
xsebek: If I read the code right, you collect list/map of modules but then add them individually. That means taking, releasing, taking, releasing,… the TVar. When you could only take once and merge the map.

If the idea was to have more atomic operations, why wait until all are collected? Wouldn’t it be better to update the global cache immediately when a module is finished, so that a (future) parallel operation can reuse it?

byorgey: > Wouldn’t it be better to update the global cache immediately when a module is finished, so that a (future) parallel operation can reuse it?

Yes, I suppose so, but that would be difficult given our current architecture.  The typechecker starts checking the top-level code and recursively checks imports as it encounters them, so it can be sure to typecheck everything in the right order.  It then returns a map containing all the typechecked modules.  Then we elaborate all of them (which is probably very fast, compared to typechecking) and add them to the global cache.  So here are some things we could do to improve the granularity of when modules get added to the global cache:

1. We could add each module to the global cache right after elaborating it, instead of elaborating all of them and then adding them all to the global cache.  This would be easy, we could do it right now; but (a) I don't think it would make much real difference, as elaboration is very quick, and (b) I'm not sure that it's not already happening anyway due to elaboration happening lazily.
2. We could inline elaboration into the typechecker, so that the typechecker performs elaboration on each module and then adds it to the global cache as soon as it completes being typechecked, even when deep inside several levels of recursive imports.  That would be a somewhat invasive change and also feels somewhat anti-compositional.
3. We could change the typechecker so that it just takes a single module and typechecks it, assuming that all its imports have already been typechecked.  This would require changing the resolution phase so that it returns not just a map of modules, but a *topological sort*.  Then we could just iterate through in topological sort order and independently call the typechecker + elaborator on each module.

Actually, now that I write that all out, (3) seems like obviously the right thing to do.

byorgey: OK, I just pushed up a commit implementing (3) and I think it's a big improvement for multiple reasons. :smile:

src/swarm-engine/Swarm/Game/Step.hs
```diff
@@ -865,9 +866,12 @@ stepCESK cesk = case cesk of
...
   -- environment we got after processing the import.
-  Suspended _ e' s (FImport loc t e : k) -> do
-    sendIO $ insertCached envCache loc e'
-    return $ In t (e <> e') s k
+  Suspended _ e' s (FImport loc vars t e : k) -> do
+    -- Restrict the environment to only that which is supposed to be
+    -- exported from the module
+    let e'' = restrictEnv vars e'
+    sendIO $ insertCached envCache loc e''
+    return $ In t (e <> e'') s k
```
xsebek: What would happen if it was not restricted? 🤔 The import value documentation did not really enlighten me as to why it should restrict itself. Is that avoiding some bug? If so, should there be a test case?

src/swarm-lang/Swarm/Language/Context.hs
```diff
@@ -254,6 +256,16 @@ delete x ctx@(Ctx m _) = case M.lookup x m of
   Nothing -> ctx
   Just t -> rollCtx $ CtxDelete x t ctx
 
+-- A \ B = A ∩ ~B
+-- so
+-- Ctx \ (Ctx \ R) = Ctx ∩ ~(Ctx ∩ ~R) = Ctx ∩ (~Ctx ∪ R) = (Ctx ∩ ~Ctx) ∪ (Ctx ∩ R) = Ctx ∩ R
```
xsebek: Maybe this should be above `keysToDelete`? It's close now, but I would not like it to be able to move away if we add more functions.

src/swarm-lang/Swarm/Language/Load.hs
```diff
@@ -39,17 +44,61 @@ import Swarm.Language.Syntax.Import hiding (ImportPhase (..))
...
 -- | A SourceMap associates canonical 'ImportLocation's to modules.
-type SourceMap phase = Map (ImportLoc (ImportPhaseFor phase)) (Module phase)
+--   It is an /ordered/ map, and has as an invariant that the modules
+--   are topologically ordered, i.e. if B imports A, then A comes
+--   before B in order.  This ensures that if we process modules in
+--   order, by the time we process a module we have already processed
+--   any modules it imports.
+type SourceMap phase = OMap (ImportLoc (ImportPhaseFor phase)) (Module phase)
+
+type ResLoc = ImportLoc Import.Resolved
```
xsebek: This type synonym is not exported, but appears in exported function signature - won't that look weird in Haddock? Maybe rename the synonym as `ResolvedImport` or `LoadedImport` with a short explanation.

src/swarm-lang/Swarm/Language/Load.hs
```diff
@@ -67,34 +116,27 @@ resolve prov s = do
         Left _ -> currentDir
         Right (loc, _) -> importDir loc
 
-  (resMap, (impSet, s')) <- runState mempty . resolveImports cur $ s
+  (resMap, (impSet, s')) <- evalState @(Set ResLoc) S.empty . runState @(SourceMap Resolved) OM.empty . resolveImports cur $ s
```
xsebek: The phantom state parameter makes this line a bit too long, how about adding a (not exported) constant:
```Haskell
emptyImports :: Set ResLoc
emptySourceMap :: SourceMap Resolved
```

src/swarm-lang/Swarm/Language/Load.hs
xsebek: Very nice! 👍

src/swarm-lang/Swarm/Language/Load.hs
```diff
@@ -103,89 +145,108 @@ checkImportCycles srcMap = do
...
 --   transitively.  Also return a canonicalized version of the import
---   location.
+--   location, and add it to the ambient @Accum@ effect which is
+--   keeping track of all imports seen in a given module.
+--
+--   Note the difference between the (State (Set ImportLoc)) and
+--   (Accum (Set ImportLoc)) effects: the former is to globally keep
+--   track of all modules seen so far, to avoid getting stuck in an
+--   import cycle; the latter is to locally accumulate the set of
+--   imports for each module.
```
xsebek: How about making type synonyms for the constraints?
```Haskell
-- | Globally keep track of all modules seen so far,
--   to avoid getting stuck in an import cycle.
type EncounteredModules = Set ResLoc

-- | Locally accumulate the set of imports for each module.
type LocalModules = Set ResLoc

resolveImport :: (Has (State EncounteredModules) sig m, Has (Accum LocalModules) sig m, ...
```

This would sort of propagate this note to the other functions sharing the constraints and make this doc string slightly shorter.

src/swarm-engine/Swarm/Game/Step.hs
byorgey: Oh, good question.  I will definitely add a note and one or more tests.  The short explanation is that a module should only export things it explicitly defines; it should *not* (by default) transitively export everything it imports.  So e.g. if we have

B.sw:

```
def b = 1 end
```

A.sw:

```
import "B"
def a = b + 1 end
```

then if you `import "A"` you will get `a = 2` in scope but you do *not* get `b` in scope.  This is the way e.g. Haskell imports work and seems like the "right" way to do things in terms of encapsulation/modularity.  There is an issue (#2664) to add a way to do controlled re-exports when you want them.

byorgey: @xsebek I added more documentation in https://github.com/swarm-game/swarm/pull/2588/changes/f1ba1d4839b640ffa7afaaa5667e48b9d882ef49 ; see what you think.  In any case I'll go ahead and merge this into #2540 .

PR #2594, opened 2025-10-17
Regularize/consolidate file I/O + take encoding into account

### LINKED ISSUE #2592: Move tutorial does not allow moving past the first room
> I'm trying the tutorial but I can't get past the 2nd one. It tells me there's a wall in the way. It doesn't matter which direction I move into, the robot can only go to the coordinates `[(0, 0), (1, 0), (2, 0)]` and that's it. Not sure what to do!  

 _Originally posted by @CrispyDrone in [#2392](https://github.com/swarm-game/swarm/issues/2392#issuecomment-3393721233)_

byorgey:
In this PR:

* Outlaw all calls to `readFile`, `writeFile`, and their variants and `Text` counterparts, forcing all file reading/writing to go through functions in `Swarm.Util`.
* Add `writeFile` and `writeFileT` functions to `Swarm.Util` in addition to the existing `readFileMay` and `readFileMayT`.
* Change `readFileMay` to read *strictly*.
* Add an extra `Encoding` argument to all four of the above functions, specifying whether to use the system locale encoding or UTF8.
* Use UTF8 for files provided with Swarm (e.g. scenarios in `data/scenarios`) and system locale for user files (e.g. a `.sw` file they wrote themselves).
    * It is not always possible to distinguish these 100%.  For example it is possible for a Windows developer to manually run `swarm format` on a UTF-8 Swarm file, which might generate a formatted version with the wrong encoding.  But that would be pretty obvious in a PR.
    * Hopefully this will "do the right thing" in 99% of cases.
    * In particular, this should fix #2592.
* Fix a couple latent crashes (e.g. it was possible to crash swarm by doing `swarm format` or `--run` on a non-existent file, because those code paths just called `readFile` which throws an exception when the file is not found).

### DIFF
--- a/src/swarm-scenario/Swarm/Game/World/DSL/Load.hs
+++ b/src/swarm-scenario/Swarm/Game/World/DSL/Load.hs
@@ -42,12 +42,12 @@ loadWorld ::
   (Has (Throw SystemFailure) sig m) =>
   FilePath ->
   TerrainEntityMaps ->
-  (FilePath, String) ->
+  (FilePath, Text) ->
   m (Text, Some (TTerm '[]))
 loadWorld dir tem (fp, src) = do
   wexp <-
     liftEither . left (AssetNotLoaded (Data Worlds) fp . SystemFailure . CanNotParseMegaparsec) $
-      runParser parseWExp (into @Text src)
+      runParser parseWExp src
   t <-
     withThrow (AssetNotLoaded (Data Worlds) fp . SystemFailure . DoesNotTypecheck . prettyText @CheckErr) $
       runReader tem . runReader @WorldMap M.empty $
--- a/src/swarm-util/Swarm/Util.hs
+++ b/src/swarm-util/Swarm/Util.hs
@@ -39,7 +39,6 @@ module Swarm.Util (
   Encoding (..),
   readFileMay,
   readFileMayT,
-  writeFile,
   writeFileT,
   findAllWithExt,
   acquireAllWithExt,
@@ -129,7 +128,6 @@ import System.IO.Error (catchIOError)
 import Witch (from)
 import Witherable (wither)
 import Prelude hiding (Foldable (..), readFile, writeFile)
-import Prelude qualified
 
 infixr 1 ?
 infix 4 %%=, <+=, <%=, <<.=, <>=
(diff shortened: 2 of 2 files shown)


src/swarm-engine/Swarm/Game/State.hs
```diff
@@ -161,8 +160,9 @@ parseCodeFile ::
   FilePath ->
   m CodeToRun
 parseCodeFile filepath = do
-  contents <- sendIO $ TIO.readFile filepath
-  pt <- either (throwError . CustomFailure) return (processTermEither contents)
+  mcontents <- sendIO $ readFileMayT SystemLocale filepath
+  contents <- maybe (throwError (AssetNotLoaded (Data Script) filepath (DoesNotExist File))) pure mcontents
```
xsebek: Is there a function to run an action and handle the alternative? 🤔 

I imagine something like:
```Haskell
contents <- sendIO (readFileMayT SystemLocale filepath)
  ??? throwError (AssetNotLoaded (Data Script) filepath (DoesNotExist File))
```

src/swarm-engine/Swarm/Game/Step/Const.hs
```diff
@@ -1227,9 +1227,10 @@ execConst runChildProg c vs s k = do
         let filePath = into @String fileName
         sData <- throwToMaybe @SystemFailure $ getDataFileNameSafe Script filePath
         sDataSW <- throwToMaybe @SystemFailure $ getDataFileNameSafe Script (filePath <> ".sw")
-        mf <- sendIO $ mapM readFileMay $ [filePath, filePath <> ".sw"] <> catMaybes [sData, sDataSW]
+        muser <- sendIO . traverse (readFileMay SystemLocale) $ [filePath, filePath <> ".sw"]
+        msys <- sendIO . traverse (readFileMay UTF8) $ catMaybes [sData, sDataSW]
```
xsebek: This was already the case, but why are we reading String and then converting `into` Text?

src/swarm-engine/Swarm/Game/State.hs
byorgey: There's no standard one that I know of, though apparently it's been redefined a bunch of times in various projects:

https://hoogle.haskell.org/?hoogle=m%20(Maybe%20a)%20-%3E%20m%20a%20-%3E%20m%20a

src/swarm-engine/Swarm/Game/Step/Const.hs
byorgey: Good point, no reason.

byorgey: Very Soon Now™ this code is going to be deleted with extreme prejudice.  But until then we might as well make it a bit nicer.

byorgey:
@CrispyDrone can you try this branch and confirm whether it solves your issue with the second tutorial level?

byorgey:
@mergifyio queue

PR #2596, opened 2025-10-18
Chore: make located more general

oliverpauffley:
Currently the comment and constructor names make located sound like it can only be used for variables, but we want to add SrcLoc's for types next and will re-use this `Located` wrapper.

### DIFF
--- a/src/swarm-scenario/Swarm/Game/Entity.hs
+++ b/src/swarm-scenario/Swarm/Game/Entity.hs
@@ -671,8 +671,7 @@ entityNameFor _ = to $ \e ->
     Just pl -> pl
     Nothing -> plural (e ^. entityName)
 
--- | A longer, free-form description of the entity.  Each 'Text' value
---   represents a paragraph.
+-- | A longer, free-form description of the entity.
 entityDescription :: Lens' Entity (Document Syntax)
 entityDescription = hashedLens _entityDescription (\e x -> e {_entityDescription = x})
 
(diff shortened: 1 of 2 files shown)


src/swarm-lang/Swarm/Language/Syntax/Loc.hs
```diff
@@ -66,10 +66,10 @@ srcLocEndsBefore :: SrcLoc -> SrcLoc -> Bool
 srcLocEndsBefore (SrcLoc _ a) (SrcLoc _ b) = a <= b
 srcLocEndsBefore _ _ = False
 
--- | A variable with associated source location, used for variable
---   binding sites. (Variable occurrences are a bare TVar which gets
---   wrapped in a Syntax node, so we don't need LocVar for those.)
-data Located v = LV {lvSrcLoc :: SrcLoc, lvVar :: v}
+-- | A type with associated source location.
+data Located t = LV {lvSrcLoc :: SrcLoc, inner :: t}
```
oliverpauffley: I thought about the `inner` name for awhile and thought it suggested the wrapping nature of `Located` but happy for other suggestions

xsebek: How about `Loc {locSrc :: SrcLoc, locVal :: a}`?

I would say that it's a _value_ with associated source location. 

To me `inner` sounds way too generic.

byorgey: I like @xsebek 's suggestion.  LV stands for "Located Variable" and is also a holdover from when it was only about variables.

PR #2597, opened 2025-10-19
Make located more general

oliverpauffley:
Sorry dropped the last PR since it was easier just to start over.

LV -> Loc

lvVar -> locVal

### DIFF
--- a/src/swarm-lang/Swarm/Language/Parser/Type.hs
+++ b/src/swarm-lang/Swarm/Language/Parser/Type.hs
@@ -73,11 +73,11 @@ parseTypeAtom =
   TyVar <$> tyVar
     <|> TyConApp <$> parseTyCon <*> pure []
     <|> TyDelay <$> braces parseType
-    <|> TyRcd <$> brackets (M.fromList . (map . first) locVal <$> parseRecord (symbol ":" *> parseType))
+    <|> TyRcd <$> brackets (toRecFieldsMap <$> parseRecord (symbol ":" *> parseType))
     <|> tyRec <$> (reserved "rec" *> tyVar) <*> (symbol "." *> parseType)
     <|> parens parseType
  where
-  toRecFieldsMap = M.fromList . (map . first) $ locVal
+  toRecFieldsMap = M.fromList . map (first locVal)
 
 -- | A type constructor.
 parseTyCon :: Parser TyCon


src/swarm-lang/Swarm/Language/Parser/Type.hs
```diff
@@ -62,27 +62,38 @@ parseType = makeExprParser parseTypeMolecule table
...
+      <$> parseTyCon
+      <*> pure []
+    <|> TyDelay
+      <$> braces parseType
+    <|> TyRcd
+      <$> brackets (M.fromList . (map . first) locVal <$> parseRecord (symbol ":" *> parseType))
+    <|> tyRec
+      <$> (reserved "rec" *> tyVar)
+      <*> (symbol "." *> parseType)
     <|> parens parseType
```
xsebek: Fourmolu seems to have broken the lines a bit here. I kind of like the neat way it was previously.

Could you please refactor the `locVal` line a little so that the whole thing can be terser?
Something like this should do it:
```Haskell
 where
  toRecFieldsMap = M.fromList . (map . first) locVal
```

Or maybe you have different version or configuration of Fourmolu, given that it changed an untouched lines below and above as well.

Not a big deal, but it would be better to not mess with whitespace unnecessarily so that I don't blame you later. 😁

src/swarm-lang/Swarm/Language/Syntax/Loc.hs
```diff
@@ -66,10 +66,10 @@ srcLocEndsBefore :: SrcLoc -> SrcLoc -> Bool
 srcLocEndsBefore (SrcLoc _ a) (SrcLoc _ b) = a <= b
 srcLocEndsBefore _ _ = False
 
--- | A variable with associated source location, used for variable
---   binding sites. (Variable occurrences are a bare TVar which gets
---   wrapped in a Syntax node, so we don't need LocVar for those.)
-data Located v = LV {lvSrcLoc :: SrcLoc, lvVar :: v}
+-- | A type with associated source location.
```
xsebek: Nitpick:
```suggestion
-- | A value with associated source location.
```

xsebek (approved):
Thanks, this even seems to read better regardless of the change to use it for other types.

I have just two small nitpicks, maybe check your Fourmolu version. 😉

src/swarm-lang/Swarm/Language/Syntax/Loc.hs
```diff
@@ -66,10 +66,10 @@ srcLocEndsBefore :: SrcLoc -> SrcLoc -> Bool
...
 
--- | A variable with associated source location, used for variable
---   binding sites. (Variable occurrences are a bare TVar which gets
---   wrapped in a Syntax node, so we don't need LocVar for those.)
-data Located v = LV {lvSrcLoc :: SrcLoc, lvVar :: v}
+-- | A value with associated source location.
+data Located v = Loc {lvSrcLoc :: SrcLoc, locVal :: v}
   deriving (Eq, Ord, Functor, Show, Data, Generic, Hashable, FromJSON, ToJSON)
 
+-- | A variable with associated source location. (Variable occurrences are a bare TVar which gets
```
byorgey: Let's put back in the phrase "used for variable binding sites".  The parenthetical comment does not make a lot of sense without that phrase.  The point is that we use `LocVar` only for places where variables are defined/bound, not where they are used.

src/swarm-lang/Swarm/Language/Parser/Type.hs
xsebek: Maybe also reorder the parens:

```Haskell
fromList . map (first locVal)
```

oliverpauffley: Hmm the formatting is annoying. I would have thought that restyled would have ensured consistency. 

I'm on `fourmolu 0.15.0.0`

xsebek: As you can see in CONTRIBUTING and restyled configuration, we are on 19.

Fourmolu gives us some limited agency in how we format code. I think you can always write multiline, but oneliners have a limit.

EDIT: Found this in [Fourmolu goals](https://github.com/fourmolu/fourmolu) (this one shared with Ourmolu):
> Let some whitespace be programmable. The layout of the input influences the layout choices in the output. This means that the choices between single-line/multi-line layouts in certain situations are made by the user, not by an algorithm. This makes the implementation simpler and leaves some control to the user while still guaranteeing that the formatted code is stylistically consistent.

PR #2599, opened 2025-10-19
Discover scenario integration tests by default instead of listing them all manually

### LINKED ISSUE #2397: Autodiscover scenarios in integration tests
Currently, our integration testing in `test/integration/Main.hs` must explicitly list every single scenario.  But that means if we ever accidentally forget to list one, it will not be tested.  In fact, it's quite possible that there already exists some scenario that is not listed --- I'm not sure how we would be able to tell!

Of course, some of the scenarios require custom configuration, e.g. a longer allowed runtime than the default, or extra custom testing.  But I propose to set up the integration tests in such a way that any scenarios in `data/scenarios/` which are not explicitly listed will be auto-discovered and run with default settings.  Or, put another way, we should auto-discover all the scenarios in `data/scenarios`, but have a way to explicitly override the default test settings for specific scenarios.

byorgey:
Previously, we had to manually list every single scenario that should be tested.  This meant in particular that if anyone created a new scenario and simply forgot to add it to the list, it would not be tested.  This happened several times in the past that I know of, and through making this PR itself we found several additional scenarios that were not listed but had a failing solution!

This PR changes the system for testing scenario solutions.  By default, every scenario in `data/scenarios` will be tested with default parameters, except: (1) those in `_Validation` are excluded; (2) there is an explicit list of scenarios that need more time; (3) there is an explicit list of scenarios with custom validation code; (4) there is an explicit list of scenarios which are expected to fail; (5) there is an explicit list of scenarios that do not have a solution.  Anything in the explicit lists overrides the default tests.  But this means that if you simply add a scenario to `data/scenarios`, it will automatically be tested in CI.  If it does not have a solution, or the solution fails, CI will fail.  You can then choose to add it to one of the explicit lists to exclude it or give it a custom configuration as necessary.

@noahyor did most of the initial coding for this PR and I helped finish it.

Closes #2397.

Note, I also want to split up `Main` into multiple modules, but I will leave that for another PR, to make the reviewing easier.


test/integration/Main.hs
```diff
@@ -197,384 +199,301 @@ time = \case
...
-              verify g
-
-  tutorialHasLog :: GameState -> Assertion
-  tutorialHasLog gs =
-    let baseDevs = gs ^?! baseRobot . equippedDevices
-     in assertBool "Base should have a logger installed!" (not . null $ lookupByName "logger" baseDevs)
-
-  testTutorialSolution t f = testSolution' t f CheckForBadErrors tutorialHasLog
-  testTutorialSolution' t f s v = testSolution' t f s $ \g -> tutorialHasLog g >> v g
+  customTests = customTestScenarios <> customTimeoutScenarios
```
xsebek, line 536: Should the lists be exclusive?

If so, could you please add test that or error out?

xsebek (approved):

test/integration/Main.hs
byorgey: Good call, done.

PR #2607, opened 2025-10-21
Fix `Testing/687-watch-command` scenario

### LINKED ISSUE #2601: `Testing/687-watch-command` scenario solution fails
**Describe the bug**
The `solution` provided with the `Testing/687-watch-command` scenario does not work.

**To Reproduce**
`swarm -i Testing/687-watch-command --autoplay`

**Expected behavior**
The scenario should be successful.  Instead, a condolences message is displayed.

**Additional context**
This was discovered while working on #2599.  This scenario was not listed in in the integration tests, and apparently the solution has not been working for some time but no one ever noticed since it was not being tested.

byorgey:
This scenario was not listed in `test/integration/Main`.  Therefore, it was not being tested.  Therefore, it bitrotted.  In particular, it was not updated when the semantics of `watch` changed in #2538. The `base` used to wake up when the `planter` robot planted a tree.  However, with the change to `watch`, the `base` woke up as soon as the `planter` moved into the cell, then immediately moved to try to `grab` the `tree`, which had not been planted yet.

The solution is to use the first `watch` to wait for the planter, then immediately execute a second `watch` and go to sleep again to wait for the tree to be planted.

Fixes #2601.

### DIFF
--- a/test/integration/Main.hs
+++ b/test/integration/Main.hs
@@ -356,7 +356,6 @@ expectFailScenarios :: M.Map FilePath String
 expectFailScenarios =
   M.fromList
     [ "Testing/231-requirements/231-command-transformer-reqs" ==> "Awaiting fix for #231"
-    , "Testing/687-watch-command" ==> "Awaiting fix for #2601"
     , "Challenges/wolf-goat-cabbage" ==> "Awaiting fix for #2602"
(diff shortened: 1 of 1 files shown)


xsebek (approved):
👍

PR #2611, opened 2025-10-25
Extract world file names in a sane way

### LINKED ISSUE #2610: The classic scenarios do not parse
**Describe the bug**

When I try to run classic or creative scenario, it fails to start. It is not even shown in the TUI menu.

**To Reproduce**

```bash
cabal run swarm -- --debug=all --scenario classic
```
```
Configuration is affected by the following files:
- cabal.project.local
Failed to acquire scenarios from path '/Users/xsebek/Git/swarm/./data/scenarios/classic.yaml':
  Parse failure:
    Aeson exception:
    Error in $.world: Import 'classic' not found
```

**Expected behavior**

Classic and creative should show up in scenario menu. They should be possible to run from command line.

**Additional context**

This could be caused by the recent fixes for paths on Windows https://github.com/swarm-game/swarm/pull/2594.

byorgey:
When loading world descriptions from `data/worlds`, we extract the base name of each `foo.world` file to serve as a key into the `WorldMap`.  However, this was done in an insane, roundabout way: taking the path of the directory used to load the world files and a world file path, splitting both into components, and dropping a number of components from the second equal to the number of components in the first.  However, when the two ended up normalized differently, (i.e. the first contained an extra `/./`) this did not work.

This PR simplifies the code to just call the standard `takeFileName` function.

Fixes #2610.

### DIFF
--- a/src/swarm-scenario/Swarm/Game/World/DSL/Load.hs
+++ b/src/swarm-scenario/Swarm/Game/World/DSL/Load.hs
@@ -21,7 +21,7 @@ import Swarm.Pretty (prettyText)
 import Swarm.ResourceLoading (getDataDirSafe)
 import Swarm.Util (acquireAllWithExt)
 import Swarm.Util.Effect (withThrow)
-import System.FilePath (dropExtension, joinPath, splitPath)
+import System.FilePath (dropExtension, takeFileName)
 import Witch (into)
 
 -- | Load and typecheck all world descriptions from `worlds/*.world`.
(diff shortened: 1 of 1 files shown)


src/swarm-scenario/Swarm/Game/World/DSL/Load.hs
```diff
@@ -33,27 +33,22 @@ loadWorlds ::
...
   t <-
     withThrow (AssetNotLoaded (Data Worlds) fp . SystemFailure . DoesNotTypecheck . prettyText @CheckErr) $
       runReader tem . runReader @WorldMap M.empty $
         infer CNil wexp
-  return (into @Text (dropExtension (stripDir dir fp)), t)
-
--- | Strip a leading directory from a 'FilePath'.
-stripDir :: FilePath -> FilePath -> FilePath
-stripDir dir fp = joinPath (drop (length (splitPath dir)) (splitPath fp))
+  return (into @Text (dropExtension (takeFileName fp)), t)
```
xsebek, line 54: Which of the paths can contain `.`? If it is the `dir` could this cause other issues in code that expects it to be normalized and not contain any `.`?

xsebek (approved):
Thanks for the quick fix. 👍

src/swarm-scenario/Swarm/Game/World/DSL/Load.hs
byorgey: Yes, the `dir` was the one which can contain `.`.  In theory it shouldn't make a difference unless we do other stupid things like this.

However, we know exactly where `dir` came from: from a call to `getDataDirSafe`.  So it probably makes sense to normalize the output of that function.  I will do that in a separate PR.

PR #2619, opened 2025-10-27
Refactor comparison using cases

### LINKED ISSUE #2618: Flipping comparison changes exception
**Describe the bug**

Flipping comparison produces different exceptions:
```c
// foobar.sw
def foo = {1} end
def bar = 1 end
```
```c
let foo = "" in let bar = "" in run "example/foobar.sw"; pure (foo > bar)
// <: Comparison is undefined for  {1} and 1

let foo = "" in let bar = "" in run "example/foobar.sw"; pure (bar < foo)
// Fatal error: Incompatible comparison of 1 and {1}
```

**Expected behavior**

Both should report that `{1}` cannot be compared.

**Additional context**

This issue is just for fun, I already have a fix.

xsebek:
* refactor comparison using [`LambdaCase`](https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/lambda_case.html) - since GHC 9.4.1 adds [Multi-way lambda expressions](https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0302-cases.rst)
* remove extra space in error message
* fix #2618

### DIFF
--- a/src/swarm-engine/Swarm/Game/Step/Arithmetic.hs
+++ b/src/swarm-engine/Swarm/Game/Step/Arithmetic.hs
@@ -59,34 +59,43 @@ compareValues = \cases
     (<>) <$> compareValues v11 v21 <*> compareValues v12 v22
   (VRcd m1) (VRcd m2) ->
     mconcat <$> (zipWithM compareValues `on` M.elems) m1 m2
-  v1 v2 -> incompatCmp v1 v2
+  v1 v2 ->
+    if incomparable v1 || incomparable v2
+      then incomparableErr v1 v2
+      else incompatCmpErr v1 v2
+
+-- | Check if comparing types which cannot be compared (e.g. functions, etc.)
+incomparable :: Value -> Bool
+incomparable = \case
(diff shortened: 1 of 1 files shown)


src/swarm-engine/Swarm/Game/Step/Arithmetic.hs
```diff
@@ -43,57 +43,50 @@ evalCmp c v1 v2 = decideCmp c $ compareValues v1 v2
...
   throwError $
     Fatal $
-      T.unwords ["Incompatible comparison of ", prettyValue v1, "and", prettyValue v2]
-
--- | Values were compared of a type which cannot be compared
---   (e.g. functions, etc.).
-incomparable :: Has (Throw Exn) sig m => Value -> Value -> m a
-incomparable v1 v2 =
-  throwError $
-    cmdExn
```
xsebek, line 94: Whoops, one is supposed to be `Fatal` other just `cmdExn`.

byorgey (approved):

PR #2620, opened 2025-10-28
Update Mergify rules to match CI changes

byorgey:
I noticed that things tagged `merge me` were just sitting there not getting merged even after all the checks passed.  After looking at the Mergify rules I realized that in #2613 we forgot to update `mergify.yml` with the names of the new CI jobs.


xsebek (approved):
Oops, thanks! 👍

PR #2622, opened 2025-10-30
Tutorial updates, + new loops tutorial

byorgey:
- Various minor tutorial improvements
- Move conditional tutorial after `world101` (world101 does not require conditionals)
- Add new loops tutorial after conditionals and before `farming`, which prompts you to implement a while loop using recursion.

### DIFF
--- a/data/scenarios/Tutorials/loops.yaml
+++ b/data/scenarios/Tutorials/loops.yaml
@@ -65,6 +65,7 @@ robots:
       - logger
       - 3D printer
       - dictionary
+      - toolkit
     inventory:
       - [8, compass]
       - [8, solar panel]
@@ -75,6 +76,7 @@ robots:
       - [8, lambda]
       - [8, strange loop]
       - [8, branch predictor]
+      - [8, toolkit]
 entities:
 known: [rock, gold]
 world:


data/scenarios/Tutorials/loops.yaml
```diff
@@ -0,0 +1,77 @@
...
+        program a robot to go and fetch it!
+      - |
+        Some kind of `while`{=snippet} loop would obviously come in
+        handy here. However, the Swarm language does not have loops,
+        only recursion, which is enabled by the `strange loop`{=entity}
+        device.  You can use recursion to implement your own control
+        structures.  For example, try filling in the `...`{=snippet}
+        below to implement your own while loop function:
+      - |
+        `def while : Cmd Bool -> Cmd Unit -> Cmd Unit = \cond. \body. ... end`{=snippet}
```
xsebek: I did not realize there wasn't a recursion tutorial. 😲 

Could you maybe add a small recursion example before the `while`? Maybe I am making it too easy, but I would add:

````markdown
For example a program that repeats forever:
```
def forever : Cmd Unit -> Cmd Unit = \body.
  body; forever body
end
```
Now try filling in the `...`{=snippet} below to implement your own while loop function
to lead you along the path of rocks.
````

File: data/scenarios/Tutorials/type-errors.yaml [hunk 2]
```diff
@@ -30,6 +32,15 @@ entities:
       char: 'W'
     description:
       - Do `place "win"` once you are done with this challenge.
+      - "Some other expressions for you to try:"
```
xsebek: What is the point of repeating optional goal in the "win device"?

If the idea is to have a shorter goal description as a reminder, shouldn't this include the initial `turn 1`?

src/swarm-lang/Swarm/Language/Typecheck.hs
```diff
@@ -518,7 +518,7 @@ instance PrettyPrec TypeErr where
     EscapedSkolem x ->
       "Skolem variable" <+> ppr x <+> "would escape its scope"
     UnboundVar x ->
-      "Unbound variable" <+> ppr x
+      "Undefined variable" <+> ppr x
```
xsebek, line 521: Out of curiosity, is there a difference between the two terms?

I guess "undefined" is more familiar and you did not rename `UnboundVar` to not have a large diff?

data/scenarios/Tutorials/loops.yaml
```diff
@@ -0,0 +1,77 @@
...
+        expressions; this one will guide you through using loops.
+      - |
+        To the east of your base lies a path of `rock`{=entity}s.  The
+        path extends in a straight line for an unknown distance, and at
+        the end of the path lies a lump of `gold`{=entity}.  You must
+        program a robot to go and fetch it!
+      - |
+        Some kind of `while`{=snippet} loop would obviously come in
+        handy here. However, the Swarm language does not have loops,
+        only recursion, which is enabled by the `strange loop`{=entity}
```
xsebek, line 18: I noticed that the strange loop is not installed if you write the `while` definition outside the `build` command. Is it worth calling out strange loop, if its description is mysterious and does not add anything?

data/scenarios/Tutorials/loops.yaml
```diff
@@ -0,0 +1,77 @@
...
+    devices:
+      - logger
+      - 3D printer
+      - dictionary
+    inventory:
+      - [8, compass]
+      - [8, solar panel]
+      - [8, logger]
+      - [8, treads]
+      - [8, grabber]
```
xsebek, line 74: Embarrassingly, I failed two times:
1. I forgot to move out of gold and could not salvage it with a robot
2. When I restarted, I found that I can not salvage with base

Could we add a `toolkit` please? 😅

data/scenarios/Tutorials/loops.yaml
```diff
@@ -0,0 +1,77 @@
...
+      - [8, scanner]
+      - [8, lambda]
+      - [8, strange loop]
+      - [8, branch predictor]
+entities:
+known: [rock, gold]
+world:
+  scrollable: false
+  dsl: |
+    let path_len = 173 in
```
xsebek, line 85: Neat use of the World DSL! 🗺️

xsebek (approved):
Looks good to me, I like the `loops` tutorial. 👍

data/scenarios/Tutorials/type-errors.yaml
byorgey: The point is so you don't have to keep re-opening the goals dialog to see what the suggestions were.  Agreed, it should include the `turn 1`.

src/swarm-lang/Swarm/Language/Typecheck.hs
byorgey: There is not really much a difference.  "Unbound" is more technically precise, because e.g. a variable could be defined *somewhere* but not bound in the specific scope where the error happened.  I kept `UnboundVar` because it is a precise description of the error that happened; I just changed the way the error is displayed to use a word that is likely to be more familiar to players who are not programming language nerds.

data/scenarios/Tutorials/loops.yaml
byorgey: > I noticed that the strange loop is not installed if you write the `while` definition outside the `build` command.

Oh, hmm, really?  I would say that is a bug. (Filed #2627.)

I called out `strange loop` because later when the player starts the classic mode game, they will want to know that they need to craft one in order to be able to use recursive functions.  Of course they can also discover this through the error message that happens when they try to run a recursive function without one; but IMO it's better if their reaction at that point is "oh, right, I remember that.  I wonder how I can craft one?" rather than "wait, what's a strange loop?".  But I could be convinced otherwise.

data/scenarios/Tutorials/loops.yaml
byorgey: Done!

PR #2632, opened 2025-11-02
Add `--v0.7` flag to `swarm format` and get rid of `--v0.6` conversion

byorgey:
The only breaking change from 0.7 to 0.8 is that the type of `atomic` and `instant` changed (see #2552).  In particular their argument type changed from `Cmd a` to `{Cmd a}`.  Therefore, `swarm format --v0.7` finds any application of `atomic` or `instant` and wraps the argument in delay brackets.

```
$ echo 'atomic move' | cabal run swarm:exe:swarm -- format --stdin
atomic move
$ echo 'atomic move' | cabal run swarm:exe:swarm -- format --stdin --v0.7
atomic {move}
```

This PR also rips out the code to convert 0.6 -> 0.7; according to convention at the moment, we only include code to upgrade one version at a time.

### DIFF
--- a/src/swarm-lang/Swarm/Language/Parser/Term.hs
+++ b/src/swarm-lang/Swarm/Language/Parser/Term.hs
@@ -103,15 +103,11 @@ parseTermAtom2 =
     <|> parseLoc (SDelay <$> braces parseTerm)
     <|> parseLoc (view antiquoting >>= (guard . (== AllowAntiquoting)) >> parseAntiquotation)
 
--- | Parse the contents of a @require@ statement: either requiring a device, or
---   requiring inventory stock.
+-- | Parse the contents of a @require@ statement.
 parseRequire :: Parser Term
-parseRequire =
-  asum
-    [ TRequire <$> (textLiteral <?> "device name in double quotes")
-    , parseStock
-    ]
+parseRequire = TRequire <$> (textLiteral <?> "device name in double quotes")
 
+-- | Parse the contents of a @stock@ statement.
 parseStock :: Parser Term
 parseStock =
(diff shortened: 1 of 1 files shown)


src/swarm-lang/Swarm/Language/Format.hs
```diff
@@ -88,17 +82,11 @@ formatSwarm (FormatConfig _ _ mWidth ver) content = case readTerm' cfg content o
...
- where
-  -- Any free occurrences of 'fst' or 'snd'?
-  freeProjs = setOf freeVarsV ast `S.intersection` S.fromList ["fst", "snd"]
-
-  addDefn :: Syntax -> Var -> Syntax
-  addDefn rest = \case
-    "fst" -> [astQ| def fst = \p. match p (\a. \_. a) end; $syn:rest |]
-    "snd" -> [astQ| def snd = \p. match p (\_. \b. b) end; $syn:rest |]
-    _ -> rest
+-- | 'instant' and 'atomic' changed types from 0.7 -> 0.8.  They used to have
```
xsebek, line 85: I forgot this needs to be added to the formatter. Though it's only for creative/system robots, so we might have never noticed.

xsebek (approved):
LGTM. 👍 Sorry this took so long.

src/swarm-lang/Swarm/Language/Format.hs
byorgey: `instant` is only for creative/system robots, but `atomic` can be used by any robot as long as it has e.g. a `rubber band` device.  I had to update some of my own personal classic mode code (and tested this formatter on it).

PR #2633, opened 2025-11-02
Update PR template to mention new `BREAKING` tag

byorgey:
I added a new `BREAKING` tag for changes that break existing swarm-lang code.  These are worth flagging specially since ideally, we try to include a flag to `swarm format` to auto-update code to the new version in each release.


xsebek (approved):
Good idea! 👍

PR #2636, opened 2025-11-05
fix: Use Named Fields for GrowthTime

### LINKED ISSUE #2626: Use named fields in GrowthTime
This is a follow-up to #2617:

https://github.com/swarm-game/swarm/blob/b2c08b652c47a45803d3fbec5c0bdf713ac21080/src/swarm-scenario/Swarm/Game/Entity.hs#L243-L244

This `newtype` needed explicit deriving strategy with GHC2021. But maybe it should **not** be a `newtype`.

1. Pairs are _lazy_. This can cause space leaks and we try to avoid those by using `StrictData`.
2. It needs a documentation to say what those values mean.

The upside is getting `ToJSON` instance which results in nice pair:
```yaml
growth:
  duration: [20, 30]
```

The same goes for combustion duration which does not use a newtype:
https://github.com/swarm-game/swarm/blob/b2c08b652c47a45803d3fbec5c0bdf713ac21080/src/swarm-scenario/Swarm/Game/Entity.hs#L261-L262

If they both used same duration type, it would be easier to reuse code for them.

stmonty:
Fixes #2626 
Hello! This is my first try at a PR on a Haskell project. My Haskell skills aren't that great, so any constructive criticism is welcome!
Also I couldn't get HLS to work so I am not sure if I found all the places where this was being used.
Another thing to call-out is that I am not too sure about the deriving strategies I used and if they are correct.

### DIFF
--- a/src/swarm-engine/Swarm/Game/Step/Combustion.hs
+++ b/src/swarm-engine/Swarm/Game/Step/Combustion.hs
@@ -94,7 +94,9 @@ addCombustionBot inputEntity combustibility ts loc = do
   let botInventory = fromMaybe [] $ do
         e <- (`lookupEntityName` em) =<< maybeCombustionProduct
         return $ pure (1, e)
-  combustionDurationRand <- uniform durationRange
+  combustionDurationRand <-
+    uniform
+      (tickRangeMin durationRange, tickRangeMax durationRange)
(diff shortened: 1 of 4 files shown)


byorgey:
@all-contributors please add @stmonty for code.

PR #2640, opened 2025-11-08
Add `swarm check` for parsing + checking `.sw` files from command-line

### LINKED ISSUE #2284: Standalone command-line option to typecheck a `.sw` file
We have `swarm format`, and we have the LSP, but sometimes it might be nice to have a quick command-line-friendly way to fully parse + typecheck a `.sw` file.  Concretely, I propose adding a new subcommand `swarm check` which is similar to `swarm format` but also typechecks the indicated file, and prints any errors (or a brief "OK" message) to stdout.

byorgey:
Add a `swarm check` subcommand, which takes the name of a `.sw` file, or else reads from standard input if given the `--stdin` flag, similarly to the `format` subcommand.  Prints `OK.` if the file parses + typechecks, otherwise prints out the same error message that would be printed at the REPL.  Also yields an appropriate exit code, so `swarm check` could be used as part of some automated shell process.

Closes #2284.

### DIFF
--- a/test/integration/TestFormat.hs
+++ b/test/integration/TestFormat.hs
@@ -11,6 +11,7 @@ import Data.Text qualified as T
 import Swarm.Language.Format
 import Swarm.Language.Parser.Core (LanguageVersion (SwarmLangLatest))
 import Swarm.Util (Encoding (..), findAllWithExt, readFileMayT)
+import Swarm.Util.InputSource (InputSource (Stdin))
 import Test.Tasty
 import Test.Tasty.HUnit
 import Witch (into)


src/swarm-util/Swarm/Util/InputSource.hs
```diff
@@ -0,0 +1,25 @@
...
+
+-- | Fetch input from the indicated location.
+getInput :: InputSource -> IO (Maybe Text)
+getInput Stdin = Just <$> T.getContents
+getInput (InputFile fp) = readFileMayT SystemLocale fp
+
+-- | Display an input source.
+showInput :: InputSource -> Text
+showInput Stdin = "(input)"
+showInput (InputFile fp) = T.pack fp
```
xsebek: Why not custom Show or Pretty instance? (Though I see this already was like that.)

src/swarm-lang/Swarm/Language/Pipeline/Cmdline.hs
```diff
@@ -0,0 +1,29 @@
...
+checkSwarmIO (CheckConfig input) = do
+  T.putStrLn $ "Checking " <> showInput input <> " ..."
+  mcontent <- getInput input
+  case mcontent of
+    Nothing -> T.hPutStrLn stderr $ "Could not read from " <> showInput input
+    Just content -> do
+      case processTerm content of
+        Right _ -> T.putStrLn "OK."
+        Left err -> do
+          T.putStrLn err
```
xsebek: I would suggest using `hPutStrLn` and error output for errors.

Some terminals color error stream automatically and even if not it's easier to script with separate errors.

src/swarm-lang/Swarm/Language/Pipeline/Cmdline.hs
```diff
@@ -0,0 +1,29 @@
...
+import Swarm.Util.InputSource
+import System.Exit (exitFailure)
+import System.IO (stderr)
+
+newtype CheckConfig = CheckConfig {checkInput :: InputSource}
+
+-- | Validate swarm-lang code.
+checkSwarmIO :: CheckConfig -> IO ()
+checkSwarmIO (CheckConfig input) = do
+  T.putStrLn $ "Checking " <> showInput input <> " ..."
```
xsebek: It's a bit of work, but it would be nice if this was just `putStr` and the OK got put on the same line.

xsebek (approved):

src/swarm-util/Swarm/Util/InputSource.hs
byorgey: Good idea, done.

src/swarm-lang/Swarm/Language/Pipeline/Cmdline.hs
byorgey: I actually had `hPutStrLn stderr` before, but then changed it to this, reasoning that showing error messages is actually part of the "normal" output of `swarm check`, not an error.  But I think you're probably right that printing to stderr is better here.

byorgey:
> Neat, what do you use it for @byorgey? 🙂

Nothing fancy. I think the other day when I wanted it, I wanted to make sure some `.sw` code still typechecked after making some changes to the codebase.  Being able to check on the command line would be quicker and easier than either opening the code in an editor (after first installing the newly updated code so my editor could find it for LSP) or starting the app, starting e.g. a creative scenario, and then trying to load the `.sw` file.

PR #2647, opened 2025-11-13
Removing `version` command in favour of `--version` flag (#2645)

### LINKED ISSUE #2645: Add `--version` flag
It would be nice to be able to query `swarm` on the command line to see what version it is.  Seems like this should be straightforward to implement.

Aster89:
With this change, I'm removing support for the `version` command, and exposing the `--version` option instead.

Closes #2645.

### DIFF
--- a/app/game/Main.hs
+++ b/app/game/Main.hs
@@ -1,7 +1,6 @@
 {-# LANGUAGE ApplicativeDo #-}
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE RecordWildCards #-}
-{-# LANGUAGE TemplateHaskell #-}
 
 -- |
 -- SPDX-License-Identifier: BSD-3-Clause
@@ -12,10 +11,9 @@ import Data.Foldable qualified
 import Data.List.Extra (enumerate)
 import Data.Set qualified as Set
 import Data.Text.IO qualified as T
-import GitHash (GitInfo, giBranch, giHash, tGitInfoCwdTry)
 import Options.Applicative
 import Options.Applicative.Help hiding (color, fullDesc)
-import Swarm.App (appMain, defaultMetrics)
+import Swarm.App (appMain, commitInfo, defaultMetrics, gitInfo)
 import Swarm.Language.Format
 import Swarm.Language.LSP (lspMain)
 import Swarm.Language.Parser.Core (LanguageVersion (..))
@@ -27,18 +25,9 @@ import Swarm.TUI.Model.KeyBindings (KeybindingPrint (..), showKeybindings)
 import Swarm.TUI.Model.UI (defaultInitLgTicksPerSecond)
 import Swarm.Util (Encoding (..), writeFileT)
 import Swarm.Util.InputSource
-import Swarm.Version
+import Swarm.Version qualified as V
 import Swarm.Web (defaultPort)
-import System.IO (hPrint, stderr)
 import Text.Read (readMaybe)
-
-gitInfo :: Maybe GitInfo
-gitInfo = either (const Nothing) Just $$tGitInfoCwdTry
-
-commitInfo :: String
-commitInfo = case gitInfo of
-  Nothing -> ""
-  Just git -> " (" <> giBranch git <> "@" <> take 10 (giHash git) <> ")"
 
 data CLI
   = Run AppOpts
@@ -47,7 +36,6 @@ data CLI
   | Format FormatConfig
   | Check CheckConfig
   | LSP
-  | Version
 
 cliParser :: Parser CLI
 cliParser =
@@ -57,7 +45,6 @@ cliParser =
         , command "format" (info (Format <$> parseFormat) (progDesc "Format a file"))
         , command "check" (info (Check <$> parseCheck) (progDesc "Parse and typecheck a file"))
         , command "lsp" (info (pure LSP) (progDesc "Start the LSP"))
-        , command "version" (info (pure Version) (progDesc "Get current and upstream version."))
         , command "keybindings" (info (ListKeybinding <$> initKeybindingConfig <*> printKeyMode <**> helper) (progDesc "List the keybindings"))
         ]
     )
@@ -66,6 +53,7 @@ cliParser =
   appOpts :: Parser AppOpts
   appOpts = do
     let repoGitInfo = gitInfo
+    version <- parseVersion
     userSeed <- seed
     userScenario <- scenario
     runOpts <- run <|> replay <|> autoplay
@@ -117,6 +105,9 @@ cliParser =
 
   seed :: Parser (Maybe Int)
   seed = optional $ option auto (long "seed" <> short 's' <> metavar "INT" <> help "Seed to use for world generation")
+
+  parseVersion :: Parser Bool
+  parseVersion = switch (long "version" <> short 'v' <> help "This is the version")
 
   webPort :: Parser (Maybe Int)
   webPort =
@@ -213,16 +204,10 @@ cliInfo :: ParserInfo CLI
 cliInfo =
   info
     (cliParser <**> helper)
-    ( header ("Swarm game - " <> version <> commitInfo)
+    ( header ("Swarm game - " <> V.version <> commitInfo)
         <> progDesc "To play the game simply run without any command."
         <> fullDesc
     )
-
-showVersion :: IO ()
-showVersion = do
-  putStrLn $ "Swarm game - " <> version <> commitInfo
-  up <- getNewerReleaseVersion gitInfo
-  either (hPrint stderr) (putStrLn . ("New upstream release: " <>)) up
 
 printKeybindings :: Bool -> KeybindingPrint -> IO ()
 printKeybindings initialize p = do
@@ -250,4 +235,3 @@ main = do
     Format cfg -> formatSwarmIO cfg
     Check cfg -> checkSwarmIO cfg
     LSP -> lspMain
-    Version -> showVersion
--- a/app/game/Swarm/App.hs
+++ b/app/game/Swarm/App.hs
@@ -1,5 +1,6 @@
 {-# LANGUAGE OverloadedRecordDot #-}
 {-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE TemplateHaskell #-}
 
 {- HLINT ignore "Use underscore" -}
 
@@ -12,6 +13,9 @@ module Swarm.App (
   app,
   appMain,
   EventHandler,
+  commitInfo,
+  gitInfo,
+  showVersion,
 
   -- * Metrics
   defaultMetrics,
@@ -34,7 +38,7 @@ import Data.List.NonEmpty qualified as NE
 import Data.Maybe (fromMaybe)
 import Data.Text qualified as T
 import Data.Text.IO qualified as T
-import GitHash (GitInfo)
+import GitHash (GitInfo, giBranch, giHash, tGitInfoCwdTry)
 import Graphics.Vty qualified as V
 import Graphics.Vty.CrossPlatform qualified as V
 import Network.Socket qualified as Net
@@ -48,6 +52,7 @@ import Swarm.TUI.Model.StateUpdate
 import Swarm.TUI.Model.UI (uiAttrMap)
 import Swarm.TUI.View
 import Swarm.Version (getNewerReleaseVersion)
+import Swarm.Version qualified as V (version)
 import Swarm.Web
 import System.Exit
 import System.IO (stderr)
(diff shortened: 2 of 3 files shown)


src/swarm-tui/Swarm/TUI/Model.hs
```diff
@@ -319,7 +319,8 @@ populateInventoryList (Just r) = do
 
 -- | Command-line options for configuring the app.
 data AppOpts = AppOpts
-  { userSeed :: Maybe Seed
+  { version' :: Maybe (IO ())
```
byorgey: Hmm, I don't like the idea of shoving an `IO ()` into a record of options.  Is there a particular reason this was necessary?  Could we just have `version :: Bool` in the record, and then in `appMain` call the appropriate `IO` action when the `Bool` is `True`?

Aster89: What you suggest was my initial attempt, but then I realized that it entailed moving more things around, so I opted for an `IO ()` in the record.

I've given another try to the `Bool` approach, and it works, but it requires moving a few helpers from `Main.hs` to `Swarm/App.hs`.

Is that ok?

byorgey: Moving stuff around is definitely OK.

app/game/Swarm/App.hs
```diff
@@ -68,9 +72,24 @@ app eventHandler =
...
+showVersion :: IO ()
+showVersion = do
+  putStrLn $ "Swarm game - " <> version <> commitInfo
+  up <- getNewerReleaseVersion gitInfo
+  either (hPrint stderr) (putStrLn . ("New upstream release: " <>)) up
+
 -- | The main @IO@ computation which initializes the state, sets up
 --   some communication channels, and runs the UI.
 appMain :: AppOpts -> IO ()
+appMain AppOpts {version' = True} = return ()
```
byorgey: So where does `showVersion` get run?  Should this really say `return ()`?

Aster89: Absolutely not! :sob:

I'm sorry, I think I've rushed a bit too much in addressing your comment of not using `IO ()`,  some excess of undo, and I ended up submitting `return ()` which is a placeholder before I imported `showVersion`. Sorry.

byorgey: No worries 😁

src/swarm-tui/Swarm/TUI/Model.hs
```diff
@@ -319,7 +319,8 @@ populateInventoryList (Just r) = do
 
 -- | Command-line options for configuring the app.
 data AppOpts = AppOpts
-  { userSeed :: Maybe Seed
+  { version' :: Bool
```
xsebek: Nitpick: I would prefer if the name did not include apostrophe.
I assume something conflicts with it, so either could change to unique name. (Or you could use GHC extensions to solve it, NoRecordFieldSelectors and OverloadedLabels can be nice.)

Another nitpick, if you put it at the end of the record, it will be closer to the related git hash and should result in marginally cleaner diff.

But feel free to ignore, I just wanted to point it out. The cleaner diff idea can help avoid merge conflicts.

app/game/Swarm/App.hs
```diff
@@ -68,9 +72,24 @@ app eventHandler =
...
+commitInfo :: String
+commitInfo = case gitInfo of
+  Nothing -> ""
+  Just git -> " (" <> giBranch git <> "@" <> take 10 (giHash git) <> ")"
+
+showVersion :: IO ()
+showVersion = do
+  putStrLn $ "Swarm game - " <> version <> commitInfo
+  up <- getNewerReleaseVersion gitInfo
+  either (hPrint stderr) (putStrLn . ("New upstream release: " <>)) up
```
xsebek: This is my mistake, but hPrint and the putStrLn are inconsistent.

Since you are touching it, could you please unwrap the tuple that goes into the first case?

```suggestion
  either (hPrint stderr . snd) (putStrLn . ("New upstream release: " <>)) up
```

app/game/Main.hs
```diff
@@ -118,6 +108,9 @@ cliParser =
   seed :: Parser (Maybe Int)
   seed = optional $ option auto (long "seed" <> short 's' <> metavar "INT" <> help "Seed to use for world generation")
 
+  ver :: Parser Bool
+  ver = switch (long "version" <> short 'v' <> help "This is the version")
```
xsebek: I think you can safely delete the version command. This is a more discoverable replacement.

If you start with the Version constructor, the compiler should guide you. 🙂

Aster89:
> Please remove the old version command, so that we don't have two ways to do the same.

@xsebek, [this comment](https://github.com/swarm-game/swarm/issues/2645#issuecomment-3529589637) from @byorgey says _we should have both_.

Can you please confirm what I should do?

Personally, I don't see why not having them both. There's long options and there's short options, so I don't see why not keeping an option given as argument. Unless the aim is precisely to simplify code.

xsebek:
@Aster89 Yes, the separate command gives us an option to do more with the version. But we never added flags to modify version output, like not checking GitHub or JSON output.

The point of the version checking GitHub is discoverability. Once the version flag is available, no one will think about looking for a version command.

So we might as well clean up the code, as it is unlikely the command will be used.

src/swarm-tui/Swarm/TUI/Model.hs
Aster89: > if you put it at the end of the record

Done.

> I would prefer if the name did not include apostrophe.

Yes, I used `version'` because of conflicts _and_ name shadownings.

I've tried renaming `version'` back to `version` throughout, and putting `{-# LANGUAGE NoFieldSelectors #-}` (I haven't found any `NoRecordFieldSelectors`, maybe it was a typo?) in the module where `AppOpts` is defined surely fixes the ambiguity that would otherwise arise in the `Swarm.Version` module. On the other hand, in the `Main` module I still have `This binding for ‘version’ shadows the existing binding   imported from ‘Swarm.Version’` at this line:
```haskell
-- in the same module as following snippet
 where
  appOpts :: Parser AppOpts
  appOpts = do
    let repoGitInfo = gitInfo
    version <- ver -- THIS LINE HERE
```
and that's because the module uses `Swarm.Version.version` unqualified here:
```haskell
-- in the same module as previous snippet
cliInfo :: ParserInfo CLI
cliInfo =
  info
    (cliParser <**> helper)
    ( header ("Swarm game - " <> version <> commitInfo) -- THIS LINE HERE
```
What should I do? Have a small qualified for this latter line? If so, what should it be? `S.version`? `V.version`?

Or does the solution lie in your suggestion to use `OverloadedLabels`? I haven't really understood yet how to use it.

app/game/Swarm/App.hs
Aster89: Ok for `. snd`, but as regards `hPrint` vs `putStrLn`, if you're suggesting to change `putStrLn` to `hPrint stdout`, it works obviously, but I also get a warning suggesting to use `print` instead. I suppose this would cause the tests to fail, no? What do you suggest?

xsebek: No it's simpler: hPrint took in input tuple, while putStrLn takes an input string. If you want, you can use `hPutStrLn` in the first case too now, that's a good idea.

Long way explanation: The result printed on output sometime looks like a string and sometimes like a tuple. The latter looks weird to show to the user. If we wanted to show the level, it would be better to custom print it. I don't think that it adds much value, so I suggested just `snd`. The custom print way could be a local helper function `printError (lvl, msg) = hPutStrLn stderr $ ...` but let's not bother. 🙂

src/swarm-tui/Swarm/TUI/Model.hs
Aster89: Oh, maybe this problem evaporates if I get  rid of the `version` command. Let me see.

xsebek: Re: the typo, yes, that was a mistake.

It's unfortunate that we have multiple things named version. You can see the same dilemma with `debug` and `debugOptions`. Maybe the simplest way is to disambiguate the imported name:
```Haskell
import Swarm.TUI.Model (AppOpts (..), ColorMode (..), RunOpts (..)) as Opts
import Swarm.Version as Executable
-- ...
Opts.version <- ver
"Swarm game - " <> Executable.version <> commitInfo
```

But this refactoring is getting a bit out of hand, so I don't require it. But if you want to play with it do go ahead, I personally enjoy these puzzles by the compiler. 🙂 

---

More advanced reading: The extensions I was referring to is a fairly new combination of not creating a getter function for each field that would cause a conflict in the global namespace ([`NoFieldSelectors`](https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/field_selectors.html#extension-FieldSelectors)).

Instead using an extension to get `x.myField` syntax to get the value ([`OverloadedRecordDot`](https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/overloaded_record_dot.html) which I mixed up with more arcane extension [`OverloadedLabels`](https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/overloaded_labels.html) used with lenses). 

To add more confusion, we use [`RecordWildCards`](https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/record_wildcards.html#extension-RecordWildCards) extension to fill the flags from the parsers in [`ApplicativeDo`](https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/applicative_do.html).

Sorry for the barrage of extensions, probably best not to worry about them now. 😅

Aster89: For now I've added some `qualified V` to disambiguate. I've not used `NoFieldSelectors` as I did in my longer comment above because that affects all other `data` in that module, and therefore it triggers errors in other files where fields are used as getters.

> But this refactoring is getting a bit out of hand, so I don't require it. But if you want to play with it do go ahead, I personally enjoy these puzzles by the compiler. 🙂
>
> [...]

Is this summary correct?

  - Do together, but possibly in a separate PR, the following
    - use `NoFieldSelectors` to avoid cluttering global namespace with field names (this way we temporarily lose a way to get fields out of objects),
    - and use `OverloadedRecordDot`  to get back access to those fields via `.field` syntax; this can be a large (if repetitive and not hard per se) task because we'd have to swtiching from `field obj` to `obj.field` syntax in all places a class is used that is defined in a module where we enable those two extensions.

xsebek: We don't need to do it for all records, just those where the name conflicts make the resulting code harder to read. 🙂 

It's fine to put those records in their own file.

Think of it as a tool rather than dogmatic new style. Those extensions are fairly new and namely setters are work in progress.

Aster89:
> Sorry @Aster89, I meant the Pull Request description.

Oh, ok, I think you referred to the fact that the description should mention that the command is being removed. Silly of me not re-reading that top message, sorry.

> Well, to me this would seem quite clear, but I leave that to you:
> 
> ```haskell
> Opts.version <- V.version -- or Executable.version
> ```

I assume by `V` you're referring to the short-name I gave to `Swarm.Version`, therefore by `V.version` you're alluding to this:
```haskell
-- app/game/Swarm/Version.hs
version :: String
version =
  let v = showVersion Paths_swarm.version
   in if v == "0.0.0.1" then "pre-alpha version" else v
```
But that's _not_ the one on the right hand side of `<-` below:
```haskell
-- app/game/Main.hs
version <- version
```
Here the conflict is between the `version` field of `Swarm.TUI.Model.AppOpts` and the `version` helper function defined in the `where` under `cliParser :: Parser CLI`:

```haskell
-- app/game/Main.hs
cliParser :: Parser CLI
cliParser =
  -- …
 where
  appOpts :: Parser AppOpts
  appOpts = do
    -- …
    version <- version
    -- …
  -- …
  version :: Parser Bool
  version = switch (long "version" <> short 'v' <> help "This is the version")
```

```haskell
-- src/swarm-tui/Swarm/TUI/Model.hs
data AppOpts = AppOpts
  { version :: Bool
  , userSeed :: Maybe Seed
```
Personally, I think renaming the function to `versionP`/`versionParser` is acceptable, if a bit ugly.

Or have I misunderstood your suggestion?

xsebek:
Ah, that makes sense. Sure, versionParser or versionFlag/Switch would both make it clear what is going on.

byorgey (approved):
@Aster89 thanks for putting up with all our questions + suggestions, looks good to me too!

byorgey:
@Aster89 , congrats on having your first PR accepted to Swarm!   We appreciate the contribution and we're really glad to welcome you as part of the [community](https://github.com/swarm-game/swarm/blob/main/COMMUNITY.md).  In a minute we will send you an invite to the repo (see https://github.com/swarm-game/swarm/blob/main/CONTRIBUTING.md#i-have-push-access-to-the-swarm-repository-now-what ).  If you use Discord, feel free to also join the Swarm server at https://discord.gg/kp8MuSgkPw .

PR #2665, opened 2025-11-24
feature: Unrecoverable Equipment Property

### LINKED ISSUE #918: Certain devices should not be restored to inventory upon being unequipped
Could indicate the ability to be restored to inventory when unequipped with a property called `reusable`.  Or, since maybe it will be more rare for devices to not be re-usable, come up with a word that means the opposite of `reusable` to indicate such.

Conceptually, something like a `compass` should be easy to remove, whereas perhaps `armor plates` would not be.

From a scenario design standpoint, this lets us devise more interesting resource sharing/distribution challenges.

stmonty:
Fixes #918 

Just adds a simple `unrecoverable` property that you can apply to an entity. If it can be equipped, and has the unrecoverable property, when unequiping it won't be added back to the inventory.

Let me know if I missed anything!

### DIFF
--- a/src/swarm-engine/Swarm/Game/Step/Const.hs
+++ b/src/swarm-engine/Swarm/Game/Step/Const.hs
@@ -440,10 +440,11 @@ execConst runChildProg c vs s k = do
       [VText itemName] -> do
         item <- ensureEquipped itemName
         myID <- use robotID
+        let recoverable = not (item `hasProperty` Unrecoverable)
 
         -- Speculatively unequip the item
         equippedDevices %= delete item
-        unless (item `hasProperty` Unrecoverable) $
+        when recoverable $
           robotInventory %= insert item
 
         -- If the base unequips the life support system, show a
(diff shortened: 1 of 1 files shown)


src/swarm-scenario/Swarm/Game/Entity.hs
```diff
@@ -183,6 +183,8 @@ data EntityProperty
     --   it disappears
     Evanescent
   | -- | Robots automatically know what this is without having to scan it.
+    Unrecoverable
```
byorgey: It looks like the comments got in the wrong order?  The "Unequipping..." comment should go *above* the `Unrecoverable` constructor, and the "Robots automatically know..." comment should go above the `Known` constructor.

stmonty: Oops you are right, put the comment in the wrong place I will fix that.

stmonty: I looked through the current entities and didn't really find anything that I think makes sense for `unrecoverable` right now. What do you think @byorgey ?

byorgey: OK, maybe you're right.  And I see that #918 does not make any such suggestion, it refers to some hypothetical entity called `armor plates`.  Just thought it was worth asking. :smile:

src/swarm-engine/Swarm/Game/Step/Const.hs
byorgey: I just realized we should also do this deletion conditionally on the item not being `unrecoverable`.  Otherwise, there's a weird case where the game doesn't allow you to unequip an unrecoverable item but one gets deleted from your inventory.

PR #2679, opened 2026-01-24
Get rid of `run` command and convert all scenarios to use `import`

### LINKED ISSUE #328: `run` prints value with bogus type
**Describe the bug**
Calling `run` on certain files causes the result of executing the file to be printed along with a unit type, regardless of the actual type of the result.

**To Reproduce**
In `true.sw`:
```
true
```
Now start swarm and at the REPL, type `run("true.sw")`  This will print out
```
true : ()
```

**Expected behavior**
It should result in
```
true : bool
```
(or perhaps it shouldn't print anything at all?).

### LINKED ISSUE #495: Change `run` to `import` and typecheck it properly

### LINKED ISSUE #1560: Don't re-parse the same import repeatedly
**Describe the bug**

It seems we might be too lazy with term processing, leading to terms being repeatedly processed during the tick:

| Inspecting profile with [Profiteur](https://jaspervdj.be/posts/2014-02-25-profiteur-ghc-prof-visualiser.html) |
|--------|
| Screenshot 2023-09-30 at 15 31 52 |

**To Reproduce**

Profile the non-inlined version in https://github.com/swarm-game/swarm/pull/1556#discussion_r1341942006.

Add a `cabal.project.local` file like this one:
```cabal
ignore-project: False

program-options
  ghc-options:
    -haddock
    -O2
    -fprof-auto
    "-with-rtsopts=-N -p -s -h -i0.1"

profiling: True
profiling-detail: all-functions
library-profiling: True
executable-profiling: True
```
Then run:
```shell
cabal run swarm -- -i data/scenarios/Challenges/wave.yaml --autoplay
profiteur swarm.prof
open swarm.prof.html
```

**Expected behavior**
I would expect there to be no (significant) `runParser` in tick profile, that should be done ages ago in `initAppState`.

### LINKED ISSUE #2068: `run` breaks type safety
**Describe the bug**
Executing `run` can overwrite existing definitions with new definitions having different types.  This can break type safety if the overwritten names were used in the same expression.

**To Reproduce**
Create a file `m.sw`:
```bash
echo "def m = move end" > m.sw
```
Now, in a Creative mode game:
```
> def m = 4 end; run "m.sw"; return (m + 1)
```
Results in a `Fatal error: bad application of execConst...`

**Expected behavior**
I don't even know what is expected.  This is just yet more evidence that `run` is broken; see #495 .   This would not be a problem with a proper `import` command since the second `m` would be typechecked with the type it has in `m.sw`, as it should.

**Additional context**
I thought of this when reading @xsebek's clever hacks in https://github.com/swarm-game/swarm/pull/2064 :smile:

### LINKED ISSUE #2182: Running a Swarm script from command line does not load definitions
**Describe the bug**

Running a Swarm script from the command line does not add the definitions to the environment.

**To Reproduce**

* run:
  ```sh
  cabal run swarm -O0 -- --scenario Blank --debug=all --run example/maybe.sw
  ```
* use one of the definitions:
  ```
  > just
  1:1: Unbound variable just
  ```
* there is also nothing in robot environment:
  ```sh
   curl -s localhost:5357/robot/0 | yq -P '.program'
   ```
  ```YAML
  Out:
    - VUnit: []
    - mu: []
      next: 0
    - []
  ```

**Expected behavior**

The player should be able to use the definitions from the run file.

### LINKED ISSUE #2602: `Challenges/wolf-goat-cabbage` solution fails
**Describe the bug**
The `solution` provided with the `Challenges/wolf-goat-cabbage` scenario does not work.

**To Reproduce**
Strangely, `swarm -i Challenges/wolf-goat-cabbage --autoplay` succeeds, yet when run via `cabal test`, it fails with the error:
```
Fatal error: Undefined variable crossRiver encountered while running the interpreter.
Please report this as a bug at
<https://github.com/swarm-game/swarm/issues/new>.
``` 

**Additional context**
This was discovered while working on #2599.  This scenario was not listed in in the integration tests, and apparently the solution has not been working for some time but no one ever noticed since it was not being tested.

byorgey:
In this PR:

* Get rid of the `run` command (good riddance!)
* Get rid of `IO` from the runtime engine, and introduce new effects for the things we were still using `IO` for: (1) mutable global cache and (2) getting the current time
* Split up `Language.Syntax.Import` into multiple modules
* Rename `getDataDirSafe` to `getDataDirThrow` since it actually throws an exception, and create an actually safe variant
* Add syntax `~swarm` as an import anchor to refer to things in the Swarm data dir, so you can write e.g. `import "~swarm/lib/control"` in order to import `data/lib/control.sw` from wherever it got installed
* Converts all scenarios to use `import`:
    *  Move a bunch of repeated code into the beginnings of a standard library
    * Typically, if we had `run "foo.sw"` and `foo.sw` contained `def ...; def ...; go` we remove `go` from the end of `foo.sw` and change `run "foo.sw"` to `import "foo"; go`

Closes #495.  Closes #328.  Closes #2068. Closes #2182. Closes #1560. Closes #2602.

### DIFF
--- a/test/unit/TestLanguagePipeline.hs
+++ b/test/unit/TestLanguagePipeline.hs
@@ -129,6 +129,12 @@ testLanguagePipeline =
                     ]
                 )
             )
+        , testCase
+            "Special warning for unbound 'run'"
+            ( process
+                "run \"foo.sw\""
+                "1:1: Undefined variable 'run'.  Note that the 'run' command was removed"
+            )
         ]
     , testCase
         "Parse pair syntax #225"


data/lib/control.sw
```diff
@@ -1,7 +1,60 @@
...
+def orC : Cmd Bool -> Cmd Bool -> Cmd Bool = \c1. \c2.
+  ifC c1 {pure true} {c2}
+end
+
+def while: ∀ a. Cmd Bool -> {Cmd a} -> Cmd Unit
+  = \test. \body.
+  ifC test {force body; while test body} {}
+end
+
+def until = \p. \c. q <- p; if q {} {c; until p c} end;
```
xsebek: DISCLAIMER: I have not tested this.
```suggestion
def until : ∀ a. Cmd Bool -> Cmd a -> Cmd Unit =
  \test. while (notC test)
end;
```

data/lib/function.sw
```diff
@@ -0,0 +1,3 @@
+def id = \t. t end
+
+def compose = \f. \g. \x. f (g x) end
```
xsebek: It would be nice if all top level functions had signatures.

Maybe we should teach `swarm format` to do that.

src/swarm-engine/Swarm/Game/Step.hs
```diff
@@ -79,28 +68,27 @@ import Swarm.Game.Step.RobotStepState
...
 import Swarm.Util.WindowedCounter qualified as WC
 import System.Clock (TimeSpec)
 import Witch (From (from), into)
 import Prelude hiding (lookup)
 
--- | GameState with support for IO and Time effect
+-- | GameState with support for Time, Metric, and Cache effects
 type HasGameStepState sig m =
   ( Has (State GameState) sig m
-  , Has (Lift IO) sig m
```
xsebek, line 101: 🥳

src/swarm-lang/Swarm/Language/Syntax/Constants.hs
```diff
@@ -782,7 +778,6 @@ constInfo c = case c of
   Random ->
     command 1 Intangible . doc (Set.singleton $ Query PRNG) "Get a uniformly random integer." $
       ["The random integer will be chosen from the range 0 to n-1, exclusive of the argument."]
-  Run -> command 1 long $ shortDoc (Set.singleton $ Mutation $ RobotChange BehaviorChange) "Run a program loaded from a file."
```
xsebek, line 785: Not to dampen the mood, but it would be nice to have some readable warning for `run`.

At least in the conversion mode from one language version to another it should say that we cannot convert automatically, but it is really easy to fix with import.

I am not sure if we want to have a syntax warning on anything named `run` or if it would be better to only customize the warning when no run definition is in scope.

src/swarm-util/Swarm/Effect/Cache.hs
```diff
@@ -0,0 +1,44 @@
+{-# LANGUAGE UndecidableInstances #-}
+
+-- |
+-- SPDX-License-Identifier: BSD-3-Clause
+-- Description: Global cache effects
```
xsebek:
```suggestion
-- Description: Global cache effect wrapper around Swarm.Util.GlobalCache which uses IO.
```

data/lib/function.sw
byorgey: Good idea: https://github.com/swarm-game/swarm/issues/2680

byorgey: I added a few type signatures and did a bit of formatting.  There's lots more stuff we could do to improve these libraries (more type signatures, comments, additional functions...) but I'd rather leave that for another PR.  And anyway I imagine scenario developers to be the primary userbase for these libraries, not players in general.  Players in general should write these themselves. =)

data/lib/control.sw
byorgey: This is behaviorally equivalent, but the reason the definition of `until` is the way it is instead of using `notC` is that using `notC` requires having a `lodestone` equipped whereas the way it is currently defined it requires only a `branch predictor`.

src/swarm-lang/Swarm/Language/Syntax/Constants.hs
byorgey: Hmm, yes, I was so excited to rip out `run` that I forgot about this. :sweat_smile:

You're right that we can't easily do automatic conversion.  But I agree it would be good to have some kind of warning.  I don't think we should make a warning on anything named `run`, because that would be annoying if anyone legitimately wants to make a function named `run`.  But a special warning when an undefined variable named `run` is encountered sounds good to me.  That will catch anyone who tries using old code, whether they specifically think to do `swarm format --v0.8` or not.  I will make a wiki page explaining how to upgrade and have the warning message link to it.

PR #2682, opened 2026-01-30
Move Fun, Vignettes, Mechanics + World Examples under Demo

### LINKED ISSUE #1088: Merge `Mechanics` folders into a single gallery of puzzle mechanics
Let's create a new toplevel scenarios category (i.e. a sibling of "Fun"/"Tutorials", etc.) to serve as a collection of demonstrations of basic building blocks for more complex challenges.  None of them need be fully-fleshed out scenarios with the obligation of polished maps and flavor text.  In fact, the more bare-bones, the better, for the purpose of copy-pasting into larger scenarios.

Things that may belong in this category:
* demonstration of "trapdoor": #976
* portal demonstration (a la [`teleport.yaml`](https://github.com/swarm-game/swarm/blob/main/data/scenarios/Challenges/teleport.yaml)/#406)
* minimal demo of [enclosure checking](https://github.com/swarm-game/swarm/blob/main/data/scenarios/Challenges/Ranching/_gated-paddock/enclosure-checking.sw) (from #835)
* invent different types of enemies/dynamic obstacles

byorgey:
All these, IMO, fall under the caetgory of "Demo" --- they are not fun or interesting scenarios to actually play, they are just demonstrating something --- they are fun to watch, demonstrating an interesting concept or mechanic, etc.

I think this closes #1088 - instead of merging the two `Mechanics` folders, I think there is a difference between a scenario merely *demonstrating* a mechanic, and one which is actually fun to play but depends on or exploits a particular mechanic.  But the former does not need to be its own independent top-level category.

### DIFF
--- a/data/scenarios/Demo/00-ORDER.txt
+++ b/data/scenarios/Demo/00-ORDER.txt
@@ -1,0 +1,4 @@
+Fun
+Vignettes
+Mechanics
+World Examples


xsebek (approved):
Sounds good to me 👍

PR #2683, opened 2026-02-01
Add `default` keyword for generating a default value of any type

### LINKED ISSUE #2192: Add generic default

I want to write a dummy implementation of a value of an unknown type and not throw an exception.

 I can not do that with `undefined` because Swarm is not lazy. It would have to be wrapped in either a function or delay.

I would like to have a new keyword `default : forall a. a`, which will provide a default value of that type:
* `0: Int`
* `"": Text`
* `noop: Cmd t`
* `inl default: (t1 + t2)`
* `\_. default: t0 -> t1`
* ...

I can write a definition like this:
```
def implementation_except_for_value = \value. [value=value, ...]
```
And provide some default at the call site, but this is not ergonomic.

**Additional context**

Admittedly, when #495 is implemented, I will not need dummy values as often.

byorgey:
`default` must be applied to a type literal, i.e. `default @Int` or `default @(Text -> Int)` or whatever.  It can also be used with type synonyms, as in `tydef X = [... record type ...] end; default @X`.  It is replaced with a default value *at compile time*, so `default` never has to be interpreted at run time.  Making it generate a default value at run time would be more flexible (it would allow e.g. higher-order uses) but getting access to a relevant context of type synonyms at runtime is tricky, so I figured static compile-time `default` value generation would be simpler.

`default` should support all inhabited types, including recursive types that have finite values.

Currently, `default` requires no special capability to use, but I could be convinced otherwise.

Closes #2192.

### DIFF
--- a/src/swarm-lang/Swarm/Language/Value.hs
+++ b/src/swarm-lang/Swarm/Language/Value.hs
@@ -31,14 +31,16 @@ module Swarm.Language.Value (
   restrictEnv,
 ) where
 
-import Control.Applicative ((<|>))
+import Control.Applicative (Alternative (..))
 import Control.Lens hiding (Const)
+import Control.Monad.Reader (ReaderT, ask, local, runReaderT)
 import Data.Bool (bool)
 import Data.Foldable (Foldable (..))
 import Data.Function (on)
 import Data.Hashable (Hashable, hash)
 import Data.Map (Map)
 import Data.Map qualified as M
+import Data.Set (Set)
 import Data.Set qualified as S
 import Data.Set.Lens (setOf)
 import Data.Strict.Tuple (Pair (..))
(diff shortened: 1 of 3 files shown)


src/swarm-lang/Swarm/Language/Value.hs
```diff
@@ -270,3 +274,45 @@ valueToTerm = \case
...
+-- This following seems like an obvious way to implement 'default'
+-- for recursive types:
+--
+--   defaultValue tdCtx (whnfType emptyTDCtx ty)
+--
+-- but it can get stuck in infinite recursion! Consider e.g. the type
+-- (rec l. (Int * l) + Unit).  Since we default to generating the left
+-- side of a sum type, this will recurse forever generating inl (0,
+-- inl (0, inl (0, ...))).  To support recursive types we would have
+-- to do more sophisticated analysis to find a finite value to pick.
```
xsebek: Would a set of encountered types (?) in recursion do the trick? 🤔 

It seems the only possible problem are loops and we should be able to detect them if types have Ord.

byorgey: Oh, good idea.  Your comment was just what I needed to see how to make it work.  It wasn't all that hard in the end, so I added it. :smile:

Issue #2504, opened 2025-06-25
More elaborate + extensible help system


byorgey:
Currently we just have a single Help dialog.  But there is so much more players might want help with:

- Documentation of the contents of other dialogs (e.g. #1495)
- A Swarm language reference (#1520)
- Tips + tricks
- Explanation of concepts like tangible + intangible commands
- Documentation of the world description format

Not to mention things *developers* might like in-game help with.  A single dialog is definitely not enough.

For starters, I'm imagining something like the following:

- A collection of help pages, each with a title and Markdown contents
- When viewing a help page, a way to navigate among and select links on the page, which could be external URLs or links to other help pages
- A way to search all help pages for a certain phrase

The pages can be stored in simple `.md` files, organized in a folder hierarchy which we can use to auto-generate a table of contents.  We could even use `00-ORDER` files to specify the order, just like we do for scenarios.  As a bonus, if all the help is stored in external `.md` files, we can easily compile it into other formats, such as static HTML pages.

PR #2688, opened 2026-02-17
Factor out generic utilities for loading recursively nested collections from disk

byorgey:
Towards #2504 .  As soon as I started working on #2504 I quickly realized that I wanted to load a whole tree of documentation from disk, which was going to involve a lot of code very similar to the code we already use to load the collection of scenarios.  In fact #2504 even mentions possibly reusing the idea of having an `00-ORDER.txt` file to specify the ordering of documentation pages within a folder.

So, as a first step, this PR factors out the generic recursively-loading-stuff-from-disk part from the scenario-specific parts.  In particular:

- Move some `OrderedMap` utilities into their own module
- Change the `ScenarioCollection` and `ScenarioItem` types into generic `Collection` and `CollectionItem` types
- Create `loadCollection` and `loadCollectionItem` functions which implement the generic logic (including reading `00-ORDER.txt` files), parameterized by a `CollectionConfig` record which specifies how to load individual items, and how to decide which items to load and which to ignore.
- Obviously we may decide to extend the `CollectionConfig` record in the future should we decide that additional configuration options are needed.
- Collection loading stuff is in `Swarm.ResourceLoading.Collection` and re-exported from `Swarm.ResourceLoading`.

### DIFF
--- a/src/swarm-util/Swarm/ResourceLoading/Collection.hs
+++ b/src/swarm-util/Swarm/ResourceLoading/Collection.hs
@@ -31,7 +31,6 @@ import Control.Effect.Lift (Lift, sendIO)
 import Control.Effect.Throw (throwError)
 import Control.Lens (Ixed (..), Traversal', makePrisms)
 import Control.Monad (filterM, forM_, when)
-import Data.Char (isSpace)
 import Data.Either (partitionEithers)
 import Data.List ((\\))
 import Data.List.NonEmpty qualified as NE
@@ -166,7 +165,7 @@ loadCollection cfg dir = do
     add (Seq.fromList itemFailures)
     return $ M.fromList okItems
 
-  -- Load a collection with items in no particular order, and
+  -- Load a collection with items sorted alphabetically by file path, and
   -- optionally warn that the ORDER file is missing.
   loadUnorderedCollection :: Map FilePath (CollectionItem a) -> m (Collection a)
   loadUnorderedCollection collectionItemMap = do


src/swarm-util/Swarm/ResourceLoading/Collection.hs
```diff
@@ -0,0 +1,200 @@
...
+  ixp [] _ col = pure col
+  ixp [s] f (Collection m) = Collection <$> ix s f m
+  ixp (d : xs) f (Collection m) = Collection <$> ix d inner m
+   where
+    inner si = case si of
+      Single {} -> pure si
+      SubCollection n' col -> SubCollection n' <$> ixp xs f col
+
+-- | The name of the special file which indicates the order of
+--   scenarios in a folder.
```
xsebek:
```suggestion
-- | The name of the special file which indicates the order of
--   items in a directory.
```

src/swarm-util/Swarm/ResourceLoading/Collection.hs
```diff
@@ -0,0 +1,200 @@
...
+orderFileName :: FilePath
+orderFileName = "00-ORDER.txt"
+
+-- | Read a special 00-ORDER.txt file indicating the order of
```
xsebek, line 119:
```suggestion
-- | Read a special 00-ORDER.txt file indicating the order of
--   items in a directory.
```

src/swarm-util/Swarm/ResourceLoading/Collection.hs
```diff
@@ -0,0 +1,200 @@
...
+  -- or has thrown SystemFailure. The following code just adds that thrown failure to others.
+  loadItems :: [FilePath] -> m (Map FilePath (CollectionItem a))
+  loadItems items = do
+    let loadItem f = runError @SystemFailure $ (f,) <$> loadCollectionItem cfg (dir </> f)
+    (itemFailures, okItems) <- partitionEithers <$> mapM loadItem items
+    add (Seq.fromList itemFailures)
+    return $ M.fromList okItems
+
+  -- Load a collection with items in no particular order, and
+  -- optionally warn that the ORDER file is missing.
```
xsebek: Isn't this an alphabetical order based on file name?

src/swarm-util/Swarm/ResourceLoading/Collection.hs
```diff
@@ -0,0 +1,200 @@
...
+  ( Has (Error SystemFailure) sig m
+  , Has (Accum (Seq SystemFailure)) sig m
+  , Has (Lift IO) sig m
+  ) =>
+  CollectionConfig a ->
+  FilePath ->
+  m (CollectionItem a)
+loadCollectionItem cfg path = do
+  isDir <- sendIO $ doesDirectoryExist path
+  let collectionName = into @Text . dropWhile isSpace . takeBaseName $ path
```
xsebek: What is `dropWhile isSpace` doing here? I would expect a `map trim nonEmptyLines` in `readOrderFile` would better take care of leading and trailing whitespace.

src/swarm-util/Swarm/ResourceLoading/Collection.hs
```diff
@@ -0,0 +1,200 @@
...
+flattenCollection :: Collection a -> [a]
+flattenCollection = concatMap flattenCollectionItem . collectionToList
+ where
+  flattenCollectionItem :: CollectionItem a -> [a]
+  flattenCollectionItem = \case
+    Single a -> [a]
+    SubCollection _ c -> flattenCollection c
+
+-- | Access and modify 'CollectionItem's based on their path.
+collectionItemByPath :: FilePath -> Traversal' (Collection a) (CollectionItem a)
```
xsebek, line 96: Most usages immediately follow up with `_Single`. Could we have an `At` instance that would get the singleton item? Then the usage would look like:
```diff
- appState ^? playState . progression . scenarios . collectionItemByPath sPath . _Single . getScenarioInfo
+ appState ^? playState . progression . scenarios . at sPath . getScenarioInfo
```

byorgey: Good idea.  I wasn't able to get an actual `At` instance to work (I couldn't get the types to line up --- not sure if there's something fundamentally impossible about it or if I just didn't hit upon the right way).  However I was able to add an `atPath` traversal so we can say `atPath sPath`.

src/swarm-util/Swarm/ResourceLoading/Collection.hs
byorgey: I mean, in practice it seems to be alphabetical, but really it depends on the order in which files are returned by `listDirectory`.  The documentation for `listDirectory` does not make any guarantees about the order of the resulting list: https://hackage-content.haskell.org/package/directory-1.3.10.1/docs/System-Directory.html#v:listDirectory  I presume it just depends on the particular order of files returned by some OS call.

I suppose we could apply an explicit `sort` in order to guarantee alphabetical order.  @xsebek do you think that would be worthwhile?

src/swarm-util/Swarm/ResourceLoading/Collection.hs
byorgey: I... am not sure.  It would be strange if a folder name started with spaces, wouldn't it?

src/swarm-util/Swarm/ResourceLoading/Collection.hs
byorgey: Oh, never mind, you're right, it will be alphabetical.  The items get stored in a `Map` after being loaded from disk, so they will end up sorted by `FilePath` no matter what order `listDirectory` returns them in.

PR #2690, opened 2026-02-20
List upcoming goals after active ones

### LINKED ISSUE #1169: First active goal should be highlighted by default
Currently, when there exist "Upcoming" goals, they are displayed by default to the user upon initial popup of the Goals dialog, since they are ordered first.

However, it is the "Active" goals that show the "immediately relevant" instructions to the user, so they should be shown by default instead.

In the example screenshot, the "Wait patiently" goal is the one that **should** be shown, but instead the "Flower curation" goal is shown (which can't be acted upon yet).

image

byorgey:
Closes #1169, by implementing the second idea I listed there: fix an order in which to display goal categories and update the display code so it explicitly uses that order, rather than implicitly relying on the order in which the `GoalStatus` constructors were declared.  The downside is that if we were to ever add a new type of `GoalStatus` we would need to manually update that list in order to display those goals.  However, if someone adds a new goal type it should be obvious if those goals are not being displayed.

To see the difference, try first running
```
cabal run -O0 swarm:exe:swarm -- -i Testing/795-prerequisite/795-prerequisite-or
```
and note that the first highlighted goal is "Upcoming":
before

Then, after applying this patch, run the same command; the first highlighted goal is now one of the "Active" ones:
after

### DIFF
--- a/src/swarm-tui/Swarm/TUI/View/Objective.hs
+++ b/src/swarm-tui/Swarm/TUI/View/Objective.hs
@@ -15,6 +15,7 @@ import Control.Lens hiding (Const, from)
 import Data.List (intercalate)
 import Data.List.NonEmpty qualified as NE
 import Data.Map.Strict qualified as M
+import Data.Maybe (mapMaybe)
 import Data.Vector qualified as V
 import Swarm.Game.Scenario.Objective
 import Swarm.Language.Syntax (Anchor, ImportPhaseFor, Phase (..), Syntax, Unresolvable)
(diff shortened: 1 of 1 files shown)


src/swarm-tui/Swarm/TUI/View/Objective.hs
```diff
@@ -31,8 +32,11 @@ makeListWidget :: GoalTracking -> BL.List Name GoalEntry
 makeListWidget (GoalTracking _announcements categorizedObjs) =
   BL.listMoveTo 1 $ BL.list (GoalWidgets ObjectivesList) (V.fromList objList) 1
  where
-  objList = intercalate [Spacer] $ map f $ M.toList categorizedObjs
-  f (h, xs) = Header h : map (Goal h) (NE.toList xs)
+  categoryDisplayOrder :: [GoalStatus]
+  categoryDisplayOrder = [Active, Upcoming, Completed, Failed]
```
kostmo, line 36: Would prefer that we had the protection of exhaustiveness checking.

What about a manual implementation of Ord?  Then the docstring of that instance could explain the purpose, and additions to the enum would not silently break the display order.

Personally I'd be fine with option 1 plus a docstring.

byorgey: Hmm, good point.  Yes, a manual implementation of `Ord` is a good idea too, I didn't think of that.

byorgey: But maybe you're right, option 1 is fine as long as we carefully document what is going on, especially since that's the only place we use that particular data type.

PR #2701, opened 2026-03-16
Immutable arrays

### LINKED ISSUE #98: Array type

byorgey:
Adds an immutable array type to the language.

- `Array T` is the type of immutable arrays with elements of type `T`.
- Literal arrays are written `[| a, b, c |]`.
- Typing `|` in the REPL now inserts a matching pair, so when you type `[|` the closing `|]` is automatically inserted.  The only other place we use `|` in the syntax that I can think of is in the boolean operator `||`, but typing `|` two times still enters `||` just as before (since the second one moves over/past the `|` already under the cursor).
- There are new primitives:
    - `arraySize : Array a -> Int`
    - `indexArray : Array a -> Int -> a`
    - `unfoldArray : b -> (b -> (Unit + (a * b))) -> Array a`
    - `unfoldArrayC : b -> (b -> Cmd (Unit + (a * b))) -> Cmd (Array a)`
- The `pigeon coop` device (recipe: 8 `box` + 2 `Elmer's glue`) allows making arrays, and using the `arraySize` and `indexArray` functions.
- The `oriagmi pigeon` decide (recipe: 1 `pigeon coop` + 1 `paper`) allows using `unfoldArray` and `unfoldArrayC`.
- Adds `lib/array.sw` with some variant introduction forms implemented in terms of `unfoldArray`, conversions to/from lists, merge sort, and Fisher-Yates shuffle.
- Adds a few scenarios that illustrate the use of arrays.

I also experimented with more complex versions of `unfoldArray` which allowed accumulating values out of order; for example, if you wanted to use an array to do a frequency count of a string.  However, doing this with immutable arrays was super tricky and I ultimately decided it would be simpler to just add mutable arrays plus a `freeze` operation.

Closes #98.


src/swarm-engine/Swarm/Game/CESK.hs
```diff
@@ -181,6 +181,11 @@ data Frame
     FRcd Env [(Var, Value)] Var [(Var, Maybe (Term Resolved))]
   | -- | We are in the middle of evaluating a record field projection.
     FProj Var
+  | -- | In the middle of evaluating an array literal.  Elements
+    --   already evaluated + environment + remaining elements.
+    FArray [Value] Env [Term Resolved]
+  | -- | In the middle of evaluating a call to unfoldArray.
```
byorgey: Should expand on this comment and explain what the fields represent.

src/swarm-lang/Swarm/Language/Syntax/AST.hs
```diff
@@ -159,6 +159,8 @@ data Term phase
     SRcd [(LocVar, Maybe (Syntax phase))]
   | -- | Record projection @e.x@
     SProj (Syntax phase) Var
+  | -- | Array literal @[e1, e2, ..., en]@
```
byorgey: Fix the syntax

data/scenarios/Challenges/Algorithmic/iterate-perm-hard.yaml
```diff
@@ -0,0 +1,270 @@
...
+        repeatedly will be MUCH better!
+      - |
+        The operation performed by the mysterious person overnight
+        consists of moving the letter in each position to a (possibly)
+        new position.  For example, if the letter A is now in the
+        third position, that means they took the letter in the first
+        position and moved it to the third position.  Your goal is to
+        perform the same operation (i.e. taking the letter that is now
+        in the first position and moving it to the third position, and
+        so on) one million more times.
```
xsebek: This is unhinged, I love it! 🤣 

How about making it an optional second goal after finishing the scenario? 🤔 I am not sure if we save a list of completed "side quests"/"bonus objectives", but if not we should!

data/scenarios/Challenges/Algorithmic/paper-sorting.yaml
```diff
@@ -0,0 +1,281 @@
...
+objectives:
+  - teaser: Sort
+    goal:
+      - |
+        Someone knocked over a stack of papers, and now they are
+        hopelessly jumbled all over the floor!
+      - |
+        The first step is to put the papers back in order.  Each paper
+        has a capital letter written on it; you should arrange them in
+        alphabetical order from left to right.
```
xsebek, line 16: Since this is about showing an algorithm, would it make sense to add a TIP about arrays?

I had to look them up in Commands and they are all the way at the bottom because of all the devices that base has.

data/scenarios/Challenges/Algorithmic/_paper-sorting/solution.sw
```diff
@@ -0,0 +1,13 @@
+import "~swarm/lib/control"
+
+def placeAll = \c.
+  let p = "paper: " ++ toChar (c + 65) in
+  while (has p) {place p; move}
+end
```
xsebek, line 21: Is this supposed to be a placeholder? I could not use `inl` or `case` and thus the "paper crane" does not seem to be the key as I did not figure out how to use `unfoldArrayC` here. 🤔

src/swarm-lang/Swarm/Language/Syntax/Constants.hs
```diff
@@ -284,6 +284,16 @@ data Const
     CharAt
   | -- | Create a singleton text value with the given character code.
     ToChar
+  | -- Array operations
+
+    -- | Create an array via an unfold.
+    UnfoldArray
+  | -- | Create an array via an effectful unfold.
+    UnfoldArrayC
```
xsebek, line 292: Is having both pure and effectful version necessary? It seems unlike other commands and would be nice to have just one.

Is it possible to define `unfoldArrayC` in swarm and not make it too inefficient?

data/entities.yaml
```diff
@@ -405,6 +405,28 @@
...
+      ```
+  properties: [pickable]
+  capabilities: [array, indexArray, arraySize]
+- name: origami pigeon
+  display:
+    attr: grey
+    char: 'P'
+  description:
+    - |
+      Unfolding this folded pigeon brings a folder of unfolding possibilities into the fold. More on this unfolding story below the fold.
```
xsebek, line 427: Originally, I was put off by the complex unfold function, but this backstory won me over.

I was even ready to propose replacing unfold with something simpler, but that would risk losing this origami piece. 😆

xsebek (commented):
I wanted to play through the challenges, but got stuck on the paper one.

data/scenarios/Challenges/Algorithmic/_paper-sorting/solution.sw
byorgey: Ah, actually, I was trying to add a scenario that would require/reward the use of arrays, but instead I realized that you could solve it in a simpler way without using arrays at all.  I should have removed the `pigeon coop` and `origami pigeon` devices but I forgot.

data/scenarios/Challenges/Algorithmic/paper-sorting.yaml
byorgey: This scenario is not supposed to use arrays, but point taken: I will add a TIP on the other scenarios.

src/swarm-lang/Swarm/Language/Syntax/Constants.hs
byorgey: I agree it would be nice to have just one version, and I thought about this very carefully.
- Of course it's not possible to define the pure version in terms of the effectful version, because we have no way to "escape" the `Cmd` monad.
- We could simply provide the effectful version, and you could still use it even if you didn't want any effects (using `pure` etc.) but that would be super annoying: any time you wanted to use an array, it would "infect" your code with `Cmd` even if everything you are doing is pure computation.
- What about defining the effectful version in terms of the pure version?  On the surface, this seems ideal.  For example, if we provided a built-in list fold, we would not need to provide a separate effectful list fold, because you can just pass `Cmd` values to the regular pure fold.  However, I don't think this works with `unfoldArray`.  In the hypothetical list fold example, you are folding a list in order to produce a single `Cmd` that you can execute afterwards.  With `unfoldArrayC`, however, each seed value produces a `Cmd` which must be executed *in order to determine* whether we are done unfolding or to produce a new seed value.  In other words, `unfoldArrayC` represents an *interleaving* of execution and computation, and there is no way to simulate that with a pure function that produces a `Cmd`.
- You also asked if we could define `unfoldArrayC` in swarm-lang instead of having it built-in, without making it too inefficient.  I am... not sure.  Perhaps, though I think it would require recursive types; the idea would be to collect a list of values and then use `unfoldArray` at the very end.  I will have to play with it and see if this works.  It would definitely be less efficient, though perhaps not asymptotically so.
- In any case I should note that at least we get `unfoldArray` and `unfoldArrayC` essentially "for free": the code to implement them in the CESK machine is mostly shared.  So it's really just an issue of complexity for the user/player, not for our implementation.

byorgey: I did manage to implement `unfoldArrayC` in terms of `unfoldArray` + recursive types (code below).  It is noticeably slower (though not asymptotically so).

*Maybe* it is possible to do it without recursive types, using continuations or Scott-encoded lists or something...?
```
import "~swarm/lib/control"
import "~swarm/lib/list"

def reverse : List a -> List a = \xs.
  let rev : List a -> List a -> List a = \acc. λcase
    (\_. acc)
    (λmatch \x. rev (inr (x, acc)))
  in  rev (inl ()) xs
end

def unfoldArrayCPrim : b -> (b -> Cmd (Unit + a * b)) -> Cmd (Array a) = \seed. \next.
  let go : List a -> b -> Cmd (Array a) = \acc. \seed.
    res <- next seed;
    case res
      (\_. pure (unfoldArray (reverse acc) (\l.l)))
      (λmatch \e. \seed'. go (inr (e, acc)) seed')
  in go (inl ()) seed
end
```

data/scenarios/Challenges/Algorithmic/iterate-perm-hard.yaml
byorgey: I don't think we do save a list of completed optional objectives.  I agree that we should!  #2754 .

In any case, yes, I think making this into an optional bonus goal instead of a separate scenario makes sense.

src/swarm-lang/Swarm/Language/Syntax/Constants.hs
byorgey: I am torn.  On the one hand, the guiding principle of providing only primitive language features and letting users build up more complex things on top suggests we should not provide `unfoldArrayC`.  On the other hand, `unfoldArray` by itself is not all that useful --- *e.g.* there are many situations in which you want to create an array by physically scanning the environment, which requires an effectful version --- and it feels annoying to make users wait until they have recursive types to be able to really use arrays.  And there is also the issue of efficiency.

byorgey: @xsebek any more thoughts on this?

PR #2714, opened 2026-05-16
constrain lsp-types

### LINKED ISSUE #2713: scripts/build-game.sh fails on lsp-types
System: Ubuntu 26.04, aarch64

From `main` branch (commit `14f80096`), ran:
```
scripts/build-game.sh
```

Tried with both GHC 9.6.7 and GHC 9.8.4, but got the same error:

```
src/Language/LSP/Protocol/Message/Method.hs:64:10: error: [GHC-39999]
    • No instance for ‘Data.EqP.EqP SMethod’
        arising from the superclasses of an instance declaration
    • In the instance declaration for ‘GEq SMethod’
   |
64 | instance GEq SMethod where
   |          ^^^^^^^^^^^

src/Language/LSP/Protocol/Message/Method.hs:72:10: error: [GHC-39999]
    • No instance for ‘Data.OrdP.OrdP SMethod’
        arising from the superclasses of an instance declaration
    • In the instance declaration for ‘GCompare SMethod’
   |
72 | instance GCompare SMethod where
   |          ^^^^^^^^^^^^^^^^

[...]
Failed to build lsp-types-2.1.0.0 (which is required by exe:swarm from swarm-0.8.0.0). See the build log above for details.
```

Incidentally, `lsp` is a rather large dependency.  Should we really require it as a dependency of `swarm-lang`?

kostmo:
Closes #2713

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -2,4 +2,5 @@ packages: .
 
 constraints:
   # Required for build success with ghc-9.6 and ghc-9.8
+  # See https://github.com/swarm-game/swarm/issues/2713
   lsp-types >= 2.1.1
--- a/cabal.project
+++ b/cabal.project
@@ -1,0 +1,5 @@
+packages: .
+
+constraints:
+  # Required for build success with ghc-9.6 and ghc-9.8
+  lsp-types >= 2.1.1


byorgey:
I was expecting to see the constraint added to the `.cabal` file.  Is there a benefit to putting it in a `cabal.project` file?

kostmo:
> I was expecting to see the constraint added to the `.cabal` file. Is there a benefit to putting it in a `cabal.project` file?

Only ignorance on my part regarding best practices.  My first thought was indeed to add it to `swarm.cabal`, but it doesn't support a `constraints` clause AFAICT.  The [documentation](https://cabal.readthedocs.io/en/3.4/cabal-project.html#cfg-field-constraints) pointed me to `cabal.project`.

byorgey:
`.cabal` doesn't support a standalone `constraints` clause that I know of, but you can specify constraints on each dependency.

Ohhh, hmm, I see what the issue is --- I didn't realize that we don't directly depend on `lsp-types`, so there's nowhere to specify a lower bound without adding it as a dependency.  In that case I think a `cabal.project` file is probably the appropriate place to specify it.

cabal.project
```diff
@@ -0,0 +1,5 @@
+packages: .
+
+constraints:
+  # Required for build success with ghc-9.6 and ghc-9.8
```
byorgey, line 4: Can we add a link to #2713 in the comment?

byorgey (approved):

whaypr:
Hello,

This fix also resolves the issue described in #2713 for me, but the build currently fails due to incorrect comment syntax. It should look like this:

```
packages: .

constraints:
  -- Required for build success with ghc-9.6 and ghc-9.8
  -- See https://github.com/swarm-game/swarm/issues/2713
  lsp-types >= 2.1.1

```

byorgey:
Hmm, does that mean our CI ignores the `cabal.project` file?  That's not great since it means any changes to that file aren't really being tested...

byorgey:
@whaypr Thanks for the report!  I've just opened #2715 which should hopefully fix this --- if you have a chance, would you mind checking if it also resolves #2713 for you?

byorgey:
@all-contributors please add @whaypr for bug, infra

whaypr:
> @whaypr Thanks for the report! I've just opened #2715 which should hopefully fix this --- if you have a chance, would you mind checking if it also resolves #2713 for you?

You're welcome, thanks for the quick reaction! I confirm, that #2715 indeed resolved #2713 for me!

PR #2727, opened 2026-05-26
Make sure changes to nested imports get picked up

### LINKED ISSUE #2725: Nested imports not checked for changes
**Problem**

`import` checks to see if the file has been changed since last `import` and doesn't reload it if it hasn't. But it doesn't check any files the imported file might `import` as well. So if I update my `commonroutines.sw`, I'll have to touch my `world.sw` to make sure the changes are loaded.

**Reproduction**

Import `a.sw` which in turn imports `b.sw`. Have a function in `a.sw` that calls a function in `b.sw`, then reimport `a.sw`. Now change the function's behaviour in `b.sw` and reimport `a.sw`. 

The new functionality will not be loaded.

**Expected behavior**

After `import "a.sw"`, I would expect all dependencies to be checked for changes, too. Or that the game just reimports everything.

byorgey:
Fixes #2725.  Every time we process some new top-level source code, first do a DFS through the module cache and remove any modules which have been modified on disk more recently than the cached version, OR which depend (transitively) on any such outdated module.  This guarantees that if e.g. `a` imports `b` and then `b` is changed, importing `a` will automatically pick up the new version of `b`.

### DIFF
--- a/src/swarm-lang/Swarm/Language/Load.hs
+++ b/src/swarm-lang/Swarm/Language/Load.hs
@@ -27,6 +27,7 @@ import Control.Monad (forM_, when)
 import Data.Function ((&))
 import Data.Map.Ordered (OMap)
 import Data.Map.Ordered qualified as OM
+import Data.Maybe (isNothing)
 import Data.Set (Set)
 import Data.Set qualified as S
 import Data.Text (Text)
@@ -129,6 +130,11 @@ resolve ::
   Syntax Raw ->
   m (SourceMap Resolved, (LocalModules, Syntax Resolved))
 resolve prov s = do
+  -- First, remove any scale module cache entries, defined as those
+  -- which have an on-disk modification time more recent than the
+  -- cache timestamp, or which have any stale dependencies.
+  cullModuleCache
+
   cur <- resolveImportDir $
     case prov of
       Nothing -> currentDir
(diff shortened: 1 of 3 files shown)


src/swarm-lang/Swarm/Language/Load.hs
```diff
@@ -151,6 +156,48 @@ resolve' ::
   (Has (Throw SystemFailure) sig m) => Syntax Raw -> m (Syntax Resolved)
 resolve' = traverseSyntax pure (throwError . DisallowedImport)
 
+-- | Do a DFS through the module cache, removing any stale modules.  A
+--   stale module is defined as one which EITHER (a) has a
+--   modification time on disk newer than the cached timestamp, or (b)
+--   has a stale dependency.
+cullModuleCache :: (Has (Lift IO) sig m, Has (Throw SystemFailure) sig m) => m ()
+cullModuleCache = dfs =<< sendIO (GC.cacheKeys moduleCache)
```
xsebek: What does cache keys return? 🤔 

I would expect it to return all cached modules, which would make it pointless to do a DFS for each.

xsebek (approved):
Nice! 👍 I skimmed through your ideas in the issue and this solution is indeed nicely noninvasive.

src/swarm-lang/Swarm/Language/Load.hs
byorgey: Oh, good catch!!  Yes, `cacheKeys` does return all cached modules, and we call `dfs` on each.  The problem was that the `evalState` was in the wrong place: it should be *outside* the calls to `dfs`, so that e.g. if we call `dfs` on a module that has already been visited we just ignore it and move on.  Should be fixed now.  I also moved the place where we add modules to the visited set to *before* recursively processing their children instead of *after*.  This shouldn't make a difference in theory, since there can't be any import cycles in the module cache... but it just seemed safer to mark a module as visited *as soon as* we start processing it.

xsebek:
@byorgey it might be too much work, but could you write a test for this?

Something along the lines of creating a loooong chain of imports from Haskell in /tmp, load, change base import and reload.

Or I could try it, but I can't promise when I will have time.

byorgey:
Hmm, I might be able to.  Maybe we can also instrument the cache somehow so we know whether we succeeded in doing only O(n) work?  I'll give it some thought.

xsebek:
@byorgey if you rebase on main once #2731 is merged, you should be able to easily benchmark the reload when you make a change in `a_000.sw`. 🙂

byorgey:
Can confirm the reload is still quick when modifying `a_000.sw` (and even quicker when not).

PR #2731, opened 2026-05-29
Fix Accum effect stack

### LINKED ISSUE #2729: Swarm freezes on chain of imports
**Describe the bug**

I tried to break swarm with a chain of imports $a_n$ imports $a_{n-1}$ and succeeded.

**To Reproduce**

Create base:
```
def a_000 = 000 end
```

Then run this script:
```python
import textwrap

for i in range(1, 1000):
    with open(f"a_{i:03}.sw", 'w') as a:
        print(textwrap.dedent(f"""
                import "a_{i-1:03}"
                def a_{i:03} = a_{i-1:03} end
                """).strip(),
            file=a)
```

**Expected behavior**

Swarm should never hang, the TUI should stay responsive even while import is processing.

**Screenshots**

I think it is running at 100% of single CPU core because it is not parallelized.

Image

**Additional context**

I originally tried making an exponential DAG with a/b files, but that was overshooting by a lot.

xsebek:
* use `AccumC` constructor directly
* add warning to docs
* add benchmarks

|  | 0 | 1 | 10 | 20 | 50 | 100 |
|:--:|-:|-:|-:|-:|-:|-:|
| before | 0.0004s | 0.001s | 0.033s | 20.722s | ? | ? |
| after | 0.0004s | 0.001s | 0.012s | 0.026s | 0.066s | 0.132s |

It seems the inner operations had to go through a stack of Accum effects that they did not need.
The solution is to use the effect just between parent and child, leaving grand children alone.

Why was this _exponentially_ slower? I am not sure, maybe it is how fused-effects work. I think they are only performant for statically known effect stacks that can be inlined and skipped. What we created was a growing runtime list of effects.

Closes #2729

### DIFF
--- a/src/swarm-lang/Swarm/Language/Load.hs
+++ b/src/swarm-lang/Swarm/Language/Load.hs
@@ -17,9 +17,8 @@ module Swarm.Language.Load (
 where
 
 import Control.Algebra (Has)
-import Control.Carrier.Accum.Strict (runAccum)
+import Control.Carrier.Accum.Strict (AccumC (..), runAccum)
 import Control.Carrier.State.Strict (evalState, runState)
-import Control.Effect.Accum (Accum, add)
 import Control.Effect.Lift (Lift, sendIO)
 import Control.Effect.State (State, get, modify)
 import Control.Effect.Throw (Throw, throwError)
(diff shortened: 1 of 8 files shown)


src/swarm-lang/Swarm/Language/Load.hs
```diff
@@ -219,20 +218,19 @@ resolveImport ::
...
   , Has (State (SourceMap Resolved)) sig m
   , Has (State VisitedModules) sig m
-  , Has (Accum LocalModules) sig m
   , Has (Lift IO) sig m
   ) =>
   ImportDir Import.Resolved ->
   ImportLoc Import.Raw ->
-  m ResolvedLoc
-resolveImport parent loc = do
+  AccumC LocalModules m ResolvedLoc
```
byorgey, line 231: Let's add a comment here explaining why we use `AccumC` directly, with a link to #2731.

byorgey (approved):
Yay, thanks!!  I like to think I might have figured this out eventually, but I don't know how long it would have taken...

byorgey (approved):

xsebek:
@Mergifyio queue

xsebek:
@Mergifyio dequeue

PR #2735, opened 2026-05-30
Clean up `termSyntax` and rename to `traverseTerm`

byorgey:
- Rename `termSyntax` to `traverseTerm`
- Remove the third argument and have it instead call `traverseSyntax` directly
- Clean up and correct some misleading/outdated comments

While trying to debug #2729 I realized that `termSyntax` was a bit strange: it took a `Syntax` traversal as a third argument, but in fact `traverseSyntax`, when calling `termSyntax`, simply passed itself as that third argument.  It's much cleaner/more sensible to just have `termSyntax` directly call `traverseSyntax` instead of taking it as a parameter.  Also, `termSyntax` should really be named `traverseTerm`.  Then we just have two mutually recursive traversals, `traverseTerm` and `traverseSyntax`, exactly mirroring the mutually recursive structure of `Term` + `Syntax`.

I have no idea why I wrote `termSyntax` the way I did originally.  I probably thought I was being terribly clever.

Of course, this did not end up being the cause of #2729 --- indeed it does not seem to make any performance difference at all --- but I think it's still a nice improvement and makes the code more understandable.

### DIFF
--- a/src/swarm-lang/Swarm/Language/Syntax.hs
+++ b/src/swarm-lang/Swarm/Language/Syntax.hs
@@ -97,8 +97,8 @@ module Swarm.Language.Syntax (
 
   -- * Traversals
 
-  -- ** Term + type traversal
-  termSyntax,
+  -- ** AST traversal
+  traverseTerm,
   traverseSyntax,
 
   -- ** Erasure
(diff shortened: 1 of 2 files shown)


xsebek (approved):
That really makes a lot of sense! 👍

PR #2738, opened 2026-06-03
Take imports into account when measuring code size

### LINKED ISSUE #2737: Code golf calculations ignore imported files
When running a challenge,  the game provides a summary on time run, code size and AST nodes. But the code size calculations ignore any imported files, only the initial script size.

For Example, the following script:
```haskell
import "def"

def givecustomer = \item.
    
    customer <- meet;
    case  customer (\_. pure()) (\c. give c item);
end;

def start =
    tr; m; m; gm; grab;

    tb; m5; grab; tb; m2; tl; m;

    equip "briefcase";

    do20(tl);

    do1000( make "scoop"; givecustomer "scoop");
    do1000( make "scoop"; givecustomer "scoop");
    do200( make "scoop"; givecustomer "scoop");
    do88( make "scoop"; givecustomer "scoop");

    givecustomer "cone";
    givecustomer "cherry";
end;

start;
```

Is 493 bytes, and the game will count it as 493 bytes:

```
last: completed in 0h 1m 7s                                     
(0:01:20.e ticks)                                        
Code: 493 chars                                           
(101 AST nodes)
```

But it ignores the 3957 bytes `def.sw` import.

byorgey:
Take transitive imports into account when measuring code size (either AST size or character count).  Fixes #2737 .

- Caches each module's *transitive* (in addition to *direct*) imports during import resolution, so we don't need to do another DFS later.  This might be useful for other things in the future as well.
- Then works by looking up the top-level module's transitive imports, applying the desired measurement to each, and adding.


src/swarm-lang/Swarm/Language/Elaborate.hs
```diff
@@ -14,7 +14,7 @@ import Swarm.Language.Syntax
 import Unsafe.Coerce (unsafeCoerce)
 
 elaborateModule :: Module Typed -> Module Elaborated
-elaborateModule (Module t ctx imps time prov) = Module (fmap elaborate t) ctx imps time prov
+elaborateModule (Module t ctx imps timps time prov) = Module (fmap elaborate t) ctx imps timps time prov
```
xsebek: Strange that we don't use lenses for Module, maybe this would still be better?
```suggestion
elaborateModule modul = modul { moduleTerm = elaborate <$> moduleTerm modul }
```

src/swarm-lang/Swarm/Language/Load.hs
```diff
@@ -329,15 +330,45 @@ importModule canonicalLoc =
...
+        (ModuleImports phase ~ Set ResolvedLoc, Functor f) =>
+        (ResolvedLoc -> f (Maybe (Module phase))) ->
+        (ResolvedLoc -> f (Set ResolvedLoc))
+      getTransImportsFrom lkup = fmap (maybe S.empty moduleTransImports) . lkup
+
+      getTransImports :: ResolvedLoc -> IO (Set ResolvedLoc)
+      getTransImports =
+        mconcat
+          [ getTransImportsFrom (GC.lookupCached moduleCache)
+          , getTransImportsFrom (pure . flip OM.lookup srcMap)
```
xsebek: The signature above looks daunting, maybe it would be less scary if the `fmap` and `(.)` were inlined? 🤔 
```Haskell
-- getTransImportsFrom :: (...) => Maybe (Module phase) -> Set ResolvedLoc
fmap getTransImportsFrom . GC.lookupCached moduleCache
```

src/swarm-lang/Swarm/Language/Pipeline.hs
```diff
@@ -128,12 +128,15 @@ processTerm prov txt menv tm = do
   -- we might as well.
   time <- sendIO getCurrentTime
 
+  -- Calculate all transitive imports
+  timps <- collectTransitiveImports srcMapRes imps
```
xsebek: nitpick: why not move this above to line 104?

src/swarm-engine/Swarm/Game/Scenario/Scoring/CodeSize.hs
```diff
@@ -5,36 +5,45 @@
...
-  { initialCode :: Maybe (Syntax Raw)
+  { initialCode :: Maybe (Module Elaborated)
   , hasUsedREPL :: Bool
   }
   deriving (Show)
 
+-- | Take an AST metric, call it on on all transitive imports, and
+--   collect the results.
+transitively :: forall metric. Monoid metric => (Syntax Elaborated -> metric) -> Module Elaborated -> IO metric
+transitively measure m = do
```
xsebek: nitpick: This name is rather generic, maybe rename it to `transitiveMetric` or `applyMetricTransitively`?

I originally thought adding explicit export list could hide it, but it is used in unit tests.

src/swarm-engine/Swarm/Game/Scenario/Status.hs
```diff
@@ -138,23 +138,24 @@ updateScenarioInfoOnFinish ::
...
+          newCompletionFlag = if completed then Completed else Attempted
+          newPlayMetric =
+            Metric newCompletionFlag $
+              ProgressStats start $
+                AttemptMetrics (DurationMetrics el ticks) cs
+      pure . ScenarioInfo p $
         Played launchParams newPlayMetric $
           updateBest newPlayMetric prevBestRecords
      where
-      el = (diffUTCTime `on` zonedTimeToUTC) z start
```
xsebek: It seems the definitions got moved to `let` but `where` was left?

xsebek: It seems the definitions got moved to `let` but `where` was left?

src/swarm-tui/Swarm/TUI/Controller/SaveScenario.hs
```diff
@@ -81,13 +79,18 @@ saveScenarioInfoOnFinish p = do
...
-    currentScenarioInfo
-      %= updateScenarioInfoOnFinish determinator t ts won
+  unless saved $ do
+    -- There's probably a nice lens-y way to do this in one line
+    mcsi <- preuse currentScenarioInfo
+    case mcsi of
+      Nothing -> pure ()
+      Just csi -> do
+        csi' <- liftIO $ updateScenarioInfoOnFinish determinator t ts won csi
+        currentScenarioInfo .= csi'
```
xsebek, line 92: I don't think there is an existing `lens` operator for it. The right side would have type `MonadState s m => m ()` which I suppose is confusing since the `MonadState` is not the effect you want to use. Also you need to either decide what should be done in traversal or constrain it to `Lens`.

Maybe you could clean it up a little with `zoom`. Or maybe there is in fact some way to use for example [`%%~`](https://hackage-content.haskell.org/package/lens-5.3.6/docs/Control-Lens-Operators.html#v:-37--37--126-) to do this magic. 🤷 

But I would suggest deleting the comment and not leading anyone else down into the "Lens rabbit hole". 😆

xsebek: I think it would be cleanest to use the "freezed cache" and pass it to pure functions. The `IO` is too general effect anyway. But that is total speculation on my part and it would need a larger (?) refactoring.

test/unit/TestScoring.hs
```diff
@@ -30,14 +31,30 @@ testHighScores =
...
+        [ compareASTChars 4 "single-move-bare.sw"
+        , compareASTChars 22 "single-move-def.sw"
+        , compareASTChars 28 "single-move-let-with-invocation.sw"
+        , compareASTChars 34 "double-move-let-with-invocation.sw"
+        , compareASTChars 32 "single-move-def-with-invocation.sw"
+        , compareASTChars 38 "double-move-def-with-invocation.sw"
+        , compareASTChars 56 "single-def-two-args-recursive.sw"
+        , compareASTChars 70 "single-def-two-args-recursive-with-invocation.sw"
+        , compareASTChars (47 + 16) "def-with-import.sw"
+        , compareASTChars (40 + 29 + 29 + 17) "diamond-import.sw"
```
xsebek, line 57: Nice tests! 👍

src/swarm-lang/Swarm/Language/Module.hs
```diff
@@ -120,7 +120,9 @@ data Module phase = Module
   --   via @def@ with their types, and type aliases defined via @tydef@
   --   with their definitions.  See Note [Module exports].
   , moduleImports :: ModuleImports phase
-  -- ^ The moduleImports are mostly for convenience, e.g. for checking modules for cycles.
+  -- ^ The set of modules directly imported by this module.
```
xsebek: The comment originally said what it is used for. Could you add a not which one is used for what?
I would assume transitive imports are easier to use for cycle detection and therefore only one would be needed.

xsebek (approved):
Nice, I particularly like the tests! 👍

I wondered if you could break this in REPL (import, edit, import) but I see we don't score the code size in that case. 😏

src/swarm-lang/Swarm/Language/Elaborate.hs
byorgey: Good idea.  Perhaps I will also make a separate PR to use lenses with `Module`, I don't know of any good reason why we don't.

src/swarm-lang/Swarm/Language/Module.hs
byorgey: Sure, I added some comments.  I think we actually need both.  We could use the set of transitive imports to detect that a cycle *exists*, but then we would not be able to report the actual cycle, i.e. "Foo -> Bar -> Baz -> Foo".  Likewise, we need to do an actual graph traversal when culling stale modules from the cache, so we need the set of direct imports, not just all transitive imports.

Technically we don't *need* the transitive imports --- we could just calculate it from scratch via a DFS each time --- but it seems worthwhile to collect this information once while doing the initial DFS for import resolution, then be able to reuse it later.

PR #2744, opened 2026-06-17
Substantially revamp Markdown processing

byorgey:
This PR represents some preparatory work for #2504, and also paves the way for #574 .  In this PR:

* Move Markdown code to `Swarm.Text.Markdown.*` instead of `Swarm.Language.Text.Markdown.*`
* Split Markdown code out into multiple modules, with the most useful stuff re-exported from `Swarm.Text.Markdown`.
* Add some Markdown processing/layout tests
* Parsing of Markdown documents is largely unchanged, but the layout algorithm is almost completely rewritten from scratch.
    - A document is first converted into a sequence of tokens, with text tokens for non-whitespace text, special space tokens, newline tokens, and push/pop tokens to indicate the start or end of special attributes (emph, strong, code, or raw).  What we get from the Commonmark parser has every single word, space, and punctuation mark as a separate token.
    - When converting a document into tokens, code blocks get pretty-printed at the chosen line width, with newline tokens inserted appropriately.  Such pretty-printed code blocks are thus already laid out before reaching the layout engine, which will mostly ignore them, unless the pretty-printer was not able to stay within the requested line width, in which case the layout engine will insert additional line breaks.
    - Some of the tokens then get "glued" into bigger lumps that should not be broken across lines.  For example, punctuation sticks to tokens on its left, quote marks stick to things on either side, etc.  But as another example, hyphens are only sticky to their left, so a hyphenated word like "self-destruct" can be broken across a line as "self-\ndestruct" (previously, we would only put line breaks at spaces).
    - The main layout algorithm then runs, trying to fit as many tokens as possible on each line, and inserting newline tokens as appropriate.  If a single token is longer than the entire line, it gets chopped into pieces as a last resort.
    - As a final processing step, the glued tokens are expanded back out, some special tokens are translated away, and consecutive text tokens are merged as much as possible (to avoid e.g. having to generate tons of tiny widgets).
* Rendering to brick widgets is not too different, except now we have to keep track of a stack of attributes which is modified every time we see a push or pop token.

The code is quite a bit longer but it has already fixed several issues, and handles code blocks much better than we did previously.  At this point it also shouldn't be too hard to get tokens directly from the pretty-printer with token type annotations, which we can use as a basis for syntax highlighting (#574).


src/swarm-lang/Swarm/Text/Markdown/Document.hs
```diff
@@ -0,0 +1,103 @@
...
+  deriving (Semigroup, Monoid) via [Paragraph c]
+
+-- | Markdown paragraphs, consisting of a list of inline leaf nodes.
+--
+--   The idea is that paragraphs do not have line breaks, and so the
+--   inline elements follow each other.  In particular inline code can
+--   be followed by text without space between them
+--   (e.g. @\`logger\`s@).
+--
+--   'Paragraph's form a 'Monoid' under concatenation.
```
xsebek: I recently read [`pretty`](https://hackage.haskell.org/package/pretty-1.1.3.6/docs/Text-PrettyPrint.html#v:-36--36-) documentation, which talks about the identity for horizontal (`<>`) and vertical  (`$$`) composition. It is probably worth calling out what the identity is here. 

Basically this means empty paragraphs are skipped, right?

src/swarm-lang/Swarm/Text/Markdown/Layout.hs
```diff
@@ -0,0 +1,192 @@
...
+
+--------------------------------------------------------------
+-- Document -> token stream conversion
+--------------------------------------------------------------
+
+-- A Document is intended to have e.g. hierarchical structure
+-- (especially once we add things like links, lists, etc.); for
+-- rendering purposes we want to first convert a structured Document
+-- into a simple token stream, split into lines at some specified
+-- maximum line width.
```
xsebek, line 58: I like reading these comments, could you turn them into haddock sections?

I think it needs to be referenced by anchor in export list.

src/swarm-lang/Swarm/Text/Markdown/Layout.hs
```diff
@@ -0,0 +1,192 @@
...
+------------------------------------------------------------
+
+-- | Split text into individual whitespace and non-whitespace tokens.
+--   Each newline is made into an individual token; other whitespace
+--   are grouped into consecutive equal characters.
+--
+-- >>> :set -XOverloadedStrings
+-- >>> tokenize "Hello   there, \n\nworld!"
+-- ["Hello","   ","there,"," ","\n","\n","world!"]
+tokenize :: Text -> [Text]
```
xsebek, line 41: This is a great usage of doctest! 👍 I skipped reading docs because I had my idea what tokens should be, but the example showed me what it actually does.

src/swarm-lang/Swarm/Text/Markdown/Parse.hs
```diff
@@ -0,0 +1,160 @@
...
+
+-- | This instance allows us to write a 'Document' as a string literal.
+instance GHC.Exts.IsString (Document (Syntax Raw)) where
+  fromString = fromText . T.pack
+
+-- | This instance allows us to write a 'Paragraph' as a string literal.
+instance GHC.Exts.IsString (Paragraph (Syntax Raw)) where
+  fromString s = case paragraphs $ GHC.Exts.fromString s of
+    [] -> mempty
+    (p : _) -> p
```
xsebek, line 72: This seems like a possible source of bugs, wouldn't it be safer to use a `headPara` function on document?

src/swarm-lang/Swarm/Text/Markdown/Token.hs
```diff
@@ -0,0 +1,192 @@
...
+data Stickiness = StickyL | StickyR | StickyLR | Neutral | Nonstick deriving (Eq, Ord, Show)
+
+-- | The "stickiness" of a token, i.e. whether it prefers to stick to
+--   tokens on its left or right.
+stickiness :: Token -> Stickiness
+stickiness = \case
+  TextToken t
+    | t `elem` T.words "( [ {" -> StickyR
+    | t `elem` T.words ". , ; : ? ! ) ] } - -- --- /" -> StickyL
+    | t `elem` T.words "\" '" -> StickyLR
```
xsebek, line 137: Neat trick!

xsebek (commented):
Nice! I will finish the review tomorrow (I hope!) but here are some notes in the meantime:

src/swarm-lang/Swarm/Text/Markdown/Document.hs
byorgey: Good point, I will clarify that.  The monoid for paragraphs means smushing two paragraphs together into a single paragraph, so the identity is an empty paragraph.  That does not necessarily mean that an empty paragraph would be skipped if a bunch of paragraphs are laid out vertically (in fact I think it would not be).  Putting a bunch of paragraphs in a list is not the same as combining them with `<>`.

src/swarm-lang/Swarm/Text/Markdown/Parse.hs
byorgey: Yes, good point.  I think that instance was there before, and I didn't look at it too closely.  Now that I think about it, I don't see any reason we would actually want this instance, so I think I will just delete it.

test/integration/TestMarkdown.hs
```diff
@@ -0,0 +1,55 @@
...
+import Data.Char (isSpace)
+import Data.Map (Map, (!?))
+import Data.Map qualified as M
+import Data.Text (Text)
+import Data.Text qualified as T
+import Swarm.Text.Markdown (fromTextM, toTextWidth)
+import Swarm.Util (acquireAllWithExt)
+import System.FilePath (dropExtension, takeExtension)
+import Test.Tasty (TestTree, testGroup)
+import Test.Tasty.HUnit (Assertion, assertEqual, testCase)
```
xsebek: There is a nice `tasty-golden` package that deals with the diff and allows updating the golden files.

xsebek (approved):
Very nice! It seems the rest of the files I did not finish before was mostly imports and beautiful documentation. 👍

byorgey:
**All** the Markdown tests are failing on Windows, probably because `\n` is not the same as `\r\n` :angry:

byorgey:
Oh my goodness, I never knew about https://hackage-content.haskell.org/package/base-4.22.0.0/docs/System-IO.html#g:24

I still don't understand what is happening though.  If the golden files contain `\n` nothing should be happening to them when they are read.

byorgey:
Or maybe it's because `tasty-golden` does its comparison *on `ByteString`s*?  I convert the produced `Text` to a `ByteString` via UTF-8, but perhaps the golden files are stored using a different encoding?

byorgey:
I gave up on `tasty-golden`, it was too fiddly to get it to work on Windows...

PR #2755, opened 2026-07-20
Handle links and lists in Markdown documents

byorgey:
Adds basic handling of links (currently, only to external URLs; a subsequent PR will add the ability to link to internal targets in the help system) and bulleted/numbered lists in Markdown documents.  This is another step towards #2504. Also in this PR:

- Get rid of the `ToStream` class in favor of simple functions to convert nodes, paragraphs, + documents into a token stream.  It seems as though the functions naturally want to take slightly different parameters, so shoehorning them all into a single type class did not make sense.
- Change `IsInline`, `IsBlock`, etc. instances to `[Node Text]` and `[Paragraph Text]` instead of `Paragraph` and `Document`.  This is more in line with how the Pandoc implementation of the `Commonmark` type classes works, and allows dealing with lists appropriately.
- Refactor Markdown pretty-printing to reuse the token stream from the layout engine instead of reimplementing it from scratch.
- Add round-trip Markdown tests, which ensure that our Markdown pretty-printing works correctly.


patrickestrada (commented):
Looks good to me.

patrickestrada (approved):

Issue #2580, opened 2025-09-08
Consider switching to a more actively maintained effects system such as https://haskell-effectful.github.io/


byorgey:
This would be a lot of work, but occasionally I wonder if it would be worth it to switch away from `fused-effects`, which sometimes gets in the way and seems to not be actively maintained any more.

xsebek:
I was a bit annoyed with the complex type errors and I really struggled to write reusable code (State GameState/World).

But the presentation about fused-effect history ([YT link](https://www.youtube.com/watch?v=vfDazZfxlNs&pp=ygUMZnVzZWQtZWZmZWN0)) made me appreciate it more and I suppose it is a pretty good fit for effects in language parser/interpreter.

But in case we wanted to try it out, here is what we use:
```bash
rg --type haskell Control.Effect -I | sed 's/[ ]*(.*)//g' | sort | uniq
```
```Haskell
import Control.Effect.Accum
import Control.Effect.Catch
import Control.Effect.Error
import Control.Effect.Lens
import Control.Effect.Lift
import Control.Effect.Reader
import Control.Effect.State
import Control.Effect.Throw
```

byorgey:
Closed by #2759 .

Issue #2669, opened 2025-12-22
Switch to `stm-containers` for `InternCache`


byorgey:
Once we merge #2588 (and subsequently #2540) we will have a `Swarm.Util.InternCache` module implementing global mutable caches.  Currently, it is implemented with a `TVar (HashMap k v)`, but according to [this recent blog post](https://www.parsonsmatt.org/2025/12/17/the_subtle_footgun_of_tvar_(map____).html), using `TVar (HashMap k v)` is a footgun waiting to happen.  In particular, it could lead to livelock if we had multiple threads all trying to access+modify the map at once, even if they are working with different keys.  (Currently, we are single-threaded so this is not even a possibility.)

I tried swapping out `TVar (HashMap k v)` for a `StmContainers.Map.Map k v` from the [`stm-containers` package](https://hackage.haskell.org/package/stm-containers), but ran into a subtle issue.

- Currently, our typechecker is carefully designed to *not* require IO, so that it can be called from places that are definitely not allowed to do IO, such as while processing code embedded in Markdown.
- Since looking up modules in a global cache requires IO, I added a `freezeCache` operation with a type like `Cache k v -> IO (k -> Maybe v)` which returns a lookup function giving a "snapshot" of the cache that can then be passed into the typechecker, which can then look things up in the cache snapshot without having to do IO.
- With `TVar (HashMap k v)` as the implementation, this is easy: just read the `TVar` as an `IO` action, and then return the partial application of the (pure) `lookup` function to the resulting `HashMap`.
- However, I don't think there is any way to implement `freezeCache` with `StmContainers.Map`, at least not without essentially traversing the entire map and building some other data structure which we can then do lookups in.

xsebek:
Cool, I also read the blog post, but did not realize it will be the case for the cache. 👍

byorgey:
Some potential solutions to the above issue:
- Figure out a good way to freeze `StmContainers.Map` (though I don't think this is possible)
- Figure out why the typechecker can't do `IO`, and see if we can relax this restriction
- Convince ourselves that `TVar (HashMap k v)` isn't going to be a problem, even when we add multithreading

Issue #1559, opened 2023-09-30
Try `eff` effect library


xsebek:
⚠️  _**Disclaimer:** This is absolutely a **huge undertaking**, but it might just be worth it. Get free experience working with effects._

---

Currently we are using the fused-effects library, which provides a nice API to work with state.

However, we read and update state a lot* so having more efficient effect library could make significant a runtime difference.

<sup>*) not sure what the exact number is, but browsing the profile it seems like most function in step spend most time in get/put/modify.</sup>

Let's try porting swarm to `eff` since fused-effects say this about it:

> [eff](https://github.com/lexi-lambda/eff) is similar in many ways to fused-effects, but is slightly more performant due to its representation of effects as typeclasses. This approach lets GHC generate better code in exchange for sacrificing the flexibility associated with effects represented as data types. `eff` also uses the monad-control package to lift effects between contexts rather than implementing an Algebra-style class itself.

byorgey:
I am pretty strongly opposed to this, for a few reasons.  First, as you say, it would be a massive undertaking.  I expect porting to a different effects library would probably take... I don't know, 20 hours of work?  It would be truly massive, not only because you would have to touch so much code, but also because different effects libraries always have slightly different ways of doing things.  When I have done such conversions before it has always required some head-scratching to figure out how to refactor things to make use of the new library's peculiar features and ways of doing things.  And after doing all that work, how much of a performance benefit will we get... who knows?  I chose `fused-effects` because it is fast, but I know `eff` is built on top of some cool new runtime support from GHC to make it really fast.  But in any case I suspect there is lots of lower-hanging fruit in terms of optimization that we can tackle first.

Second, `eff` is still "under construction" and it is not even on Hackage.  So building and distributing Swarm would get much harder.  At ICFP I talked with several people working in this space, including Lexi (the author of `eff`) and Nick Wu (one of the authors of the paper on which `fused-effects` is based) and my sense is that we are getting close to having the "one true effects library" which is blazing fast and has all the nice features, but we're not there yet.  If we switch to `eff` we might just have to switch to something else again soon after.

After having said all this, of course, if someone really wants to put in the enormous amount of work needed to try porting to `eff`, I can't stop them, and I'm sure we would learn something interesting.  But because of reason (2) I would not be in favor of merging such a change.

xsebek:
@byorgey Ah, I was not aware of the (2), that is a blocker. The changes are in merged GHC now though, so I am hopeful it will be released some day soon.

[...]

PR #2759, opened 2026-08-05
Switch effect system to effectful

patrickestrada:
This pull request switches all usages of `fused-effects` for `effectful`. All tests pass and the change results in a significant speedup according to the benchmark.

Almost all of the changes in this PR amount to changing type signatures. As for changes outside of type signatures, we also define a few functions to replace the ones from fused-effects-lens and add some extra `runEff` or `runPureEff` calls as needed, among other relatively minor changes. All usages of the `Throw` effect have been replaced by usages of the `Error` effect due to the lack of a `Throw` effect in `effectful`. 

A long-term downside of this change is that we will lose access to more complicated control effects like `NonDet` that do backtracking. We don't currently use any such effects, but if we need to do something similar, we could use `conduit` or `list-t`.

This change addresses #2580.


byorgey:
> we will lose access to more complicated control effects like `NonDet` that do backtracking.

I seem to recall that `NonDet` from `fused-effects` doesn't work properly anyway, though I can't find a link to a relevant discussion right now.  In any case I don't think this is a big loss.

Also, for some related history of our use of effects systems in Swarm, see https://github.com/swarm-game/swarm/issues/228 and https://github.com/swarm-game/swarm/issues/1559 .

byorgey:
Thanks so much for putting this together!  On an initial glance it looks very nice.  I will look through it more carefully over the next few days.

byorgey:
I ran the integration test suite and this does indeed seem to be *much* faster!

patrickestrada:
As an aside, we only use the `add` operation from `Accum` and not the `look` or `looks` operators. So we could potentially axe `Accum` and replace it with an `Output` effect that only supports adding elements. We could even remove the Monoid constraint and use of sequences from `Data.Sequence` in favor of a simple list. An `Output` effect is actually scheduled to be released with the next release of `effectful-core` (see [here](https://github.com/haskell-effectful/effectful/blob/master/effectful-core/src/Effectful/Output/Static/Local/List.hs)) but until then we could just implement our own `Output` effect.

src/swarm-lang/Swarm/Language/Load.hs
```diff
@@ -122,14 +120,14 @@ type LocalModules = Set ResolvedLoc
 --
 --   See Note [Module loading] for an overview.
 resolve ::
-  (Has (Lift IO) sig m, Has (Throw SystemFailure) sig m) =>
-  -- | Provenance of the source, and location relative to which
+  (IOE :> es, Error SystemFailure :> es) =>
+  -- \| Provenance of the source, and location relative to which
```
byorgey: Stray backslash?  This should be a real Haddock comment, not escaped.

src/swarm-lang/Swarm/Language/Load.hs
```diff
@@ -267,30 +262,25 @@ resolveImports parent = runAccum emptyImportSet . traverseSyntax pure (resolveIm
...
-  forall sig m.
-  ( Has (Throw SystemFailure) sig m
-  , Has (State (SourceMap Resolved)) sig m
-  , Has (State VisitedModules) sig m
-  , Has (Lift IO) sig m
+  forall es.
+  ( Error SystemFailure :> es
+  , State (SourceMap Resolved) :> es
+  , State VisitedModules :> es
+  , Accum LocalModules :> es
```
byorgey, line 270: Do we know for sure that the performance issues we had here with `Accum` are not a problem with `effectful`?

src/swarm-util/Swarm/Effect/Metric.hs
```diff
@@ -10,72 +10,54 @@ module Swarm.Effect.Metric (
...
-    L (GaugeSet g v) -> (<$ ctx) <$> liftIO (Gauge.set g $ fromIntegral v)
-    R other -> MetricIOC (alg (runMetricIO . hdl) other ctx)
-  {-# INLINE alg #-}
-
-newtype FakeMetric m a = FakeMetric (m a)
-  deriving newtype (Applicative, Functor, Monad)
-
-runFakeMetric :: FakeMetric m a -> m a
-runFakeMetric (FakeMetric m) = m
+runMetricIO :: (IOE :> es) => Eff (Metric : es) a -> Eff es a
```
byorgey, line 50: The reduced amount of boilerplate is fantastic!

src/swarm-util/Swarm/Util/Effect.hs
```diff
@@ -1,60 +1,58 @@
...
---   by supplying an adapter function of type @(e1 -> e2)@.
-withThrow :: (Has (Throw e2) sig m) => (e1 -> e2) -> ThrowC e1 m a -> m a
-withThrow f = runThrow >=> either (throwError . f) return
-
 -- | Transform an @Error e1@ constraint into a @Error e2@ constraint,
 --   by supplying an adapter function of type @(e1 -> e2)@.
-withError :: (Has (Error e2) sig m) => (e1 -> e2) -> ErrorC e1 m a -> m a
-withError f = runError >=> either (throwError . f) return
+-- withError :: HasCallStack => (e1 -> e2) -> Eff (Error e : es) a -> Eff es a
+-- withError f = runErrorNoCallStackWith (throwError_ . f)
```
byorgey: Stray commented-out code?

src/swarm-util/Swarm/Util/Effect.hs
```diff
@@ -1,60 +1,58 @@
...
 --   to convert from one style to the other while we are in the middle
 --   of incrementally converting.  Eventually this should not be needed.
-asExceptT :: ErrorC e m a -> ExceptT e m a
-asExceptT (ErrorC m) = m
+asExceptT :: Eff [Error e, IOE] a -> ExceptT e IO a
+asExceptT = ExceptT . (runEff . runErrorNoCallStack)
 
 -- | Log a single failure as a warning.
-warn :: Has (Accum (Seq w)) sig m => w -> m ()
+warn :: (Accum (Seq w) :> es) => w -> Eff es ()
```
byorgey, line 53: Many (but not all) of the uses of `Accum`, especially `Accum (Seq w)`, are for accumulating warnings.  Would it make sense to make our own custom `Warning` effect?

src/swarm-util/Swarm/Util.hs
```diff
@@ -624,25 +624,25 @@ liftText :: T.Text -> Q Exp
 liftText txt = AppE (VarE 'T.pack) <$> lift (T.unpack txt)
 
 ------------------------------------------------------------
--- Fused-Effects Lens utilities
+-- Effectful Lens utilities
```
byorgey: Should these move to `Swarm.Util.Lens`?

swarm.cabal
```diff
@@ -182,6 +182,12 @@ common directory
 common dotgen
   build-depends: dotgen >=0.4 && <0.5
 
+common effectful-core
+  build-depends: effectful-core >=2.6.0.0 && <2.7
+
+common effectful
```
byorgey: Do we actually use this one?

byorgey (approved):
Looks great to me!  The boilerplate is much cleaner, and it's much faster --- I ran the whole test suite locally and it is significantly faster (at least 5x, maybe 10x, maybe more; I'm not sure how long it took before).  Just a few comments/questions.

byorgey:
> So we could potentially axe `Accum` and replace it with an `Output` effect that only supports adding elements.

That makes sense to me.  See also my comment about possibly adding a custom `Warnings` effect (though that is not the only use of `Accum`).  I'll leave it up to you whether it makes more sense to make any such changes as part of this PR, or merge this as-is and do the switch away from `Accum` as a separate PR later.

byorgey:
@patrickestrada Adding a custom `ModuleCacheLookup` effect makes sense, but can you elaborate on how it would help with `IO` during typechecking (and hence with #2669)?

swarm.cabal
patrickestrada: We indeed don't use the full effectful library.

src/swarm-lang/Swarm/Language/Load.hs
patrickestrada: I don't think it should cause problems. The "StaticRep" used to implement the Accum effect functionally amounts to an IORef. Technically speaking, the StaticRep for our effect (Accum w) is actually cast to Any via unsafeCoerce and then stored in a mutable array, but that's an implementation detail. The StaticRep is always evaluated to whnf by design, so I also don't think there should be any space leaks coming from using the Accum effect here.

src/swarm-util/Swarm/Util/Effect.hs
patrickestrada: I think that would make sense. We could implement it by piggybacking off the same code used to implement the Error effect, and have its interface essentially consist of the same throwing functions but with no catching functions.

As an example of how we might implement and use it, see below:
```haskell
data Warn w :: Effect where
  MkWarn :: w -> Eff es a
type instance DispatchOf (Warn w) = Dynamic

-- Should be called in the same thread as runWarn or otherwise propagated
-- to the thread in which runWarn is called
warn :: (Warn w :> es) => w -> Eff es a
warn w = send . MkWarn

runWarn :: Eff (Warn w : es) a -> Eff es (Either w a)
runWarn = reinterpret runErrorNoCallStack $ \case ->
  MkWarn w -> throwError_ w

-- New version of traverseW
-- Once we're done collecting warnings, we can use runOutput
-- to collect all the warnings
traverseW :: 
  (Output w :> es, Witherable t) =>
  (a -> Eff (Warn w : es) b) ->
  t a ->
  Eff es (t b)
traverseW f = do
  wither $ \a -> case runWarn (f a) of
  Left warning -> output warning >> pure Nothing
  Right b -> pure (Just b)

-- Version of traverseW that automatically collects all the warnings
traverseW' :: 
  (Witherable t) => 
  (a -> Eff (Warn w : es) b) ->
  t a ->
  Eff es (t b, [w])
traverseW' f = runOutput . traverseW f
  
```
Note: Haven't tested this code yet

I'm thinking it might be better to put this into a slightly later PR.

patrickestrada:
Instead of passing ModuleCacheLookup as a function in the `Swarm.Language.Typecheck` module, we could instead turn it into a (dynamically dispatched) effect
`data (ModuleCacheLookup :: Effect) where ...`
and add the `ModuleCacheLookup :> es` constraint to functions that do typechecking. We already use the `Error ContextualTypeErr` effect in many of those functions, so this wouldn't be too intrusive a change to the types.

To that effect we could add an operation
 `lookupModule :: (ModuleCacheLookup :> es) => ImportLoc Import.Resolved -> Eff es (Maybe (Module Elaborated))`.

In the effect handler we could then actually do the IO. For example, the type signature of `runModCacheLookupIO` might be 
`runModCacheLookupIO :: IOE :> es => Eff (ModuleCacheLookup : es) a -> Eff es a`. We'd still be doing IO in typechecking, but the only IO we'll actually be allowed to do is looking up modules in the cache.

In cases where we don't want to allow imports or we want to only allow imports from a static cache of modules, we could add another handler that doesn't have an IOE constraint and whose implementation of `lookupModule` always returns Nothing (for the first case) or always tries to lookup from the given static cache (for the second case).

src/swarm-util/Swarm/Util/Effect.hs
byorgey: I was thinking that `warn` would correspond to an output effect, rather than an error effect.  The idea behind `traverseW` is that it takes a bunch of individual possibly-failing computations and collects up the failures as warnings.  In any case, I agree putting this off to a later PR makes sense.

byorgey:
Ah, that makes sense.  Yes, making a `ModuleCacheLookup` effect sounds like a good idea, though I am still unsure how to resolve the issue of allowing `IO` during typechecking.

patrickestrada:
The `lookupModule` operation (which is what we will use to actually lookup modules) doesn't have an IOE constraint. The fact that `lookupModule` will be implemented using IO is opaque to its users. It's only when we try to run the effect that we require IO (in the scope in which the handler is called). So the question becomes where we can actually insert `runModCacheLookupIO` when calling into the type checker.

Looking at the non-testing/benchmark components, based on the exports in `Swarm.Language.Typecheck`, it seems like there are only two relevant "outside" places in which we call into the type checker. One of them is `Swarm.Language.Pipeline` and the other is `Swarm.Language.Parser.Value` (for `readValue`). 

In `Swarm.Language.Pipeline`, there are two functions that (directly) try to do typechecking: `processTerm` and `processTermNoImports`. For `processTerm`, it so happens that it already has an `IOE` constraint, so we can call the `runModCacheLookupIO` function from there. As for `processTermNoImports` and `readValue`, they don't actually use the module cache, and so we can use a handler that makes the operation `lookupModule` always return a `Nothing` in those places.

PR #2761, opened 2026-08-07
Refactor: Add Warn effect

patrickestrada:
This change adds a warn effect which has a warn operation to replace the `warn` function in `Swarm.Util.Effect`. This allows us to replace usages of `Accum (Seq w))` with `Warn w` and avoid leaking details about how we're accumulating warnings when we use `warn`.


src/swarm-doc/Swarm/Doc/Pedagogy.hs
```diff
@@ -176,8 +175,8 @@ loadScenarioCollection = simpleErrorHandle $ do
   -- 'loadScenarios' below.  Any warnings will be caught when loading
   -- all the scenarios via the usual code path; we do not need to do
   -- anything with them here while simply rendering pedagogy info.
-  worlds <- ignoreWarnings @(Seq SystemFailure) $ loadWorlds tem
-  ignoreWarnings @(Seq SystemFailure) $ loadScenarios (ScenarioInputs worlds tem) True
+  worlds <- loadWorlds tem
```
byorgey, line 178: Hah, I didn't notice this `ignoreWarnings` call was doing nothing before!

src/swarm-util/Swarm/Util/Effect.hs
```diff
@@ -30,7 +28,7 @@ liftEither = either throwError_ pure
 
 -- | Transform a @Throw e@ constraint into a concrete @Maybe@,
```
byorgey: This should say `Error e` instead of `Throw e`.  In fact, maybe we ought to rename this function as `errorToWarning`?

src/swarm-util/Swarm/Effect/Warn/Local.hs
```diff
@@ -0,0 +1,38 @@
...
+import Effectful.Dispatch.Static
+
+-- | The Warn effect
+data Warn (w :: Type) :: Effect
+
+type instance DispatchOf (Warn w) = Static NoSideEffects
+newtype instance StaticRep (Warn w) = Warn [w]
+
+-- | Log a single failure as a warning.
+warn :: forall w es. (Warn w :> es) => w -> Eff es () -- (Warn w :> es) => w -> Eff es ()
```
byorgey: Stray commented-out type?

src/swarm-util/Swarm/Effect/Warn/Local.hs
```diff
@@ -0,0 +1,38 @@
+{-# LANGUAGE TypeFamilies #-}
+
+-- | Accumulates thread-local warnings.
+-- Code adapted from the as-of-yet unreleased Output effect in
+-- the effectful repo
```
byorgey: I suppose once `Output` is released, we should refactor `Warn` to just be implemented in terms of `Output`.  Perhaps we should create an issue to track that and link to it here.

src/swarm-util/Swarm/ResourceLoading/Collection.hs
```diff
@@ -158,9 +155,8 @@ loadCollection cfg dir = do
   -- or has thrown SystemFailure. The following code just adds that thrown failure to others.
   loadItems :: [FilePath] -> Eff es (Map FilePath (CollectionItem a))
   loadItems items = do
-    let loadItem f = runErrorNoCallStack @SystemFailure $ (f,) <$> loadCollectionItem cfg (dir </> f)
-    (itemFailures, okItems) <- partitionEithers <$> mapM loadItem items
-    add (Seq.fromList itemFailures)
+    let loadItem fp = runErrorNoCallStack @SystemFailure $ (fp,) <$> loadCollectionItem cfg (dir </> fp)
+    okItems <- traverseW loadItem items
```
byorgey, line 159: Good catch!

src/swarm-util/Swarm/Util/Effect.hs
```diff
@@ -49,17 +47,13 @@ ignoreWarnings = evalAccum mempty
...
-
 -- | A version of 'traverse'/'mapM' that also accumulates warnings.
 --
 --   Note that we can't generalize this to work over any 'Traversable'
 --   because it also needs to have a notion of "filtering".
 --   'Witherable' provides exactly the right abstraction.
 traverseW ::
-  (Accum (Seq w) :> es, Witherable t) =>
+  (Warn w :> es, Witherable t) =>
   (a -> Eff es (Either w b)) ->
```
byorgey, line 57: Would it make sense to change this to take an argument of type `a -> Eff (Error w : es) b`?   I am not 100% sure if this would work or how it would affect uses of `traverseW`.  But, for example, I notice that in the use of `traverseW` in `ResourceLoading.Collection`, we get the `Either`s by running `runErrorNoCallStack` on some computations with an `Error` effect, so I think it would simplify the code in that case.  I don't know whether there are any calls to `traverseW` on things that "natively" return an `Either`, but we could easily use `liftEither` on those.

byorgey (approved):

src/swarm-util/Swarm/Util/Effect.hs
patrickestrada: As far as I can tell, there are two uses of traverseW in the codebase: the one in `ResourceLoading.Collection` and the one in `Swarm.Game.Achievement.Persistence` via `forMW`. In the second setting we already "natively" return on an either on account of it being obtained by `Data.Yaml.decodeFileEither` and then wrapping `Left` in an error of our own choosing. We could use `throwError` there, but I'm leaning towards keeping the Either version as it doesn't require a `Error` constraint and making users of `traverseW` throw.

byorgey: Makes sense!

src/swarm-util/Swarm/Effect/Warn/Local.hs
patrickestrada: I'm thinking that the implementation of the `Warn` effect may be subject to changes that make it more distinct from the implementation of the `Output` effect. For example, if we want to use `warn` in a multi-threaded setting, we might want to group related warnings so that when we display the warnings, each group appears contiguously. Even if we don't end up using `Warn` in a multi-threaded setting, its implementation is so straightforward that I think basing it off the Output effect wouldn't have much benefit.

PR #2763, opened 2026-08-18
Add import highlighting

### LINKED ISSUE #2728: Import syntax highlighting
I'm always frustrated when writing swarm code and the `import` keyword is the same color.

I would like it to use the configured editor color for import/include/using/whatever it is called in given language.

None. It has to be done. 😆 

**Additional context**

Found while thinking about a test for #2727:

Image

xsebek:
* Closes #2728


byorgey (approved):
LGTM!

PR #2765, opened 2026-08-19
Shorten parse error "expected..."

### LINKED ISSUE #2676: Cut down length of parse errors
Parse errors can be extremely long, with a giant list of possible next tokens that were expected.  e.g.:
```
> def x = 3
unexpected end of input                                                             
expecting "!=", "&&", "()", "++", "<=", "==", ">=", "def", "false", "import", "let",
  "require", "requirements", "stock", "true", "tydef", "||", '"', '$', '(', '*',    
  '+', '-', '.', '/', ':', ';', '<', '>', '@', '[', '\', '^', 'end' keyword for     
  definition of 'x', '{', built-in user function, digit, direction constant, integer
  literal, or variable name                                                         
```
This is so long that it's not actually helpful.  There must be some way we could make these error messages more reasonable.  e.g. perhaps we can give more things names with the `<?>` operator, so we get helpful names for categories of things rather than a giant list of possible tokens.

xsebek:
* label operators just as "operator"
* use double quotes for `end` so that it gets sorted first
* label unit literal `()` (would have been before `"end"`)
* label `let`/`def` bindings (would have been before `"end"`)
* Closes #2676

### DIFF
--- a/src/swarm-lang/Swarm/Language/Parser/Term.hs
+++ b/src/swarm-lang/Swarm/Language/Parser/Term.hs
@@ -65,13 +65,13 @@ parseTermAtom = do
 parseTermAtom2 :: Parser (Syntax Raw)
 parseTermAtom2 =
   parseLoc
-    ( TUnit <$ symbol "()"
+    ( TUnit <$ (symbol "()" <?> "unit literal")
         <|> TConst <$> parseConst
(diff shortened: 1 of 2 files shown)


byorgey (approved):
It's still rather long, but it's much better!

PR #2766, opened 2026-08-19
Remove LSP hover Pandoc markdown annotations

### LINKED ISSUE #2590: LSP hover shows internal markdown annotation
**Describe the bug**

The LSP hover shows `{=type}` annotation:

Image

**To Reproduce**

Hover over `pure` command.

**Expected behavior**

The `{=type}` annotation should only be used for coloring in swarm TUI.

xsebek:
* Closes #2590


byorgey (approved):
I do hope we will eventually use this documentation in the help system within the TUI in the future.  At that point we may want to revisit this and actually have the documentation be processed via our Markdown framework, and process it for sending via the LSP appropriately.  But I think this is a good simple fix for now.

PR #2767, opened 2026-08-19
Add const long doc to LSP hover

xsebek:
* Found while reading through #2590

| image |
|:-:|
| <sup>I don't know why we didn't do this from the start.</sup> |


byorgey (approved):
Nice idea!

PR #2773, opened 2026-08-20
Skip loading scenarios on --scenario

### LINKED ISSUE #2772: Skip loading scenarios with `--scenario` flag

When you run `swarm --scenario Blank` the main menu is skipped. But swarm still takes long time loading all scenarios.

Skip loading scenarios in this case.

Make loading scenarios so fast it is not noticeable.

xsebek:
* refactor `initPersistentState`
* skip loading scenario and achievement collections if `--scenario` was set on command line
* closes #2772

### DIFF
--- a/src/swarm-engine/Swarm/Game/ScenarioInfo.hs
+++ b/src/swarm-engine/Swarm/Game/ScenarioInfo.hs
@@ -109,19 +109,19 @@ normalizeScenarioPath col p =
 loadScenarios ::
   (Warn SystemFailure :> es, IOE :> es) =>
   ScenarioInputs ->
-  Bool ->
+  (String -> Bool) ->
   Eff es (ScenarioCollection ScenarioInfo)
-loadScenarios scenarioInputs loadTestScenarios = do
+loadScenarios scenarioInputs filterScenarioDir = do
   res <- runErrorNoCallStack @SystemFailure $ getDataDirThrow Scenarios "scenarios"
   case res of
     Left err -> warn err >> pure emptyCollection
-    Right dataDir -> loadCollection (scenarioCollectionConfig scenarioInputs loadTestScenarios) dataDir
+    Right dataDir -> loadCollection (scenarioCollectionConfig scenarioInputs filterScenarioDir) dataDir
 
 -- | Configuration record for recursively loading the scenario collection.
-scenarioCollectionConfig :: ScenarioInputs -> Bool -> CollectionConfig (ScenarioWith ScenarioInfo)
-scenarioCollectionConfig scenarioInputs loadTestScenarios =
+scenarioCollectionConfig :: ScenarioInputs -> (String -> Bool) -> CollectionConfig (ScenarioWith ScenarioInfo)
+scenarioCollectionConfig scenarioInputs filterScenarioDir =
   CollectionConfig
-    { shouldLoad = isYamlOrPublicDirectory
+    { shouldLoad = isYamlOrAllowedDirectory
     , warnUnordered = True
     , loadItem = loadScenarioItem scenarioInputs
     }
@@ -129,16 +129,14 @@ scenarioCollectionConfig scenarioInputs loadTestScenarios =
   -- Keep only files which are .yaml files or directories not starting
   -- with an underscore.  Marked directories contain scenarios that
   -- can't be parsed (failure tests) or only script solutions.
-  isYamlOrPublicDirectory :: FilePath -> FilePath -> IO Bool
-  isYamlOrPublicDirectory d f = do
+  -- Uses provided filter for the directory.
+  isYamlOrAllowedDirectory :: FilePath -> FilePath -> IO Bool
+  isYamlOrAllowedDirectory d f = do
     isDir <- doesDirectoryExist $ d </> f
     return $
       if isDir
-        then not ("_" `isPrefixOf` f || isHiddenDir f)
+        then not ("_" `isPrefixOf` f) && filterScenarioDir f
         else takeExtensions f == ".yaml"
-
-  isHiddenDir :: String -> Bool
-  isHiddenDir f = not loadTestScenarios && f == "Testing"
 
 -- | Load a single scenario from a path, returning either a loading
 --   error, or a scenario along with a list of warnings.
--- a/src/swarm-tui/Swarm/TUI/Model/StateUpdate.hs
+++ b/src/swarm-tui/Swarm/TUI/Model/StateUpdate.hs
@@ -160,8 +160,8 @@ initPersistentState opts@(AppOpts {..}) =
     ui <- initUIState UIInitOptions {..}
     ks <- initKeyHandlingState
 
-    loadedScenarios <- whenC showMainMenu $ loadScenarios (getScenarioInputs rs) skipLoadingTests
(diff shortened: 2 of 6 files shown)


xsebek:
What a weird failure on Windows:

```
    data\test\language-snippets\format\parens.sw:                                                                                  OK

divide by zero

 Attempting to reconstruct a stack trace...

   Frame	Code address
 * 0x344f3fdc30	0x7ff6d854750e D:\a\swarm\swarm\dist-newstyle\build\x86_64-windows\ghc-9.12.1\swarm-0.8.0.0\t\swarm-integration\build\swarm-integration\swarm-integration.exe+0x195750e
 * 0x344f3fdc38	0x7ff6d72d4106 D:\a\swarm\swarm\dist-newstyle\build\x86_64-windows\ghc-9.12.1\swarm-0.8.0.0\t\swarm-integration\build\swarm-integration\swarm-integration.exe+0x6e4106
 * 0x344f3fdc40	0x7ff42bb043b0
 * 0x344f3fdc48	0x7ff42b58e2f0
 * 0x344f3fdc50	0x7ff423c6b011

    data\test\language-snippets\format\list.sw:                                                                                    Test suite swarm-integration: FAIL
```

src/swarm-tui/Swarm/TUI/Model/StateUpdate.hs
```diff
@@ -156,31 +154,38 @@ initPersistentState ::
...
             , _scenarioSequence = mempty
-            , _uiPopupAnimationState = animState
+            , _uiPopupAnimationState = AnimInactive
             }
     return $ PersistentState rs ui ks progState
-  let initRS' = addWarnings initRS (F.toList warnings)
-  return $ PersistentState initRS' initUI initKs initProg
+ where
+  addWarningsToRuntimeState :: (PersistentState, [SystemFailure]) -> PersistentState
+  addWarningsToRuntimeState (PersistentState initRS initUI initKs initProg, warnings :: [SystemFailure]) =
```
byorgey: The `[SystemFailure]` type annotation on `warnings` should no longer be necessary.

src/swarm-tui/Swarm/TUI/Model/StateUpdate.hs
```diff
@@ -156,31 +154,38 @@ initPersistentState ::
...
+    let initRS' = addWarnings initRS (F.toList warnings)
+     in PersistentState initRS' initUI initKs initProg
+  -- helpers for getting inner data
+  showMainMenu = not (skipMenu opts)
+  skipLoadingTests = Set.member LoadTestingScenarios debugOptions
+  getScenarioInputs rs = rs ^. stdGameConfigInputs . to initState . to gsiScenarioInputs
+  -- skip loading collections
+  whenC b a = if b then a else pure emptyCollection
```
byorgey, line 185: Hmm, is `Collection` not a `Monoid`?  Should we make it one?  If we did then we could just use `whenM`.

src/swarm-tui/Swarm/TUI/Model/StateUpdate.hs
```diff
@@ -156,31 +154,38 @@ initPersistentState ::
...
-  let initRS' = addWarnings initRS (F.toList warnings)
-  return $ PersistentState initRS' initUI initKs initProg
+ where
+  addWarningsToRuntimeState :: (PersistentState, [SystemFailure]) -> PersistentState
+  addWarningsToRuntimeState (PersistentState initRS initUI initKs initProg, warnings :: [SystemFailure]) =
+    let initRS' = addWarnings initRS (F.toList warnings)
+     in PersistentState initRS' initUI initKs initProg
+  -- helpers for getting inner data
+  showMainMenu = not (skipMenu opts)
+  skipLoadingTests = Set.member LoadTestingScenarios debugOptions
```
byorgey: `skipLoadingTests` seems like a confusing/misleading name for this variable, since as far as I can tell it means the opposite: when it is true it means we *should* load the tests, not skip them.

src/swarm-tui/Swarm/TUI/Model/StateUpdate.hs
```diff
@@ -156,31 +154,38 @@ initPersistentState ::
...
     ui <- initUIState UIInitOptions {..}
     ks <- initKeyHandlingState
 
-    s <-
-      loadScenarios
-        (gsiScenarioInputs $ initState $ rs ^. stdGameConfigInputs)
-        (loadTestScenarios $ mkRuntimeOptions opts)
-    achievements <- loadAchievementsInfo
+    loadedScenarios <- whenC showMainMenu $ loadScenarios (getScenarioInputs rs) skipLoadingTests
+    achievements <- whenM showMainMenu $ categorizeAchievement <$> loadAchievementsInfo
```
byorgey: If we skip loading achievements, does that affect the player's ability to earn/unlock achievements?  Even though we won't need to show achievements in the menu, it should still be possible to unlock new achievements when playing via `--scenario`.

byorgey: I tested it out, and it seems like when playing with `--scenario` you can indeed still earn achievements, but the problem is that the game thinks you have not achieved anything previously, so you can even earn achievements that you already obtained previously, and then the achievement save info gets overwritten with the new time.

I think we should not skip loading achievements; fortunately I think loading achievement info probably takes almost no time whatsoever, so no one will notice.  The important thing is to skip loading scenarios.

byorgey:
> What a weird failure on Windows

That is very odd.  I can't think of any reason that this change would cause a code formatting test to fail with division by zero... does it happen consistently every time the test suite runs?

src/swarm-tui/Swarm/TUI/Model/StateUpdate.hs
xsebek: ups, my bad

src/swarm-tui/Swarm/TUI/Model/StateUpdate.hs
xsebek: Ah, I did not think about it comparing achievements for current scenario, I thought its just for menu.

src/swarm-tui/Swarm/TUI/Model/StateUpdate.hs
xsebek: I looked and it did not seem like a reasonable semigroup. Its map values are single or lists, but the lists need a group name. Default instance would keep the right map values, which does not seem right.

But it's irrelevant since I wont need the other helper for achievements, which I will revert.

xsebek:
@byorgey no, it happened once only.

byorgey:
> @byorgey no, it happened once only.

:shrug:

src/swarm-engine/Swarm/Game/ScenarioInfo.hs
```diff
@@ -109,37 +109,35 @@ normalizeScenarioPath col p =
 loadScenarios ::
   (Warn SystemFailure :> es, IOE :> es) =>
   ScenarioInputs ->
-  Bool ->
+  (String -> Bool) ->
```
byorgey, line 112: Oh, passing a filter to `loadScenarios` instead of a flag is a nice idea.

byorgey (approved):

PR #2776, opened 2026-08-22
Use Vector.Strict in TyConF to avoid space leak

xsebek:
* requires #2779 to avoid having to write ToJSON(1) instance for vectors

Co-Authored-By @patrickestrada

### DIFF
--- a/src/swarm-lang/Swarm/Language/SmallVector.hs
+++ b/src/swarm-lang/Swarm/Language/SmallVector.hs
@@ -1,0 +1,57 @@
+-- |
+-- SPDX-License-Identifier: BSD-3-Clause
+--
+-- Strict small vector type to save on allocations, specifically for type parameters.
+-- Most swarm functions will not have more than two parameters.
+module Swarm.Language.SmallVector (
+  SmallVector,
+  toList,
+  fromList,
+) where
+
+import Data.Data (Data)
+import Data.Functor.Classes (Eq1 (..), Ord1 (liftCompare), Show1 (..))
+import Data.Hashable (Hashable (..))
+import Data.Hashable.Lifted (Hashable1 (..))
+import Data.Vector.Instances ()
+import Data.Vector.Strict (Vector)
+import Data.Vector.Strict qualified as Vec
+import GHC.Generics (Generic, Generic1)
+
+data SmallVector a = Nil | One a | Two a a | Three a a a | Many (Vector a)
+  deriving (Functor, Foldable, Traversable, Generic, Generic1, Data, Hashable, Hashable1)
+
+toList :: SmallVector a -> [a]
+toList = \case
+  Nil -> []
+  One a -> [a]
+  Two a b -> [a, b]
+  Three a b c -> [a, b, c]
+  Many v -> Vec.toList v
+
+fromList :: [a] -> SmallVector a
+fromList = \case
+  [] -> Nil
+  [a] -> One a
+  [a, b] -> Two a b
+  [a, b, c] -> Three a b c
+  l -> Many $ Vec.fromList l
+
+instance Eq a => Eq (SmallVector a) where
+  v1 == v2 = toList v1 == toList v2
+
+instance Ord a => Ord (SmallVector a) where
+  compare v1 v2 = compare (toList v1) (toList v2)
+
+instance Show a => Show (SmallVector a) where
+  show = show . toList
+
+instance Eq1 SmallVector where
+  liftEq p v1 v2 = liftEq p (toList v1) (toList v2)
+
+instance Ord1 SmallVector where
+  liftCompare p v1 v2 = liftCompare p (toList v1) (toList v2)
+
+instance Show1 SmallVector where
+  liftShowsPrec a b c v = liftShowsPrec a b c (toList v)
+  liftShowList a b vs = liftShowList a b (map toList vs)
--- a/src/swarm-lang/Swarm/Language/Types.hs
+++ b/src/swarm-lang/Swarm/Language/Types.hs
@@ -164,10 +164,10 @@ import Effectful.Error.Static
 import Effectful.Reader.Static
 import GHC.Generics (Generic, Generic1)
 import Prettyprinter (align, braces, brackets, concatWith, flatAlt, hsep, pretty, punctuate, softline, (<+>))
+import Swarm.Data.SmallVector (SmallVector)
+import Swarm.Data.SmallVector qualified as SVec
 import Swarm.Language.Context (Ctx)
 import Swarm.Language.Context qualified as Ctx
-import Swarm.Language.SmallVector (SmallVector)
-import Swarm.Language.SmallVector qualified as SVec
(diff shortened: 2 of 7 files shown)


src/swarm-lang/Swarm/Language/SmallVector.hs
```diff
@@ -0,0 +1,57 @@
...
+import Data.Data (Data)
+import Data.Functor.Classes (Eq1 (..), Ord1 (liftCompare), Show1 (..))
+import Data.Hashable (Hashable (..))
+import Data.Hashable.Lifted (Hashable1 (..))
+import Data.Vector.Instances ()
+import Data.Vector.Strict (Vector)
+import Data.Vector.Strict qualified as Vec
+import GHC.Generics (Generic, Generic1)
+
+data SmallVector a = Nil | One a | Two a a | Three a a a | Many (Vector a)
```
byorgey: I really don't think we need a special case for `Three`.  It will almost never be used and only adds a bit of complexity to the implementation.

src/swarm-util/Swarm/Data/SmallVector.hs
```diff
@@ -0,0 +1,57 @@
...
+
+fromList :: [a] -> SmallVector a
+fromList = \case
+  [] -> Nil
+  [a] -> One a
+  [a, b] -> Two a b
+  [a, b, c] -> Three a b c
+  l -> Many $ Vec.fromList l
+
+instance Eq a => Eq (SmallVector a) where
```
byorgey, line 50: Would it be worth implementing the `Eq` and `Ord` instances by hand rather than going through `toList`?  I don't know how often the instances will be used, but it seems like unnecessary overhead to generate a bunch of cons cells every time we need to compare two `SmallVector`s.

src/swarm-lang/Swarm/Language/SmallVector.hs
```diff
@@ -0,0 +1,57 @@
+-- |
+-- SPDX-License-Identifier: BSD-3-Clause
+--
+-- Strict small vector type to save on allocations, specifically for type parameters.
+-- Most swarm functions will not have more than two parameters.
+module Swarm.Language.SmallVector (
```
byorgey: Perhaps this should go in `swarm-util/Swarm/Util/SmallVector`?  Sure, we're only using it for `TypeF` now, but it doesn't really have anything in particular to do with the Swarm language.

src/swarm-lang/Swarm/Language/Types.hs
```diff
@@ -307,34 +304,30 @@ data TypeF t
...
+--   saturated (higher kinds are not supported), so we just
+--   directly store a list of all arguments (as opposed to
+--   iterating binary application).
+--
+--   Note that this is a pattern for easier matching on lists.
+--   See 'TyConVF' which uses 'SmallVector' for storage.
 pattern TyConF :: TyCon -> [t] -> TypeF t
-pattern TyConF t ts <- TyConVF t (Vec.toList -> ts)
+pattern TyConF t ts <- TyConVF t (SVec.toList -> ts)
   where
```
byorgey, line 318: Simply making a `TyConF` synonym that goes through a list is fine for now, since it means we don't have to change much existing code.  But I wonder if it would eventually be worth making some argument-number-specific synonyms, e.g.
```
pattern TyConF2 :: TyCon -> t -> t -> TypeF t
pattern TyConF2 t t1 t2 = TyConVF t (Two t1 t2)
```
(and similar for `TyConF1`) and then using these in as many places as possible to again avoid the intermediate list conversions.

src/swarm-util/Swarm/Data/SmallVector.hs
patrickestrada: I think the laziness of the `Eq` and `Ord` instances for `[a]` should prevent the generation of an intermediate list.

byorgey: Hmm, maybe so.

src/swarm-lang/Swarm/Language/Types.hs
xsebek: I would prefer INLINE pragma and have compiler deal with it.

Currently I don't export the constructors, so toList is the only option.

But it would be an interesting experiment to try to do it manually in each place and see how much we can save.

src/swarm-util/Swarm/Data/SmallVector.hs
xsebek: Would it sort some Two before One? That would require a quadratic number of cases.

If it's just comparing One and One, then that might be worth testing the performance.

EDIT: But I would prefer leaving it for someone else to test. 😅

byorgey: I don't think you need a quadratic number of cases... you just need 2 per constructor, something like
```
compare = \cases
  Nil Nil -> EQ
  Nil _ -> LT
  (One x) (One y) -> compare x y
  (One _) _ -> LT
  etc.
```
Leaving it to test later is totally fine though!

xsebek: Isn't it lexicographic though?
```Haskell
ghci> compare [1] [0, 0]
GT
ghci> compare [1] [2, 2]
LT
```

byorgey: Oh, I guess so, but I don't think the exact semantics of the `Ord` instance matters.  As long as there is *some* instance (so `SmallVector` values can be used as `Map` keys, e.g.) it doesn't matter if we use lexicographic order or shortlex or whatever.

xsebek:
@byorgey a weird error again:
```
    data\test\language-snippets\format\parens.sw:                                                                                  OK

divide by zero

 Attempting to reconstruct a stack trace...

   Frame	Code address
 * 0x66087fd670	0x7ff753cd7e4e D:\a\swarm\swarm\dist-newstyle\build\x86_64-windows\ghc-9.12.4\swarm-0.8.0.0\t\swarm-integration\build\swarm-integration\swarm-integration.exe+0x1d07e4e
 * 0x66087fd678	0x7ff7528b45d6 D:\a\swarm\swarm\dist-newstyle\build\x86_64-windows\ghc-9.12.4\swarm-0.8.0.0\t\swarm-integration\build\swarm-integration\swarm-integration.exe+0x8e45d6
 * 0x66087fd680	0x7ff4908f3db0
 * 0x66087fd688	0x7ff49166c9f0
 * 0x66087fd690	0x7ff49112c999
```

byorgey (approved):

byorgey:
That is really a weird error.  Is it even coming from the Haskell runtime?  We might need to ask a Windows expert.  It's hard for me to imagine how we would get a divide by zero while formatting code.

xsebek:
@mergify queue

byorgey:
I think when we changed the GHC versions in #2786 we forgot to also update `mergify.yml` to match, so it is waiting for actions to complete with out-of-date names.

PR #2777, opened 2026-08-22
Add eventlog traceMarkerIO

xsebek:
* to be merged after #2773

### DIFF
--- a/src/swarm-engine/Swarm/Game/ScenarioInfo.hs
+++ b/src/swarm-engine/Swarm/Game/ScenarioInfo.hs
@@ -42,6 +42,7 @@ import Data.Map.Ordered qualified as OM
 import Data.Maybe (isJust)
 import Data.Text (Text)
 import Data.Yaml as Y
+import Debug.Trace (traceMarkerIO)
 import Effectful
 import Effectful.Error.Static
 import Swarm.Effect.Warn.Local (Warn, warn)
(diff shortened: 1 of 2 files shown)


src/swarm-engine/Swarm/Game/ScenarioInfo.hs
```diff
@@ -111,11 +112,17 @@ loadScenarios ::
...
   (String -> Bool) ->
   Eff es (ScenarioCollection ScenarioInfo)
-loadScenarios scenarioInputs filterScenarioDir = do
+loadScenarios scenarioInputs filterScenarioDir = marked $ do
   res <- runErrorNoCallStack @SystemFailure $ getDataDirThrow Scenarios "scenarios"
   case res of
     Left err -> warn err >> pure emptyCollection
     Right dataDir -> loadCollection (scenarioCollectionConfig scenarioInputs filterScenarioDir) dataDir
+ where
+  marked a = do
```
byorgey, line 121: Is it worth factoring out a utility function in `Swarm.Util` (or `Swarm.Util.X` for some appropriate `X`) like
```
markedWith label a = do
  liftIO . traceMarkerIO $ "Begin " <> label
  r <- a
  liftIO . traceMarkerIO $ "End " <> label
  pure r
```
?

byorgey (approved):

src/swarm-engine/Swarm/Game/ScenarioInfo.hs
xsebek: Maybe. 🙂 I wrote some ideas in Doscord but did not have time to try them out:

> You could also put the `traceFunction` inside the function it is tracing and have it pull the caller name from callstack: https://stackoverflow.com/a/60292991
But that might prevent inlining it entirely.

> I meant to save having to write `traceFunction "START <funcname>" $` and instead just write `traceFunction $` 🙂

But at least in eventlog2html you can't even see most markers because they overlap. So maybe it would be better to just have start markers.

PR #2781, opened 2026-08-22
[PATCH] Fix a large space leak

xsebek:
* add manual `plate` implementation
* fix HLint and reformat

According to ghc-debug, quite a few thunks were being allocated to the definition of plate via Data and a lack of specialization. Adding an INLINE pragma to plate and implementing it manually seems to have stopped these thunks.

Co-Authored-By: Patrick Estrada <212604433+patrickestrada@users.noreply.github.com>


xsebek:
@patrickestrada this really helps a lot with the allocations, so I hope you wont mind that I pushed your work. 🙂

I set you as the author, so git blame should find you. 😉 

| Before | After |
| - | - |
| before | after |

byorgey (approved):
🎉

PR #2790, opened 2026-08-24
Update haskell-ci

xsebek:
* bump GHC minor versions
  * see #2786
* regenerate haskell-ci
  * this likely fixes some Node 20 action warnings

PR #2791, opened 2026-08-28
Fix mergify config and test future changes

xsebek:


.github/mergify.yml
```diff
@@ -14,6 +14,8 @@ queue_rules:
           - -files~=cabal.project.*
           - -files~=^[^/]*\.yaml$
           - -files~=^data/.*\.yaml$
+          - -files~=^\.github/workflows/haskell-ci.yaml$
+          - -files~=^\.github/workflows/haskell-ci-simple.yaml$
```
xsebek, line 18: With this change, https://github.com/swarm-game/swarm/pull/2790 should NOT pass, because it does not update its check-success yet.

byorgey (approved):

PR #2792, opened 2026-08-28
Use pooledMapConcurrently for scenario loading

### LINKED ISSUE #2774: Speed up scenario loading

Loading scenarios on startup takes a long time. 

Loading scenarios should not look like the game froze for 20 seconds. Possible solutions:

* Add loading screen
* Loadsthe scenarios in the background
* Load the scenarios in parallel to make it _faster_
* Load the scenarios lazily when the scenario menu is opened

**Additional context**

From Patrik on Discord:

> Does it usually take a few seconds to start the swarm executable?

> Well unfortunately quite a bit longer than a few seconds

> @xsebek: Yes, it is loading all scenarios eagerly. We should add a loading screen or load them lazily.

> Based on some profiling I'm doing, it seems like populating comments is taking more time than parsing terms, by a factor of between 2x and 3.5x. This might be worth improving in the future, though I suspect that just loading scenarios lazily will make this much less impactful.

> Compiled with late-cost-center profiling and looked at the resulting profile in speedscope:
>  Image

> @byorgey: I also wonder if it wouldn't be too bad to fork off a thread that loads the scenarios in the background and sends a message when they have been loaded, like we currently do for checking whether there is a more recent upstream version.

Related issues:
* #2772

xsebek:
* closes #2774

### DIFF
--- a/src/swarm-util/Swarm/ResourceLoading/Collection.hs
+++ b/src/swarm-util/Swarm/ResourceLoading/Collection.hs
@@ -33,7 +33,6 @@ import Data.Map qualified as M
 import Data.Map.Ordered (OMap)
 import Data.Map.Ordered qualified as OM
 import Data.Text (Text)
-import Debug.Trace (traceEventIO)
 import Effectful
 import Effectful.Concurrent.Async (pooledMapConcurrently, runConcurrent)
 import Effectful.Error.Static
--- a/src/swarm-util/Swarm/ResourceLoading/Collection.hs
+++ b/src/swarm-util/Swarm/ResourceLoading/Collection.hs
@@ -33,7 +33,9 @@ import Data.Map qualified as M
 import Data.Map.Ordered (OMap)
 import Data.Map.Ordered qualified as OM
 import Data.Text (Text)
+import Debug.Trace (traceEventIO)
 import Effectful
+import Effectful.Concurrent.Async (pooledMapConcurrently, runConcurrent)
 import Effectful.Error.Static
 import Swarm.Effect.Warn.Local (Warn, warn)
 import Swarm.Failure (
@@ -156,7 +158,8 @@ loadCollection cfg dir = do
   loadItems :: [FilePath] -> Eff es (Map FilePath (CollectionItem a))
   loadItems items = do
     let loadItem fp = runErrorNoCallStack @SystemFailure $ (fp,) <$> loadCollectionItem cfg (dir </> fp)
-    okItems <- traverseW loadItem items
+    eItems <- runConcurrent $ pooledMapConcurrently loadItem items
+    okItems <- traverseW pure eItems
     return $ M.fromList okItems
 
   -- Load a collection with items sorted alphabetically by file path, and
@@ -188,7 +191,7 @@ loadCollectionItem ::
   CollectionConfig a ->
   FilePath ->
   Eff es (CollectionItem a)
-loadCollectionItem cfg path = do
+loadCollectionItem cfg path = marked $ do
   isDir <- liftIO $ doesDirectoryExist path
   let collectionName = into @Text . takeBaseName $ path
   case isDir of
@@ -198,3 +201,9 @@ loadCollectionItem cfg path = do
       case eitherItem of
         Right (ws, item) -> mapM_ warn ws >> pure (Single item)
         Left loadFailure -> throwError loadFailure
+ where
+  marked a = do
+    liftIO $ traceEventIO $ "START loadCollectionItem " <> path
+    r <- a
+    liftIO $ traceEventIO $ "STOP loadCollectionItem " <> path
+    pure r
--- a/src/swarm-util/Swarm/ResourceLoading/Collection.hs
+++ b/src/swarm-util/Swarm/ResourceLoading/Collection.hs
@@ -191,7 +191,7 @@ loadCollectionItem ::
   CollectionConfig a ->
   FilePath ->
   Eff es (CollectionItem a)
-loadCollectionItem cfg path = marked $ do
+loadCollectionItem cfg path = do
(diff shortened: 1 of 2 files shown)


src/swarm-util/Swarm/ResourceLoading/Collection.hs
```diff
@@ -156,7 +158,8 @@ loadCollection cfg dir = do
   loadItems :: [FilePath] -> Eff es (Map FilePath (CollectionItem a))
   loadItems items = do
     let loadItem fp = runErrorNoCallStack @SystemFailure $ (fp,) <$> loadCollectionItem cfg (dir </> fp)
-    okItems <- traverseW loadItem items
+    eItems <- runConcurrent $ pooledMapConcurrently loadItem items
```
byorgey, line 160: What's the thinking behind using `pooledMapConcurrently` here?  I didn't even know about that function previously; why did you choose it instead of any of the 147 other similar functions?  Just curious.

xsebek: @patrickestrada mentioned it on Discord:

https://discord.com/channels/1260717118576525372/1260717118576525376/1541034655275425902

I just chose the other one that sets the number based on capabilities/CPU cores.

I would like to measure it better, I only started playing with ghc-events-analyze.

https://well-typed.com/blog/2014/02/ghc-events-analyze/

xsebek:
| | Before | After |
| - | - | - |
| Startup `-O1` <sup>(1)</sup> | 2.530 s ±  0.030 s | 0.798 s ±  0.039 s |
| Profiling startup | 18.5 s | 5.5 s |
| Heap |  |  |

1) I added `exitSuccess` at the end of `initPersistentState` to make it possible to measure with `hyperfine`.

xsebek:
@byorgey I think this already helps, but maybe it could be even better. What I don't like is that the concurrent map is over directories as well. I think it would be better in two phases, first load `Collection FilePath` and then load scenarios in parallel.

So something like this:
```Haskell
loadCollection' cfg dir = do
   -- Collection FilePath
   collectedPaths <- loadCollection cfg { loadItem = \p -> pure (Right ([], p)) } dir
   -- Collection a
   traverseW (liftIO . loadItem cfg) collectedPaths
```

But I struggle to write `Witherable` for `Collection`. @byorgey any tips how to do that?

byorgey:
@xsebek splitting it into loading `Collection FilePath` + then loading scenarios sounds like I good idea, I think that should work.  Here's what I came up with for `Witherable Collection`.  We can derive `Foldable` and `Traversable` for `Collection` and `CollectionItem`.  Then we can implement `Filterable`.  Then `Witherable` comes for free:

```
-- | A collection of @a@ is a tree, where at each level we map
--   FilePaths to either singleton items of type @a@, or nested
--   subcollections.
newtype Collection a = Collection
  {collectionMap :: OMap FilePath (CollectionItem a)}
  deriving (Functor, Foldable, Traversable)

-- | Either a singleton item, or a nested subcollection with a label.
data CollectionItem a = Single a | SubCollection Text (Collection a)
  deriving (Functor, Foldable, Traversable)

makePrisms ''CollectionItem

instance Filterable Collection where
  catMaybes :: Collection (Maybe a) -> Collection a
  catMaybes (Collection m) = Collection (mapMaybe filterItem m)

instance Ord k => Filterable (OMap k) where
  catMaybes :: OMap k (Maybe a) -> OMap k a
  catMaybes = OM.fromList . mapMaybe strength . OM.assocs
   where
    strength (k, ma) = (k,) <$> ma

filterItem :: CollectionItem (Maybe a) -> Maybe (CollectionItem a)
filterItem = \case
  Single ma -> Single <$> ma
  SubCollection label c -> Just $ SubCollection label (catMaybes c)

instance Witherable Collection
```
Unfortunately this depends on an orphan `Filterable` instance for `OMap`, but we could contribute one upstream.  With access to the internals there is surely a better implementation than going via `assocs` and `fromList`.

xsebek:
@byorgey thanks a lot! I finished it in `task/xsebek/pooled-scenario-wither` and measured:

| | Before | After | After Two Phases |
|--| -- | -- | -- |
| Startup `-O1` | 2.530 s ± 0.030 s | 0.798 s ± 0.039 s | 0.867 s ± 0.006 s |

It took a lot of effort and it is **worse**. 🤦

My worry was that starting too many actions would be bad with scenarios fighting for CPU, but apparently it is fine.

byorgey:
> It took a lot of effort and it is worse. 🤦

Hah, well, this is why we do profiling!

src/swarm-util/Swarm/ResourceLoading/Collection.hs
```diff
@@ -198,3 +201,9 @@ loadCollectionItem cfg path = do
       case eitherItem of
         Right (ws, item) -> mapM_ warn ws >> pure (Single item)
         Left loadFailure -> throwError loadFailure
+ where
+  marked a = do
+    liftIO $ traceEventIO $ "START loadCollectionItem " <> path
```
byorgey: Was this just for debugging or is it intended to be permanent?

byorgey (approved):
Tried it out and it is indeed quite a bit faster!

src/swarm-util/Swarm/ResourceLoading/Collection.hs
xsebek: I think we could keep it, it only gets called once for each scenario.

But it is easy to add those START/STOP depending on what you are measuring and keeping them around might just add noise.

So on second thought let's delete it.

xsebek: traceMarkerIO is the one thats permanent, that one shows as line in eventlog heap profile.

